PR #10, opened 2016-11-08
Fix authentication issues

ydogandjiev:
This change addresses two major issues we've seen with the current implementation of the authentication APIs:
1. The call to microsoftTeams.authentication.authenticate is often triggering the browser's popup blocker because we are calling window.open from the Microsoft Teams window in response to a posted message as opposed to the click handler of the element responsible for kicking off the authentication flow.
2. Authentication is failing in Edge and IE11 for certain tabs that trigger redirects in the authentication popup across different security zones.

The change fixes these issues as follows:
1. The call to window.open will now happen inside the iframe and as long as microsoftTeams.authentication.authenticate is called from a click handler browsers will not block the popup.
2. Instead of relying on window.opener the authentication flow now uses localStorage to communicate success/failure as well as flow completion between the popup and the iframe. Additionally, the authentication popup now closes itself at the end of the flow instead of relying on its opener to do so.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -535,6 +535,7 @@ namespace microsoftTeams
             }
 
             delete handlers["initialize"];
+            delete handlers["navigateCrossDomain"];
         }
 
         function startAuthenticationWindowMonitor(): void
@@ -572,6 +573,15 @@ namespace microsoftTeams
             handlers["initialize"] = () =>
             {
                 return [ frameContexts.authentication, hostClientType ];
+            };
+
+            // Set up a navigateCrossDomain message handlers that will block cross-domain re-navigation attempts
+            // in the authentication window. We could at some point choose to implement this method via a call to
+            // authenticationWindow.location.href = url; however, we would first need to figure out how to
+            // validate the url against the tab's list of valid domains.
+            handlers["navigateCrossDomain"] = (url: string) =>
+            {
+                return false;
             };
         }
 
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -550,7 +550,7 @@ namespace microsoftTeams
                 {
                     handleFailure("CancelledByUser");
                 }
-                else if (!childOrigin)
+                else
                 {
                     try
                     {
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -85,7 +85,7 @@ namespace microsoftTeams
         // Undocumented field used to mock the window for unit tests
         currentWindow = this._window as Window || window;
 
-        // Listen for messages post to our window (in a way that works for all browsers)
+        // Listen for messages post to our window
         let messageListener = (evt: MessageEvent) => processMessage(evt);
         currentWindow.addEventListener("message", messageListener, false);
 
@@ -446,6 +446,13 @@ namespace microsoftTeams
             // Open an authentication window with the parameters provided by the caller
             openAuthenticationWindow(authenticateParameters.url, authenticateParameters.width, authenticateParameters.height);
 
+            // If we failed to open the window fail the authentication flow
+            if (!childWindow)
+            {
+                handleFailure("FailedToOpenWindow");
+                return;
+            }
+
             authParams = authenticateParameters;
         }
 
@@ -515,11 +522,6 @@ namespace microsoftTeams
 
                 // Open the window with a desired set of standard browser features
                 childWindow = currentWindow.open(link.href, "_blank", "toolbar=no, location=yes, status=no, menubar=no, top=" + top + ", left=" + left + ", width=" + width + ", height=" + height);
-            }
-
-            if (!childWindow)
-            {
-                throw new Error("Failed to open authentication window.");
             }
 
             // Start monitoring the authentication window so that we can detect if it gets closed before the flow completes
@@ -746,12 +748,12 @@ namespace microsoftTeams
             return;
         }
 
-        // Process only if the message is coming from a valid origin
+        // Process only if the message is coming from a different window and a valid origin
         let messageSource = evt.source || evt.originalEvent.source;
         let messageOrigin = evt.origin || evt.originalEvent.origin;
-        if ((validOrigins.indexOf(messageOrigin.toLowerCase()) === -1) &&
-            (messageOrigin !== currentWindow.location.origin) ||
-            (messageSource === currentWindow))
+        if (messageSource === currentWindow ||
+            (messageOrigin !== currentWindow.location.origin &&
+             validOrigins.indexOf(messageOrigin.toLowerCase()) === -1))
         {
             return;
         }
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -48,12 +48,14 @@ namespace microsoftTeams
     // It does not indicate whether initialization is complete. That can be inferred by whether parentOrigin is set.
     let initializeCalled = false;
 
+    let currentWindow: Window;
     let parentWindow: Window;
     let parentOrigin: string;
     let messageQueue: MessageRequest[] = [];
     let nextMessageId = 0;
     let callbacks: {[id: number]: Function} = {};
     let frameContext: string;
+    let loadedInDesktopClient: boolean;
 
     let themeChangeHandler: (theme: string) => void;
     handlers["themeChange"] = handleThemeChange;
@@ -72,19 +74,34 @@ namespace microsoftTeams
         initializeCalled = true;
 
         // Undocumented field used to mock the window for unit tests
-        let currentWindow = this._window as Window || window;
+        currentWindow = this._window as Window || window;
+        let messageListener: (evt: MessageEvent) => void;
 
-        // If we are in an iframe then our parent window is the one hosting us (i.e. window.parent); otherwise,
-        // it's the window that opened us (i.e. window.opener)
+        // If we are in an iframe then our parent window is the one hosting us (i.e. window.parent)
         parentWindow = (currentWindow.parent !== currentWindow.self) ? currentWindow.parent : currentWindow.opener;
-        if (!parentWindow)
+        if (parentWindow)
         {
-            throw new Error("This page must be loaded in an iframe");
-        }
+            // Listen for messages post to our window (in a way that works for all browsers)
+            messageListener = (evt: MessageEvent) => processMessage(evt);
+            currentWindow.addEventListener("message", messageListener, false);
 
-        // Listen for messages post to our window (in a way that works for all browsers)
-        let messageListener = (evt: MessageEvent) => processMessage(evt);
-        currentWindow.addEventListener("message", messageListener, false);
+            try
+            {
+                // Send the initialized message to any origin since at this point we most likely don't know what our
+                // parent window's origin is yet and this message contains no data that could pose a security risk.
+                parentOrigin = "*";
+                let messageId = sendMessage("initialize");
+                callbacks[messageId] = (context: string, inDesktopClient: boolean) =>
+                {
+                    frameContext = context;
+                    loadedInDesktopClient = inDesktopClient;
+                };
+            }
+            finally
+            {
+                parentOrigin = null;
+            }
+        }
 
         // Undocumented function used to clear state between unit tests
         this._uninitialize = () =>
@@ -109,22 +126,6 @@ namespace microsoftTeams
 
             currentWindow.removeEventListener("message", messageListener, false);
         };
-
-        try
-        {
-            // Send the initialized message to any origin since at this point we most likely don't know what our
-            // parent window's origin is yet and this message contains no data that could pose a security risk.
-            parentOrigin = "*";
-            let messageId = sendMessage("initialize");
-            callbacks[messageId] = (context: string) =>
-            {
-                frameContext = context;
-            };
-        }
-        finally
-        {
-            parentOrigin = null;
-        }
     }
 
     /**
@@ -421,22 +422,62 @@ namespace microsoftTeams
             let link = document.createElement("a");
             link.href = authenticateParameters.url;
 
-            let messageId = sendMessage(
-                "authentication.authenticate",
-                link.href,
-                authenticateParameters.width,
-                authenticateParameters.height);
-            callbacks[messageId] = (success: boolean, response: string) =>
+            // Clear any previous success/failure values
+            localStorage.removeItem("authentication.success");
+            localStorage.removeItem("authentication.failure");
+
+            // Open a popup window to the specified authentication URL
+            let authWindow: Window;
+            if (loadedInDesktopClient)
             {
-                if (success)
+                let messageId = sendMessage("authentication.authenticate", link.href);
+                callbacks[messageId] = (allowed: boolean) =>
                 {
-                    authenticateParameters.successCallback(response);
-                }
-                else
+                    if (allowed)
+                    {
+                        authWindow = currentWindow.open(link.href, "Microsoft Teams", "width = " + authenticateParameters.width + ", height = " + authenticateParameters.height);
+                    }
+                    else
+                    {
+                        throw new Error("Authentication is only supported for URLs matching the pattern registered in the manifest.");
+                    }
+                };
+            }
+            else
+            {
+                authWindow = currentWindow.open(link.href, "Microsoft Teams", "width = " + authenticateParameters.width + ", height = " + authenticateParameters.height);
+            }
+
+            if (!authWindow)
+            {
+                throw new Error("Failed to open authentication window.");
+            }
+
+            // Wait for the authentication window to close
+            let interval = setInterval(() =>
+            {
+                let result: string = localStorage.getItem("authentication.success");
+                let reason: string = localStorage.getItem("authentication.failure");
+                if (result || reason)
                 {
-                    authenticateParameters.failureCallback(response);
+                    clearInterval(interval);
+                    authWindow.close();
+
+                    // Notify caller of the authentication result
+                    if (result)
+                    {
+                        authenticateParameters.successCallback(result);
+                    }
+                    else
+                    {
+                        authenticateParameters.failureCallback(reason);
+                    }
+
+                    // Clear the success/failure values
+                    localStorage.removeItem("authentication.success");
+                    localStorage.removeItem("authentication.failure");
                 }
-            };
+            }, 200);
         }
 
         /**
@@ -449,7 +490,7 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.authentication);
 
-            sendMessage("authentication.authenticate.success", result);
+            localStorage.setItem("authentication.success", result);
         }
 
         /**
@@ -462,7 +503,7 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.authentication);
 
-            sendMessage("authentication.authenticate.failure", reason);
+            localStorage.setItem("authentication.failure", reason);
         }
 
         export interface AuthenticateParameters
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -525,7 +525,10 @@ namespace microsoftTeams
             }
 
             // Start monitoring the authentication window so that we can detect if it gets closed before the flow completes
-            startAuthenticationWindowMonitor();
+            if (childWindow)
+            {
+                startAuthenticationWindowMonitor();
+            }
         }
 
         function stopAuthenticationWindowMonitor(): void
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -34,7 +34,7 @@ namespace microsoftTeams
     const hostClientTypes =
     {
         desktop: "desktop",
-        webb: "web",
+        web: "web",
     };
 
     interface MessageRequest
@@ -133,6 +133,7 @@ namespace microsoftTeams
             nextMessageId = 0;
             callbacks = {};
             frameContext = null;
+            hostClientType = null;
 
             currentWindow.removeEventListener("message", messageListener, false);
         };
@@ -171,7 +172,7 @@ namespace microsoftTeams
 
         if (childWindow)
         {
-            sendMessageRequest(childWindow, "themeChange", theme);
+            sendMessageRequest(childWindow, "themeChange", [ theme ]);
         }
     }
 
@@ -186,7 +187,7 @@ namespace microsoftTeams
     {
         ensureInitialized();
 
-        let messageId = sendMessageRequest(parentWindow, "navigateCrossDomain", url);
+        let messageId = sendMessageRequest(parentWindow, "navigateCrossDomain", [ url ]);
         callbacks[messageId] = (success: boolean) =>
         {
             if (!success)
@@ -216,7 +217,7 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.settings, frameContexts.remove);
 
-            sendMessageRequest(parentWindow, "settings.setValidityState", validityState);
+            sendMessageRequest(parentWindow, "settings.setValidityState", [ validityState ]);
         }
 
         /**
@@ -241,7 +242,7 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.settings);
 
-            sendMessageRequest(parentWindow, "settings.setSettings", settings);
+            sendMessageRequest(parentWindow, "settings.setSettings", [ settings ]);
         }
 
         /**
@@ -363,7 +364,7 @@ namespace microsoftTeams
             {
                 this.ensureNotNotified();
 
-                sendMessageRequest(parentWindow, "settings.save.failure", reason);
+                sendMessageRequest(parentWindow, "settings.save.failure", [ reason ]);
 
                 this.notified = true;
             }
@@ -408,7 +409,7 @@ namespace microsoftTeams
             {
                 this.ensureNotNotified();
 
-                sendMessageRequest(parentWindow, "settings.remove.failure", reason);
+                sendMessageRequest(parentWindow, "settings.remove.failure", [ reason ]);
 
                 this.notified = true;
             }
@@ -468,7 +469,7 @@ namespace microsoftTeams
             }
         }
 
-        function openAuthenticationWindow(url: string, width?: number, height?: number): void
+        function openAuthenticationWindow(url: string, width: number, height: number): void
         {
             // Close the previously opened window if we have one
             closeAuthenticationWindow();
@@ -490,7 +491,7 @@ namespace microsoftTeams
                 // We are running in our desktop app so give our main process a heads up on the URL we are about
                 // to call window.open with; this will allow the window.open call to pass through without getting
                 // kicked out into the browser (which is the default behavior)
-                let messageId = sendMessageRequest(parentWindow, "authentication.allowWindowOpen", link.href);
+                let messageId = sendMessageRequest(parentWindow, "authentication.allowWindowOpen", [ link.href ]);
                 callbacks[messageId] = (allowed: boolean) =>
                 {
                     if (allowed)
@@ -507,8 +508,8 @@ namespace microsoftTeams
             else
             {
                 // We are running in the browser so we need to center the new window ourselves
-                let left: number = (typeof currentWindow.screenLeft !== undefined) ? currentWindow.screenLeft : currentWindow.screenX;
-                let top: number = (typeof currentWindow.screenTop !== undefined) ? currentWindow.screenTop : currentWindow.screenY;
+                let left: number = (typeof currentWindow.screenLeft !== "undefined") ? currentWindow.screenLeft : currentWindow.screenX;
+                let top: number = (typeof currentWindow.screenTop !== "undefined") ? currentWindow.screenTop : currentWindow.screenY;
                 left += (currentWindow.outerWidth / 2) - (width / 2);
                 top += (currentWindow.outerHeight / 2) - (height / 2);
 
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -421,22 +422,61 @@ namespace microsoftTeams
...
-            let messageId = sendMessage(
-                "authentication.authenticate",
-                link.href,
-                authenticateParameters.width,
-                authenticateParameters.height);
-            callbacks[messageId] = (success: boolean, response: string) =>
+            // Clear any previous success/failure values
+            localStorage.removeItem("authentication.success");
+            localStorage.removeItem("authentication.failure");
+
```
dfederm: Isn't local storage per domain? This forces the auth window and content/settings frames to be on the same domain, which breaks our validDomains scenario.

src/MicrosoftTeams.ts
```diff
@@ -421,22 +422,61 @@ namespace microsoftTeams
...
-                if (success)
+                let messageId = sendMessage("authentication.authenticate", link.href);
+                callbacks[messageId] = (allowed: boolean) =>
                 {
-                    authenticateParameters.successCallback(response);
-                }
-                else
+                    if (allowed)
+                    {
+                        authWindow = currentWindow.open(link.href, "Microsoft Teams", "width = " + authenticateParameters.width + ", height = " + authenticateParameters.height);
```
dfederm: >width = " + authenticateParameters.width + ", height = " + authenticateParameters.height); [](start = 87, length = 90)

I can't remember, but I'm pretty sure we don't do param validation in the client today.

src/MicrosoftTeams.ts
```diff
@@ -72,19 +74,34 @@ namespace microsoftTeams
...
-        // Listen for messages post to our window (in a way that works for all browsers)
-        let messageListener = (evt: MessageEvent) => processMessage(evt);
-        currentWindow.addEventListener("message", messageListener, false);
+            try
+            {
+                // Send the initialized message to any origin since at this point we most likely don't know what our
+                // parent window's origin is yet and this message contains no data that could pose a security risk.
+                parentOrigin = "*";
+                let messageId = sendMessage("initialize");
+                callbacks[messageId] = (context: string, inDesktopClient: boolean) =>
```
ydogandjiev: This should become version 0.3 of our SDK and we will only ask tabs to switch to it once the corresponding FE client changes have been made and deployed to PROD. The SDK will actually continue to work but desktop auth will be broken until the FE client changes have been implemented.

src/MicrosoftTeams.ts
```diff
@@ -449,7 +489,8 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.authentication);
 
-            sendMessage("authentication.authenticate.success", result);
+            localStorage.setItem("authentication.success", result);
```
ydogandjiev: Yes, the auth completion endpoint will have to be in the same domain as the endpoint that kicked off the auth flow. I will make sure to document this before asking developers to switch to it. At this point I don't know of a way to make this work across domains without message posting which depends on window.opener. Once auth completes the tab content/config page would be able to use navigateCrossDomain to move to a different domain if it wants (and that's what VSTS does today).

src/MicrosoftTeams.ts
ydogandjiev: Yes, the auth completion endpoint will have to be in the same domain as the endpoint that kicked off the auth flow. I will make sure to document this before asking developers to switch to it. At this point I don't know of a way to make this work across domains without message posting which depends on window.opener. Once auth completes the tab content/config page would be able to use navigateCrossDomain to move to a different domain if it wants (and that's what VSTS does today).

src/MicrosoftTeams.ts
```diff
@@ -421,22 +422,61 @@ namespace microsoftTeams
...
+            }
+
+            if (!authWindow)
+            {
+                throw new Error("Failed to open authentication window.");
+            }
+
+            // Wait for the authentication window to close
+            let interval = setInterval(() =>
+            {
```
ydogandjiev: Good catch; I thought checking the success/failure status was sufficient here but if the user closes the window before the authentication flow completes then we will keep spinning indefinitely. I added a check for the authWindow.closed.

src/MicrosoftTeams.ts
ydogandjiev: I will clean this up so that it follows the same logic as what we have in the FE client code today.

src/MicrosoftTeams.ts
```diff
@@ -421,22 +422,61 @@ namespace microsoftTeams
...
+            localStorage.removeItem("authentication.success");
+            localStorage.removeItem("authentication.failure");
+
+            // Open a popup window to the specified authentication URL
+            let authWindow: Window;
+            if (loadedInDesktopClient)
             {
-                if (success)
+                let messageId = sendMessage("authentication.authenticate", link.href);
+                callbacks[messageId] = (allowed: boolean) =>
```
ydogandjiev: We will never even make this call until the client starts returning isDesktopClient as part of the initialization. I will make all the necessary FE client changes in a single check-in to get the desktop client auth flow work end-to-end with this new version of the SDK.

src/MicrosoftTeams.ts
```diff
@@ -48,12 +48,16 @@ namespace microsoftTeams
     // It does not indicate whether initialization is complete. That can be inferred by whether parentOrigin is set.
     let initializeCalled = false;
 
+    let currentWindow: Window;
     let parentWindow: Window;
     let parentOrigin: string;
+    let childWindow: Window;
+    let childOrigin: string;
     let messageQueue: MessageRequest[] = [];
```
dfederm: Would we ever have more than one child?

src/MicrosoftTeams.ts
```diff
@@ -599,6 +712,7 @@ namespace microsoftTeams
             if (handler)
             {
                 // We don't expect any handler to respond at this point
+                console.log("Invoking handler");
                 handler.apply(this, message.args);
```
dfederm: Remove

src/MicrosoftTeams.ts
```diff
@@ -463,6 +501,43 @@ namespace microsoftTeams
             ensureInitialized(frameContexts.authentication);
 
             sendMessage("authentication.authenticate.failure", reason);
+            window.close();
+        }
+
+        function handleSuccess(result?: string): void
+        {
+            console.log("In handleSuccess");
+            if (authInterval)
```
dfederm: Remove

src/MicrosoftTeams.ts
ydogandjiev: Good catch. Removed.

src/MicrosoftTeams.ts
```diff
@@ -463,6 +501,43 @@ namespace microsoftTeams
             ensureInitialized(frameContexts.authentication);
 
             sendMessage("authentication.authenticate.failure", reason);
+            window.close();
+        }
```
dfederm: Didn't we discover a race condition here where the message wasn't actually sent since it destroyed itself?

src/MicrosoftTeams.ts
```diff
@@ -463,6 +501,43 @@ namespace microsoftTeams
...
+            }
+
+            authParams = null;
+            authWindow = null;
+        }
+
+        function handleFailure(reason?: string): void
+        {
+            console.log("In handleFailure");
+            if (authInterval)
```
dfederm: Remove

src/MicrosoftTeams.ts
```diff
@@ -463,6 +501,43 @@ namespace microsoftTeams
...
+
+        function handleSuccess(result?: string): void
+        {
+            console.log("In handleSuccess");
+            if (authInterval)
+            {
+                clearInterval(authInterval);
+                authInterval = 0;
+            }
+
```
dfederm: Should this happen in the init handler? Also, don't we need an init handler?

src/MicrosoftTeams.ts
dfederm: This is now moot since we're not using localStorage

---
In reply to: [87050549](https://github.com/OfficeDev/microsoft-teams-library-js/pull/10#discussion_r87050549) [](ancestors = 87050549)

src/MicrosoftTeams.ts
dfederm: Yea, I guess I was just thinking we would make the client change first, but I guess since this is a versioned library, it doesn't really matter as long as the client changes happen before we actually release the new version. It does mean that we're deciding that the latest in master may not actually work against any current app, but I think that's probably fine since no one should be taking the latest (unversioned) bits especially since they wouldn't be on any CDN.

src/MicrosoftTeams.ts
dfederm: This was related to my other comment about sequencing the changes with the client first. No longer is an issue.

src/MicrosoftTeams.ts
```diff
@@ -48,12 +48,14 @@ namespace microsoftTeams
...
     let initializeCalled = false;
 
+    let currentWindow: Window;
     let parentWindow: Window;
     let parentOrigin: string;
     let messageQueue: MessageRequest[] = [];
     let nextMessageId = 0;
     let callbacks: {[id: number]: Function} = {};
     let frameContext: string;
+    let loadedInDesktopClient: boolean;
```
ydogandjiev: Yeah, I will change it to an enum on my next iteration. Right now it will just have Web and Desktop but we can add more values to it later.

src/MicrosoftTeams.ts
ydogandjiev: That was just a red herring from what I remember but I will add a timeout around the window.close() just in case.

dfederm:
As a general comment, and perhaps one out of scope for this specific change, but the unit test file is getting a little large. We should perhaps break it up. For example there are some pure declarations that could be its own file, maybe a file per sub-namespace ("base", settings, and auth), and maybe a separate file for mocking and setting up/resetting the library. This perhaps needs some thought though as we should avoid a generic "helpers" file :)

src/MicrosoftTeams.ts
ydogandjiev: Yeah, good point. I was originally thinking that the init would just bubble up to our top-level window and get handled there but that window would have no way of knowing who is calling it or how to give it a proper context. I will also stop the ping loop when the init comes in since at that point it will no longer be necessary.

src/MicrosoftTeams.ts
```diff
@@ -409,6 +414,12 @@ namespace microsoftTeams
 
     export namespace authentication
     {
+        let authWindow: Window;
+        let authInterval: number;
+        let authParams: AuthenticateParameters;
+        handlers["authentication.authenticate.success"] = handleSuccess;
+        handlers["authentication.authenticate.failure"] = handleFailure;
+
```
dfederm: Don't we actually need to proxy *all* calls? For example if the child asks for the context

src/MicrosoftTeams.ts
```diff
@@ -417,26 +428,52 @@ namespace microsoftTeams
...
+                    {
+                        authParams = null;
+                        throw new Error("Authentication is only supported for URLs matching the pattern registered in the manifest.");
+                    }
+                };
+            }
+            else
+            {
+                authWindow = currentWindow.open(link.href, "Microsoft Teams", "width = " + authParams.width + ", height = " + authParams.height);
+            }
```
dfederm: This is duped above. I wonder if it's worth it to extract to a separate function to avoid duplication/potential divergence.

src/MicrosoftTeams.ts
```diff
@@ -417,26 +428,52 @@ namespace microsoftTeams
...
+
+            if (!authWindow)
+            {
+                authParams = null;
+                throw new Error("Failed to open authentication window.");
+            }
```
dfederm: Haven't thought about this too much, but I wonder if instead of throwing we should just immediately invoke the failure callback.

src/MicrosoftTeams.ts
```diff
@@ -417,26 +428,52 @@ namespace microsoftTeams
...
+                if (!authWindow || authWindow.closed)
                 {
-                    authenticateParameters.successCallback(response);
+                    handleFailure();
                 }
                 else
                 {
-                    authenticateParameters.failureCallback(response);
+                    authWindow.postMessage("ping", "*");
                 }
```
dfederm: Does this new message type need a handler? I would have imagined that it would to complete the wire up on the child side.

src/MicrosoftTeams.ts
```diff
@@ -417,26 +428,52 @@ namespace microsoftTeams
...
+
+            // Wait for the authentication window to close
+            authInterval = setInterval(() =>
             {
-                if (success)
+                if (!authWindow || authWindow.closed)
                 {
-                    authenticateParameters.successCallback(response);
+                    handleFailure();
                 }
```
dfederm: Is there a reason we can include to the caller to help them understand what happened?

src/MicrosoftTeams.ts
```diff
@@ -417,26 +428,52 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
 
+            authParams = authenticateParameters;
+
```
dfederm: THink about if authenticate is called twice. Shoudl we force-close the existing windows and/or call notifyFailure?

src/MicrosoftTeams.ts
```diff
@@ -409,6 +414,12 @@ namespace microsoftTeams
 
     export namespace authentication
     {
+        let authWindow: Window;
+        let authInterval: number;
```
dfederm: just use childWindow?

src/MicrosoftTeams.ts
ydogandjiev: No handler is needed but the payload does need to be an object; otherwise it gets rejected by the child window. Fix.

src/MicrosoftTeams.ts
```diff
@@ -31,6 +31,12 @@ namespace microsoftTeams
         remove: "remove",
     };
 
+    const hostClientTypes =
+    {
+        desktop: "desktop",
+        webb: "web",
```
dfederm: >webb [](start = 8, length = 4)

spelling

src/MicrosoftTeams.ts
```diff
@@ -102,29 +125,16 @@ namespace microsoftTeams
             initializeCalled = false;
             parentWindow = null;
             parentOrigin = null;
-            messageQueue = [];
+            parentMessageQueue = [];
+            childWindow = null;
+            childOrigin = null;
+            childMessageQueue = [];
             nextMessageId = 0;
```
dfederm: Null out hostClientType too

src/MicrosoftTeams.ts
```diff
@@ -417,25 +441,132 @@ namespace microsoftTeams
...
+                    {
+                        childOrigin = null;
+                    }
                 }
+            }, 100);
+
+            // Set up an initialize message handler that will give the authentication window its frame context
+            handlers["initialize"] = () =>
+            {
+                return frameContexts.authentication;
```
dfederm: >return frameContexts.authentication; [](start = 16, length = 36)

Missing the host type.

src/MicrosoftTeams.ts
```diff
@@ -417,25 +441,132 @@ namespace microsoftTeams
...
+        function stopAuthenticationWindowMonitor(): void
+        {
+            if (authWindowMonitor)
+            {
+                clearInterval(authWindowMonitor);
+                authWindowMonitor = 0;
+            }
+
+            handlers["initialize"] = null;
+        }
```
dfederm: Consider: `delete handlers["initialize"];` to just remove it from the object completely.

src/MicrosoftTeams.ts
```diff
@@ -417,25 +441,132 @@ namespace microsoftTeams
...
+                }
+            }
+            finally
+            {
+                childWindow = null;
+                childOrigin = null;
+            }
+        }
+
+        function openAuthenticationWindow(url: string, width?: number, height?: number): void
```
dfederm: >width?: number, height?: number [](start = 55, length = 31)

These two params actually aren't optional.

Note that there's a difference b/t optional params (not provided in the invocation) and passing null/undefined as the value.

src/MicrosoftTeams.ts
```diff
@@ -417,25 +441,132 @@ namespace microsoftTeams
...
+                    else
+                    {
+                        throw new Error("Authentication is only supported for URLs matching the pattern registered in the manifest.");
+                    }
+                };
+            }
+            else
+            {
+                // We are running in the browser so we need to center the new window ourselves
+                let left: number = (typeof currentWindow.screenLeft !== undefined) ? currentWindow.screenLeft : currentWindow.screenX;
```
dfederm: >(typeof currentWindow.screenLeft !== undefined) [](start = 35, length = 47)

the typeof operator always returns a string, so this never passes. You should use `"undefined"` (the string) instead of `undefined` (the literal).

Alternately, you can just remove the typeof operator completely and check if `currentWindow.screenLeft !== undefined`

src/MicrosoftTeams.ts
```diff
@@ -417,25 +442,132 @@ namespace microsoftTeams
...
                 }
-                else
+                else if (!childOrigin)
                 {
-                    authenticateParameters.failureCallback(response);
+                    try
+                    {
+                        childOrigin = "*";
+                        sendMessageRequest(childWindow, "ping");
+                    }
```
dfederm: Instead of pinging until the success/fail/window close, should this the child respond with something and the callback do the teardown of the interval?

src/MicrosoftTeams.ts
```diff
@@ -417,25 +442,132 @@ namespace microsoftTeams
...
+            // Stop monitoring the authentication window
+            stopAuthenticationWindowMonitor();
+
+            // Try to close the authentication window and clear all properties associated with it
+            try
+            {
+                if (childWindow)
+                {
+                    childWindow.close();
+                }
```
dfederm: Looks like the child window is trying to close itself as well as the the parent trying to close the child. Should we be content with one or the other?

src/MicrosoftTeams.ts
```diff
@@ -604,32 +817,113 @@ namespace microsoftTeams
...
+        if (targetOrigin)
         {
-            parentWindow.postMessage(request, parentOrigin);
+            targetWindow.postMessage(request, targetOrigin);
         }
         else
         {
-            messageQueue.push(request);
+            let targetMessageQueue = getTargetMessageQueue(targetWindow);
+            if (targetMessageQueue)
```
dfederm: >targetMessageQueue [](start = 16, length = 18)

getTargetMessageQueue can't return null, so no need for this check.

src/MicrosoftTeams.ts
```diff
@@ -564,23 +736,64 @@ namespace microsoftTeams
...
+        {
+            parentWindow = messageSource;
+            parentOrigin = messageOrigin;
+        }
+        else if (!childWindow || (messageSource === childWindow))
+        {
+            childWindow = messageSource;
+            childOrigin = messageOrigin;
+        }
+
```
dfederm: Shouldn't we only update either origin if it's null? It seems like if the child or parent are compromised, we loose security because we will listen to anything that sends us a message.

dfederm: This is interesting. If the parent or child sites are compromised, we'll still listen to them regardless with the next message they send us. navigateCrossDomain seems only necessary if the parent is the client? I'm not sure if this is strong enough security.

If the parent is the client, parentOrigin should never change. If the parent is not the client and the parent navigates cross origin (telling its parent), without this communication would break.

If the child navigates cross origin, they would call navigateCrossDomain which the parent can handle just as the client does.

So the non-client cross origin navigate while the popup is up is the only case that might not work.

Thoughts?

src/MicrosoftTeams.ts
ydogandjiev: Fixed

src/MicrosoftTeams.ts
ydogandjiev: Good catch, added

src/MicrosoftTeams.ts
ydogandjiev: Ah yes, good point

src/MicrosoftTeams.ts
ydogandjiev: As discussed offline, we only call updateRelationships if the messageOrigin matches one of our validDomains or is equal to our current origin (i.e. window.location.origin). This is enforced by processMessage. Therefore, even if the parent or child get compromised and navigate to some different non-valid domain then we will ignore their messages.

src/MicrosoftTeams.ts
ydogandjiev: Simplified down to a single line

src/MicrosoftTeams.ts
ydogandjiev: Well in certain cases (e.g. where authenticate gets called again before the previous auth flow completes) we would need to be the ones to close the popup. Since closeAuthenticationWindow does more than just close the window it needs to be called in several places where a call to childWindow.close() isn't necessarily needed; however, I think it's simplest and safest to just do it in all cases. Worst case the browser gets two requests to close the popup which should be just fine.

src/MicrosoftTeams.ts
ydogandjiev: I've seen this call throw before but I can't seem to reproduce it now. I mostly added the try...finally because I wanted to make sure that we always clean up the childWindow and childOrigin regardless of whether closing the window was successful or not. At this point I would prefer to let the exception bubble out so that I can make sure this call isn't failing in some unexpected cases. I might just completely remove the try...finally at some point.

src/MicrosoftTeams.ts
```diff
@@ -449,7 +581,11 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.authentication);
 
-            sendMessage("authentication.authenticate.success", result);
+            sendMessageRequest(parentWindow, "authentication.authenticate.success", result);
+
+            setTimeout(() => {
+                window.close();
+            }, 100);
         }
```
ydogandjiev: The timeout is actually unnecessary. I previously thought that postMessage was failing if the window got closed too quickly after it was called; however, I later on figured out that the IE/Edge issue was actually caused by window.opener getting reset in certain auth flows and had nothing to do with the window.close() call. I've removed the timeout.

src/MicrosoftTeams.ts
ydogandjiev: Ah, oops. I copied this from stack overflow but clearly I didn't do it right. Fixed.

src/MicrosoftTeams.ts
```diff
@@ -417,25 +441,132 @@ namespace microsoftTeams
...
+        }
+
+        function closeAuthenticationWindow(): void
+        {
+            // Stop monitoring the authentication window
+            stopAuthenticationWindowMonitor();
+
+            // Try to close the authentication window and clear all properties associated with it
+            try
+            {
```
ydogandjiev: I've seen this call throw before but I can't seem to reproduce it now. I mostly added the try...finally because I wanted to make sure that we always clean up the childWindow and childOrigin regardless of whether closing the window was successful or not. At this point I would prefer to let the exception bubble out so that I can make sure this call isn't failing in some unexpected cases. I might just completely remove the try...finally at some point.

src/MicrosoftTeams.ts
ydogandjiev: Done

src/MicrosoftTeams.ts
ydogandjiev: Good catch, done

src/MicrosoftTeams.ts
ydogandjiev: So I did at one point have logic here that stopped pinging once the childOrigin had been established; however, I found out that this approach doesn't work for VSTS since it initializes the SDK in several different pages that are part of the auth flow. Therefore, we need to keep pinging the popup until the auth flow completes and/or the popup closes.

src/MicrosoftTeams.ts
```diff
@@ -417,25 +443,145 @@ namespace microsoftTeams
...
+
+            // Set up a navigateCrossDomain message handlers that will block cross-domain re-navigation attempts
+            // in the authentication window. We could at some point choose to implement this method via a call to
+            // authenticationWindow.location.href = url; however, we would first need to figure out how to
+            // validate the url against the tab's list of valid domains.
+            handlers["navigateCrossDomain"] = (url: string) =>
+            {
+                return false;
             };
         }
```
dfederm: Can you log a work item for this if there isn't already one?

src/MicrosoftTeams.ts
```diff
@@ -564,23 +746,64 @@ namespace microsoftTeams
...
 
-        // Process only if the message is coming from a valid origin or the origin used for local testing
+        // Process only if the message is coming from a valid origin
+        let messageSource = evt.source || evt.originalEvent.source;
         let messageOrigin = evt.origin || evt.originalEvent.origin;
-        if (validOrigins.indexOf(messageOrigin.toLowerCase()) === -1)
+        if ((validOrigins.indexOf(messageOrigin.toLowerCase()) === -1) &&
+            (messageOrigin !== currentWindow.location.origin) ||
+            (messageSource === currentWindow))
         {
```
dfederm: Should these be grouped for clarity? Right now you have A && B || C but it'd be clearer to have A && (B || C)

Also a nit: You don't need the parens around each clause. You actually have (A) && (B) || (C), so those parens don't do anything.

So all in all, something like...
        if (validOrigins.indexOf(messageOrigin.toLowerCase()) === -1
            && (messageOrigin !== currentWindow.location.origin || messageSource === currentWindow))

src/MicrosoftTeams.ts
```diff
@@ -72,20 +76,32 @@ namespace microsoftTeams
...
-        // If we are in an iframe then our parent window is the one hosting us (i.e. window.parent); otherwise,
-        // it's the window that opened us (i.e. window.opener)
-        parentWindow = (currentWindow.parent !== currentWindow.self) ? currentWindow.parent : currentWindow.opener;
-        if (!parentWindow)
-        {
-            throw new Error("This page must be loaded in an iframe");
-        }
+        currentWindow = this._window as Window || window;
 
         // Listen for messages post to our window (in a way that works for all browsers)
```
ydogandjiev: Deleted that part of the comment

src/MicrosoftTeams.ts
ydogandjiev: Yup, that's what I am doing now. I still don't like the fact that there is a generic set of childWindow handling that lives in the microsoftTeams namespace and a further set of auth-specific logic that also applies to the child window and lives in the microsofTeams.authentication namespace. I will have to think about how we can clean this up in the future.

src/MicrosoftTeams.ts
ydogandjiev: I do close the existing window if there is one; however, I am not certain that we want to call notifyFailure because it might confuse the hell out of developers who reuse the same error handler across authenticate calls.

src/MicrosoftTeams.ts
ydogandjiev: Done

src/MicrosoftTeams.ts
```diff
@@ -417,25 +443,147 @@ namespace microsoftTeams
...
+                left += (currentWindow.outerWidth / 2) - (width / 2);
+                top += (currentWindow.outerHeight / 2) - (height / 2);
+
+                // Open the window with a desired set of standard browser features
+                childWindow = currentWindow.open(link.href, "_blank", "toolbar=no, location=yes, status=no, menubar=no, top=" + top + ", left=" + left + ", width=" + width + ", height=" + height);
+            }
+
+            // Start monitoring the authentication window so that we can detect if it gets closed before the flow completes
+            startAuthenticationWindowMonitor();
+        }
```
dfederm: I think we'll actually want to avoid this if childWindow is null. Else, unless I didn't follow this correctly, you'll start the interval and in the first iteration notice that the window was closed and fail with "CancelledByUser".

src/MicrosoftTeams.ts
```diff
@@ -417,25 +443,147 @@ namespace microsoftTeams
...
+            // - Notifies the caller of failure if it detects that the authentication window is closed
+            // - Keeps pinging the authentication window while its open in order to re-establish
+            //   contact with any pages along the authentication flow that need to communicate
+            //   with us
+            authWindowMonitor = setInterval(() =>
+            {
+                if (!childWindow || childWindow.closed)
+                {
+                    handleFailure("CancelledByUser");
                 }
```
dfederm: Should the interval be stopped at this point? I feel like if you close the window, this interval never stops.

ydogandjiev: The call to handleFailure stops the interval

dfederm:
:shipit:

src/MicrosoftTeams.ts
ydogandjiev: Done

samgavinio:
👍  for this. Any idea when a new version containing the fixes in this PR could possibly be available?

Also, not sure about the state of the master branch but I tried to use it for a bit to verify an unrelated issue and kept receiving the failure notification even if I called notifySuccess. Adding a small delay around `window.close()` as discussed here

https://github.com/OfficeDev/microsoft-teams-library-js/pull/10/files/e6fea2da0119e89e128a29a76ec396ad31cbd35d#r87310519

resolved the issue. Has it been resolved for good without the need for a delay? Experienced the issue with both Chrome 54.0.2840.98 (64-bit) and FE 49.0.2.

ydogandjiev:
Ah, good catch @samgavinio! In cases where the initialize response or parent ping haven't come in before notifySuccess/notifyFailure are called, the authentication window will close itself before it has managed to post the message to its parent. I've sent out a follow-up PR with the fix for this.

dfederm:
@samgavinio to comment about a new version, we'll probably release a new version sometime this week (no promises). The master branch is just the latest unreleased changes and generally may not actually work in cases where we haven't made the client (the MS Teams app) changes required to support new calls yet.

samgavinio:
Thanks guys. We'll sit tight for now until these changes gets released. Thanks!

PR #12, opened 2016-11-14
Send the SDK version with the initialize call

dfederm:
In the short term this is mainly for instrumentation so we can easily see how many tabs are using which version. Longer term we might have breaking changes which we'll need to hand back-compat support for, so this proactively helps with that if we can't just infer the version by the way the call looks.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -12,6 +12,8 @@ namespace microsoftTeams
 {
     "use strict";
 
+    const version = "0.2";
+
     const validOrigins = [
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -12,6 +12,8 @@ namespace microsoftTeams
 {
     "use strict";
 
+    const version = "0.2";
```
ydogandjiev: How are we going to rev this going forward? Also, how are you planning to publish this to v2 of our SDK without pulling in all of my authentication changes?

PR #14, opened 2016-11-15
Allow multiple initialize calls

dfederm:
This is based on feedback from @WrathOfZombies. I would still imagine that most users would want to call initialize onready, but I can imagine scenarios where there is a valid reason to potentially call it more than once. In any case I'd rather err on the side of flexibility for the developer.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -79,7 +79,9 @@ namespace microsoftTeams
     {
         if (initializeCalled)
         {
-            throw new Error("initialize must not be called more than once.");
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -79,7 +79,9 @@ namespace microsoftTeams
     {
         if (initializeCalled)
         {
-            throw new Error("initialize must not be called more than once.");
```
ydogandjiev: Do we still need to throw here if we expect developers to do this? I think we should just get rid of this exception.

PR #17, opened 2016-11-21
Don't close the authentication window until it has finished sending the notifySuccess or notifyFailure message to its parent

ydogandjiev:
In cases where the initialize response or parent ping haven't come in before notifySuccess/notifyFailure are called, the authentication window will close itself before it has managed to post the message to its parent. This is a common scenario since code in the authentication window will typically call notifySuccess/notifyFailure immediately after initialize.

This change fixes the issue by ensuring that the authentication window waits for the parent message queue to be empty (i.e. all messages sent) before closing itself.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -606,7 +606,14 @@ namespace microsoftTeams
 
             sendMessageRequest(parentWindow, "authentication.authenticate.success", [result]);
 
-            window.close();
+            // Wait for the message to be sent before closing the window
+            setInterval(() =>
+            {
+                if (getTargetMessageQueue(parentWindow).length === 0)
+                {
+                    window.close();
+                }
+            }, 100);
         }
 
         /**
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -621,7 +628,14 @@ namespace microsoftTeams
...
-            window.close();
+            // Wait for the message to be sent before closing the window
+            setInterval(() =>
+            {
+                if (getTargetMessageQueue(parentWindow).length === 0)
+                {
+                    window.close();
+                }
+            }, 100);
         }
```
dfederm: Nit: extract to a separate func to reduce duplication

dfederm:
A different (not sure if it's better) approach would be for notifySuccess/notifyFailure to get an ack from the parent and the lib register a callback which does the close.

Although I suppose that requires a client change, and may be overkill.

dfederm:
:shipit:

ydogandjiev:
Waiting for parent window ack is an interesting option but I think I prefer the simplicity of the interval loop. I added unit tests to validate the logic which makes me pretty confident that it will work as expected. Either approach will fail in the case where we fail to establish a connection to the parent window but at that point the user can just manually close the window as a workaround.

PR #18, opened 2016-11-23
Add showDeepLinkDialog

dfederm:


src/MicrosoftTeams.ts
```diff
@@ -203,6 +203,20 @@ namespace microsoftTeams
     }
 
     /**
+     * Displays a dialog to the user with a url they can use to navigate back to a specific state in this page.
+     */
+    export function showDeepLinkDialog(deepLinkDialogParameters: DeepLinkDialogParameters): void
```
ydogandjiev: Now that I see it in the SDK I am wondering if showDeepLinkDialog is a bit too implementation specific of a name. Something like shareEntityReference or shareItemLink might actually be better and would allow us to rev the UX going forward to eventually become the full-blown share experience that we desire (which would still give the user the option to copy a deep link but would also do so much more). I don't feel super strongly about this but just wanted to have us think about it some more.

PR #20, opened 2016-12-01
Add new authentication APIs in order to enable AAD SSO for Teams Apps

ydogandjiev:
This is the initial proposal for the API surface needed to enable AAD single sign on for Teams Apps. I've tried to closely mirror the corresponding APIs in the AuthenticationService of Microsoft Teams and underlying ADAL library:
https://github.com/AzureAD/azure-activedirectory-library-for-js

The AAD user object definition is based on the spec found here:
https://docs.microsoft.com/en-us/azure/active-directory/active-directory-token-and-claims

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -475,6 +475,47 @@ namespace microsoftTeams
             authParams = authenticateParameters;
         }
 
+        /**
+         * Requests an AAD token to be issued on behalf of the app. The token is acquired token from the
+         * cache if it is not expired. Otherwise a request will be sent to AAD to obtain a new token.
+         * @param authTokenRequest A set pf values that configure the token request.
+         */
+        export function getAuthToken(authTokenRequest: AuthTokenRequest): void
+        {
+            ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
+
+            let messageId = sendMessageRequest(parentWindow, "authentication.getAuthToken", [ authTokenRequest.resources ]);
+            callbacks[messageId] = (error: string, token: string) =>
+            {
+                if (error)
+                {
+                    authTokenRequest.failureCallback(error);
+                }
+                else
+                {
+                    authTokenRequest.successCallback(token);
+                }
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.ts
```diff
@@ -475,6 +475,47 @@ namespace microsoftTeams
...
 
+        /**
+         * Requests an AAD token to be issued on behalf of the app. The token is acquired token from the
+         * cache if it is not expired. Otherwise a request will be sent to AAD to obtain a new token.
+         * @param authTokenRequest A set pf values that configure the token request.
+         */
+        export function getAuthToken(authTokenRequest: AuthTokenRequest): void
+        {
+            ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
+
```
dfederm: So this isn't allowed in the auth context because we're assuming they're doing their own auth?

src/MicrosoftTeams.ts
```diff
@@ -475,6 +475,47 @@ namespace microsoftTeams
...
+            };
+        }
+
+        /**
+         * Requests the decoded AAD user identity on behalf of the app.
+         */
+        export function getUser(userRequest: UserRequest): void
+        {
+            ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
+
```
dfederm: Is there a particular reason we wouldn't allow this call in the auth context?

src/MicrosoftTeams.ts
```diff
@@ -475,6 +475,50 @@ namespace microsoftTeams
...
+            let messageId = sendMessageRequest(parentWindow, "authentication.getUser");
+            callbacks[messageId] = (success: boolean, result: UserProfile | string) =>
+            {
+                if (success && typeof(result) === "UserProfile")
```
dfederm: >typeof(result) === "UserProfile" [](start = 31, length = 32)

Pretty sure this doesn't work. JS will just see this as "object"

dfederm:
:shipit:

PR #22, opened 2017-01-04
Fix desktop client auth

ydogandjiev:
This change addresses a couple of issues that were preventing the popup authentication flow in v0.3 of the SDK from working correctly in the Microsoft Teams desktop client. It also adds a unit test case that exercises the end-to-end authentication flow in the desktop client in order to catch issues like these in the future.

I verified the end-to-end authentication flow in the desktop client and all supported browsers (i.e. Chrome, Edge, Firefox, and IE).

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -462,17 +462,10 @@ namespace microsoftTeams
         {
             ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
 
-            // Open an authentication window with the parameters provided by the caller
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -568,6 +561,17 @@ namespace microsoftTeams
...
+                        if (childWindow)
+                        {
+                            startAuthenticationWindowMonitor();
+                        }
+                        else
+                        {
+                            // If we failed to open the window fail the authentication flow
+                            handleFailure("FailedToOpenWindow");
+                        }
                     }
```
dfederm: Not: Besides the args to window.open, this is duped below. Might be worth considering extracting out to a separate function.

PR #23, opened 2017-01-05
Add teamId and channelId fields to the Context interface

ydogandjiev:


src/MicrosoftTeams.ts
```diff
@@ -870,6 +870,16 @@ namespace microsoftTeams
         groupId?: string;
 
         /**
+         * The Microsoft Teams id for the team with which the content is associated.
+         */
+        teamId?: string;
+
```
dfederm: Not sure we really needed this.

PR #24, opened 2017-01-06
Allow developers to provide their own entity id

dfederm:
This change allows developers to provide their own entity id as opposed to just always using the content url as the unique identifier.

This allows developers to use multiple urls which may have minor differences (param ordering, presentation-related query params, etc.) which all should be considered to point to the same underlying entity. This also makes it easier for things like bots to create deep links to existing tabs without knowing the exact content url for example.

As a reminder, we had originally decided that we wouldn't need an entity id since we could generate it from the content url, and if we had the content url we wouldn't need an entity id, but this makes assumptions about a url exactly and uniquely identifying an entity, which as described above isn't necessarily a good assumption.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -210,7 +210,8 @@ namespace microsoftTeams
         ensureInitialized(frameContexts.content);
 
         sendMessageRequest(parentWindow, "shareDeepLink", [
-            deepLinkParameters.deepLinkContext,
+            deepLinkParameters.entityId,
+            deepLinkParameters.subEntityId,
             deepLinkParameters.label,
             deepLinkParameters.webUrl,
         ]);
@@ -330,18 +331,9 @@ namespace microsoftTeams
             websiteUrl?: string;
 
             /**
-             * The custom settings for this content instance.
-             * The developer may use this for generic storage specific to this instance,
-             * for example a JSON blob describing the previously selected options used to pre-populate the UI.
-             * The string must be less than 1kb.
-             */
-            customSettings?: string;
-
-            /**
              * The developer-defined unique id for the entity this content points to.
-             * If not provided, one will be generated from the contentUrl.
              */
-            entityId?: string;
+            entityId: string;
         }
 
         export interface SaveEvent
@@ -878,13 +870,18 @@ namespace microsoftTeams
         /**
          * The Microsoft Teams id for the channel with which the content is associated.
          */
-        channelId?: string;
+        channelId: string;
 
         /**
          * The developer-defined unique id for the entity this content points to.
-         * If none was provided, this value is generated from the contentUrl.
          */
         entityId: string;
+
+        /**
+         * The developer-defined unique id for the sub-entity this content points to.
+         * This field should be used to restore to a specific state within an entity, for example scrolling to or activating a specific piece of content.
+         */
+        subEntityId?: string;
 
         /**
          * The current locale that the user has configured for the app formatted as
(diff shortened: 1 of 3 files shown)


src/MicrosoftTeams.ts
```diff
@@ -875,6 +881,12 @@ namespace microsoftTeams
         channelId?: string;
 
         /**
+         * The developer-defined unique id for the entity this content points to.
+         * If none was provided, this value is generated from the contentUrl.
+         */
+        entityId: string;
```
ShreeBalasubramanian: why not optional here ?

dfederm: We'll always provide it. It may just be a hash of the content url though at this point, but it'll always be given.

src/MicrosoftTeams.ts
```diff
@@ -872,7 +870,18 @@ namespace microsoftTeams
         /**
          * The Microsoft Teams id for the channel with which the content is associated.
          */
-        channelId?: string;
+        channelId: string;
```
dfederm: Based on discussion with Yuri, this will always be provided by the client.

dfederm:
Current summary of the changes:
- Replaced customSettings with entityId and expose it as part of the Context (as opposed to only the settings frame)
- Replaced deepLinkContext with subEntityId
- Made label required for deep links

src/MicrosoftTeams.ts
```diff
@@ -330,12 +331,9 @@ namespace microsoftTeams
             websiteUrl?: string;
 
             /**
```
ydogandjiev: Wait, are we completely getting rid of customSettings?

test/MicrosoftTeams.spec.ts
```diff
@@ -444,6 +448,7 @@ describe("MicrosoftTeams", () =>
             contentUrl: "someContentUrl",
             websiteUrl: "someWebsiteUrl",
             customSettings: "someCustomSettings",
```
ydogandjiev: If customSettings are going away then you should delete this.

src/MicrosoftTeams.ts
dfederm: Yup, it's being replaced with entity id. I know they're not entirely the same thing, but they serve the same basic intent. Entity Id is actually slightly more powerful since it'll be exposed to the content frame and not just the settings frame.

src/MicrosoftTeams.ts
```diff
@@ -210,7 +210,8 @@ namespace microsoftTeams
         ensureInitialized(frameContexts.content);
 
         sendMessageRequest(parentWindow, "shareDeepLink", [
-            deepLinkParameters.deepLinkContext,
+            deepLinkParameters.entityId,
+            deepLinkParameters.subEntityId,
             deepLinkParameters.label,
```
dfederm: This is a breaking change but we haven't released this yet.

src/MicrosoftTeams.ts
```diff
@@ -210,7 +210,8 @@ namespace microsoftTeams
         ensureInitialized(frameContexts.content);
 
         sendMessageRequest(parentWindow, "shareDeepLink", [
-            deepLinkParameters.deepLinkContext,
+            deepLinkParameters.entityId,
+            deepLinkParameters.subEntityId,
```
dfederm: Hmm, we actually don't need this since it's this tab's entity id.

src/MicrosoftTeams.ts
richardtaylorrt: Do we still document the 1kb restriction somewhere?

src/MicrosoftTeams.ts
```diff
@@ -900,25 +908,20 @@ namespace microsoftTeams
...
+         * so this field may be used to restore to a specific state within an entity, for example scrolling to or activating a specific piece of content.
          */
-        deepLinkContext?: string;
+        subEntityId?: string;
 
         /**
          * The label which should be displayed for this deep link when the link is rendered in a client.
          */
-        label?: string;
+        label: string;
```
richardtaylorrt: I wonder if this should be subEntityLabel and subEntityWebUrl?  Or if the interface should be SubEntityDeepLinkParameters?  To avoid confusion elsewhere with the main (pinned) entity's label and webUrl.

richardtaylorrt: ...and does the resulting deeplink contain both the labels as it contains both the IDs?  Might need the entity label itself if user decides to pin it as a tab.

dfederm: I don't think we need those. I think we're expecting the sub-entity to be renderable using the parent entity url + subEntityId provided via the SDK. Else we're going to start hitting the url length limit very soon with 2 full urls.

dfederm: The label is for the deeplink, not the entity right?

PR #28, opened 2017-01-19
Schema changes to rationalize gallery tabs, static tabs, and scopes

akinolugbade:
Refactored to make gallery tabs, static tabs and bots as top level app properties, each of which also have a defined scope.

### DIFF
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -182,7 +182,6 @@
                                 "personal"
                             ],
                         "description": "Specifies whether the bot offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
-
                         }
                     }
                 },
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -106,18 +106,13 @@
                     "scope": {
                         "type": "array",
                         "items": {
-                            "teams": {
-                                "type": "string",
-                                "description": "A value of 'teams' indicates the tab offers an experience in the context of a channel in a team"
-                            },
-                            "people": {
-                                "type": "string",
-                                "description": "A value of 'people' indicates the tab offers an experience in the context of a chat between one or more people"
-                            },
-                            "personal": {
-                                "type": "string",
-                                "description": "A value of 'personal' indicates the tab offers an experience scoped to an individual user alone"
-                            }
+                            "enum": [
+                                "team",
+                                "people",
+                                "personal"
+                            ],
+                        "description": "Specifies whether the tab offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
+
                         }
                     }
                 },
@@ -134,44 +129,37 @@
                 "properties": {
                     "entityId": {
                         "type": "string",
-                        "description": "A unique identifier for the entity which the tab displays. The id must use reverse domain name notation.",
-                        "maxLength": 64
+                        "description": "A unique identifier for the entity which the tab displays."
                     },
                     "name": {
                         "type": "string",
-                        "description": "The display name of the tab.",
-                        "maxLength": 16
+                        "description": "The display name of the tab."
                     },
-                    "canvasUrl": {
+                    "contentUrl": {
                         "type": "string",
                         "description": "The url which points to the entity UI to be displayed in the Teams canvas."
                     },
                     "websiteUrl": {
                         "type": "string",
-                        "description": "The url to point at if a user opts to view in a browser. If this is not supplied the browswer will be pointed at the canvasUrl"
+                        "description": "The url to point at if a user opts to view in a browser."
                     },
                     "scope": {
                         "type": "array",
                         "items": {
-                            "teams": {
-                                "type": "string",
-                                "description": "A value of 'teams' indicates the tab offers an experience in the context of a channel in a team"
-                            },
-                            "people": {
-                                "type": "string",
-                                "description": "A value of 'people' indicates the tab offers an experience in the context of a chat between one or more people"
-                            },
-                            "personal": {
-                                "type": "string",
-                                "description": "A value of 'personal' indicates the tab offers an experience scoped to an individual user alone"
-                            }
+                            "enum": [
+                                "team",
+                                "people",
+                                "personal"
+                            ],
+                        "description": "Specifies whether the tab offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
+
                         }
                     }
                 },
                 "required": [
                     "entityId",
                     "name",
-                    "canvasUrl",
+                    "contentUrl",
                     "scope"
                 ]
             }
@@ -189,18 +177,13 @@
                     "scope": {
                         "type": "array",
                         "items": {
-                            "teams": {
-                                "type": "string",
-                                "description": "A value of 'teams' indicates the bot offers an experience in the context of a channel in a team"
-                            },
-                            "people": {
-                                "type": "string",
-                                "description": "A value of 'people' indicates the bot offers an experience in the context of a chat between one or more people"
-                            },
-                            "personal": {
-                                "type": "string",
-                                "description": "A value of 'personal' indicates the bot offers an experience scoped to an individual user alone"
-                            }
+                            "enum": [
+                                "team",
+                                "people",
+                                "personal"
+                            ],
+                        "description": "Specifies whether the bot offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
+
                         }
                     }
                 },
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
                     },
-                    "description": {
-                        "type": "object",
-                        "properties": {
-                            "short": {
+                    "scope": {
+                        "type": "array",
+                        "items": {
+                            "teams": {
                                 "type": "string",
```
dfederm: I don't think this is how an array of pre-defined strings is specified in a schema.

I think items should be like:
"items": {
  "enum": [ "teams", "people", "personal" ]
}

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
+            }
+        },
+        "staticTabs": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "entityId": {
+                        "type": "string",
+                        "description": "A unique identifier for the entity which the tab displays. The id must use reverse domain name notation.",
```
dfederm: >The id must use reverse domain name notation. [](start = 99, length = 45)

This isn't true for entity ids, it's up to the developer.

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
+                        "maxLength": 64
```
dfederm: >"maxLength": 64 [](start = 24, length = 15)

We don't enforce this when configuring a tab. Should we?

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
                     },
-                    "accentColor": {
+                    "name": {
                         "type": "string",
-                        "description": "A color to use in conjunction with the extension's icons. The value must be a valid HTML color code starting with '#', for example `#4464ee`."
+                        "description": "The display name of the tab.",
+                        "maxLength": 16
```
dfederm: >"maxLength": 16 [](start = 24, length = 15)

We don't enforce this when configuring a tab. Should we?

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
                     },
-                    "configUrl": {
+                    "canvasUrl": {
```
dfederm: >canvasUrl [](start = 21, length = 9)

We actually call this "contentUrl" in the SDK when configuring a tab (see: [https://github.com/OfficeDev/microsoft-teams-library-js/blob/master/src/MicrosoftTeams.ts#L320](https://github.com/OfficeDev/microsoft-teams-library-js/blob/master/src/MicrosoftTeams.ts#L320)). Should we use a consistent term here?

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
-                        "description": "The url to use when configuring the extension."
+                        "description": "The url which points to the entity UI to be displayed in the Teams canvas."
                     },
-                    "canUpdateConfig": {
-                        "type": "boolean",
-                        "description": "A value indicating whether an instance of the extension's config can be updated by the user after creation.",
-                        "default": true
+                    "websiteUrl": {
+                        "type": "string",
+                        "description": "The url to point at if a user opts to view in a browser. If this is not supplied the browswer will be pointed at the canvasUrl"
```
dfederm: >If this is not supplied the browswer will be pointed at the canvasUrl [](start = 97, length = 69)

This isn't the case. We just don't show the "go to website" button if this isn't provided.

Note that this is the same as the websiteUrl in the tab settings when configuring a tab. See: [https://github.com/OfficeDev/microsoft-teams-library-js/blob/master/src/MicrosoftTeams.ts#L330](https://github.com/OfficeDev/microsoft-teams-library-js/blob/master/src/MicrosoftTeams.ts#L330)

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,136 @@
...
+                    },
+                    "scope": {
+                        "type": "array",
+                        "items": {
+                            "teams": {
```
dfederm: Same comment regarding how an enum is specified.

src/MicrosoftTeams.schema.json
```diff
@@ -129,56 +186,37 @@
...
-                                "url": {
-                                    "type": "string",
-                                    "description": "The content url for this tab instance"
-                                },
-                                "websiteUrl": {
-                                    "type": "string",
-                                    "description": "The website url of the content displayed in this tab instance"
-                                }
+                            "teams": {
+                                "type": "string",
```
dfederm: Same enum comment

akinolugbade:
made changes incorporating David's feedback

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,124 @@
...
+                        "type": "array",
+                        "items": {
+                            "enum": [
+                                "team",
+                                "people",
+                                "personal"
+                            ],
+                        "description": "Specifies whether the tab offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
+
+                        }
```
dfederm: Nit: remove extra newline

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,124 @@
...
```
dfederm: Nit: remove extra newline

src/MicrosoftTeams.schema.json
```diff
@@ -129,56 +174,32 @@
...
-                                "url"
-                            ]
+                            "enum": [
+                                "team",
+                                "people",
+                                "personal"
+                            ],
+                        "description": "Specifies whether the bot offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
+
                         }
```
dfederm: Nit: remove extra newline

dfederm:
Looks good, but I think we should get @ydogandjiev's approval as well since he's put more thought into the implementation details.

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,123 @@
...
+            "type": "string",
+            "description": "A color to use in conjunction with the app's icons. The value must be a valid HTML color code starting with '#', for example `#4464ee`."
+        },
+        "galleryTabs": {
             "type": "array",
             "items": {
                 "type": "object",
                 "properties": {
-                    "id": {
+                    "configUrl": {
```
ydogandjiev: I suggest we call this configurationUrl since I'm not a fan of using abbreviations in public-facing schemas/APIs. What do you think?

src/MicrosoftTeams.schema.json
```diff
@@ -43,79 +43,123 @@
...
                         "type": "string",
-                        "description": "A unique identifier for this extension. The id must use reverse domain name notation.",
-                        "maxLength": 64
+                        "description": "The url to use when configuring the tab."
                     },
-                    "name": {
-                        "type": "string",
-                        "description": "The display name of the extension.",
-                        "maxLength": 16
+                    "canUpdateConfig": {
```
ydogandjiev: Same as above; I would rather we call this canUpdateConfiguration.

src/MicrosoftTeams.schema.json
```diff
@@ -129,56 +173,31 @@
...
                     }
                 },
                 "required": [
-                    "mri"
+                    "mri",
+                    "scope"
                 ]
             }
         },
         "needsIdentity": {
```
ydogandjiev: The "needsIdentity" field needs to be replaced with an array of enum values that represent the various capabilities/permissions that apps will need. I suggest we rename this to "capabilities" or "permissions" and change it to an array. The possible values in this array for now will just be "identity" or nothing; however, we will add more options once we fully define our capabilities model. I suggest you work with Larry and Richard to figure out what all of the possible capabilities will need to be.

src/MicrosoftTeams.schema.json
```diff
@@ -129,56 +173,31 @@
...
+                                "people",
+                                "personal"
+                            ],
+                        "description": "Specifies whether the bot offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
                         }
                     }
                 },
                 "required": [
-                    "mri"
+                    "mri",
```
ydogandjiev: Sujeet suggested that we move away from using "mri" in the schema since that's an internal acronym we use for any contact identifier. We should just call this "id" or "botId".

src/MicrosoftTeams.schema.json
akinolugbade: fixed

src/MicrosoftTeams.schema.json
akinolugbade: fixed

src/MicrosoftTeams.schema.json
akinolugbade: fixed. went with botId

src/MicrosoftTeams.schema.json
akinolugbade: I called this permissions and so far only have identities. We need to close on the set of permissions this week.

PR #29, opened 2017-02-01
bot commands, app mentions, and incoming webhooks

akinolugbade:
This should be close to final except for adding permissions when they come. Also made some style changes  and added some missing descriptions. Please validate those as well.

### DIFF
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -172,9 +172,9 @@
                         "description": "A unique identifier for the bot that matches its ID in the bot framework.",
                         "maxLength": 64
                     },
-                    "hasIncomingWebhook": {
+                    "needsChannelSelctor": {
                         "type": "boolean",
-                        "description": "This value describes whether or not the bot utilizes an incoming webhook",
+                        "description": "This value describes whether or not the bot utilizes a user hint to add the bot to a specific channel.",
                         "default": "false"
                     },
                     "scope": {
@@ -256,16 +256,17 @@
                 "required": [
                     "botId",
                     "scope",
-                    "hasIncomingWebhook"
+                    "needsChannelSelctor"
                 ]
             }
         },
-        "capabilities": {
+        "permissions": {
             "type": "array",
-            "description": "Specifies which capabilties the bots or tabs request",
+            "description": "Specifies which permissions the bots or tabs request.",
             "items": {
                 "enum": [
-                    "identity"
+                    "identity",
+                    "sendToTeamMembers"
                 ]
             }
         },
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -182,7 +182,6 @@
                                 "personal"
                             ],
                         "description": "Specifies whether the bot offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
-
                         }
                     }
                 },
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -12,12 +12,12 @@
         },
         "id": {
             "type": "string",
-            "description": "A human readable unique identifier for this set of tabs and bots. This id should be in reverse domain notation.",
-            "maxLength": 64
-        },
-        "guid": {
-            "type": "string",
             "description": "A unique identifier for this set of tabs and bots. This id must be a guid.",
+            "maxLength": 36
+        },
+        "reverseDomainId": {
+            "type": "string",
+            "description": "A unique identifier for this set of tabs and bots in reverse domain notation.",
             "maxLength": 64
         },
         "developer": {
@@ -288,7 +288,7 @@
         "manifestVersion",
         "version",
         "id",
-        "guid",
+        "reverseDomainId",
         "developer",
         "icons",
         "accentColor"
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -172,7 +172,7 @@
                         "description": "A unique identifier for the bot that matches its ID in the bot framework.",
                         "maxLength": 64
                     },
-                    "needsChannelSelctor": {
+                    "needsChannelSelector": {
                         "type": "boolean",
                         "description": "This value describes whether or not the bot utilizes a user hint to add the bot to a specific channel.",
                         "default": "false"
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.schema.json
```diff
@@ -173,36 +172,107 @@
                         "description": "A unique identifier for the bot that matches its ID in the bot framework.",
                         "maxLength": 64
                     },
+                    "needsChannelSelctor": {
```
ydogandjiev: TYPO: "needsChannelSelctor" --> "needsChannelSelector"

src/MicrosoftTeams.schema.json
```diff
@@ -173,36 +172,107 @@
...
             }
         },
         "permissions": {
             "type": "array",
+            "description": "Specifies which permissions the bots or tabs request.",
             "items": {
                 "enum": [
-                    "identity"
+                    "identity",
+                    "sendToTeamMembers"
```
ydogandjiev: NIT: I think "messageTeamMembers" makes a little more sense than "sendToTeamMembers"

src/MicrosoftTeams.schema.json
```diff
@@ -173,36 +172,107 @@
...
                             "enum": [
                                 "team",
-                                "people",
                                 "personal"
-                            ],
-                        "description": "Specifies whether the bot offers an experience in the context of a channel in a team, a chat between one or more people, or an experience scoped to an individual user alone. These options are non-exclusive"
+                            ]
+                        }
+                    },
+                    "commandList": {
```
ydogandjiev: It would still be good to decide now whether we will actually be able to make bot input menus work end-to-end for GA because if the answer is no then we should hold off on adding support for them to our public manifest. Otherwise, we will get flack from developers when they try to specify this for their bots but it doesn't work.

src/MicrosoftTeams.schema.json
```diff
@@ -173,36 +172,107 @@
...
+                        "type": "array",
```
ydogandjiev: We should add a maximum here; since we only have 2 scopes then at most we should support 2 command lists.

src/MicrosoftTeams.schema.json
```diff
@@ -173,36 +172,107 @@
...
+                                    "description": "Specifies the scopes for which the command list is valid",
+                                    "items": {
+                                        "enum": [
+                                            "team",
+                                            "personal"
+                                        ]
+                                    }
+                                },
+                                "commands": {
+                                    "type": "array",
```
ydogandjiev: We should add a maximum here; what's the largest number of commands that we can realistically render for a bot?

src/MicrosoftTeams.schema.json
```diff
@@ -173,36 +172,107 @@
...
+                                "commands"
+                            ]
                         }
                     }
                 },
                 "required": [
                     "botId",
-                    "scope"
+                    "scope",
+                    "needsChannelSelctor"
```
ydogandjiev: The "needsChannelSelctor" property shouldn't be required; if not specified we will just assume it's false.

src/MicrosoftTeams.schema.json
akinolugbade: Fixed.

src/MicrosoftTeams.schema.json
akinolugbade: I think it should be 3. A set that is just personal, a set just teams, and a set for both. Alternatively, we could ask developers to duplicate commands that exist in both scopes. 

Let me know if you disagree, and I'll change the maxitems to 2. For now, making it 3.

src/MicrosoftTeams.schema.json
akinolugbade: I'd defer to @larryjin or @richardtaylorrt on that max as it depends on the ui here.

src/MicrosoftTeams.schema.json
akinolugbade: fixed.

src/MicrosoftTeams.schema.json
akinolugbade: agreed and fixed.

src/MicrosoftTeams.schema.json
akinolugbade: @larryjin @richardtaylorrt

src/MicrosoftTeams.schema.json
ydogandjiev: Since we went with distinct commandLists per scope I don't think we should be trying to merge commands across different menus. In other words, you can have no command list, a command list per scope, or a unified command list that works in all scopes. In any case, the maximum count here isn't a big deal; what's more important is that we document what we expect here from developers.

src/MicrosoftTeams.schema.json
ydogandjiev: Once you close on this with Larry and Richard feel free to commit your changes. I am signing off.

src/MicrosoftTeams.schema.json
```diff
@@ -87,10 +87,11 @@
         },
         "accentColor": {
             "type": "string",
-            "description": "A color to use in conjunction with the app's icons. The value must be a valid HTML color code starting with '#', for example `#4464ee`."
+            "description": "A color to use in conjunction with the icon. The value must be a valid HTML color code starting with '#', for example `#4464ee`."
         },
         "galleryTabs": {
             "type": "array",
+            "description": "The set of tabs that users can optionally add to Team's as a presentation canvas.",
```
richardtaylorrt: Teams not Team's.   Also, "presentation canvas" is internal terminology.  Happy to brainstorm how to present this.

src/MicrosoftTeams.schema.json
```diff
@@ -173,15 +172,85 @@
                         "description": "A unique identifier for the bot that matches its ID in the bot framework.",
                         "maxLength": 64
                     },
+                    "needsChannelSelector": {
+                        "type": "boolean",
+                        "description": "This value describes whether or not the bot utilizes a user hint to add the bot to a specific channel.",
```
richardtaylorrt: Only valid when the scope of the bot includes 'team'.

src/MicrosoftTeams.schema.json
richardtaylorrt: @larryjin  to decide all questions to do with bot input menus.

src/MicrosoftTeams.schema.json
```diff
@@ -193,16 +262,17 @@
         },
         "permissions": {
             "type": "array",
+            "description": "Specifies which permissions the bots or tabs request.",
             "items": {
                 "enum": [
-                    "identity"
+                    "identity",
+                    "messageTeamMembers"
```
richardtaylorrt: Let's document what each of these permissions mean.

src/MicrosoftTeams.schema.json
ydogandjiev: I don't think this is something that we can express in the schema but we could add it to the property description as documentation for developers.

PR #36, opened 2017-03-22
Add schema constraints to more closely match what we validate in Teams

dfederm:


src/MicrosoftTeams.schema.json
```diff
@@ -92,17 +98,20 @@
         },
         "accentColor": {
             "type": "string",
-            "description": "A color to use in conjunction with the icon. The value must be a valid HTML color code starting with '#', for example `#4464ee`."
+            "description": "A color to use in conjunction with the icon. The value must be a valid HTML color code starting with '#', for example `#4464ee`.",
+            "pattern": "^#[0-9a-fA-F]{6}$"
```
ydogandjiev: Nice, we should also update this in the settings store which today only enforces a max length of 7 for the accentColor.

src/MicrosoftTeams.schema.json
```diff
@@ -92,17 +98,20 @@
...
             "description": "The set of tabs that users can optionally add to Teams",
+            "maxItems": 1,
             "items": {
                 "type": "object",
                 "properties": {
                     "configurationUrl": {
                         "type": "string",
-                        "description": "The url to use when configuring the tab."
+                        "description": "The url to use when configuring the tab.",
+                        "maxLength": 2048
```
ydogandjiev: I wonder if we can do some basic URL validation using regex here. Nothing as complex at what we use in the FE client but at least a check that the URL starts with https: would be great.

src/MicrosoftTeams.schema.json
```diff
@@ -92,17 +98,20 @@
...
-            "description": "A color to use in conjunction with the icon. The value must be a valid HTML color code starting with '#', for example `#4464ee`."
+            "description": "A color to use in conjunction with the icon. The value must be a valid HTML color code starting with '#', for example `#4464ee`.",
+            "pattern": "^#[0-9a-fA-F]{6}$"
         },
         "galleryTabs": {
             "type": "array",
             "description": "The set of tabs that users can optionally add to Teams",
+            "maxItems": 1,
             "items": {
                 "type": "object",
```
ydogandjiev: We should probably also add "additionalProperties": false for all objects in the schema to discourage developers from adding useless chunks to their manifest. Settings store does for all of its schemas.

ydogandjiev:
You should also add a maxLength of 16 to the manifestVersion field.

src/MicrosoftTeams.schema.json
```diff
@@ -8,7 +8,8 @@
         },
         "version": {
             "type": "string",
-            "description": "The version of the manifest you are submitting. Changes to your manifest should cause a version change. This version string must follow the semver standard."
+            "description": "The version of the manifest you are submitting. Changes to your manifest should cause a version change. This version string must follow the semver standard.",
+            "maxLength": 256
         },
         "id": {
```
ydogandjiev: Could we validate that the ID is a GUID via regex?

src/MicrosoftTeams.schema.json
```diff
@@ -279,8 +298,10 @@
...
             "type": "array",
             "description": "A list of valid domains from which the tabs expect to load any content. Domain listings can include wildcards, for example `*.example.com`. If your tab configuration or content UI needs to navigate to any other domain besides the one use for tab configuration, that domain must be specified here.",
+            "maxItems": 16,
             "items": {
-                "type": "string"
+                "type": "string",
+                "maxLength": 2048
             }
         }
     },
```
ydogandjiev: Please add "name" and "description" to the list of required attributes below. I think we accidentally got rid of it when we went from v0.3 --> v0.4 --> v0.5.

PR #37, opened 2017-03-24
Update bot input menu defintion

XinGao:
Update the bot input menu definition, remove unnecessary fields.

### DIFF
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -220,7 +220,7 @@
                             ]
                         }
                     },
-                    "commandList": {
+                    "commandLists": {
                         "type": "array",
                         "maxItems": 2,
                         "description": "The list of commands that the bot supplies, including their usage, description, and the scope for which the commands are valid. A seperate command list should be used for each scope.",
@@ -231,6 +231,7 @@
                                 "scopes": {
                                     "type": "array",
                                     "description": "Specifies the scopes for which the command list is valid",
+                                    "maxItems": 2,
                                     "items": {
                                         "enum": [
                                             "team",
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.schema.json
```diff
@@ -247,38 +248,18 @@
...
-                                            "usageHint": {
-                                                "type": "string",
-                                                "description": "A short usage template for the command",
-                                                "maxLength": 32
-                                            },
-                                            "shortDescription": {
-                                                "type": "string",
-                                                "description": "A short description of the command",
+                                                "description": "The bot command name",
                                                 "maxLength": 32
```
ydogandjiev: Are you sure that 32 characters is right for the maxLength here? Will the input menu in our UI actually be able to fit a command name this long?

src/MicrosoftTeams.schema.json
```diff
@@ -247,38 +248,18 @@
...
-                                                "enum" : ["imBack", "invoke", "openURL", "prompt", "gotoTab"],
-                                                "description": "The type of bot interaction the command defines.",
-                                                "maxLength": 16
-                                            },
-                                            "image": {
+                                            "description": {
                                                 "type": "string",
-                                                "description": "An optional URL to an image to use for the command entry"
+                                                "description": "A simple text description or an example of the command syntax and its arguments.",
+                                                "maxLength": 128
```
ydogandjiev: Same comment as above applies here? Where will this string be displayed in our UI and will it look good even at the max length?

src/MicrosoftTeams.schema.json
XinGao: Larry and I discussed about the max length of the command name and description and we came out these numbers. Haven't got UX redline yet. Let's keep them for the moment.

PR #39, opened 2017-04-11
input extensions, app concept, icons, tab names, bot type, style

akinolugbade:


src/MicrosoftTeams.schema.json
```diff
@@ -5,22 +5,22 @@
...
+            "description": "The version of your app you are submitting. Changes to your manifest should cause a version change. This version string must follow the semver standard.",
             "maxLength": 256
         },
         "id": {
             "type": "string",
             "description": "A unique identifier for this set of tabs and bots. This id must be a guid.",
             "pattern": "^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$"
         },
-        "reverseDomainId": {
+        "packageName": {
```
dfederm: We don't use or plan to use this field at all. Do we need it?

src/MicrosoftTeams.schema.json
```diff
@@ -209,6 +223,14 @@
                         "description": "This value describes whether or not the bot utilizes a user hint to add the bot to a specific channel.",
                         "default": false
                     },
+                    "botType": {
```
dfederm: Gunjan added just a boolean flag for this. Can y'all sync on the right thing to do here?

src/MicrosoftTeams.schema.json
```diff
@@ -57,22 +57,36 @@
...
+                    "maxLength": 50
+                },
+                "full": {
+                    "type": "string",
+                    "description": "The full name of the app, used if the full app name exceeds 50 characters",
+                    "maxLength": 100
+                }
+            },
+            "required": [
+                "short"
```
dfederm: Why is only the short one required?

src/MicrosoftTeams.schema.json
akinolugbade: We need to add this into our telemetry pipeline. @ydogandjiev @richardtaylorrt  please correct me if i'm wrong.

src/MicrosoftTeams.schema.json
akinolugbade: Many app names will fully fit within the 50 character limit. Seems weird to require it in that case.

src/MicrosoftTeams.schema.json
ydogandjiev: Let's just leave this in here for now. We can choose to consume it later.

PR #40, opened 2017-04-21
URL validation, edits

billbliss:
Added URL validation for URLs that didn't have them, grammar tweaks, wording clarifications.

### DIFF
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -36,19 +36,19 @@
                     "type": "string",
                     "description": "The url to the page that provides support information for the app.",
                     "maxLength": 2048,
-                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]*://"
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://"
                 },
                 "privacyUrl": {
                     "type": "string",
                     "description": "The url to the page that provides privacy information for the app.",
                     "maxLength": 2048,
-                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]://"
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://"
                 },
                 "termsOfUseUrl": {
                     "type": "string",
                     "description": "The url to the page that provides the terms of use for the app.",
                     "maxLength": 2048,
-                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]://"
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://"
                 }
             },
             "required": [
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -10,12 +10,12 @@
         },
         "version": {
             "type": "string",
-            "description": "The version of your app you are submitting. Changes to your manifest should cause a version change. This version string must follow the semver standard.",
+            "description": "The version of the app. Changes to your manifest should cause a version change. This version string must follow the semver standard (http://semver.org).",
             "maxLength": 256
         },
         "id": {
             "type": "string",
-            "description": "A unique identifier for this set of tabs and bots. This id must be a guid.",
+            "description": "A unique identifier for this app. This id must be a GUID.",
             "pattern": "^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$"
         },
         "packageName": {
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.schema.json
```diff
@@ -35,17 +35,20 @@
                 "websiteUrl": {
                     "type": "string",
                     "description": "The url to the page that provides support information for the app.",
-                    "maxLength": 2048
+                    "maxLength": 2048,
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]://"
```
ydogandjiev: We don't ever load any of these URLs in our iframes; are you sure that they need to be https?

src/MicrosoftTeams.schema.json
```diff
@@ -157,7 +160,7 @@
         },
         "staticTabs": {
             "type": "array",
-            "description": "A set of tabs that may be 'pinned' by default, without the user adding them manually.  Static tabs declared in personal scope are always pinned to the app's personal experience.  Static tabs declared in the team scope are candidates to be pinned automatically to channels, subject to approval.",
+            "description": "A set of tabs that may be 'pinned' by default, without the user adding them manually.  Static tabs declared in personal scope are always pinned to the app's personal experience.  Static tabs declared in the team scope are candidates to be pinned automatically to channels, subject to a user's approval.",
```
akinolugbade: @richardtaylorrt  should comment on this but i think the approving entity is Microsoft, and in the future a tenant admin, not a user.

richardtaylorrt: @akinolugbade @billbliss - yes, today it is Microsoft (and we'll only approve 1P apps like Assignments and Wiki etc).  In future, it might be tenant admin or even team owner/member.  After you install Evernote, we ask "do you want Evernote tab in each channel in this team?"

src/MicrosoftTeams.schema.json
billbliss: Good point, I'll change that to make the [Ss] optional.

src/MicrosoftTeams.schema.json
billbliss: OK, makes sense, but what should we say instead of "a user's approval"? Simply saying "approval" isn't terribly helpful: whose approval?

src/MicrosoftTeams.schema.json
```diff
@@ -35,17 +35,20 @@
                 "websiteUrl": {
                     "type": "string",
                     "description": "The url to the page that provides support information for the app.",
-                    "maxLength": 2048
+                    "maxLength": 2048,
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]*://"
```
ydogandjiev: I think "?" is more appropriate than "*" since you want 0 or 1 rather than 0 to N instances of the letter "s" or "S" in there, i.e.:
"pattern": "^[Hh][Tt][Tt][Pp][Ss]?://"

src/MicrosoftTeams.schema.json
```diff
@@ -35,17 +35,20 @@
...
                 },
                 "privacyUrl": {
                     "type": "string",
                     "description": "The url to the page that provides privacy information for the app.",
-                    "maxLength": 2048
+                    "maxLength": 2048,
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]://"
```
ydogandjiev: Use "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://" since this does't have to be an https URL

src/MicrosoftTeams.schema.json
```diff
@@ -35,17 +35,20 @@
...
                 },
                 "termsOfUseUrl": {
                     "type": "string",
                     "description": "The url to the page that provides the terms of use for the app.",
-                    "maxLength": 2048
+                    "maxLength": 2048,
+                    "pattern": "^[Hh][Tt][Tt][Pp][Ss]://"
```
ydogandjiev: Use "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://" since this does't have to be an https URL

src/MicrosoftTeams.schema.json
```diff
@@ -182,7 +185,8 @@
                     "websiteUrl": {
                         "type": "string",
                         "description": "The url to point at if a user opts to view in a browser.",
-                        "maxLength": 2048
+                        "maxLength": 2048,
+                        "pattern": "^[Hh][Tt][Tt][Pp][Ss]://"
```
ydogandjiev: Use "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://" since this does't have to be an https URL

ydogandjiev:
@dfederm, do you have any idea why the CI build is failing? Looks like a TS compilation issue despite the fact that Bill's change doesn't touch any of our code. Did Travis upgrade compilers or something?

dfederm:
Looking. It seems like the @types/node is being installed automatically by npm and that's causing setInterval to de defined as returning a NodeJS.Timer instead of a number. I'll keep digging.

---
In reply to: [297270011](https://github.com/OfficeDev/microsoft-teams-library-js/pull/40#issuecomment-297270011) [](ancestors = 297270011)

src/MicrosoftTeams.schema.json
RichMoeMSFT: @akinolugbade @richardtaylorrt Not sure how Microsoft approval works.  Does this mean you whitelist behind the scenes for now?

src/MicrosoftTeams.schema.json
billbliss: Ah yes I misread the regex rules.

PR #43, opened 2017-05-03
Initial version of msteams-app.css

billbliss:


src/msteams-app.css
```diff
@@ -0,0 +1,163 @@
...
+.theme-contrast .ts-btn-primary:focus,
+.theme-contrast .ts-btn-primary:hover {
+  background: #ffff01;
+  color: #000;
+  cursor: pointer;
+}
+.theme-contrast .section-caption {
+  color: #fff;
+}
+
```
ydogandjiev: How are we generating all of this CSS? How are we going to maintain it in the future and make sure that we don't break tabs that have taken dependencies on it?

src/msteams-app.css
```diff
@@ -0,0 +1,163 @@
...
+.form-field-title {
+  display: flex;
+  margin: 1rem 0 .3rem;
+}
+.form-control:focus {
+  border-bottom-color: #5558af;
+  box-shadow: none;
+  outline: 0;
+}
+.form-control {
```
ydogandjiev: Providing styles with generic class names like "form-control" is going to make it really difficult for developers to pull these into their pages without risking collisions with their existing CSS. Consider using more unique class names throughout.

src/msteams-app.css
ydogandjiev: Also, how are you planning to distribute it to our partners? Do you expect this css file to go on our CDN?

billbliss: 1. It was generated manually based on what we did for the Trello tab.
2. It should be versioned just like the SDK. 
3. Yes it would be on our CDN just like our SDK which also answers the versioning question because all the CDN paths are versioned.

src/msteams-app.css
billbliss: This is intended for new code, our config experience, so there shouldn't be conflicting styles. I will try hierarchical styles for the default theme (which is what causes the problem because there's no class for it).

billbliss: Actually I was halfway through doing something like this but it makes the css a lot more unwieldy. 

Plus, .form-control is CSS that's specific to Teams. The only stock CSS that we style is <a> and <input> and <textarea> and this is about giving the config experience a look and feel that matches Teams.

PR #50, opened 2017-05-31
adding isFullScreen attribute to Context interface

liyipui:


src/MicrosoftTeams.ts
```diff
@@ -936,6 +936,11 @@ namespace microsoftTeams
          * The current UI theme the user is using.
          */
         theme?: string;
+
+        /**
+         * The current tab is full screen.
```
ydogandjiev: NIT: I think a slightly better description of this would go something like:
"Indication whether the tab is in full screen mode."

PR #51, opened 2017-06-15
Add compose extension namespace and helpers

maoyu-dev:
This change adds the compose extension namespace and its owner helpers (notifySuccess & notifyFailure) for the authentication and configuration flow of compose extension.

Previously we have been reusing authentication.notifySuccess and authentication.notifyFailure, however it's not only confusing, but also limiting the ability to add future compose extension specific helpers, e.g. we are thinking about a signout callback.

By having a separate namespace for compose extension, it's cleaner and more extensible.


maoyu-dev:
We discussed and decided to not make this change.

PR #52, opened 2017-07-17
Add openStage API and unit tests

ydogandjiev:
Kicking this PR off so we can discuss how to best structure this API. @larryjin, how were you thinking the equivalent bot schema will be structured? Should we model this API after it?

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -228,7 +228,7 @@ namespace microsoftTeams
     /**
      * Shares a deep link a user can use to navigate back to a specific state in this page.
      */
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -963,6 +986,29 @@ namespace microsoftTeams
...
+        /**
+         * Display name for the stage content. If the display name isn't provided the app's name is used.
+         */
+        displayName?: string;
+
+        /**
+         * The developer-defined unique id for the entity the stage content points to.
+         */
+        entityId?: string;
+
```
dfederm: double check that this is optional

dfederm:
:shipit:

PR #55, opened 2017-09-22
Jayongg/addnewproperties2

jayongg:
Update based on new API design

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -53,7 +53,25 @@ namespace microsoftTeams
         args?: any[]; // tslint:disable-line:no-any:The args here are a passthrough from OnMessage where we do receive any[]
     }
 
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -54,14 +54,15 @@ namespace microsoftTeams
...
-        tabList?: TabInstance[];
+        teamTabs?: TabInstance[];
     }
 
     export interface TabInstance {
         tabName: string;
         entityId?: string;
         channelId?: string;
         channelName?: string;
+        channelIsFavorite?: boolean;
```
ydogandjiev: Do we need a teamIsFavorite property as well or did we decide to not enumerate non-favorited teams at all?

jayongg: great point. let me just add it to future proof this.

PR #58, opened 2017-09-25
Add a teamType parameter for context.

jayongg:


src/MicrosoftTeams.ts
```diff
@@ -970,6 +970,11 @@ namespace microsoftTeams
          * Indication whether the tab is in full screen mode.
          */
         isFullScreen?: boolean;
+
+        /**
+         * The type of the team.  A standard team is ""
```
ydogandjiev: Can we mention the possible types here?

PR #59, opened 2017-09-28
behavior getTabInstances() will change - we will return all tabs, not…

jayongg:
behavior getTabInstances() will change - we will return all tabs, not just tabs in favorite channels.

Updating documentation


jayongg:
Thanks!

PR #60, opened 2017-10-02
add "TabInstanceParameters" so getTabInstances() will be able to get …

jayongg:
…them. Also, rename the message request to "getTabInstances" since I'll change the front end to that name. Makes things consistent.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -75,6 +75,18 @@ namespace microsoftTeams {
         Staff = 4
     }
 
+    export interface TabInstanceParameters {
+        /**
+         * Flag indicating to return all channels
+         */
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.ts
```diff
@@ -75,6 +75,18 @@ namespace microsoftTeams {
         Staff = 4
     }
 
+    export interface TabInstanceParameters {
+        /**
+         * Flag indicating to return all channels
```
tacri: Nit: At least for me these flag names suggests some different from description. Maybe something like "flag allowing to select favorite teams/channels only"?

jayongg: oops.  ok :)

PR #62, opened 2017-10-03
Missed some parameters in recent sdk update.  Adding them

jayongg:


src/MicrosoftTeams.ts
```diff
@@ -55,6 +55,7 @@ namespace microsoftTeams {
 
     export interface TabInstance {
         tabName: string;
+        tabId: string;
```
ydogandjiev: Hmm, why are we exposing this? Not a good idea; this is an internal detail of Teams and this ID is subject to mutation which is not made obvious here

jayongg:
Necessary for navigation to work.  We can document it as internal only. But my understanding is there is no way to navigate from the other data.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

src/MicrosoftTeams.ts
tacri: To navigate to a channel we need the channelId and the tabid. We can try to find the tabid when the method is invoked based on the tab properties passed and rely that there is only a tab with that name. 
Or we could also rename this to tabInstanceId

ydogandjiev: As per our offline conversation we will indeed need this; however, I don't think we should expose it to the developers so it doesn't need to be on the interface.

jayongg:
created a different PR - abandoning this one

PR #65, opened 2017-10-06
Add "appId" to TabInstance.

jayongg:


src/MicrosoftTeams.ts
```diff
@@ -54,6 +54,7 @@ namespace microsoftTeams {
     }
 
     export interface TabInstance {
+        appId: string;
```
tacri: Do we need to add this or should we just return the appid as part of the result but not add it to the official TabInstance  interface?

jayongg:
rethinking approach.

PR #75, opened 2017-11-01
Adding @private attribute

charlesprakash:
Added @private attribute for elements that don't need to be shown in the documentation.


charlesprakash:
Travis continuously erring out looking for a non-existent branch. Moving these changes to a different PR.

PR #80, opened 2017-11-08
Add support for connectors inline configuration experience

nigarg1490:
**Align the configuration experience of 3rd party developed connectors with the existing configuration experience of Tabs across all connector clients.**

**Changes:**
Connectors will return a newly added `SaveResult` object as argument to `settings.save` event which will contain the `webhookUrl` that can be accessed by the 3rd party in their `registerOnSaveHandler` like such: 
```
microsoftTeams.settings.registerOnSaveHandler(function (saveEvent) {
 var webhookUrl = saveEvent.result["webhookUrl"];
...... 
}
```

For **Outlook Desktop**, the authentication flow does not contain a pop up because of technical limitations of the client itself. The user experience would be:

> 1. User clicks on Authenticate/Login button on the 3rd party iframed page
> 2. Instead of a pop up, the complete connector's page hosting the iframe redirects to the 3rd party authentication url
> 3. The user completes the authentication
> 4. The window redirects back to the connector's url and now the post-login configuration page loads in the iframe.

**The changes to support this experience are:**

> 1. Similar to Teams desktop, the SDK calls `authentication.authenticate` on it's parent (connector's page)
> 2. The connector's page figures out the client as **Outlook Desktop** and instead of opening a pop up, redirects the complete window to the authentication url after appending a `callbackUrl` query string parameter (with value being the url of the connector's page hosting the iframe).
> 3. Since authentication url is a page hosted on the 3rd party, they will preserve the value of this QSP on their end and the user will complete the authentication process.
> 4. In their post-consent page where today they call `authentication.notifySuccess()/authentication.notifyFailure()` with an optional string `result/reason` param, they will now pass an additional string param `callbackUrl` containing the preserved value from above.
> 5. The SDK will look at the `callbackUrl` param(if present), validate it and redirect the (post-consent)page back to the connector's page by appending the `result/reason` (if supplied) along with `authSuccess/authFailure` as fragments to the connector's URL. 
> 6. When we redirect to the connector's page(which would reload the 3rd party iframe as well), the authentication parameters(success and failure callback methods) need to be registered in the SDK again. To solve this,
> 7. We introduce a `registerAuthenticationParameters()` method to be called on page load which would save the authentication parameters in a global variable in the SDK . This method needs to be called on the configuration page alongside `initialize()`.  So when the configuration page reloads in the iframe, this method would register the callback methods again with the SDK.
> 8. The connector's page will then read the `authSuccess/authFailure` and `result/reason` from URL fragments and call the `authentication.authenticate.success/authentication.authenticate.failure` events with the corresponding values of `result/reason`. 
> 9. The `authentication.authenticate()` no longer needs to pass the authentication parameters if `registerAuthenticationParameters` is called.  

**To summarize, with the new flow developers need to:**

> 1. Call `initialize()` as before
> 2. Call `authentication.registerAuthenticationParameters({url: <auth URL>, width: <width>, height: <height>, successCallback: <successCallback>, failureCallback: <failureCallback>})`
> 3. Call `authentication.authenticate()` as before but without the need to pass authentication parameters.
> 4. Preserve the `callbackUrl` QSP appended to the authentication url when loaded.
> 5. In their post-consent page, instead of `authentication.notifySuccess(result)`, call
```
authentication.notifySucess(result,callbackUrl)
```
> Or, instead of `authentication.notifyFailure(reason)`, call
```
authentication.notifyFailure(reason,callbackUrl))
```

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -230,7 +230,7 @@ namespace microsoftTeams {
      * thrown. This function needs to be used only when navigating the frame to a URL in a different domain 
      * than the current one in a way that keeps the app informed of the change and allows the SDK to 
      * continue working.
-     * @param {string} url The URL to navigate the frame to.
+     * @param url The URL to navigate the frame to.
      */
     export function navigateCrossDomain(url: string): void {
         ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
@@ -247,7 +247,7 @@ namespace microsoftTeams {
      * Allows an app to retrieve for this user tabs that are owned by this app.
      * If no TabInstanceParameters are passed, the app defaults to favorite teams and favorite channels.
      * @param callback The callback to invoke when the {@link TabInstanceParameters} object is retrieved.
-     * @param {TabInstanceParameters} tabInstanceParameters OPTIONAL Flags that specify whether to scope call to favorite teams or channels.
+     * @param tabInstanceParameters OPTIONAL Flags that specify whether to scope call to favorite teams or channels.
      */
     export function getTabInstances(callback: (tabInfo: TabInformation) => void, tabInstanceParameters?: TabInstanceParameters): void {
         ensureInitialized();
@@ -259,7 +259,7 @@ namespace microsoftTeams {
     /**
      * Allows an app to retrieve the most recently used tabs for this user.
      * @param callback The callback to invoke when the {@link TabInformation} object is retrieved.
-     * @param {TabInstanceParameters} tabInstanceParameters OPTIONAL Ignored, kept for future use
+     * @param tabInstanceParameters OPTIONAL Ignored, kept for future use
      */
     export function getMruTabInstances(callback: (tabInfo: TabInformation) => void, tabInstanceParameters?: TabInstanceParameters): void {
         ensureInitialized();
@@ -270,7 +270,7 @@ namespace microsoftTeams {
 
     /**
      * Shares a deep link that a user can use to navigate back to a specific state in this page.
-     * @param {DeepLinkParameters} deepLinkParameters ID and label for the link and fallback URL.
+     * @param deepLinkParameters ID and label for the link and fallback URL.
      */
     export function shareDeepLink(deepLinkParameters: DeepLinkParameters): void {
         ensureInitialized(frameContexts.content);
@@ -284,7 +284,7 @@ namespace microsoftTeams {
 
     /**
      * Navigates the Microsoft Teams app to the specified tab instance.
-     * @param {TabInstance} tabInstance The tab instance to navigate to.
+     * @param tabInstance The tab instance to navigate to.
      */
     export function navigateToTab(tabInstance: TabInstance): void {
         ensureInitialized();
@@ -310,7 +310,7 @@ namespace microsoftTeams {
         /**
          * Sets the validity state for the settings.
          * The initial value is false, so the user cannot save the settings until this is called with true.
-         * @param {boolean} validityState Indicates whether the save or remove button is enabled for the user.
+         * @param validityState Indicates whether the save or remove button is enabled for the user.
          */
         export function setValidityState(validityState: boolean): void {
             ensureInitialized(frameContexts.settings, frameContexts.remove);
@@ -332,7 +332,7 @@ namespace microsoftTeams {
         /**
          * Sets the settings for the current instance.
          * This is an asynchronous operation; calls to getSettings are not guaranteed to reflect the changed state.
-         * @param {Settings} settings The desired settings for this instance.
+         * @param settings The desired settings for this instance.
          */
         export function setSettings(settings: Settings): void {
             ensureInitialized(frameContexts.settings);
@@ -413,7 +413,7 @@ namespace microsoftTeams {
 
             /**
              * Indicates that creation of the underlying resource failed and that the settings cannot be saved.
-             * @param {string} reason Specifies a reason for the failure. If provided, this string is displayed to the user; otherwise a generic error is displayed.
+             * @param reason Specifies a reason for the failure. If provided, this string is displayed to the user; otherwise a generic error is displayed.
              */
             notifyFailure(reason?: string): void;
         }
@@ -426,7 +426,7 @@ namespace microsoftTeams {
 
             /**
              * Indicates that removal of the underlying resource failed and that the content cannot be removed.
-             * @param {string} reason Specifies a reason for the failure. If provided, this string is displayed to the user; otherwise a generic error is displayed.
+             * @param reason Specifies a reason for the failure. If provided, this string is displayed to the user; otherwise a generic error is displayed.
              */
             notifyFailure(reason?: string): void;
         }
@@ -507,7 +507,7 @@ namespace microsoftTeams {
 
         /**
          * Initiates an authentication request, which opens a new window with the specified settings.
-         * @param {AuthenticateParameters} authenticateParameters A set of values that configure the authentication pop-up.
+         * @param authenticateParameters A set of values that configure the authentication pop-up.
          */
         export function authenticate(authenticateParameters: AuthenticateParameters): void {
             ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove);
@@ -541,7 +541,7 @@ namespace microsoftTeams {
         /**
          * Requests an Azure AD token to be issued on behalf of the app. The token is acquired from the cache
          * if it is not expired. Otherwise a request is sent to Azure AD to obtain a new token.
-         * @param {AuthTokenRequest} authTokenRequest A set of values that configure the token request.
+         * @param authTokenRequest A set of values that configure the token request.
          */
         export function getAuthToken(authTokenRequest: AuthTokenRequest): void {
             ensureInitialized();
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -795,6 +836,33 @@ namespace microsoftTeams {
...
+        /**
+         * Appends either result or reason as a hash param to the 'state' url 
+         * @param uri - the url to modify
+         * @param key - the QSP key
+         * @param value - the QSP value
+         */
+        function updateUrlParameter(uri: string, key: string, value: string): string {
+            let i = uri.indexOf("#");
+            let hash = i === -1 ? "" : uri.substr(i);
+            hash = hash + "&" + key + "=" + value;
```
ydogandjiev: If there is no # in the uri this code will just append a new QSP - &<key>=<value> - whereas if there is a # in the uri the key-value pair will be added after it which has completely different behavior. Is this intentional? Maybe the "" should actually be "#"?

src/MicrosoftTeams.ts
```diff
@@ -822,6 +890,17 @@ namespace microsoftTeams {
             failureCallback?: (reason?: string) => void;
         }
 
+        export interface AuthenticationResultParameters {
+            /* reason Specifies a reason for the authentication failure.If specified, the frame that initiated the authentication pop- up receives this value in its callback*/
```
ydogandjiev: NIT: Please fix typos and use standard /** */ comment style as described here:
https://github.com/airbnb/javascript#comments

ydogandjiev: Same for other comments below.

src/MicrosoftTeams.ts
```diff
@@ -745,12 +770,20 @@ namespace microsoftTeams {
          * Notifies the frame that initiated this authentication request that the request was successful.
          * This function is usable only on the authentication window.
          * This call causes the authentication window to be closed.
-         * @param result Specifies a result for the authentication. If specified, the frame that initiated the authentication pop-up receives this value in its callback.
+         * @param authenticationResultParams Specifies a result for the authentication. It contains the optional values of state(URL to redirect back after authentication in Outlook Desktop), reason(in case of failure), result(in case of success).
          */
-        export function notifySuccess(result?: string): void {
+        export function notifySuccess(authenticationResultParams: AuthenticationResultParameters): void {
+            if (authenticationResultParams.state) {
```
ydogandjiev: I don't see a unit test for this code path (i.e. the one where auth happens via redirect). Consider adding one.

src/MicrosoftTeams.ts
```diff
@@ -760,12 +793,20 @@ namespace microsoftTeams {
          * Notifies the frame that initiated this authentication request that the request failed.
          * This function is usable only on the authentication window.
          * This call causes the authentication window to be closed.
-         * @param reason Specifies a reason for the authentication failure. If specified, the frame that initiated the authentication pop-up receives this value in its callback.
+         * @param authenticationResultParams Specifies a result for the authentication. It contains the optional values of state(URL to redirect back after authentication in Outlook Desktop), reason(in case of failure), result(in case of success).
          */
-        export function notifyFailure(reason?: string): void {
+        export function notifyFailure(authenticationResultParams: AuthenticationResultParameters): void {
+            if (authenticationResultParams.state) {
```
ydogandjiev: I don't see a unit test for this code path (i.e. the one where auth happens via redirect). Consider adding one.

src/MicrosoftTeams.ts
```diff
@@ -795,6 +836,33 @@ namespace microsoftTeams {
...
         }
 
+        /*
+        * Validates that the state param is a valid url with host as outlook.office.com and client_type correctly set to Outlook Desktop
+        * @param uri - the url to validate
+        */
+        function isValidState(state: string): boolean {
+            let link = document.createElement("a");
+            link.href = state;
+            if (link.host && link.host !== window.location.host && link.host === "outlook.office.com" && link.search.indexOf("client_type=Win32_Outlook") > -1) {
```
ydogandjiev: This code seems very fragile and risky. Our SDK already has a concept of a hostClientType which tells us what kind of client we're being hosted in and that's sent down from the parent and not inferred ad-hoc from data fed to us by the 3rd party. Could we do something similar here where we introduce a win32outlook or win32desktop client type and the SDK decides on how to handle the notifySuccess/notifyFailure calls based on that?

src/MicrosoftTeams.ts
```diff
@@ -745,12 +770,20 @@ namespace microsoftTeams {
...
-         * @param result Specifies a result for the authentication. If specified, the frame that initiated the authentication pop-up receives this value in its callback.
+         * @param authenticationResultParams Specifies a result for the authentication. It contains the optional values of state(URL to redirect back after authentication in Outlook Desktop), reason(in case of failure), result(in case of success).
          */
-        export function notifySuccess(result?: string): void {
+        export function notifySuccess(authenticationResultParams: AuthenticationResultParameters): void {
+            if (authenticationResultParams.state) {
+                let decodedUrl = decodeURIComponent(authenticationResultParams.state);
+                if (isValidState(decodedUrl)) {
+                    let link = document.createElement("a");
+                    link.href = updateUrlParameter(decodedUrl, "result", authenticationResultParams.result);
```
ydogandjiev: This is dangerous because the result might be a token and this code would be putting it in the QSPs which are generally not secure (even on an https URL). When would a caller provide both a state and a result? In those cases what does the result typically contain?

src/MicrosoftTeams.ts
nigarg1490: When we redirect the connector's page to the authentication url, there is no parent page reference to call the initialize() method which sets the hostClientType. 
We can add more checks here if required to ensure the 3rd party has not tempered with the value of the state param and we don't redirect to a malicious url.

src/MicrosoftTeams.ts
nigarg1490: state should be passed when the client is Outlook desktop, result depends on the 3rd party if they want any information back in their successCallback method they can choose to pass it in  result.
we can either
1. Document that it would be part of QSP so they should not assign sensitive information to it.
2. We can remove support for result altogether for connectors but we need to make sure the solution is consistent for all clients.

src/MicrosoftTeams.ts
nigarg1490: You're right. It should be "#". Will update in the next iteration.

src/MicrosoftTeams.ts
nigarg1490: Added Unit tests

src/MicrosoftTeams.ts
nigarg1490: added unit tests

src/MicrosoftTeams.ts
```diff
@@ -572,35 +589,43 @@ namespace microsoftTeams {
         handlers["authentication.authenticate.failure"] = handleFailure;
 
         /**
-         * Initiates an authentication request, which opens a new window with the specified settings.
+         * Registers the authentication parameters
          * @param authenticateParameters A set of values that configure the authentication pop-up.
+	     */
+        export function registerAuthenticationParameters(authenticateParameters: AuthenticateParameters): void {
```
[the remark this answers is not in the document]
nigarg1490: The authentication parameters(successCallback method etc) are registered via the authenticate() call but in Win32 Outlook when we redirect the page back to connector's url, the iframe page also reloads. 
This means the 3rd party needs to register the authenticate parameters again so that the connector's page can call the callback methods. Since authenticate() call only happens when the user clicks on Authenticate button, we need an API to register the authentication params automatically on page load. 
So developers now need to call registerAuthenticationParameters() in addition to initialize() on the configuration page to ensure the successCallback/failureCallback methods are registered in the SDK on page reload. This also means they no longer need to pass these parameters in the authenticate() call.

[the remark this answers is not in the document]
ydogandjiev: Why not just make this an optional parameter to the initialize call then? Also, I think a better name for this would be something like registerAuthenticationHandlers. What happens if the parent Connectors page invokes the callbacks before this code has run? Will the app just fail to get notified?

src/MicrosoftTeams.ts
```diff
@@ -486,8 +486,8 @@ namespace microsoftTeams {
             removeHandler = handler;
         }
 
-        function handleSave(): void {
-            let saveEvent = new SaveEventImpl();
+        function handleSave(result: SaveResult): void {
```
ydogandjiev: The result parameter should be marked as optional since it won't always be provided (e.g. in the case of tabs).

src/MicrosoftTeams.ts
```diff
@@ -551,12 +556,24 @@ namespace microsoftTeams {
             notifyFailure(reason?: string): void;
         }
 
+        export interface SaveResult {
```
ydogandjiev: I don't think SaveResult is the right name for this; a result implies something that comes out of the save operation but actually this is passed in at the start of it. I think SaveParameters might be amore appropriate name for this.

src/MicrosoftTeams.ts
```diff
@@ -551,12 +556,24 @@ namespace microsoftTeams {
...
+            webhookUrl?: string;
+        }
+
         /**
          * @private
          * Hide from docs, since this class is not directly used.
          */
         class SaveEventImpl implements SaveEvent {
             public notified: boolean = false;
+            public result: SaveResult = {};
```
ydogandjiev: It's pointless to initialize the result here given that you are setting its value in the constructor.

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
nigarg1490: If we make it optional parameter in the initialize call, we'd have to expose the AuthenticateParameters interface outside the Authentication namespace. Although that would ensure that callbacks would be invoked only after the handlers are registered but it doesn't look like an ideal design.
Yes, the successCallback/failureCallback methods would not be called if the handlers are invoked by connector's page before this API is called. We can document that this API needs to be called even before initialize() on this page to prevent this.
I'm fine with either approach.

PR #82, opened 2017-11-16
Add back support for back button handling and navigation

ydogandjiev:
This change is needed in order to bring the Assignments app to our Mobile clients.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -169,6 +169,9 @@ namespace microsoftTeams {
     let fullScreenChangeHandler: (isFullScreen: boolean) => void;
     handlers["fullScreenChange"] = handleFullScreenChange;
 
+    let backButtonPressHandler: () => void;
+    handlers["backButtonPress"] = handleBackButtonPress;
+
     /**
      * Initializes the library. This must be called before any other SDK calls
      * but after the frame is loaded successfully.
@@ -209,13 +212,12 @@ namespace microsoftTeams {
 
         // Undocumented function used to clear state between unit tests
         this._uninitialize = () => {
-            if (frameContext === frameContexts.settings) {
-                settings.registerOnSaveHandler(null);
-            }
+            settings.registerOnSaveHandler(null);
+            settings.registerOnRemoveHandler(null);
 
-            if (frameContext === frameContexts.remove) {
-                settings.registerOnRemoveHandler(null);
-            }
+            registerOnThemeChangeHandler(null);
+            registerFullScreenHandler(null);
+            registerBackButtonHandler(null);
 
             initializeCalled = false;
             parentWindow = null;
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -282,6 +284,42 @@ namespace microsoftTeams {
...
+         } else {
+             navigateBack();
+         }
+    }
+
+    /**
+     * Navigates back in the Teams client. See registerBackButtonHandler for more information on when
+     * it's appropriate to use this method.
+     */
+    export function navigateBack(): void {
```
joeuy: This will always mean close the opened tab right?

src/MicrosoftTeams.ts
```diff
@@ -282,6 +284,42 @@ namespace microsoftTeams {
...
+     * method to ask the Teams client to handle it instead.
+     * @param handler The handler to invoke when the user presses their Team client's back button.
+     */
+    export function registerBackButtonHandler(handler: () => void): void {
+        ensureInitialized();
+
+        backButtonPressHandler = handler;
+    }
+
+    function handleBackButtonPress(): void {
```
joeuy: it would be nice if ** backButtonPressHandler** could signal whether to perform "navigateBack" or suppress it. so the client of the SDK does not have to go ahead and call navigateBack();

similar to when you have a click event and you say: event.preventDefault() [maybe returning false works too];

src/MicrosoftTeams.ts
ydogandjiev: Yes

src/MicrosoftTeams.ts
ydogandjiev: Done; if the tab returns false from its back button handler we will automatically call navigateBack(); if it returns true then we assume the back button press was handled and we shouldn't do anything.

src/MicrosoftTeams.ts
ydogandjiev: Yup

ydogandjiev:
@jayongg, @tacri, can one of you guys please take a look at this PR when you get a chance?

jayongg:
Yes will do

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

src/MicrosoftTeams.ts
```diff
@@ -209,6 +212,12 @@ namespace microsoftTeams {
 
         // Undocumented function used to clear state between unit tests
         this._uninitialize = () => {
+            if (frameContext) {
+                registerOnThemeChangeHandler(null);
```
jayongg: whoa, were we leaking before??

ydogandjiev: This code only runs during unit testing and it was resulting in certain state to persist across test cases. It would not have had an impact on the use of our library in production.

PR #92, opened 2018-02-08
Add sessionId to SDK context to allow for better correlation of telemetry events between Teams and tabs

ydogandjiev:
In order to debug various issues our developers have encountered with their apps we have found the need to correlate events between the Teams telemetry and that of the app. We plan to implement this by sharing the Teams sessionId with the tabs we host via the getContext API of our SDK.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -200,7 +200,6 @@ namespace microsoftTeams {
             // Send the initialized message to any origin, because at this point we most likely don't know the origin
             // of the parent window, and this message contains no data that could pose a security risk.
             parentOrigin = "*";
-
             let messageId = sendMessageRequest(parentWindow, "initialize", [version]);
             callbacks[messageId] = (context: string, clientType: string) => {
                 frameContext = context;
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -175,8 +176,9 @@ namespace microsoftTeams {
     /**
      * Initializes the library. This must be called before any other SDK calls
      * but after the frame is loaded successfully.
+     * @param {string} [appCorrelationId] Optional unique ID for use in correlating telemetry data.
      */
-    export function initialize(): void {
+    export function initialize(appCorrelationId?: string): void {
```
ydogandjiev: I am considering removing the option to provide a correlationId. I think we have 3 options here and I am still trying to evaluate which one is the best:
1. Have the developer provide a correlation ID on initialization and simply pass it back to the Teams client.
2. Always generate a unique ID on initialization and pass it back to the Teams client.
3. Have the Teams client either generate a unique ID or pass its session ID to the SDK.

ydogandjiev: I need to also think through the implications for this when the parent isn't Teams but the tab content/configuration page instead. This happens when the SDK is loaded inside the authentication popup and the app is running in our web client.

PR #94, opened 2018-04-10
updating package scope to @microsoft

sachingupta:
This change will make package name under Microsoft scope to be publish as npm packages through Microsoft account.


ydogandjiev:
@charlesprakash, can you please double-check that this change won't mess with our auto-generated documentation. Also, be aware that @sachingupta is working towards a plan to publish v1.3 of the SDK as an npm package. We will then do that for every release going forward.

PR #99, opened 2018-04-13
Update manifest schema to support apps in group chat

XinGao:
Add a new scope "groupchat" for configurableTabs and bots.

### DIFF
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -128,7 +128,7 @@
         },
         "configurableTabs": {
             "type": "array",
-            "description": "These are tabs users can optionally add to their channels and require extra configuration before they are added. Configurable tabs are not supported in the personal scope. Currently only one configurable tab per app is supported.",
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.schema.json
```diff
@@ -147,11 +147,12 @@
...
                         "type": "array",
-                        "description": "Specifies whether the tab offers an experience in the context of a channel in a team, or an experience scoped to an individual user alone. These options are non-exclusive. Currently, configurable tabs are only supported in the teams scope.",
-                        "maxItems": 1,
+                        "description": "Specifies whether the tab offers an experience in the context of a channel in a team, in a 1:1 or group chat, or in an experience scoped to an individual user alone. These options are non-exclusive. Currently, configurable tabs are only supported in the teams and groupchats scopes.",
+                        "maxItems": 2,
                         "items": {
                             "enum": [
-                                "team"
+                                "team",
+                                "groupchat"
```
ydogandjiev: What about staticTabs? We should probably support groupchat there as well.

XinGao: We're not planning to support it for the moment, but I'll talk to PM and let him figure out.

PR #106, opened 2018-05-03
Changes to the sdk to support the launch and dismissal of the task module

debajyoti2510:
Adding a new namespace under microsoftTeams.task. 
Adding a couple of functions: start() to launch the task module and complete() when the task is done

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -151,6 +151,12 @@ namespace microsoftTeams {
     favoriteTeamsOnly?: boolean;
   }
 
+  export const enum TaskModuleDimension {
+    Large = "80px",
+    Medium = "60px",
+    Small = "40px"
+  }
+
   // This indicates whether initialize was called (started).
   // It does not indicate whether initialization is complete. That can be inferred by whether parentOrigin is set.
   let initializeCalled = false;
@@ -1585,12 +1591,12 @@ namespace microsoftTeams {
     /**
      * The requested height of the webview/iframe in pixels.
      */
-    height?: string;
+    height?: TaskModuleDimension | Number;
 
     /**
      * The requested width of the webview/iframe in pixels.
      */
-    width?: string;
+    width?: TaskModuleDimension | Number;
 
     /**
      * Title of the task module.
@@ -1620,7 +1626,9 @@ namespace microsoftTeams {
       // Ensure that the tab content is initialized
       ensureInitialized(frameContexts.content);
 
-      let messageId = sendMessageRequest(parentWindow, "tasks.startTask", [taskInfo]);
+      let messageId = sendMessageRequest(parentWindow, "tasks.startTask", [
+        taskInfo
+      ]);
       callbacks[messageId] = completionHandler;
     }
 
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -1607,13 +1607,13 @@ namespace microsoftTeams {
    * Namespace to interact with the task module-specific part of the SDK.
    * This object is usable only on the content frame.
    */
-  export namespace task {
+  export namespace tasks {
     /**
      * Allows an app to open the task module.
      * @param taskInfo An object containing the parameters of the task module
      * @param completionHandler Handler to call when the task module is completed
      */
-    export function start(
+    export function startTask(
       taskInfo: TaskInfo,
       completionHandler?: (err: string, result: string) => void
     ): void {
@@ -1629,7 +1629,7 @@ namespace microsoftTeams {
      * @param result Contains the result to be sent to the bot or teh app. Typically a JSON object or a serialized version of it
      * @param appId Helps to validate that the call originates from the same appId as the one that invoked the task module
      */
-    export function complete(
+    export function completeTask(
       result?: string | object,
       appIds?: string[]
     ): void {
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -152,9 +152,9 @@ namespace microsoftTeams {
   }
 
   export const enum TaskModuleDimension {
-    Large = "80px",
-    Medium = "60px",
-    Small = "40px"
+    Large = "large",
+    Medium = "medium",
+    Small = "small"
   }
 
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.ts
```diff
@@ -1566,4 +1570,73 @@ namespace microsoftTeams {
...
+      ensureInitialized(frameContexts.content);
+
+      let messageId = sendMessageRequest(parentWindow, "start", [taskInfo]);
+      callbacks[messageId] = completionHandler;
+    }
+
+    /**
+     * Complete the task module.
+     * @param result Contains the result to be sent to the bot or teh app. Typically a JSON object or a serialized version of it
+     * @param appId Helps to validate that the call originates from the same appId as the one that invoked the task module
```
mevinay: nit: teh -> the

src/MicrosoftTeams.ts
```diff
@@ -1566,4 +1570,73 @@ namespace microsoftTeams {
...
+     * @param completionHandler Handler to call when the task module is completed
+     */
+    export function startTask(
+      taskInfo: TaskInfo,
+      completionHandler?: (err: string, result: string) => void
+    ): void {
+      // Ensure that the tab content is initialized
+      ensureInitialized(frameContexts.content);
+
+      let messageId = sendMessageRequest(parentWindow, "start", [taskInfo]);
```
mevinay: I think "start" will not give sufficient context to the message receiver and we should ideally use the function name "startTask". Same for "completeTask".

src/MicrosoftTeams.ts
mevinay: We are documenting for appId here but actually using appIds in the function. I am not clear on the need for appIds, but please check and correct.

src/MicrosoftTeams.ts
ydogandjiev: @debajyoti2510, please make sure the messages we send match the namespace + API name. This should send "tasks.startTask" instead of just "start".

src/MicrosoftTeams.ts
```diff
@@ -1566,4 +1570,73 @@ namespace microsoftTeams {
...
+     * @param appId Helps to validate that the call originates from the same appId as the one that invoked the task module
+     */
+    export function completeTask(
+      result?: string | object,
+      appIds?: string[]
+    ): void {
+      // Ensure that the tab content is initialized
+      ensureInitialized(frameContexts.content);
+
+      sendMessageRequest(parentWindow, "complete", [result, appIds]);
```
ydogandjiev: Please make the message name here "tasks.completeTask" instead of just "complete". That's the pattern we've established in the rest of the SDK and makes it clear which API call corresponds to the message.

src/MicrosoftTeams.ts
```diff
@@ -1566,4 +1570,89 @@ namespace microsoftTeams {
...
+
+    /**
+     * The requested height of the webview/iframe in pixels.
+     */
+    height?: string;
+
+    /**
+     * The requested width of the webview/iframe in pixels.
+     */
+    width?: string;
```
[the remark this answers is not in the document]
ydogandjiev: If we are going to use small, medium, large here then please use en enum and update the comments.

src/MicrosoftTeams.ts
debajyoti2510: Just made the change to the app Ids yesterday. Missed changing it on the parameter. The need for appIds is if the developer has multiple apps using the same task module pages

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
debajyoti2510: I think @billbliss wanted to support both strings (small, large, medium) and numbers. Thats why I kept it as a string to account for both.

src/MicrosoftTeams.ts
```diff
@@ -1566,4 +1570,89 @@ namespace microsoftTeams {
...
+
+    /**
+     * Invoke message sent to the bot to fect the URL or Adaptive card body dynamically.
+     */
+    fetchtask?: boolean;
+
+    /**
+     * The requested height of the webview/iframe in pixels.
+     */
+    height?: string;
```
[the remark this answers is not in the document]
debajyoti2510: As mentioned in the height case, Bill wanted to support both numbers and string. Keeping this as string because of that

src/MicrosoftTeams.ts
```diff
@@ -151,6 +151,12 @@ namespace microsoftTeams {
     favoriteTeamsOnly?: boolean;
   }
 
+  export const enum TaskModuleDimension {
+    Large = "80px",
+    Medium = "60px",
+    Small = "40px"
```
ydogandjiev: I'm not sure we want to list explicit px sizes here for the enum. I think the Large, Medium, and Small should essentially be snap points that we can choose to map to sizes on our end and change them in the future if needed. Have you verified what this looks like in the message? Do we send over the actual values (i.e. 40, 60, 80) or the enum names (i.e. Small, Medium, Large)?

aosolis: The last time I checked, the spec said these are percentages of the available space. 40, 60 and 80 pixels are *really* small dimensions.

src/MicrosoftTeams.ts
```diff
@@ -1566,4 +1576,75 @@ namespace microsoftTeams {
...
+     * Complete the task module.
+     * @param result Contains the result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
+     * @param appIds Helps to validate that the call originates from the same appId as the one that invoked the task module
+     */
+    export function completeTask(
+      result?: string | object,
+      appIds?: string[]
+    ): void {
+      // Ensure that the tab content is initialized
+      ensureInitialized(frameContexts.content);
```
ydogandjiev: Hmm, we should probably add a new frameContext for the task module (e.g. frameContextx.taskModule) and validate that this call is only happening from there. Otherwise, someone might try to call this API from their content page which would have unpredictable behavior.

billbliss: Nice catch

debajyoti2510: sure will do

PR #107, opened 2018-05-14
Added support for ViewConfiguration, Navigation Bar Menu and Action Menu. Also, added support for SDK integration in a iframe-less scenario.

harjaMsft:


src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
         args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
     }
 
+     /**
```
mevinay: nit: extra space before /

src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
...
+     * Namespace to send message to SDK from native apps.
+     */
+    export namespace native {
+        export function postMessage(msg: MessageEvent): void {
+            handleParentMessage(msg);
+        }
+    }
+
+     /**
```
mevinay: nit: extra space before /

src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
...
+
+        function handleViewConfigItemPress(id: String): void {
+            if (!viewConfigItemPressHandler || !viewConfigItemPressHandler(id)) {
+                ensureInitialized();
+                sendMessageRequest(parentWindow, "viewConfigItemPress", [id]);
+            }
+        }
+
+        /**
+         * Used to set menu items on the Navigation Bar. If icon is available, icon will be shown, other title will be shown.
```
mevinay: other -> otherwise

src/MicrosoftTeams.ts
```diff
@@ -134,6 +304,10 @@ namespace microsoftTeams {
         Plc = 3,
         Staff = 4,
     }
+    interface Window {
```
mevinay: nit: need a newline before this interface.

src/MicrosoftTeams.ts
```diff
@@ -1389,17 +1572,25 @@ namespace microsoftTeams {
...
     function sendMessageRequest(targetWindow: Window, actionName: string, args?: any[]): number {
         let request = createMessageRequest(actionName, args);
-        let targetOrigin = getTargetOrigin(targetWindow);
-
-        // If the target window isn't closed and we already know its origin, send the message right away; otherwise,
-        // queue the message and send it after the origin is established
-        if (targetWindow && targetOrigin) {
-            targetWindow.postMessage(request, targetOrigin);
+        // tslint:disable-next-line: no-any
+        if (currentWindow["teamsNativeClient"]){
```
mevinay: nit: space before {

src/MicrosoftTeams.ts
```diff
@@ -188,14 +362,23 @@ namespace microsoftTeams {
...
         // If we are in an iframe, our parent window is the one hosting us (i.e., window.parent); otherwise,
         // it's the window that opened us (i.e., window.opener)
         parentWindow = (currentWindow.parent !== currentWindow.self) ? currentWindow.parent : currentWindow.opener;
 
+        // Listen for messages post to our window
+        let messageListener = (evt: MessageEvent) => processMessage(evt);
+
+        // tslint:disable-next-line: no-any
+        if (!parentWindow && currentWindow["teamsNativeClient"]) {
+            // assign the parent
```
mevinay: Please add a more elaborate comment to clarify the usage of this for frame-less scenario.

src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
         args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
     }
 
+     /**
+     * Namespace to send message to SDK from native apps.
```
mevinay: This comment should be more generic and mention about the use in frame-less scenario. And you should have another comment before postMessage which mentions about its use for sending messages.

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
harjaMsft: done

src/MicrosoftTeams.ts
```diff
@@ -188,14 +362,23 @@ namespace microsoftTeams {
...
-
         // If we are in an iframe, our parent window is the one hosting us (i.e., window.parent); otherwise,
         // it's the window that opened us (i.e., window.opener)
         parentWindow = (currentWindow.parent !== currentWindow.self) ? currentWindow.parent : currentWindow.opener;
 
+        // Listen for messages post to our window
+        let messageListener = (evt: MessageEvent) => processMessage(evt);
+
+        // tslint:disable-next-line: no-any
+        if (!parentWindow && currentWindow["teamsNativeClient"]) {
```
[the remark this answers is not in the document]
mevinay: This is needed for all Web Modules (both iOS & Android). It is very much on the lines of the earlier changes done for Assignments. We will work with the Assignments team to switch to this.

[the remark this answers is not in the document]
mevinay: Assignments on iOS is the only user of this functionality as of now, and they are not using the SDK for this anyway, so there is no need of compatibility. We went with the name changes which are more suitable in the SDK and also we didn't want to do any special handling on Android to keep the same names. Transition to the SDK needs very minor changes.

src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
...
+             */
+            listItems: MenuItem[];
+        }
+
+        /**
+         * Represents information about type of list to display in Navigation Bar Menu.
+         */
+        export enum MenuListType {
+            dropDown = "dropDown",
+            actionSheet = "actionSheet",
```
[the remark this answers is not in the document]
harjaMsft: 'actionSheet' is mentioned in specs. If required, it can be changed.

@mevinay What is your suggestion?

src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
...
+        export function postMessage(msg: MessageEvent): void {
+            handleParentMessage(msg);
+        }
+    }
+
+     /**
+     * Namespace to interact with the menu-specific part of the SDK.
+     * This object is used to show View Configuration, Action Menu and Navigation Bar Menu.
+     */
+    export namespace menu {
```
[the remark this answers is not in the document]
harjaMsft: done

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
mevinay: "actionSheet" comes from https://developer.apple.com/ios/human-interface-guidelines/views/action-sheets/. But i am fine with "popOver" as well.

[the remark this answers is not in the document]
ydogandjiev: OK, let's go with "popOver" then. Feel free to describe the specific iOS/Android semantics in the comments.

src/MicrosoftTeams.ts
```diff
@@ -182,6 +377,16 @@ namespace microsoftTeams {
   let backButtonPressHandler: () => boolean;
   handlers["backButtonPress"] = handleBackButtonPress;
 
+  handlers["nativeJsLoaded"] = handleNativeJsLoaded;
```
mevinay: nativeJsLoaded -> nativeClientInitialize
handleNativeJsLoaded -> handleNativeClientInitialize

src/MicrosoftTeams.ts
```diff
@@ -200,7 +405,17 @@ namespace microsoftTeams {
 
     // Listen for messages post to our window
     let messageListener = (evt: MessageEvent) => processMessage(evt);
-    currentWindow.addEventListener("message", messageListener, false);
+
+    // tslint:disable-next-line: no-any
+    if (!parentWindow) {
```
mevinay: I think these lines should not be here and should be before the "try" block.

src/MicrosoftTeams.ts
```diff
@@ -200,7 +405,17 @@ namespace microsoftTeams {
...
+      isFramelessWindow = true;
+      // For frame-less scenario, assign parent window for communication
+      // tslint:disable-next-line: no-any
+      parentWindow = currentWindow.teamsNativeClient;
```
mevinay: I think we don't need this parentWindow initialization now.

src/MicrosoftTeams.ts
```diff
@@ -1389,17 +1572,25 @@ namespace microsoftTeams {
     // tslint:disable-next-line:no-any
     function sendMessageRequest(targetWindow: Window, actionName: string, args?: any[]): number {
         let request = createMessageRequest(actionName, args);
-        let targetOrigin = getTargetOrigin(targetWindow);
-
-        // If the target window isn't closed and we already know its origin, send the message right away; otherwise,
-        // queue the message and send it after the origin is established
-        if (targetWindow && targetOrigin) {
-            targetWindow.postMessage(request, targetOrigin);
+        // tslint:disable-next-line: no-any
```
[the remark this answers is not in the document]
mevinay: done

src/MicrosoftTeams.ts
```diff
@@ -1389,17 +1572,25 @@ namespace microsoftTeams {
...
+        if (currentWindow["teamsNativeClient"]){
+            // tslint:disable-next-line: no-any
+            setTimeout(function(): void {
+                currentWindow["teamsNativeClient"]["framelessPostMessage"](request);
```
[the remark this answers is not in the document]
mevinay: done

src/MicrosoftTeams.ts
```diff
@@ -1389,17 +1572,25 @@ namespace microsoftTeams {
...
-        let targetOrigin = getTargetOrigin(targetWindow);
-
-        // If the target window isn't closed and we already know its origin, send the message right away; otherwise,
-        // queue the message and send it after the origin is established
-        if (targetWindow && targetOrigin) {
-            targetWindow.postMessage(request, targetOrigin);
+        // tslint:disable-next-line: no-any
+        if (currentWindow["teamsNativeClient"]){
+            // tslint:disable-next-line: no-any
+            setTimeout(function(): void {
```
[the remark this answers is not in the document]
mevinay: sendMessageRequest needs to return before the request is processed.

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
mevinay: done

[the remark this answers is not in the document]
harjaMsft: done

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
harjaMsft: Messages need to be send to frameless window in asynchronous manner. Hence I have used setTimeout here.

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
harjaMsft: done

src/MicrosoftTeams.ts
```diff
@@ -49,6 +49,176 @@ namespace microsoftTeams {
         args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
     }
 
+     /**
+     * Namespace to send message to SDK from native apps.
+     */
+    export namespace native {
+        export function postMessage(msg: MessageEvent): void {
```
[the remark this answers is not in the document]
harjaMsft: postMessage is a way of communication from native client to SDK. 
You have rightly pointed, that this may confuse developers. I have added a comment to address this confusion.

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
harjaMsft: done

src/MicrosoftTeams.ts
```diff
@@ -47,6 +55,191 @@ namespace microsoftTeams {
...
+
+    /**
+     * Used to show Action Menu.
+     * @param title Title for Action Menu
+     * @param items List of MenuItems for Action Menu.
+     * @param handler The handler to invoke when the user selects menu item.
+     */
+    export function showActionMenu(
+      title: string,
+      items: MenuItem[],
```
ydogandjiev: Might be better to wrap the title and items into their own class (e.g. ActionMenuConfiguration or ActionMenuParameters) so that we can easily add to them later without re-ordering the parameters of the API. For example, when we implement this in web/desktop we would need to ask the developer to provide relative coordinates for the menu.

src/MicrosoftTeams.ts
mevinay: done

src/MicrosoftTeams.ts
```diff
@@ -36,7 +44,7 @@ namespace microsoftTeams {
     web: "web"
   };
 
-  interface MessageRequest {
+  export interface MessageRequest {
```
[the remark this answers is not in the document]
mevinay: This is needed by the framelessPostMessage api as argument. We will have to use any in the argument if we don't export.

src/MicrosoftTeams.ts
```diff
@@ -47,6 +55,191 @@ namespace microsoftTeams {
     args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
   }
 
+  /**
+   * Namespace to send message to SDK from native apps, incase of frameless scenario.
+   */
+  export namespace native {
```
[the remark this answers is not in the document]
mevinay: If we don't export, we will have no way to call the method from the mobile app code. Do you have any alternative in mind?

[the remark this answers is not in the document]
ydogandjiev: Not sure I follow; the registration for messages should always be done from the SDK side which can reference all of its non-exported functions.

src/MicrosoftTeams.ts
```diff
@@ -182,6 +387,16 @@ namespace microsoftTeams {
   let backButtonPressHandler: () => boolean;
   handlers["backButtonPress"] = handleBackButtonPress;
 
+  handlers["nativeClientInitialize"] = handleNativeClientInitialize;
```
ydogandjiev: What is this needed for? Initialization should always happen in response to the SDK’s initalize call; not in a handler.

src/MicrosoftTeams.ts
```diff
@@ -1547,20 +1772,32 @@ namespace microsoftTeams {
...
     let request = createMessageRequest(actionName, args);
-    let targetOrigin = getTargetOrigin(targetWindow);
-
-    // If the target window isn't closed and we already know its origin, send the message right away; otherwise,
-    // queue the message and send it after the origin is established
-    if (targetWindow && targetOrigin) {
-      targetWindow.postMessage(request, targetOrigin);
+    if (isFramelessWindow) {
+      if (
+        currentWindow.teamsNativeClient &&
```
ydogandjiev: Why can’t you ensure that these injected objects are available from the get-go. This asynchronous initialization seems unnecessary.

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
mevinay: In a iframe-less scenario, as window.postMessage is not usable to deliver the messages, some method is required to receive the messages from client app. This is that namespace/method.

src/MicrosoftTeams.ts
mevinay: In a iframe-less scenario, Android client app needs to inject js into the page to receive even the first "initialize" message. This handler is invoked when the client app is done with injection and is ready to receive the messages. Till that time the messages are held in a queue in the SDK.

src/MicrosoftTeams.ts
mevinay: There is no way on Android to ensure the objects are available from the start. On iOS there is a way, but unfortunately none on Android.

src/MicrosoftTeams.ts
ydogandjiev: But you have no guarantee that the loaded page has even initialized the SDK. More importantly, they might be using it via the npm package at which point it would just be part of their app bundle and not in the global namespace. The SDK has to always be the one that hooks up the message handler to some global object (e.g. the Window). Your injected code must not make direct calls to the SDK; it must only post messages via the registered handler.

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
ydogandjiev: Please see my other comment. The SDK needs to register this handler. Your injected code cannot directly call methods on the SDK.

src/MicrosoftTeams.ts
mevinay: I didn't get you here. Can you share some code snippet of what you have in mind?

ydogandjiev: It seems like your plan is to have the injected code call microsofTeams.native.postMessage in order to send messages to the SDK. What I am trying to say is that this will not work in cases where the microsoftTeams object isn't available in the global namespace. That could happen because the <script> tag that pulls in the SDK is being delay loaded or the developer has bundled the SDK with their app's code.

You must instead follow the pattern that was established by the iOS version of the SDK that Assignments uses, i.e. register a handler on some global object when the developer calls microsoftTeams.initialize(). That would look something like:
```
window.onNativeMessage = handleParentMessage;
```

Then in the injected code you just call onNativeMessage when you want to post something to the SDK.

src/MicrosoftTeams.ts
ydogandjiev: This is going to be very problematic since it means you need to handle both the case where the SDK loads first and the case where the injected code load first. Please think through both scenarios and make sure your solution works for them. You might need to set up a poller in the SDK to wait for the injected code to become available or store all the messages in a global queue that the injected code can pump once loaded.

mevinay: Yes, the current code works for both the scenarios. Also, the changes already have the queue approach and all the messages are kept in the queue till the native client is ready. Once the native client is ready, it is currently flushing the queue inside the handleNativeClientInitialize. The poller will work as well, but i guess this is cleaner than that.

src/MicrosoftTeams.ts
ydogandjiev: I don't see you registering  window.microsoftTeams.native.postMessage anywhere. And yes, please remove this namespace and method from the SDK and don't export it. All you need to do is register the handler inside of microsoftTeams.initialize. I am going to repeat what I said earlier; you **cannot** rely on the microsoftTeams object to be globally available on the window; you must explicitly register the global handler at runtime when the hosted web app calls microsoftTeams.initialize().

ydogandjiev: If you want to see what I mean just build a web app using typescript or webpack+babel and use the SDK via its npm package. In that case you wouldn't reference the SDK from our CDN but instead you would first:
```
yarn add @microsoft/teams-js
```

And then inside of your code you would do something like:
```
import * as microsoftTeams from '@microsoft/teams-js';
microsoftTeams.initialize();
```

For an app built in this way window.microsoftTeams would always be undefined.

src/MicrosoftTeams.ts
```diff
@@ -182,6 +374,8 @@ namespace microsoftTeams {
   let backButtonPressHandler: () => boolean;
   handlers["backButtonPress"] = handleBackButtonPress;
 
+  window.onNativeMessage = handleParentMessage;
```
ydogandjiev: Please move this into the initialize() API and only set it when you detect that you’re running in frameless mode (i.e. when you don’t have a parent or opener).

harjaMsft: done

src/MicrosoftTeams.ts
```diff
@@ -1547,20 +1748,25 @@ namespace microsoftTeams {
...
     let request = createMessageRequest(actionName, args);
-    let targetOrigin = getTargetOrigin(targetWindow);
-
-    // If the target window isn't closed and we already know its origin, send the message right away; otherwise,
-    // queue the message and send it after the origin is established
-    if (targetWindow && targetOrigin) {
-      targetWindow.postMessage(request, targetOrigin);
+    if (isFramelessWindow) {
+      currentWindow.nativeInterface.framelessPostMessage(
+        JSON.stringify(request)
```
ydogandjiev: Are your sure that you don’t need a queue here? According to @mevinay the injection of the framelessPostMessage API in Android would be asynchronous and might not be in place when the SDK starts making calls.

mevinay: @ydogandjiev we found a way on Android which seems to work till now. If at all it fails on any devices, we will have to go back to the earlier approach.

harjaMsft: Yes, the queue is not needed anymore. We changed approach on the client side. Instead of injecting JS, we have exposed JavaInterface for JS.

PR #108, opened 2018-05-14
adding isteamArchived to context

sachingupta:
This will add isTeamArchived paramter to get context that will be used to get information about if team is archived or not


src/MicrosoftTeams.ts
```diff
@@ -1291,6 +1291,11 @@ namespace microsoftTeams {
      * The Microsoft Teams ID for the chat with which the content is associated.
      */
     chatId?: string;
+
+    /**
+     * Indicate wheather team is archived
```
ydogandjiev: Consider adding something along the lines of:
"apps should use this as a signal to prevent any changes to content associated with archived teams"

PR #111, opened 2018-06-22
Fix 60+ tslint errors

billbliss:
Clearly we don't run our standard tslint when we build the SDK, but if you do directly build it in a project that is, you see over 60 errors. This fixes all of them.


ydogandjiev:
@billbliss, you can't make these changes without updating our tslint.json rules since these changes break our build:
https://travis-ci.org/OfficeDev/microsoft-teams-library-js/builds/395522046?utm_source=github_status&utm_medium=notification

PR #112, opened 2018-06-29
Saching/sharepoint support

sachingupta:
Adding regex support for valid host domains.
added sharepoint domain in valid host regex

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -21,14 +21,26 @@ namespace microsoftTeams {
 
   const version = "1.2";
 
-  const validOrigins = [
-    "https://teams.microsoft.com",
-    "https://teams.microsoft.us",
-    "https://int.teams.microsoft.com",
-    "https://devspaces.skype.com",
-    "https://ssauth.skype.com",
+  const validHostRegEx = new RegExp(
+    "https://teams.microsoft.com"
+    + "|" +
+    "https://teams.microsoft.us"
+    + "|" +
+    "https://int.teams.microsoft.com"
+    + "|" +
+    "https://devspaces.skype.com"
+    + "|" +
+    "https://ssauth.skype.com"
+    + "|" +
     "http://dev.local" // local development
-  ];
+    + "|" +
+    "https://msft.spoppe.com"
+    + "|" +
+    "http*.*sharepoint.com"
+    + "|" +
+    "http*.*sharepoint-df.com"
+    , "i"
+  );
 
   const handlers: { [func: string]: Function } = {};
 
@@ -1062,13 +1074,13 @@ namespace microsoftTeams {
         link.href,
         "_blank",
         "toolbar=no, location=yes, status=no, menubar=no, scrollbars=yes, top=" +
-          top +
-          ", left=" +
-          left +
-          ", width=" +
-          width +
-          ", height=" +
-          height
+        top +
+        ", left=" +
+        left +
+        ", width=" +
+        width +
+        ", height=" +
+        height
       );
       if (childWindow) {
         // Start monitoring the authentication window so that we can detect if it gets closed before the flow completes
@@ -1647,7 +1659,7 @@ namespace microsoftTeams {
     if (
       messageSource === currentWindow ||
       (messageOrigin !== currentWindow.location.origin &&
-        validOrigins.indexOf(messageOrigin.toLowerCase()) === -1)
+        validHostRegEx.test(messageOrigin.toLowerCase()))
     ) {
       return;
     }
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -1,3 +1,13 @@
+declare interface String {
+  startsWith(search: string, pos: number): boolean;
+}
+
+if (!String.prototype.startsWith) {
+  String.prototype.startsWith = function (search: string, pos: number): boolean {
+    return this.substr(!pos || pos < 0 ? 0 : +pos, search.length) === search;
+  };
+}
+
 // Shim in definitions used for browser-compat
 interface MessageEvent {
   // Needed for Chrome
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -21,14 +21,26 @@ namespace microsoftTeams {
 
   const version = "1.2";
 
-  const validOrigins = [
-    "https://teams.microsoft.com",
-    "https://teams.microsoft.us",
-    "https://int.teams.microsoft.com",
-    "https://devspaces.skype.com",
-    "https://ssauth.skype.com",
+  const validHostRegEx = new RegExp(
```
ydogandjiev: NIT: Consider making this a little easier to extend by creating the regex dynamically in code out of an array

src/MicrosoftTeams.ts
```diff
@@ -21,14 +21,26 @@ namespace microsoftTeams {
...
 
-  const validOrigins = [
-    "https://teams.microsoft.com",
-    "https://teams.microsoft.us",
-    "https://int.teams.microsoft.com",
-    "https://devspaces.skype.com",
-    "https://ssauth.skype.com",
-    "http://dev.local" // local development
-  ];
+  const validHostRegEx = new RegExp(
```
aosolis: Can we name this back to validOriginsRegExp, since it's actually checking the origin, not just the host?

src/MicrosoftTeams.ts
```diff
@@ -21,14 +21,26 @@ namespace microsoftTeams {
...
+    "|" +
+    "https://devspaces.skype.com" +
+    "|" +
+    "https://ssauth.skype.com" +
+    "|" +
+    "http://dev.local" + // local development
+    "|" +
+    "https://msft.spoppe.com" +
+    "|" +
+    "http*.*sharepoint.com" +
```
aosolis: This looks wrong. I think this clause of the regexp means
    `"htt" + (zero or more "p"s) + (zero or more of any characters) + "sharepoint" + (any character) + com`

which would match the following:
- htt://x.phishing-site-sharepoint.com
- httpppppp://x.malicious-sharepoint.com
- httpppppp://x.malicious-sharepoint-com
- httpmalicious-sharepointxcom

src/MicrosoftTeams.ts
sachingupta: sure

src/MicrosoftTeams.ts
ydogandjiev: Consider re-using the logic we have for domain validation in extensionTabService.ts.

src/MicrosoftTeams.ts
```diff
@@ -1,3 +1,13 @@
+declare interface String {
```
ydogandjiev: Hmm... Why are these needed?

test/MicrosoftTeams.spec.ts
```diff
@@ -169,20 +181,63 @@ describe("MicrosoftTeams", () => {
...
+          id: getContextMessage.id,
+          args: [
+            {
+              groupId: "someMaliciousValue"
+            }
+          ]
+        } as MessageResponse
+      } as MessageEvent);
+
+      expect(callbackCalled).toBe(true);
```
aosolis: What resets callbackCalled to false for each item in supportedDomains? I think this ends up checking only that the callback was called for at least *one* of the domains, not each one.

test/MicrosoftTeams.spec.ts
```diff
@@ -169,20 +181,63 @@ describe("MicrosoftTeams", () => {
...
+          id: getContextMessage.id,
+          args: [
+            {
+              groupId: "someMaliciousValue"
+            }
+          ]
+        } as MessageResponse
+      } as MessageEvent);
+
+      expect(callbackCalled).toBe(false);
```
aosolis: Similarly here it checks that the callback was never called (which is correct), but should it fail and a callback does end up getting called, we don't keep enough information to identify which one was missed.

src/MicrosoftTeams.ts
```diff
@@ -21,15 +31,45 @@ namespace microsoftTeams {
...
   const validOrigins = [
     "https://teams.microsoft.com",
     "https://teams.microsoft.us",
     "https://int.teams.microsoft.com",
     "https://devspaces.skype.com",
     "https://ssauth.skype.com",
-    "http://dev.local" // local development
+    "http://dev.local", // local development
+    "https://msft.spoppe.com",
+    "*.sharepoint.com",
```
aosolis: The mix of explicit origins and wildcarded hostnames makes this hard to maintain-- someone will be tempted to add an entry like `https://*.teams.microsoft.com` which will not work as expected.

test/MicrosoftTeams.spec.ts
```diff
@@ -169,20 +181,63 @@ describe("MicrosoftTeams", () => {
...
+  it("should allow messages from supported domains", () => {
+    initializeWithContext("content");
+    const supportedDomains = [
+      "https://teams.microsoft.com",
+      "https://teams.microsoft.us",
+      "https://int.teams.microsoft.com",
+      "https://devspaces.skype.com",
+      "http://dev.local",
+      "https://microsoft.sharepoint.com",
+      "https://a.b.sharepoint.com",
```
aosolis: Do these pass? I think "*" in the pattern matches *exactly* one subdomain.

test/MicrosoftTeams.spec.ts
sachingupta: thanks i am fixing Unit tests logic

test/MicrosoftTeams.spec.ts
sachingupta: Looks like the way i have setup Unit test is wrong. i am working on fixing it.

src/MicrosoftTeams.ts
sachingupta: updated with domain validation logic in extensiontabService

PR #114, opened 2018-07-04
HostClientType exposed

harjaMsft:
Description : 
1. Added 2 new Client Types : android, ios
2. Exposed HostClientType

Test Done :
Tested on sample website with android.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -42,7 +42,9 @@ namespace microsoftTeams {
 
   const hostClientTypes = {
     desktop: "desktop",
-    web: "web"
+    web: "web",
+    android: "android",
+    ios: "ios"
   };
 
   interface MessageRequest {
@@ -453,6 +455,16 @@ namespace microsoftTeams {
       hostClientType = null;
       isFramelessWindow = false;
     };
+  }
+
+  /**
+   * Retrieves client type of the host.
+   * Possible values are : desktop, web, android, ios
+   */
+  export function getHostClientType(): string {
+    ensureInitialized();
+
+    return hostClientType;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.ts
```diff
@@ -455,6 +457,16 @@ namespace microsoftTeams {
     };
   }
 
+  /**
+   * Retrieves client type of the host.
+   * Possible values are : desktop, web, android, ios
+   */
+  export function getHostClientType(): string {
```
ydogandjiev: I think Vinay and I discussed adding this property to the context. Any particular reason you chose to add a dedicated getter for this? What happens if the initialize response hasn’t come back yet when this method is called? How does a dev know when it’s OK to call the API?

src/MicrosoftTeams.ts
```diff
@@ -42,7 +42,9 @@ namespace microsoftTeams {
 
   const hostClientTypes = {
```
ydogandjiev: Might be better to convert this to an enum and export it

src/MicrosoftTeams.ts
```diff
@@ -1531,6 +1538,11 @@ namespace microsoftTeams {
      * Apps should use this as a signal to prevent any changes to content associated with archived teams.
      */
     isTeamArchived?: boolean;
+
+    /**
+     * The type of the host client. Possible values are : android, ios, web, desktop
+     */
+    hostClientType?: string;
```
ydogandjiev: It's probably best to use an enum for the hostClientType instead of a simple string

src/MicrosoftTeams.ts
```diff
@@ -463,7 +465,12 @@ namespace microsoftTeams {
     ensureInitialized();
 
     let messageId = sendMessageRequest(parentWindow, "getContext");
-    callbacks[messageId] = callback;
+    callbacks[messageId] = (context: Context) => {
+      if (hostClientType != null) {
```
ydogandjiev: This is a nice stop-gap but eventually it might be best if the Teams client just added this to the context

harjaMsft: Mobile Teams client has added in context too.
This was done to support hostClientType on web/desktop too. In case, web/desktop can also add in the context we don't need this.

harjaMsft: done

PR #115, opened 2018-07-11
Add configurationUrl field to connector definition

ydogandjiev:
This change also auto-formats the schema using Prettier.

### DIFF
--- a/src/MicrosoftTeams.schema.json
+++ b/src/MicrosoftTeams.schema.json
@@ -1,460 +1,441 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
-    "type": "object",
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.schema.json
```diff
@@ -1,460 +1,441 @@
...
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "connectorId": {
+            "type": "string",
+            "description":
+              "A unique identifier for the connector which matches its ID in the Connectors Developer Portal.",
+            "maxLength": 64
+          },
+          "configurationUrl": {
```
ydogandjiev: This is the only new field

PR #118, opened 2018-07-24
Adding app sideload api for 1P

goelashish7:
@ydogandjiev

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -688,13 +688,13 @@ namespace microsoftTeams {
    * @private
    * Hide from docs.
    * ------
-   * Sideload an App-manifest directly to a team.
+   * Upload a custom App manifest directly to both team and personal scopes.
    * This method works just for the first party Apps.
    */
-  export function sideLoadAppManifestPackage(manifestBlob: Blob): void {
+  export function uploadACustomApp(manifestBlob: Blob): void {
     ensureInitialized();
 
-    let messageId = sendMessageRequest(parentWindow, "sideLoadApp", [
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.ts
```diff
@@ -684,6 +684,26 @@ namespace microsoftTeams {
     sendMessageRequest(parentWindow, "openFilePreview", params);
   }
 
+  /**
+   * @private
+   * Hide from docs.
+   * ------
+   * Sideload an App-manifest directly to a team.
```
ydogandjiev: Wouldn't this allow for apps to be sideloaded both personally or into a team? If so, please update the comment.

src/MicrosoftTeams.ts
```diff
@@ -684,6 +684,26 @@ namespace microsoftTeams {
...
+   * This method works just for the first party Apps.
+   */
+  export function sideLoadAppManifestPackage(manifestBlob: Blob): void {
```
ydogandjiev: We don't really use the term "sideloading" in any of our public-facing documentation. I think it might be better if we called this API something like "createCustomApp" or "uploadCustomApp".

src/MicrosoftTeams.ts
```diff
@@ -684,6 +684,26 @@ namespace microsoftTeams {
...
+    ensureInitialized();
+
+    let messageId = sendMessageRequest(parentWindow, "sideLoadApp", [
```
ydogandjiev: NIT: let --> const

src/MicrosoftTeams.ts
ydogandjiev: @aosolis, @MattSFT, thoughts?

PR #124, opened 2018-08-03
Updates to taskInfo structure

billbliss:
1. Added completionBotId
2. Added fetchTask to match spec
3. Changed taskInfo.card to be an any type because it's a JSON object, not a string.
4. Added embedCode

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -1987,7 +1987,17 @@ namespace microsoftTeams {
     /**
      * JSON defining an adaptive card.
      */
-    card?: string;
+    card?: any;
+
+    /**
+     * HTML for an embeddable <iframe>.
+     */
+    embedCode?: string;
+
+    /**
+     * An invoke message is sent to the bot to fetch the URL or Adaptive Card to appear in the task module.
+     */
+    fetchTask?: boolean;
 
     /**
      * The requested height of the webview/iframe.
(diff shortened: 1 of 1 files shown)


billbliss:
Before accepting this, let's have a discussion about taskInfo. Now there are four mutually exclusive types of taskInfo invocations - url, card, fetchTask, and embedCode.

We should talk about an approach where we do 

type: "url" | "card" | "fetchTask" | "embedCode"
value?: any // would have the URL, the card body, or the embedCode

Thoughts?

src/MicrosoftTeams.ts
```diff
@@ -1987,7 +1987,17 @@ namespace microsoftTeams {
     /**
      * JSON defining an adaptive card.
      */
-    card?: string;
+    card?: any;
+
+    /**
```
ydogandjiev: This is feature creep and not in scope of this feature. Please remove. This needs to go thorough a thorough design + security review before we even attempt to build it.

src/MicrosoftTeams.ts
```diff
@@ -1987,7 +1987,17 @@ namespace microsoftTeams {
...
+     * HTML for an embeddable <iframe>.
+     */
+    embedCode?: string;
```
ydogandjiev: This is feature creep and not in scope of this feature. Please remove. This needs to go thorough a thorough design + security review before we even attempt to build it.

PR #127, opened 2018-08-06
Fix HostClientType enum and prepare to release v1.3.3 of the SDK

ydogandjiev:
The HostClientType must be exported as a const enum so that it gets compiled down to vanilla strings.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -3,7 +3,10 @@ declare interface String {
 }
 
 if (!String.prototype.startsWith) {
-  String.prototype.startsWith = function (search: string, pos?: number): boolean {
+  String.prototype.startsWith = function(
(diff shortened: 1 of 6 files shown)


src/MicrosoftTeams.ts
```diff
@@ -76,7 +80,7 @@ namespace microsoftTeams {
     remove: "remove"
   };
 
-  export enum HostClientTypes {
+  export const enum HostClientType {
```
sachingupta: If app developer who is using HostClientTypes need to update teams js version then they need to update their code. should we keep name same ?

ydogandjiev: Yeah, but we just added this with v1.3 so I think it's still safe to fix it. I would rather do that now before too many folks have taken a dependency on the old name.

PR #129, opened 2018-08-17
Rename completeTask to submitTask

ydogandjiev:
This change renames completeTask to submitTask and allows developers to pass either a single appId or an array of appIds to the function.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -2035,11 +2035,11 @@ namespace microsoftTeams {
     /**
      * Allows an app to open the task module.
      * @param taskInfo An object containing the parameters of the task module
-     * @param completionHandler Handler to call when the task module is completed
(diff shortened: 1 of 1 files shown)


src/MicrosoftTeams.ts
```diff
@@ -2066,34 +2067,35 @@ namespace microsoftTeams {
...
      */
-    export function completeTask(
+    export function submitTask(
       result?: string | object,
-      appIds?: string[]
+      appIds?: string | string[]
     ): void {
-      // Ensure that the tab content is initialized
-      ensureInitialized(frameContexts.content);
+      ensureInitialized(frameContexts.task);
```
mevinay: This change of using "task" frameContext will cause another incompatibility on mobile clients. Can we have this as ensureInitialized(frameContexts.content, frameContexts.task); till we don't have "task" frameContext support on mobile clients?

PR #138, opened 2018-09-21
adding print capability to client sdk, App can register custom print handler.

sachingupta:
adding print capability to client sdk, 
1. By default we it will be disabled App need to call microsoftTeams.enablePrintCapability()

2. microsoftTeams.printHandler: 
App can register custom print handler otherwise by default it will call window.print.
It support
2.1. Ctrl+P 
2.2. Cmd+P (mac)
and 2.3. by calling explicit microsoftTeams.printHandler

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -610,27 +610,12 @@ namespace microsoftTeams {
   }
 
   /**
-   * Registers a handler for print.
    * default print handler
    */
   export function print(): void {
     if (printCapabilityEnabled) {
-      if (customPrintHandler) {
-        customPrintHandler();
-      } else {
-        window.print();
-      }
+      window.print();
     }
-  }
-
-  /**
-   * Registers a custom handler for print.
-   * @param handler The handler to invoke when printHandler is called.
-   */
-  export function registerCustomPrintHandler(handler: () => void): void {
-    ensureInitialized();
-
-    customPrintHandler = handler;
   }
 
   /**
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -39,18 +39,6 @@ interface Window {
 }
 
 /**
- * adding ctrl+P and cmd+P handler
- */
-document.addEventListener("keydown", function(event: KeyboardEvent): void {
-  if ((event.ctrlKey || event.metaKey) && event.keyCode === 80) {
-    microsoftTeams.print();
-    event.cancelBubble = true;
-    event.preventDefault();
-    event.stopImmediatePropagation();
-  }
-});
-
-/**
  * This is the root namespace for the JavaScript SDK.
  */
 namespace microsoftTeams {
@@ -606,7 +594,19 @@ namespace microsoftTeams {
    * enable print capability
    */
   export function enablePrintCapability(): void {
+    ensureInitialized();
     printCapabilityEnabled = true;
+    /**
+     * adding ctrl+P and cmd+P handler
+     */
+    document.addEventListener("keydown", function(event: KeyboardEvent): void {
+      if ((event.ctrlKey || event.metaKey) && event.keyCode === 80) {
+        microsoftTeams.print();
+        event.cancelBubble = true;
+        event.preventDefault();
+        event.stopImmediatePropagation();
+      }
+    });
   }
 
   /**
@@ -615,7 +615,6 @@ namespace microsoftTeams {
    */
   export function print(): void {
     if (printCapabilityEnabled) {
-      ensureInitialized();
       if (customPrintHandler) {
         customPrintHandler();
       } else {
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -40,7 +40,7 @@ interface Window {
 
 document.addEventListener("keydown", function(event: KeyboardEvent): void {
   if ((event.ctrlKey || event.metaKey) && event.keyCode === 80) {
-    print();
+    microsoftTeams.printHandler();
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -38,6 +38,18 @@ interface Window {
   onNativeMessage(evt: MessageEvent): void;
 }
 
+/**
+ * adding ctrl+P and cmd+P handler
+ */
+document.addEventListener("keydown", function(event: KeyboardEvent): void {
```
ydogandjiev: @sachingupta, I think it would be better to only register this handler when printing is enabled both for performance reasons and also to avoid interfering with any similar logic the developer might already have on their page.

src/MicrosoftTeams.ts
```diff
@@ -588,6 +602,38 @@ namespace microsoftTeams {
...
+    printCapabilityEnabled = true;
+  }
+
+  /**
+   * Registers a handler for print.
+   * default print handler
+   */
+  export function print(): void {
+    if (printCapabilityEnabled) {
+      ensureInitialized();
```
ydogandjiev: Not sure if this is needed. If we only register the print handler when enablePrintCapability is called then all we need to do is call ensureInitialized() in there.

src/MicrosoftTeams.ts
sachingupta: sounds good

src/MicrosoftTeams.ts
sachingupta: thanks done.

src/MicrosoftTeams.ts
```diff
@@ -588,6 +590,49 @@ namespace microsoftTeams {
...
+        microsoftTeams.print();
+        event.cancelBubble = true;
+        event.preventDefault();
+        event.stopImmediatePropagation();
+      }
+    });
+  }
+
+  /**
+   * Registers a handler for print.
```
aosolis: Is this comment correct? It looks like this actually calls (not registers) the custom print handler or window.print.

src/MicrosoftTeams.ts
```diff
@@ -588,6 +590,49 @@ namespace microsoftTeams {
...
+   * default print handler
+   */
+  export function print(): void {
+    if (printCapabilityEnabled) {
```
aosolis: Is printCapabilityEnabled meant to gate the entire printing feature, or just so we don't register ctrl-P/cmd-P unnecessarily? Not sure why we need to check for printCapabilityEnabled in print(), when the page can easily bypass it by calling window.print() (or the custom handler) directly.

src/MicrosoftTeams.ts
```diff
@@ -588,6 +590,49 @@ namespace microsoftTeams {
...
+        window.print();
+      }
+    }
+  }
+
+  /**
+   * Registers a custom handler for print.
+   * @param handler The handler to invoke when printHandler is called.
+   */
+  export function registerCustomPrintHandler(handler: () => void): void {
```
aosolis: What might an app do in a custom print handler?

src/MicrosoftTeams.ts
sachingupta: thanks. will update

src/MicrosoftTeams.ts
sachingupta: I was thinking to make it backward compatible so that to get default Print support app need to call enablePrintCapability. I am adding check in print function to make sure app developer call enablePrintCapability before using it either explicit call or Ctr+P

src/MicrosoftTeams.ts
sachingupta: as discussed offline will remove this support as of now

src/MicrosoftTeams.ts
```diff
@@ -588,6 +588,31 @@ namespace microsoftTeams {
     };
   }
 
+  /**
+   * enable print capability
+   */
```
ydogandjiev: NIT: Capitalize the comment and make it a little more detailed.

src/MicrosoftTeams.ts
```diff
@@ -588,6 +588,31 @@ namespace microsoftTeams {
...
+  export function enablePrintCapability(): void {
+    ensureInitialized();
+    /**
```
ydogandjiev: NIT: Generally the `/** ... */` multi-line comments only appear at the top of functions. In here you can just use a single line comment `// ...`

src/MicrosoftTeams.ts
```diff
@@ -588,6 +588,31 @@ namespace microsoftTeams {
...
+        event.preventDefault();
+        event.stopImmediatePropagation();
+      }
+    });
+  }
+
+  /**
+   * default print handler
+   */
+  export function print(): void {
```
ydogandjiev: Does this need to get exported? Seems like something we only use internally.

src/MicrosoftTeams.ts
```diff
@@ -588,6 +588,31 @@ namespace microsoftTeams {
...
 
+  /**
+   * enable print capability
+   */
+  export function enablePrintCapability(): void {
+    ensureInitialized();
+    /**
+     * adding ctrl+P and cmd+P handler
+     */
+    document.addEventListener("keydown", function(event: KeyboardEvent): void {
```
ydogandjiev: If a developer makes the mistake of calling `enablePrintCapability` twice would we register two of these handlers? Might be a good idea to somehow protect against that.

ydogandjiev: NIT: Consider using a lambda instead of function for the event listener

PR #139, opened 2018-09-25
Adding GetChatMembers API

XinGao:
Adding a new GetChatMembers API to return all members in the chat.

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -1238,13 +1238,13 @@ namespace microsoftTeams {
         link.href,
         "_blank",
         "toolbar=no, location=yes, status=no, menubar=no, scrollbars=yes, top=" +
-        top +
-        ", left=" +
-        left +
-        ", width=" +
-        width +
-        ", height=" +
-        height
+          top +
+          ", left=" +
+          left +
+          ", width=" +
+          width +
+          ", height=" +
+          height
       );
       if (childWindow) {
         // Start monitoring the authentication window so that we can detect if it gets closed before the flow completes
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -2149,7 +2149,7 @@ namespace microsoftTeams {
    * ------
    * Allows an app to retrieve information of all chat members
    * Because a malicious party run your content in a browser, this value should
-   * be used only as a hint as to who the members are and never as proof of identity.
+   * be used only as a hint as to who the members are and never as proof of membership.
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -2119,4 +2119,43 @@ namespace microsoftTeams {
...
+    members: ChatMember[];
+  }
+
+  /**
+   * @private
+   * Hide from docs
+   * --------
+   * Information about a chat member
+   */
+  export interface ChatMember {
```
ydogandjiev: As mentioned in the WebClient PR, I think it might be good to name this something more generic (e.g. ThreadMember or simply User).

src/MicrosoftTeams.ts
```diff
@@ -2119,4 +2119,43 @@ namespace microsoftTeams {
...
+     * The user name of the user principal.
+     */
+    upn: string;
+  }
+
+  /**
+   * @private
+   * Hide from docs
+   * ------
+   * Allows an app to retrieve information of all chat members
```
aosolis: This value can be spoofed by a malicious attacker, right? I think it needs the same dire warnings as with the identity hints in tab context: apps should not make security decisions based on this value without considering the ramifications of the spoofing threat.

src/MicrosoftTeams.ts
```diff
@@ -2119,4 +2119,43 @@ namespace microsoftTeams {
...
+
+  /**
+   * @private
+   * Hide from docs
+   * --------
+   * Information about a chat member
+   */
+  export interface ChatMember {
+    /**
+     * The user name of the user principal.
```
aosolis: Let's change this to "the member's user principal name in the current tenant"
"User name" could be confusing, because in a UPN like xxx@example.com, technically "xxx" is the "user name".

src/MicrosoftTeams.ts
XinGao: Added warning

src/MicrosoftTeams.ts
XinGao: Updated

src/MicrosoftTeams.ts
XinGao: Changed to ThreadMember

unckleg:
@ydogandjiev @aosolis @sachingupta @XinGao 
- Is there a way to get values from this function in public API?
- What will be the approach?  

Currently, I'm able to get chat members values by bloating the codebase with overriding the library itself so Is there any clean way of doing this. 
Thanks!

PR #144, opened 2018-10-10
Saching/release 1.3.7

sachingupta:
Adding these change to the code of branch release-1.3.6 to prepare release-1.3.7


ydogandjiev:
@sachingupta, we usually update the version that the SDK reports from initialize too. Seems like that's missing from the PR.

PR #147, opened 2018-10-17
Adding webpack module build support

sachingupta:
This change will add webpack support in teams-js code to support
modules (Currently we have one js file where we add all interfaces and code).

Followup change will be to break existing JS file into different module files.

To test please try
Option 1. npm package
https://www.npmjs.com/package/@microsoft/teams-js/v/1.4.0-beta.3.9
import { microsoftTeams } from '@microsoft/teams-js'

2. Option 2: as standalone JS
 <script src="https://unpkg.com/@microsoft/teams-js@1.4.0-beta.3.9/dist/MicrosoftTeams.min.js"></script>


package.json
```diff
@@ -1,47 +1,50 @@
 {
   "name": "@microsoft/teams-js",
   "author": "Microsoft Teams",
-  "version": "1.3.6",
+  "version": "1.4.0-beta.3.9",
```
sachingupta: will revert this

src/MicrosoftTeams.ts
```diff
@@ -485,11 +488,11 @@ namespace microsoftTeams {
   let initializeCalled = false;
 
   let isFramelessWindow = false;
-  let currentWindow: Window;
-  let parentWindow: Window;
+  let currentWindow: any | any;
```
ydogandjiev: Shouldn't this be `Window | any`?

src/MicrosoftTeams.ts
```diff
@@ -485,11 +488,11 @@ namespace microsoftTeams {
...
+  let parentWindow: Window | any;
   let parentOrigin: string;
   let parentMessageQueue: MessageRequest[] = [];
-  let childWindow: Window;
+  let childWindow: Window | any;
```
ydogandjiev: Why do you need to change the type here (and above)? In what scenario does this need to be `any`?

src/MicrosoftTeams.ts
```diff
@@ -830,6 +838,29 @@ namespace microsoftTeams {
     sendMessageRequest(parentWindow, "openFilePreview", params);
   }
 
+  export interface ShowNotificationParameters {
+    message: string;
+    isDownloadComplete: boolean;
```
ydogandjiev: I assume you will change this to an enum, right?

src/MicrosoftTeams.ts
```diff
@@ -830,6 +838,29 @@ namespace microsoftTeams {
...
+  }
+
+  /**
+   * @private
+   * Hide from docs.
+   * ------
+   * download file.
+   * @param file The file to download.
```
ydogandjiev: Update the comments once you make this API more generic.

tsconfig.json
```diff
@@ -1,15 +1,20 @@
 {
   "compilerOptions": {
+    "outDir": "./dist",
+    "declarationDir": "./dts",
+    "target": "es6",
```
ydogandjiev: Hmm... Shouldn't the target be es5? Some of the browsers we need to support can't handle es6 yet.

src/MicrosoftTeams.ts
sachingupta: yes

src/MicrosoftTeams.ts
sachingupta: yes

src/MicrosoftTeams.ts
sachingupta: Looks like we are adding new Window interface that override WIndow interface. will fix it

PR #155, opened 2018-11-13
Saching/adding deep link api

sachingupta:
Changes include. 1. adding deepLink API 
2. refactoring few Interfaces to separate module file (will do more in separate CR)

### DIFF
--- a/src/MicrosoftTeams.interface.ts
+++ b/src/MicrosoftTeams.interface.ts
@@ -7,7 +7,3 @@ export interface ShowNotificationParameters {
   message: string;
   notificationType: NotificationTypes;
 }
-
-export interface ExecuteDeepLinkParameters {
-  deepLink: string;
-}
--- a/src/MicrosoftTeams.interface.ts
+++ b/src/MicrosoftTeams.interface.ts
@@ -1,9 +1,0 @@
-export const enum NotificationTypes {
-  fileDownloadStart = "fileDownloadStart",
-  fileDownloadComplete = "fileDownloadComplete"
-}
-
-export interface ShowNotificationParameters {
-  message: string;
-  notificationType: NotificationTypes;
-}
--- a/src/MicrosoftTeams.interface.ts
+++ b/src/MicrosoftTeams.interface.ts
@@ -1,0 +1,13 @@
+export const enum NotificationTypes {
+  fileDownloadStart = "fileDownloadStart",
+  fileDownloadComplete = "fileDownloadComplete"
+}
+
+export interface ShowNotificationParameters {
+  message: string;
+  notificationType: NotificationTypes;
+}
+
+export interface ExecuteDeepLinkParameters {
+  deepLink: string;
+}
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -861,13 +861,11 @@ export function showNotification(
  * Hide from docs.
  * ------
  * execute deep link API.
- * @param deepLinkParameters deep link parameters.
+ * @param deepLink deep link.
  */
-export function executeDeepLink(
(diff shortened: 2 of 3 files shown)


src/MicrosoftTeams.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export interface ShowNotificationParameters {
+  message: string;
+  notificationType: NotificationTypes;
+}
+
+export interface ExecuteDeepLinkParameters {
+  deepLink: string;
+}
```
ydogandjiev: Why only move these as opposed to all the enums/interfaces from MicrosoftTeams.ts?

src/MicrosoftTeams.ts
```diff
@@ -861,6 +856,21 @@ export function showNotification(
...
 
+/**
+ * @private
+ * Hide from docs.
+ * ------
+ * execute deep link API.
+ * @param deepLink deep link.
+ */
+export function executeDeepLink(
+  deepLinkParameters: ExecuteDeepLinkParameters
```
ydogandjiev: Do we really need an ExecuteDeepLinkParameters object? Couldn't we just have the developer pass the URL?

src/MicrosoftTeams.ts
```diff
@@ -861,6 +856,21 @@ export function showNotification(
   sendMessageRequest(parentWindow, "showNotification", params);
 }
 
+/**
+ * @private
+ * Hide from docs.
+ * ------
+ * execute deep link API.
+ * @param deepLink deep link.
```
ydogandjiev: The comment doesn't match the actual parameters of the API.

src/MicrosoftTeams.ts
```diff
@@ -861,6 +856,21 @@ export function showNotification(
   sendMessageRequest(parentWindow, "showNotification", params);
 }
 
+/**
+ * @private
+ * Hide from docs.
```
ydogandjiev: Why mark this as @private? I think this is an API we actually want made available to all apps.

src/MicrosoftTeams.ts
```diff
@@ -861,6 +856,21 @@ export function showNotification(
...
+ * ------
+ * execute deep link API.
+ * @param deepLink deep link.
+ */
+export function executeDeepLink(
+  deepLinkParameters: ExecuteDeepLinkParameters
+): void {
+  ensureInitialized(frameContexts.content);
+  const params = [deepLinkParameters.deepLink];
+  sendMessageRequest(parentWindow, "executeDeepLink", params);
```
ydogandjiev: Seems like you are missing the error handling for this.

src/MicrosoftTeams.interface.ts
sachingupta: Will move other in separate PR

src/MicrosoftTeams.ts
sachingupta: I am planning to keep it private till this is available in teams Public Ring

src/MicrosoftTeams.ts
sachingupta: will do that now.

src/MicrosoftTeams.ts
```diff
@@ -861,6 +861,24 @@ export function showNotification(
...
+/**
+ * @private
+ * Hide from docs.
+ * ------
+ * execute deep link API.
+ * @param deepLink deep link.
+ */
+export function executeDeepLink(deepLink: string): void {
+  ensureInitialized(frameContexts.content);
+  const params = [deepLink];
```
ydogandjiev: NIT: You could just put this inline below rather than using an intermediate `params` variable.

PR #160, opened 2018-11-15
Add tasks.updateTask support

cjsheets:
This change adds support for tasks.updateTask commands. The only TaskInfo properties supported will be width and height.


src/MicrosoftTeams.ts
```diff
@@ -2165,6 +2165,24 @@ export namespace tasks {
...
   }
 
+  /**
+   * Update height/width task info properties.
+   * @param taskInfo An object containing width and height properties
+   * @param appIds Helps to validate that the call originates from the same appId as the one that invoked the task module
+   */
+  export function updateTask(
+    taskInfo?: object,
+    appIds?: string | string[]
```
ydogandjiev: I don't think the appIds are necessary here. They are only needed in submitTask to secure the channel we use to send data to the app's bot. In this case there is no data flowing back to the bot and the origin of these calls is already secured via the validDomains.

cjsheets: Gotcha, that's good to know. I'll adjust the PR accordingly.

src/MicrosoftTeams.ts
```diff
@@ -2168,18 +2168,11 @@ export namespace tasks {
   /**
    * Update height/width task info properties.
    * @param taskInfo An object containing width and height properties
-   * @param appIds Helps to validate that the call originates from the same appId as the one that invoked the task module
    */
-  export function updateTask(
-    taskInfo?: object,
-    appIds?: string | string[]
-  ): void {
+  export function updateTask(taskInfo?: object): void {
```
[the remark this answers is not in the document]
cjsheets: That's a good idea. A check here would be helpful. I improved the interface and added a validation step.

PR #162, opened 2018-12-06
Added ios and android hostClientTypes to authenticate() whitelist

### LINKED ISSUE #163: microsoftTeams.authenticate() does not propagate the event to mobile clients
`microsoftTeams.authenticate()` is currently whitelisted to allow only the `desktop` client to pass the event to the Teams Shell. For other clients (web, ios, android), it triggers `window.open()`. Mobile will now support `microsoftTeams.authenticate()` and hence will need the function to propagate to the Host.

msmoosa:
**Description**
`microsoftTeams.authenticate()` is currently whitelisted to allow only the `desktop` client to pass the event to the Teams Shell. For other clients (web, ios, android), it triggers `window.open()`. Mobile will now support `microsoftTeams.authenticate()` and hence will need the function to propagate to the Host.

**Fix**
Added `ios` and `android` clients to the whitelist.

Fixes #163


src/MicrosoftTeams.ts
```diff
@@ -1200,7 +1200,9 @@ export namespace authentication {
       frameContexts.task
     );
 
-    if (hostClientType === HostClientType.desktop) {
+    if (hostClientType === HostClientType.desktop
```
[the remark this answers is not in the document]
msmoosa: Or it could be 
!!hostClientType && hostClientType !== HostClientType.web

test/MicrosoftTeams.spec.ts
```diff
@@ -1138,6 +1138,46 @@ describe("MicrosoftTeams", () => {
       expect(message.args[2]).toBe(authenticationParams.height);
     });
 
+    it("should successfully pop up the auth window in the android client", () => {
```
ydogandjiev: NIT: You can implement the android, ios, and desktop test cases using a single code block inside of a loop, i.e.:
```
["android", "ios", "desktop].forEach(hostClientType => {
  it(`should successfully handle auth success in the ${hostClientType} client`, () => {
    initializeWithContext("content", hostClientType);
    ...
  });
});
```

src/MicrosoftTeams.ts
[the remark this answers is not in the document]
ydogandjiev: I like the way you have it. I think it's better to be explicit here since the else case uses vanilla window.open and can work in any browser environment whereas the android/ios/desktop case requires the host to handle most of the auth flow logic and would fail in any other environment.

test/MicrosoftTeams.spec.ts
msmoosa: Nice !

msmoosa: Done.
Should i also do it for the auth success and auth failure cases?

msmoosa: I've added them for auth success and failure, since the tests do fail if the client types are removed

test/MicrosoftTeams.spec.ts
```diff
@@ -1118,70 +1118,77 @@ describe("MicrosoftTeams", () => {
...
-      expect(successResult).toBe("someResult");
-      expect(failureReason).toBeUndefined();
+        expect(successResult).toBe("someResult");
+        expect(failureReason).toBeUndefined();
+      });
     });
 
-    it("should successfully handle auth failure in the desktop client", () => {
-      initializeWithContext("content", "desktop");
+    ["android", "ios", "desktop"].forEach(hostClientType => {
```
ydogandjiev: NIT: You don’t need to repeat the loops; just move all of these test cases into the first loop above.

msmoosa: true... was just thinking that

PR #170, opened 2019-01-18
- added GetContextAsync with a Promise

baywet:
- fixed compiler options to reflect needed es2015 and dom

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -629,6 +629,21 @@ export function getContext(callback: (context: Context) => void): void {
 }
 
 /**
+ * Retrieves the current context the frame is running in and returns it with a promise.
+ */
+export function getContextAsync(): Promise<Context> {
+  return new Promise<Context>((resolve, reject) => {
+    try {
+      getContext(context => {
+        resolve(context);
+      });
+    } catch (error) {
+      reject(error);
+    }
+  });
+}
+
+/**
  * Registers a handler for theme changes.
  * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
  * @param handler The handler to invoke when the user changes their theme.
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -629,21 +629,6 @@ export function getContext(callback: (context: Context) => void): void {
 }
 
 /**
- * Retrieves the current context the frame is running in and returns it with a promise.
- */
-export function getContextAsync(): Promise<Context> {
-  return new Promise<Context>((resolve, reject) => {
-    try {
-      getContext(context => {
-        resolve(context);
-      });
-    } catch (error) {
-      reject(error);
-    }
-  });
-}
-
-/**
  * Registers a handler for theme changes.
  * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
  * @param handler The handler to invoke when the user changes their theme.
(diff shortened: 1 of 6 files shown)


baywet:
anyone @sachingupta  @asnow003 @ydogandjiev ?

src/MicrosoftTeams.ts
```diff
@@ -628,6 +628,21 @@ export function getContext(callback: (context: Context) => void): void {
   callbacks[messageId] = callback;
 }
 
+/**
```
ydogandjiev: We've considered creating a promise-based version of our library but it will probably be modeled as a separate wrapper that adds promises for all async APIs. Have you considered building something like that instead of just adding this one-off method?

baywet: Do you mean you'd like to do everything at one? Is there any problem with doing it one (or a couple of) method at a time? (besides more PRs to review, which will be smaller so easier to review)
The reason why I'm pushing this one specifically is because we're starting to see SPFx and Teams tab integration and the number 3 of this doc is a lot of "plumbing" code for consumers when they could just return the promise if the Teams SDK had an async method.
https://docs.microsoft.com/en-us/sharepoint/dev/spfx/web-parts/get-started/using-web-part-as-ms-teams-tab#updating-code-to-be-aware-of-the-microsoft-teams-context

Doing everything at once for async methods seems like a lot of work to write, review and merge, although I haven't looked in details, and I don't think I'd be able to take it on my own.
Maybe the first step would be to inventory which methods need to be ported async as issues (display of work principle), create a project and either assign people or let them self assign, bit by bit. That looks easier from an external perspective to contribute to rather than a huge chunk.

baywet: any thoughts?

ydogandjiev: I don't think adding promised-based APIs to the existing library is the way to go here. For one, it's inconsistent since most of the methods use callbacks instead of promises. For two, this SDK once compiled is supposed to work in all browsers and Promises are only available in ES6 which is not universally supported yet (e.g. IE11 doesn't support them). We need an alternative approach here that doesn't suffer from these issues.

baywet: What do you think of a MicrosoftTeamsAsync file/class instead that takes a dependency on MicrosoftTeams and is basically a bunch of wrappers?

ydogandjiev: Yeah, I think that approach would work better. @sachingupta, what do you think?

sachingupta: Agree either we should do it once or we should create a MicrosoftTeamsAsync file and that will have wrapper and take dependency on MicrosoftTeams.

baywet: I just updated the pull request to reflect our discussions, but I only implemented one method to be sure we're on the same page before moving to anything else.
The only thing I'm not a big fan off so far is the seperation of the imports (no way to extend the existing namespace definition as far as I know as it is defined today)

baywet: any thoughts on my update @ydogandjiev @sachingupta ?

lokimeyburg: @baywet we are discussing using Promises in Teams and @sachingupta wrote a small pull request so we can get a better idea of the pros/cons of moving forward with this approach: https://github.com/OfficeDev/microsoft-teams-library-js/pull/200

baywet: thanks for getting back to me!
So do you guys want me to abandon this one? port some changes to this new PR?

sachingupta: @baywet You can abandon this PR. We are still comparing two approach. Hopefully we will finalize decision and changes end of this week then i will also add you in the PR.

baywet:
replaced by #200

PR #175, opened 2019-02-12
Saching/inline configuration handler

sachingupta:
Tab can register onSettingsHandler and when user click on settings to edit tab teams client will notify tab instead of opening tab config dialog

### DIFF
--- a/src/MicrosoftTeams.ts
+++ b/src/MicrosoftTeams.ts
@@ -564,7 +564,6 @@ export function initialize(hostWindow: any = window): void {
       registerFullScreenHandler(null);
       registerBackButtonHandler(null);
       registerBeforeUnloadHandler(null);
-      registerChangeSettingsHandler(null);
     }
 
     if (frameContext === frameContexts.settings) {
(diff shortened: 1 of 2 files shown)


src/MicrosoftTeams.ts
```diff
@@ -741,6 +745,26 @@ function handleBeforeUnload(): void {
   }
 }
 
+/**
+ * Registers a handler for when the user reconfigurated tab
+ * @param handler The handler to invoke when the user click on Settings.
+ */
+export function registerOnSettingsChangeHandler(
```
ydogandjiev: I think this name is slightly confusing since in my mind it implies that some settings have changed and this handler is the way that the tab can get notified. I wonder if a better name would be `registerChangeSettingsHandler` or `registerSettingsButtonClickedHandler` or something like that.

ydogandjiev:
@sachingupta, please consider adding some unit tests for the new functionality.

src/MicrosoftTeams.ts
sachingupta: I like `registerChangeSettingsHandler` better

PR #182, opened 2019-02-27
Added support for conversational subEnities inside tabs

debajyoti2510:
We are adding 3 new functions to support conversational subEntities inside tabs.

- startConversation(): This will be used to start a new conversation. The conversation Id will be sent back to the developer.

- showConversation(): This will be used to show an existing conversation in the right pane. The subEntity Id and the conversation Id will need to be passed in.

- closeConversation(): This will be used to close an existing conversation in the right pane. The subEntity Id and the conversation Id will need to be passed in.


src/MicrosoftTeams.ts
```diff
@@ -2329,3 +2329,90 @@ export function getChatMembers(
...
+   * @param conversationId The Id of the conversation
+   * @param conversationalSubEntity Callback containing if the tab pane was closed
+   */
+  export function showConversation(
+    subEntityId: string,
+    title: string,
+    conversationId: string,
+    conversationalSubEntity: ConversationalSubEntity
+  ): void {
+    const messageId = sendMessageRequest(parentWindow, "showConversation", [
```
asnow003: What happens when two or more showConversations are called executed without a close?  Just thinking developers might put code in the onCloseConversation and are depending on that being called if a new conversation it opened.  Would we also need to provide the conversationId and subEntityId on the onClose callback also?

asnow003: Just thinking of the example where a user clicks the "X" button and we fire the onCloseConversation.  Maybe in my app I want to do some visual thing like change an icon state or even do some logging.  So maybe having that information in the onClose would help?

src/MicrosoftTeams.ts
```diff
@@ -2329,3 +2329,90 @@ export function getChatMembers(
...
+  /**
+   * @private
+   * Hide from docs
+   * --------------
+   * Allows the user to start a conversation with each subentity inside a tab
+   * @param subEntityId The Id of the subEntity where the conversation is taking place
+   * @param title The title of the conversation
+   * @param conversationalSubEntity Callback containing the conversation Id and if the tab pane was closed
+   */
+  export function startConversation(
```
sachingupta: Can we merge these two function into on lets say startConversation(conversationRequest: ConversationRequest)  where conversational Id will be optional 
1. if provided we will open pane
2. If null we will create one and open pane return conversation Id.


  export interface ConversationRequest {
   subEntityId: string;
   title: string;
  conversationId?: string
    successCallback?: (conversationId: string) => void;

    failureCallback?: (reason: string) => void;
  }

src/MicrosoftTeams.ts
debajyoti2510: - We expect two or more conversations to be just called using showConversations. When we receive consecutive calls, we would treat it as the previous conversation has been closed. And the pane would be opened with the new conversation. This is so that the developer doesnt need to call the close if a user just opens a new conversation.

- The onclose would be just to tell us that the right pane doesnt need to be kept open. Hence subEntityId or conversation Id is not required. Thats why I removed it.

- The developer would know as soon as the  onCloseConversation callback is fired. This is fired only for the two cases - X and opening the tab pane. That should be enough if the developer wants to take some action. I can pass in a string or something. Will run this by Loki

src/MicrosoftTeams.ts
debajyoti2510: We did discuss about that in the dev design review. But we decided to avoid method overloading to reduce confusion of the developer. And in the final version of the feature we might need to have separate functions.

debajyoti2510: Like the idea of adding a reason in the onCloseConversation callback. Added it

sachingupta: Lets use this kind of request object.
export interface ConversationRequest {
   subEntityId: string;
   title: string;
  conversationId?: string
    successCallback/onStartConversation?: (conversationId: string) => void;
    onCloseConversation?: (reason: string) => void;
    failureCallback?: (reason: string) => void;
  }

sachingupta: we can discuss about happening one or two separate function

src/MicrosoftTeams.ts
```diff
@@ -2329,3 +2329,90 @@ export function getChatMembers(
...
+   * Hide from docs
+   * --------------
+   * Allows the user to show the conversation in the right pane
+   * @param subEntityId The Id of the subEntity where the conversation is taking place
+   * @param title The title of the conversation
+   * @param conversationId The Id of the conversation
+   * @param conversationalSubEntity Callback containing if the tab pane was closed
+   */
+  export function showConversation(
+    subEntityId: string,
```
sachingupta: It would be better to combine all parameter into one request interface like ShowConversationRequest

src/MicrosoftTeams.ts
```diff
@@ -2329,3 +2329,90 @@ export function getChatMembers(
...
+   * --------------
+   * Allows the user to start a conversation with each subentity inside a tab
+   * @param subEntityId The Id of the subEntity where the conversation is taking place
+   * @param title The title of the conversation
+   * @param conversationalSubEntity Callback containing the conversation Id and if the tab pane was closed
+   */
+  export function startConversation(
+    subEntityId: string,
+    title: string,
+    conversationalSubEntity: ConversationalSubEntity
```
sachingupta: It would be better to combine all parameter into one request interface like StartConversationRequest that will help if we add any new parameter then caller don't need to worry about.

src/MicrosoftTeams.ts
```diff
@@ -2312,30 +2321,38 @@ export namespace conversations {
...
-    subEntityId: string,
-    title: string,
-    conversationalSubEntity: ConversationalSubEntity
+    startConversationRequest: StartConversationRequest
   ): void {
     ensureInitialized(frameContexts.content);
     const messageId = sendMessageRequest(parentWindow, "startConversation", [
-      subEntityId,
-      title
+      startConversationRequest
```
sachingupta: [
startConversationRequest.subEntityId,
startConversationRequest.title
]

src/MicrosoftTeams.ts
```diff
@@ -2345,25 +2362,17 @@ export namespace conversations {
...
-    conversationId: string,
-    conversationalSubEntity: ConversationalSubEntity
+    showConversationRequest: ShowConversationRequest
   ): void {
     ensureInitialized(frameContexts.content);
     const messageId = sendMessageRequest(parentWindow, "showConversation", [
-      subEntityId,
-      title,
-      conversationId
+      showConversationRequest
```
sachingupta: [
showConversationRequest.subEntityId,
showConversationRequest.title,
showConversationRequest.conversationId
]

src/private/conversations.ts
```diff
@@ -0,0 +1,66 @@
+import { ensureInitialized, sendMessageRequest } from "../internal/internalAPIs";
```
sachingupta: export all these function to public by adding _export * from "./conversation"; inside ./src/private/index.ts file.

sachingupta: we need to this for all new file added inside private or public folder.

src/private/conversations.ts
```diff
@@ -0,0 +1,66 @@
...
+  */
+  export function startConversation(
+    startConversationRequest: StartConversationRequest
+  ): void {
+    ensureInitialized(frameContexts.content);
+    const messageId = sendMessageRequest(GlobalVars.parentWindow, "conversations.startConversation", [{
+      title: startConversationRequest.title,
+      subEntityId: startConversationRequest.subEntityId
+    }]);
+    GlobalVars.callbacks[messageId] = (conversationId?: string, reason?: string) => {
```
ydogandjiev: This callback only gets invoked a single time when the handler for this message on the Teams side completes; however, my expectation is that we might have to call both `onStartConversation` when the user posts the first message and then again `onCloseConversation` some time later when the user dismisses the chat pane. How do you plan to trigger both of these handlers for a single call to startConversation?

src/private/conversations.ts
```diff
@@ -0,0 +1,66 @@
...
+export namespace conversations {
+
+  /**
+  * @private
+  * Hide from docs
+  * --------------
+  * Allows the user to start a conversation with each subentity inside a tab
+  * @param startConversationRequest Callback containing the conversation Id and if the tab pane was closed
+  */
+  export function startConversation(
```
ydogandjiev: I am still wondering if we should just merge `startConversation` and `showConversation`. They just seem too similar in functionality with the only difference being whether a `conversationId` is provided. We could just make the `conversationId` an optional parameter on the `ShowConversationRequest` structure.

ydogandjiev: Do you remember why we decided to go w/ two different methods here?

src/private/conversations.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+  * @private
+  * Hide from docs
+  * --------------
+  * Allows the user to show the conversation in the right pane
+  * @param showConversationRequest Callback containing if the tab pane was closed
+  */
+  export function showConversation(
```
ydogandjiev: Let's get some unit tests added for all of this new functionality once we've finalized the interface.

src/private/conversations.ts
debajyoti2510: We went to and fro on this. The decision was that we would keep it as separate functions because

1. overloading the method makes the callback handling messy
2. "title" not making sense in the showConversation at the moment.

src/internal/internalAPIs.ts
```diff
@@ -9,6 +9,13 @@ GlobalVars.handlers["fullScreenChange"] = handleFullScreenChange;
 GlobalVars.handlers["backButtonPress"] = handleBackButtonPress;
 GlobalVars.handlers["beforeUnload"] = handleBeforeUnload;
 GlobalVars.handlers["changeSettings"] = handleChangeSettings;
+GlobalVars.handlers["getConversationId"] = handleConversationId;
```
ydogandjiev: Wouldn't a more appropriate name for this be `conversationCreated` or something like that? I was pretty confused when I first saw this code and had to go read the contents of the handler to understand what the purpose of this is.

src/private/conversations.ts
```diff
@@ -12,44 +12,20 @@ export namespace conversations {
...
-    const messageId = sendMessageRequest(GlobalVars.parentWindow, "conversations.showConversation", [{
-      title: showConversationRequest.title,
-      subEntityId: showConversationRequest.subEntityId,
-      conversationId: showConversationRequest.conversationId
-    }]);
-    GlobalVars.callbacks[messageId] = (reason?: string) => {
-      showConversationRequest.onCloseConversation(reason);
+    GlobalVars.getConversationIdHandler = openConversationRequest.onCloseConversation;
+    GlobalVars.callbacks[messageId] = (conversationId?: string) => {
+      openConversationRequest.onStartConversation(conversationId);
```
ydogandjiev: Hmm, I think you flipped the `onStartConversation` and `onCloseConversation` calls. Like we discussed offline, the openConversation method should return only once the conversation closes. Therefore, the callbacks[messageId] handler should invoke the `onCloseConversation` handler.

src/public/interfaces.ts
```diff
@@ -364,6 +364,11 @@ export interface StartConversationRequest {
   */
   title: string;
 
+  /**
+  * The Id of the conversation
```
ydogandjiev: NIT: Add a comment indicating that this is optional and should be specified whenever a previous conversation about a specific sub-entity has already been started before

src/internal/internalAPIs.ts
sachingupta: it should be onCloseConversation

src/private/conversations.ts
sachingupta: @ydogandjiev as close conversation can happen any time it should not be response of openConversation call. I think we should not wait till conversation is close to return response from openConversation call. I like debo existing changes. Let me know if i am missing anything.

src/internal/internalAPIs.ts
```diff
@@ -9,11 +9,18 @@ GlobalVars.handlers["fullScreenChange"] = handleFullScreenChange;
 GlobalVars.handlers["backButtonPress"] = handleBackButtonPress;
 GlobalVars.handlers["beforeUnload"] = handleBeforeUnload;
 GlobalVars.handlers["changeSettings"] = handleChangeSettings;
-GlobalVars.handlers["getConversationId"] = handleConversationId;
+GlobalVars.handlers["onStartConversation"] = handleStartConversation;
```
ydogandjiev: NIT: For consistency consider following the existing naming pattern here for the event + handler. Either:
```
GlobalVars.handlers["startConversation"] = handleStartConversation;
GlobalVars.handlers["closeConversation"] = handleCloseConversation;
```

Or:
```
GlobalVars.handlers["onStartConversation"] = handleOnStartConversation;
GlobalVars.handlers["onCloseConversation"] = handleOnCloseConversation;
```

src/private/conversations.ts
```diff
@@ -23,9 +23,12 @@ export namespace conversations {
       subEntityId: openConversationRequest.subEntityId,
       conversationId: openConversationRequest.conversationId
     }]);
-    GlobalVars.getConversationIdHandler = openConversationRequest.onCloseConversation;
-    GlobalVars.callbacks[messageId] = (conversationId?: string) => {
-      openConversationRequest.onStartConversation(conversationId);
+    GlobalVars.onCloseConversationHandler = openConversationRequest.onCloseConversation;
+    GlobalVars.onStartConversationHandler = openConversationRequest.onStartConversation;
```
ydogandjiev: Should we at some point clear out these handlers? Maybe when the closeConversation event comes in?

src/internal/internalAPIs.ts
sachingupta: for consistency use this 
GlobalVars.handlers["startConversation"] = handleStartConversation;
GlobalVars.handlers["closeConversation"] = handleCloseConversation;

src/private/conversations.ts
sachingupta: agree with @ydogandjiev

src/private/conversations.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export namespace conversations {
+
+  /**
+  * @private
+  * Hide from docs
+  * --------------
+  * Allows the user to start or continue a conversation with each subentity inside a tab
+  */
+  export function openConversation(
```
sachingupta: thanks @debajyoti2510 now it looks very clean.

src/private/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./menus";
 export * from "./interfaces";
-export * from "./privateAPIs";
\ No newline at end of file
+export * from "./privateAPIs";
+export * from "./conversations";
```
sachingupta: please merge with latest change and export conversation API and interface explicit

src/private/index.ts
```diff
@@ -19,4 +19,5 @@ export {
   sendCustomMessage,
   showNotification,
   uploadCustomApp
-} from "./privateAPIs";
\ No newline at end of file
+} from "./privateAPIs";
+export * from "./conversations";
```
sachingupta: Please use explicit variable here export { Conversations } from "./conversations";

debajyoti2510: Sachin both the functions here are exported

PR #184, opened 2019-02-28
Adding 'Rigel' as a supported HostClientType

### LINKED ISSUE #183: Add Rigel as one of the supported HostClientTypes
I'm working on a website which needs to show different UI when projected from a Rigel device versus Teams from a desktop/other device.  I'd like to add Rigel as one of the supported HostClientTypes so we can detect if we're running on a Rigel at runtime.

GSloves:
Adding 'Rigel' as a supported HostClientType to allow consumers to detect if they are being projected by a Rigel device.  

This will allow users to detect if they are running on a Rigel device by checking the HostClientType returned in GetContext().  

Fix #183

### DIFF
--- a/src/public/authentication.ts
+++ b/src/public/authentication.ts
@@ -31,7 +31,8 @@ export namespace authentication {
     ensureInitialized(frameContexts.content, frameContexts.settings, frameContexts.remove, frameContexts.task);
     if (GlobalVars.hostClientType === HostClientType.desktop ||
       GlobalVars.hostClientType === HostClientType.android ||
-      GlobalVars.hostClientType === HostClientType.ios) {
+      GlobalVars.hostClientType === HostClientType.ios ||
+      GlobalVars.hostClientType === HostClientType.rigel) {
       // Convert any relative URLs into absolute URLs before sending them over to the parent window.
       const link = document.createElement("a");
       link.href = authenticateParams.url;
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -2,7 +2,8 @@ export const enum HostClientType {
   desktop = "desktop",
   web = "web",
   android = "android",
-  ios = "ios"
+  ios = "ios",
+  rigel = "rigel"
 }
 
 /**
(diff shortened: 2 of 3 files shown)


GSloves:
Working with Monica Dumencu to support this from Teams end, sorry not sure if it's possible to link/mention her here.

src/MicrosoftTeams.ts
```diff
@@ -1251,7 +1252,8 @@ export namespace authentication {
     if (
       hostClientType === HostClientType.desktop ||
       hostClientType === HostClientType.android ||
-      hostClientType === HostClientType.ios
+      hostClientType === HostClientType.ios ||
+      hostClientType === HostClientType.rigel
```
GSloves: Can someone confirm that we want to do the same URL conversion on Rigel as on desktop devices?  I assumed that Rigel was previously treated as a desktop device.

GSloves:
@ydogandjiev any update on this PR? No rush, just curious if anything else needs to be done to complete it.

src/MicrosoftTeams.ts
sachingupta: @ydogandjiev

src/MicrosoftTeams.ts
```diff
@@ -84,7 +84,8 @@ export const enum HostClientType {
   desktop = "desktop",
   web = "web",
   android = "android",
-  ios = "ios"
+  ios = "ios",
```
sachingupta: Please sync latest changes from master and move your changes to ./src/public/constants.ts file and ./src/public/authentication.ts

src/MicrosoftTeams.ts
ydogandjiev: Yes, that's the safest option; to be honest we should probably do this URL expansion in all cases.

GSloves: Great, thanks.  Yeah I agree that you might as well do it everywhere, but not a big deal either way.

src/MicrosoftTeams.ts
GSloves: Just updated the PR, also updated the comment in interface.ts, just fyi.

GSloves: Sorry probably should have merged instead of trying to rebase.

GSloves:
Thanks for pointing me to where the code should be in the new refactored code and merging in the PR, @sachingupta

PR #194, opened 2019-03-12
Saching/taskmodule message passing support from task module(child window) to tab(parent window) and vice versa

sachingupta:
This change support message passing from task module(child window) to tab(parent window) and vice versa using generic window postMessage handlers.


src/public/windowObject.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    GlobalVars.callbacks[messageId] = (success: boolean, result: string) => {
+      if (!success) {
+        throw new Error(result);
+      }
+    };
+  }
+
+  public addEventListener(type, listener) {
+    if (type == "message") {
```
ydogandjiev: NIT: `==` --> `===`

src/public/windowObject.ts
```diff
@@ -0,0 +1,60 @@
...
+    ]);
+
+    GlobalVars.callbacks[messageId] = (success: boolean, result: string) => {
+      if (!success) {
+        throw new Error(result);
+      }
+    };
+  }
+
+  public addEventListener(type, listener) {
```
ydogandjiev: NIT: Add types for `type` and `listener`

ydogandjiev: And a return type for the `addEventListener` method

src/public/windowObject.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (type == "message") {
+      GlobalVars.handlers["messageForParent"] = listener;
+    }
+  }
+}
+
+export class ParentWindowObject implements IWindowObject {
+  private static _instance: ParentWindowObject;
+  public static get Instance() {
```
ydogandjiev: NIT: Add return type for `Instance()`

src/public/windowObject.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    GlobalVars.callbacks[messageId] = (success: boolean, result: string) => {
+      if (!success) {
+        throw new Error(result);
+      }
+    };
+  }
+
+  public addEventListener(type, listener) {
+    if (type == "message") {
```
ydogandjiev: NIT: `==` --> `===`

src/public/windowObject.ts
```diff
@@ -0,0 +1,60 @@
...
+    ]);
+
+    GlobalVars.callbacks[messageId] = (success: boolean, result: string) => {
+      if (!success) {
+        throw new Error(result);
+      }
+    };
+  }
+
+  public addEventListener(type, listener) {
```
ydogandjiev: NIT: Add types for `type` and `listener` as well as return type for `addEventListener` method

PR #196, opened 2019-03-14
Refactoring teams-js code into modules/logical files

sachingupta:
Motivation: code organizing into separate logical module, improve code quality, reduce maintenance.
organizing sdk into 3 parts
1.Internal : code that we don't export
2. private: code that we export as private API
3. public: code that we export as public

This is 2nd step. Only remaining part is breaking Unit tests into different logical files. Pr related to 1st step https://github.com/OfficeDev/microsoft-teams-library-js/pull/193


src/public/authentication.ts
```diff
@@ -0,0 +1,411 @@
...
+  GlobalVars.handlers["authentication.authenticate.success"] = handleSuccess;
+  GlobalVars.handlers["authentication.authenticate.failure"] = handleFailure;
+  /**
+   * Registers the authentication GlobalVars.handlers
+   * @param authenticateParameters A set of values that configure the authentication pop-up.
+   */
+  export function registerAuthenticationHandlers(authenticateParameters: AuthenticateParameters): void {
+    authParams = authenticateParameters;
+  }
+  /**
```
ydogandjiev: NIT: Add an empty line between methods for some better visual separation and readability

src/index.ts
```diff
@@ -1 +1,2 @@
-export * from "./MicrosoftTeams";
+export * from "./private/index";
+export * from "./public/index";
```
ydogandjiev: Did you figure out how to generate a single .d.ts file for the SDK after these changes or are we deciding that's no longer necessary?

sachingupta: We can do it if partner can not consume npm package for typing.

PR #200, opened 2019-03-22
adding promise based api

sachingupta:
We have two option to support promise bases API in sdk
1. Convert all existing callback based API to promise bases and release new version with major version change(breaking change)
  - Benefit: no need to maintain two version of API
  - drawback: user need to update their code if they want to upgrade API version.
2. Keep existing api as it is and add Promise base api that user can import using {MicrosoftTeams} from "@microsoft/teams-js"
  - Benefit: user don't need to update their code if they want to upgrade API version.
  - drawback: need to maintain two version of API
3. hybrid of option 1 and 2 - will not use
https://github.com/OfficeDev/microsoft-teams-library-js/pull/201

Open questions:

- 1. Going forward we can release all new API as promise based.
- 2. We can deprecate existing callback APIs in future

This is draft PR with option 2

this is how it can be use https://github.com/ydogandjiev/microsoft-teams-test-tab/pull/13/files


baywet:
I'd argue for option 1 for the following reasons:
- The web has moved to promises for the most part almost 2 years ago to avoid callback hell
- The API surface is going to be leaner and cleaner, even from a consumer point of view
- It's ok to having breaking changes as long as they are versioned properly
- If the consumer wants callback for some reasons, they can always leverage the .then syntax

PR #201, opened 2019-03-25
supporting Promise from callback API option 3

sachingupta:


sachingupta:
we will not use this approach

PR #212, opened 2019-04-30
Adding SDK API for tab ready success and fail notification

sachingupta:
Adding SDK API for tab ready success and fail notification
Tab can notify to host if tab is ready to interact or failed with reason.

### DIFF
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -32,3 +32,9 @@ export const enum TaskModuleDimension {
   Medium = "medium",
   Small = "small"
 }
+
+export const enum PageLoadFailReason {
+  AuthFailed = "AuthFailed",
+  TimeOut = "TimeOut",
+  Other = "Other"
+}
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -35,6 +35,6 @@ export const enum TaskModuleDimension {
 
 export const enum AppInitializationFailedReason {
   AuthFailed = "AuthFailed",
-  TimeOut = "Timeout",
+  Timeout = "Timeout",
   Other = "Other"
 }
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -33,8 +33,8 @@ export const enum TaskModuleDimension {
   Small = "small"
 }
 
-export const enum AppLoadFailReasonTypes {
+export const enum AppInitializationFailedReason {
   AuthFailed = "AuthFailed",
-  TimeOut = "TimeOut",
+  TimeOut = "Timeout",
   Other = "Other"
 }
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -33,7 +33,7 @@ export const enum TaskModuleDimension {
   Small = "small"
 }
 
-export const enum PageLoadFailReason {
+export const enum AppLoadFailReasonTypes {
   AuthFailed = "AuthFailed",
   TimeOut = "TimeOut",
   Other = "Other"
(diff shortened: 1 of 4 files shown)


src/public/publicAPIs.ts
```diff
@@ -3,92 +3,93 @@ import { GlobalVars } from "../internal/globalVars";
...
-    return;
-  }
-
-  GlobalVars.initializeCalled = true;
+export function initialize(hostWindow: any = window): AppLoadEvent {
+  // Independent components might not know whether the SDK is initialized so might call it to be safe.
+  // Just no-op if that happens to make it easier to use.
+  if (!GlobalVars.initializeCalled) {
+    GlobalVars.initializeCalled = true;
+
```
ydogandjiev: NIT: Delete extra empty line

src/public/publicAPIs.ts
```diff
@@ -317,3 +318,13 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
   GlobalVars.callbacks[messageId] = onComplete ? onComplete : getGenericOnCompleteHandler(errorMessage);
 }
 
+export class AppLoadEvent implements IAppLoadEvent {
+  public notifySuccess(): void {
+    ensureInitialized();
+    sendMessageRequest(GlobalVars.parentWindow, "onTabShow.success", [version]);
```
ydogandjiev: NIT: I think it would be better if we called this `appLoadEvent.success` or `initialize.success`

ydogandjiev: Same for the failure case below

src/public/interfaces.ts
```diff
@@ -394,3 +394,20 @@ export interface OpenConversationRequest {
...
+export interface IAppLoadEvent {
+  /**
+   * Indicates that the underlying resource has been created and the settings can be saved.
+   */
+  notifySuccess(): void;
+  /**
+   * Indicates that creation of the underlying resource failed and that the settings cannot be saved.
+   * @param reason Specifies a reason for the failure. If provided, this string is displayed to the user; otherwise a generic error is displayed.
+   */
+  notifyFailure(appLoadFailReason: IAppLoadFailReason): void;
```
ydogandjiev: Can we add an optional string parameter called something like `correlationTag` or `correlationId` that we can use to correlate telemetry between us and our partners? We can sanitize it for PII data and choose to only log this for 1P/2P apps to reduce the risk of PII data getting into our pipeline.

ydogandjiev: Oh, oops... I see that's what the `message` is for. That should be fine.

src/public/interfaces.ts
```diff
@@ -394,3 +394,20 @@ export interface OpenConversationRequest {
   */
   onCloseConversation?: (subEntityId: string, conversationId?: string) => void;
 }
+
+export interface IAppLoadFailReason {
+  reason: AppLoadFailReasonTypes;
+  message?: string;
+}
+
+export interface IAppLoadEvent {
```
ydogandjiev: NIT: I think `IAppInitEvent` or `IAppInitializationEvent` might be a better name for this

src/public/constants.ts
```diff
@@ -31,4 +31,10 @@ export const enum TaskModuleDimension {
   Large = "large",
   Medium = "medium",
   Small = "small"
+}
+
+export const enum AppLoadFailReasonTypes {
+  AuthFailed = "AuthFailed",
+  TimeOut = "TimeOut",
```
aosolis: "timeout" is a single word, so I think this should be cased "Timeout"

PR #214, opened 2019-05-24
adding explicit app load, initialization success/fail api

sachingupta:
adding explicit app load, initialization success/fail api
This api can be use by tab to hide loading indicator before notifying appInitialization success/fail.

### DIFF
--- a/src/public/publicAPIs.ts
+++ b/src/public/publicAPIs.ts
@@ -3,7 +3,7 @@ import { GlobalVars } from "../internal/globalVars";
 import { version, frameContexts } from "../internal/constants";
 import { ExtendedWindow, MessageEvent } from "../internal/interfaces";
 import { settings } from "./settings";
-import { TabInformation, TabInstanceParameters, TabInstance, DeepLinkParameters, Context, IAppInitializationFailedRequest } from "./interfaces";
+import { TabInformation, TabInstanceParameters, TabInstance, DeepLinkParameters, Context } from "./interfaces";
 import { getGenericOnCompleteHandler } from "../internal/utils";
 
 // ::::::::::::::::::::::: MicrosoftTeams SDK public API ::::::::::::::::::::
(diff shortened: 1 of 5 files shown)


src/public/publicAPIs.ts
```diff
@@ -3,16 +3,15 @@ import { GlobalVars } from "../internal/globalVars";
...
-import { AppInitializationFailedReason } from "./constants";
 
 // ::::::::::::::::::::::: MicrosoftTeams SDK public API ::::::::::::::::::::
 /**
  * Initializes the library. This must be called before any other SDK calls
  * but after the frame is loaded successfully.
  */
-export function initialize(hostWindow: any = window): AppInitializationEvent {
+export function initialize(hostWindow: any = window) {
   // Independent components might not know whether the SDK is initialized so might call it to be safe.
```
ydogandjiev: NIT: Add a return type to initialize (void)

src/public/publicAPIs.ts
```diff
@@ -87,8 +86,30 @@ export function initialize(hostWindow: any = window): AppInitializationEvent {
       GlobalVars.isFramelessWindow = false;
     };
   }
+}
+
+/**
+ * To notify app loaded to hide loading indicator
+ */
+export function notifyAppLoad(): void {
```
ydogandjiev: NIT: I think `notifyAppLoaded` or `showApp` would be a better name for this API.

src/public/publicAPIs.ts
```diff
@@ -87,8 +86,30 @@ export function initialize(hostWindow: any = window): AppInitializationEvent {
       GlobalVars.isFramelessWindow = false;
     };
   }
+}
+
+/**
```
ydogandjiev: NIT: Consider putting all of the notifyApp*** APIs in their own namespaces (e.g. like settings, tasks, etc.)

PR #217, opened 2019-06-04
Tejab/log handler

tburra:
Add log handling for extension apps


src/private/privateAPIs.ts
```diff
@@ -164,3 +164,30 @@ export function getConfigSetting(
...
+}
+
+/**
+ * @private
+ * Hide from docs
+ * ------
+ * Sends the log mesasges to the Teams client. Should be used with registerGetLogHandler
+ * @param log app log to be sent to Teams client
+ */
+export function logMessages(log: string): void {
```
ydogandjiev: Instead of exposing this as a stand-alone function I think it would be better if you just define this logic directly within `handleGetLogRequest` above. If we haven't asked for logs from the app I don't think it shouldn't be possible for it to send them to us.

src/private/privateAPIs.ts
```diff
@@ -164,3 +164,30 @@ export function getConfigSetting(
...
+  ensureInitialized();
+
+  sendMessageRequest(GlobalVars.parentWindow, "logMessages", [log]);
```
ydogandjiev: Consider naming this message "sendLog" instead of "logMessages" just so it matches the "getLog" request above.

src/private/privateAPIs.ts
tburra: fixed

src/internal/internalAPIs.ts
```diff
@@ -62,6 +63,13 @@ function handleChangeSettings(): void {
   }
 }
 
+function handleGetLogRequest(): void {
+  if (GlobalVars.getLogHandler) {
+    const log: string = GlobalVars.getLogHandler();
+    sendMessageRequest(GlobalVars.parentWindow, "sendLog", [log]);
```
sachingupta: We should create a separate function call 
export function sendLog(log: String) {
          sendMessageRequest(GlobalVars.parentWindow, "sendLog", [log]);
 }

that can be use by app to send log to teams client. getLogHandler should call microsoftTeams.sendLog function to send call.

src/internal/internalAPIs.ts
```diff
@@ -62,6 +63,13 @@ function handleChangeSettings(): void {
   }
 }
 
+function handleGetLogRequest(): void {
```
sachingupta: We can move all functions related to log into its separate namespace/file similar to menus.ts

src/internal/internalAPIs.ts
tburra: i had the method logMessages earlier.
Yuri's suggestion above was to be able to send logs only when asked from app

src/private/files.ts
```diff
@@ -0,0 +1,35 @@
...
+import { GlobalVars } from "../internal/globalVars";
+
+/**
+ * Namespace to interact with the files part of the SDK.
+ * This object is used to handle logs
+ *
+ * @private
+ * Hide from docs
+ */
+export namespace files {
```
sachingupta: please rename this file and namespace to more logical with logs ? May be AppLogs @ydogandjiev  what you think?

src/private/files.ts
```diff
@@ -0,0 +1,35 @@
+import { ensureInitialized, sendMessageRequest } from "../internal/internalAPIs";
+import { GlobalVars } from "../internal/globalVars";
+
+/**
+ * Namespace to interact with the files part of the SDK.
+ * This object is used to handle logs
```
sachingupta: we can update description about sending onDemand app logs to host client?

src/private/files.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @private
+ * Hide from docs
+ */
+export namespace files {
+
+  GlobalVars.handlers["log.request"] = handleGetLogRequest;
+
+  export function handleGetLogRequest(): void {
```
ydogandjiev: Does this function need to get exported? As far as I can tell it's internal to just this file and isn't something our developers would even be calling.

src/private/files.ts
ydogandjiev: Probably `log` or `logs` would be best and that's kind of the name we've chosen by using `log.request` as the message name.

src/private/index.ts
```diff
@@ -1,4 +1,5 @@
 export { menus } from './menus';
+export { files } from './files';
```
ydogandjiev: Rename this to match whatever you decide to call the namespace (i.e. `log` or `logs`).

src/private/files.ts
tburra: renamed to logs

PR #218, opened 2019-06-05
add json tab bot handler and corresponding unit tests

FaizChishtie:


src/public/publicAPIs.ts
```diff
@@ -312,4 +312,22 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
...
-}
\ No newline at end of file
+}
+
+/**
+ * Sends query to bot in order to retrieve data.
+ * @param botRequest query to send to bot.
+ * @param callback callback to invoke when data is retrieved from bot
+ */
+export function getBotData(botRequest: BotAPIRequest, callback?: (botResponse: any) => void): void { // void for now
```
sachingupta: rename this to executeBotQuery

aosolis: Can we make it more generic, say sendBotRequest or sendBotMessage()? I can see this being used for more than queries.

aosolis: How does this function report errors?

src/public/publicAPIs.ts
```diff
@@ -312,4 +312,22 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
...
+    ensureInitialized();
+    // send request to teams
+
+    // SHOULD BE executeBotQuery - testing with getContext
```
aosolis: Comment out of date

src/public/interfaces.ts
```diff
@@ -394,3 +394,10 @@ export interface OpenConversationRequest {
   */
   onCloseConversation?: (subEntityId: string, conversationId?: string) => void;
 }
+
+export interface BotAPIRequest {
```
aosolis: Could we name this BotRequestMessage (echoing .NET's HttpRequestMessage or Node.js http.IncomingMessage)
The "API" part feels redundant, and a bit in-apt, as it's a pretty generic mechanism, not a specific API.

src/public/publicAPIs.ts
sachingupta: Lets use callback?: (status: boolean, data?: string | BotResponse) => void;

1. In case of success
status will be true and data will be response
2. in case of fail
status will be false and data will be response

sachingupta: sendBotRequest ?

src/public/interfaces.ts
sachingupta: BotRequest?

src/public/publicAPIs.ts
```diff
@@ -312,4 +312,21 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
 
   const errorMessage = "Invalid internalTabInstanceId and/or channelId were/was provided";
   GlobalVars.callbacks[messageId] = onComplete ? onComplete : getGenericOnCompleteHandler(errorMessage);
-}
\ No newline at end of file
+}
+
+/**
```
ydogandjiev: I think it might be best if we marked this as @private for now so that it doesn't end up in our documentation until we are ready to ship the feature.

test/MicrosoftTeams.spec.ts
```diff
@@ -1972,6 +1972,26 @@ describe("MicrosoftTeams", () => {
...
+    });
+    it("should allow calls after initialization", () => {
+      initializeWithContext("content");
+      expect(() =>
+      microsoftTeams.sendBotRequest({ query: "" }, () => {
+        return;
+      })
+    ).not.toThrowError();
+    });
+
```
ydogandjiev: Consider adding some more unit test cases to check that:
1. The query is correctly sent to the parent (i.e. Teams)
2. The response correctly invokes the callback with the right data

FaizChishtie: Will work on this in a separate PR

src/public/publicAPIs.ts
```diff
@@ -312,4 +312,25 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
...
+ * @private
+ * Hide from docs until release.
+ * ------
+ * 
+ * Sends query to bot in order to retrieve data.
+ * @param botRequest query to send to bot.
+ * @param onBotResponse callback to invoke when data is retrieved from bot
+ */
+export function sendBotRequest(botRequest: BotRequest, onBotResponse?: (status: boolean, data: string | BotResponse) => void): void { // void for now
```
sachingupta: Lets add two handler onBotResponse(date) and onError(string) => {}

FaizChishtie: sure

test/MicrosoftTeams.spec.ts
ydogandjiev: Without corresponding unit tests it's generally difficult to ensure your code is actually doing what it's supposed to. In a true test-driven development process it's actually the unit tests that get written first:
https://en.wikipedia.org/wiki/Test-driven_development

As a best practice, I would urge you to add a full set of unit tests as part of your current PR.

test/MicrosoftTeams.spec.ts
```diff
@@ -1972,6 +1972,83 @@ describe("MicrosoftTeams", () => {
...
+      microsoftTeams.sendBotRequest(request, handleBotResponse, handleError);
+
+      // find message request in jest
+      const message = findMessageByFunc("executeBotQuery");
+
+      // check message is sending correct data
+      expect(message).not.toBeUndefined();
+      expect(message.args).toContain(request);
+
+      const _botResponse: BotResponse = { data: ["some", "queried", "items"] };
```
ydogandjiev: NIT: The `_botResponse` variable is not necessary here; consider defining it directly inline within the `data` object below.

test/MicrosoftTeams.spec.ts
```diff
@@ -1972,6 +1972,83 @@ describe("MicrosoftTeams", () => {
...
+
+      const handleBotResponse = (response: BotResponse) => (botResponse = response);
+      const handleError = (_error: string): any => (error = _error);
+
+      microsoftTeams.sendBotRequest(request, handleBotResponse, handleError);
+      const message = findMessageByFunc("executeBotQuery");
+      expect(message).not.toBeUndefined();
+      expect(message.args).toContain(request);
+
+      const _botResponse: string = "Something went wrong...";
```
ydogandjiev: NIT: The `_botResponse` variable is not necessary here; consider defining it directly inline within the `data` object below.

src/public/publicAPIs.ts
```diff
@@ -312,4 +312,31 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
...
+ */
+export function sendBotRequest(botRequest: BotRequest, onBotResponse?: (data: BotResponse) => void, onError?: (error: string) => {}): void { // void for now
+    ensureInitialized();
+
+    // send request to teams
+    const messageId = sendMessageRequest(GlobalVars.parentWindow, "executeBotQuery", [
+      botRequest
+    ]);
+
+    // register handler for callback id
```
sachingupta: Lets put bot related functions and interfaces inside its own namespace (take a look on task.ts file)

sachingupta: export namespace bot {
    sendQueryRequest(botRequest: BotRequest, onBotResponse?: (data: string | BotResponse) => void, onError?: (error: string) => {}): void {
  ....
// send request to teams
    const messageId = sendMessageRequest(GlobalVars.parentWindow, "bot.executeQuery", [
      botRequest
    ]);

...
}

sachingupta:
Moving this change to feature branch https://github.com/OfficeDev/microsoft-teams-library-js/pull/228

PR #220, opened 2019-06-06
Make regsiterBeforeUnloadHandler hidden from doc generation

asnow003:


src/public/publicAPIs.ts
```diff
@@ -188,6 +188,7 @@ export function navigateBack(onComplete?: (status: boolean, reason?: string) =>
 }
 
 /**
+ * @private
  * Registers a handler to be called before the page is unloaded.
```
ydogandjiev: NIT: Consider updating the comment to indicate that the page will be hidden and it should only use this handler to save its state and any user-entered data. This handler cannot be used to block the unload/destruction of the hosted experience.

PR #222, opened 2019-06-07
Add the capability to use conversational subentity APIs in a personal app scope

debajyoti2510:
Changes to allow the API to be used in the personal app scope:

- The developer can now send in the thread Id and the entity Id while calling openConversation. This is so that in a personal scope we can start / open a conversation for the subentity in a tab in a particular team.

- We will send back the threadId and the entity Id along with the conversation Id and the subentity Id so that the developer can use it when opening the conversation in a personal app scope.

### DIFF
--- a/src/internal/globalVars.ts
+++ b/src/internal/globalVars.ts
@@ -24,6 +24,6 @@ export class GlobalVars {
   public static backButtonPressHandler: () => boolean;
   public static beforeUnloadHandler: (readyToUnload: () => void) => boolean;
   public static changeSettingsHandler: () => void;
-  public static onStartConversationHandler: (sunEntityId: string, conversationId: string, threadId: string, entityId: string) => void;
-  public static onCloseConversationHandler: (sunEntityId: string, conversationId?: string, threadId?: string, entityId?: string) => void;
+  public static onStartConversationHandler: (subEntityId: string, conversationId: string, threadId: string, entityId: string) => void;
+  public static onCloseConversationHandler: (subEntityId: string, conversationId?: string, threadId?: string, entityId?: string) => void;
 }
--- a/src/internal/globalVars.ts
+++ b/src/internal/globalVars.ts
@@ -24,6 +24,6 @@ export class GlobalVars {
   public static backButtonPressHandler: () => boolean;
   public static beforeUnloadHandler: (readyToUnload: () => void) => boolean;
   public static changeSettingsHandler: () => void;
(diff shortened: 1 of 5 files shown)


src/internal/globalVars.ts
```diff
@@ -24,6 +24,6 @@ export class GlobalVars {
   public static backButtonPressHandler: () => boolean;
   public static beforeUnloadHandler: (readyToUnload: () => void) => boolean;
   public static changeSettingsHandler: () => void;
-  public static onStartConversationHandler: (sunEntityId: string, conversationId: string) => void;
-  public static onCloseConversationHandler: (sunEntityId: string, conversationId?: string) => void;
+  public static onStartConversationHandler: (sunEntityId: string, conversationId: string, threadId: string, entityId: string) => void;
```
sachingupta: make sure you are returning these fields in same order in teams code

aosolis: subEntityId, right? (typo)

src/public/interfaces.ts
```diff
@@ -384,6 +384,16 @@ export interface OpenConversationRequest {
   */
   conversationId?: string;
 
+  /**
+   * The Id of the channel. This is optional and should be specified whenever a conversation is started or opened in a personal app scope
+   */
+  threadId?: string;
```
aosolis: Is this truly only for channel tabs? Does not work for tabs configured in group chat?

src/public/interfaces.ts
```diff
@@ -384,6 +384,16 @@ export interface OpenConversationRequest {
...
+
+  /**
+   * The Id of the tab. This is optional and should be specified whenever a conversation is started in a personal scope
+   */
+  entityId: string;
```
aosolis: If it’s optional should this be “entityId?”, like threadId above

src/public/interfaces.ts
debajyoti2510: Only for channel tabs

src/internal/globalVars.ts
debajyoti2510: subEntityId was already there @aosolis . Adding entity Id now, because from the personal scope we will need to know which tab the user wants to start a conversation in

src/public/interfaces.ts
debajyoti2510: Its not optional. We changed it. Forgot to update the comment. Done

PR #224, opened 2019-06-10
Changes to the SDK to make the entity Id optional and make the response send an object instead of individual parameters

debajyoti2510:
Made the entity Id in the open conversation response as optional.
Also now sending back a ConversationRequest object instead of individual parameters. 

This will be a breaking change but we should be good since the conversational subentity APIs are still in beta and is not being currently used by a lot of devs.

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -397,10 +397,38 @@ export interface OpenConversationRequest {
   /**
   * A function that is called once the conversation Id has been created
   */
-  onStartConversation?: (subEntityId: string, conversationId: string) => void;
+  onStartConversation?: (conversationSubEntityCallback: ConversationSubEntityCallback) => void;
 
   /**
   * A function that is called if the pane is closed
   */
-  onCloseConversation?: (subEntityId: string, conversationId?: string) => void;
+  onCloseConversation?: (conversationSubEntityCallback: ConversationSubEntityCallback) => void;
+}
+
+/**
+ * @private
+ * Hide from docs.
+ * ------
+*/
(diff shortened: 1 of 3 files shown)


src/public/interfaces.ts
```diff
@@ -397,10 +397,38 @@ export interface OpenConversationRequest {
...
-  onCloseConversation?: (subEntityId: string, conversationId?: string) => void;
+  onCloseConversation?: (conversationSubEntityCallback: ConversationSubEntityCallback) => void;
+}
+
+/**
+ * @private
+ * Hide from docs.
+ * ------
+*/
+export interface ConversationSubEntityCallback {
```
ydogandjiev: A callback is a function that gets invoked when an async operation completes. This is not a callback. Consider naming it something like `OpenConversationResponse` or something along those lines.

sachingupta: may be ConversationInfo or ConversationResponse ? something generic that we can use in onStartConversation and onCloseConversation

debajyoti2510: Changed it to ConversationSubEntityResponse

ydogandjiev: But why "ConversationSubEntity"? That name isn't mentioned anywhere in the SDK library? It's only an internal name we use for this feature.

debajyoti2510: Yeah you are right. When I was making this change, I felt that there should be a way to distinguish this from regular conversation in a tab. The tab conversation that we had before this feature.

But since I already had the previous object  named as ConversationRequest, it makes sense to just change it to ConversationResponse.

PR #231, opened 2019-06-13
Adding linting support to teams sdk with option to auto fix using prettier and fixing existing tslint errors

sachingupta:
Adding linting support to teams sdk with option to auto fix using prettier


src/internal/utils.ts
```diff
@@ -1,29 +1,29 @@
...
 
 export function getGenericOnCompleteHandler(errorMessage?: string): (success: boolean, reason?: string) => void {
-  return (success: boolean, reason: string) => {
-    if (!success) {
-      throw new Error(errorMessage ? errorMessage : reason);
-    }
-  };
-}
\ No newline at end of file
+    return (success: boolean, reason: string) => {
```
ydogandjiev: Any chance we can stick with 2 spaces of indentation? That's what our other code bases use and I think it's good to be consistent.

sachingupta: updated with 2 space indentation

PR #237, opened 2019-06-25
enable exucute deepLink API in task module

sachingupta:
Enable execute deep Link API in task module


src/public/publicAPIs.ts
```diff
@@ -271,7 +271,7 @@ export function shareDeepLink(deepLinkParameters: DeepLinkParameters): void {
  * @param deepLink deep link.
  */
 export function executeDeepLink(deepLink: string, onComplete?: (status: boolean, reason?: string) => void): void {
-  ensureInitialized(frameContexts.content);
+  ensureInitialized(frameContexts.content, frameContexts.task);
```
ydogandjiev: NIT: Consider adding a unit test for this case.

PR #242, opened 2019-07-08
add new interfaces for authorization implementation

FaizChishtie:


src/private/bot.ts
```diff
@@ -66,11 +66,21 @@ export namespace bot {
   }
 
   export interface QueryResponse {
+    data: BotResults | BotAuth;
+    type: string;
```
sachingupta: create an string enum with values for different response type.

src/private/bot.ts
```diff
@@ -66,11 +66,21 @@ export namespace bot {
   }
 
   export interface QueryResponse {
+    data: BotResults | BotAuth;
+    type: BotResponseType;
+  }
+
+  export interface BotResults {
```
sachingupta: we can remove Bot prefix from all interface as all these interfaces are already inside bot namespace

PR #248, opened 2019-07-31
Added the tab session Id in the sdk Context interface

debajyoti2510:
We will stop using the app session Id for telemetry purpose. We will generate a new session Id everytime a tab is created. This will be used for telemetry investigation

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -309,9 +309,9 @@ export interface Context {
   ringId?: string;
 
   /**
-   * Unique ID for the current Tab session for use in correlating telemetry data.
+   * Unique ID for the current session for use in correlating telemetry data.
    */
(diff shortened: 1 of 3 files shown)


src/public/interfaces.ts
```diff
@@ -307,6 +307,11 @@ export interface Context {
    * Current ring ID
    */
   ringId?: string;
+
+  /**
+   * Unique ID for the current Tab session for use in correlating telemetry data.
+   */
+  tabSessionId?: string;
```
ydogandjiev: Could we call this `appSessionId`? We generally don't explicitly refer to tabs in the SDK since it supports many more scenarios than just tabs (e.g. task modules, connectors, etc.).

debajyoti2510: Changed it Yuri

Issue #172, opened 2019-01-21
initialize is async, it should return a Promise


JSteunou:
Have the case sometimes with authentification process. Calling initialize, then authenticate, but `hostClientType` is not set yet because called to soon. Adding a setTimeout do the trick but that's not pretty nor reliable.

ydogandjiev:
Thanks @JSteunou for reporting this issue! We’ve been aware of it for sometime and will attempt to address it in the future; however, the recommended approach for tabs is to always call microsoftTeams.initialize() in their page’s onload method or shortly thereafter. On the other hand, the microsoftTeams.authentication.authenticate() call should generally be made only in response to the user clicking on a “Sign in” button (in order to avoid triggering browsers’ popup blocker). This typically provides more than enough time for the hostClientType to get determined.

JSteunou:
Yes, but I have this case #171 where I'm using `authenticate` to open a popup (because `window.open`) does not work in Electron, and I do not want to spread the MS logic everywhere in my app, so I'm calling `initialize` in the same logic code that is using `authenticate`, like just a line before. 

In that specific case I would need a Promise back :)

Tyrcheg:
@JSteunou Hi, why not to wrap initialize inside of new Promise()?

JSteunou:
I could, but when I should call `resolve` ? Using a `setTimeout` ? This is not realistic.

ydogandjiev:
I've sent out a PR to address this issue here - #249. @JSteunou, please take a look and let me know if you think that it will suit your needs.

ydogandjiev:
Apps can now wait for initialization to complete as follows:
```
microsoftTeams.initialize(() => {
  // Initialization complete
});
```

Or if a promise-based solution is required like so:
```
return new Promise(resolve => {
  microsoftTeams.initialize(() => {
    resolve();
  });
});
```

This change will be released in v1.5 of our SDK.

PR #249, opened 2019-08-01
#172 - Add callback to initialize

ydogandjiev:
This change adds a callback to the initialize API that will be called once initialization completes. This can be used as follows:
```
microsoftTeams.initialize(() => {
  // Initialization complete
});
```

Or turned into a promise-based API like so:
```
return new Promise(resolve => {
  microsoftTeams.initialize(() => {
    resolve();
  });
});
```

### DIFF
--- a/src/public/publicAPIs.ts
+++ b/src/public/publicAPIs.ts
@@ -12,20 +12,18 @@ import { logs } from '../private/logs';
  * Initializes the library. This must be called before any other SDK calls
  * but after the frame is loaded successfully.
  */
-export function initialize(hostWindow: any = window): void {
+export function initialize(callback?: () => void): void {
   // Independent components might not know whether the SDK is initialized so might call it to be safe.
   // Just no-op if that happens to make it easier to use.
   if (!GlobalVars.initializeCalled) {
     GlobalVars.initializeCalled = true;
 
-    // Undocumented field used to mock the window for unit tests
-    GlobalVars.currentWindow = hostWindow;
-
     // Listen for messages post to our window
-    const messageListener = (evt: MessageEvent) => processMessage(evt);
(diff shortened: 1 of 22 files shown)


JSteunou:
Nice!

Why not hide the Promise into the `initialize`? It would be a better DX as anyone is familiar with it now.

```js
function initialize() {
  return new Promise(resolve => {

    // all init code

    GlobalVars.initializeCompleted ? resolve() : GlobalVars.initializeCallbacks.push(resolve);
  })
}
```

Or even a Deferred promised like if you re not ok with wrapping the all initialize code into a promise.

As of 2019 I prefer true async function which can be used with callbacks for the rest of users than the inverse, but that's my though :)

src/public/publicAPIs.ts
```diff
@@ -12,20 +12,18 @@ import { logs } from '../private/logs';
  * Initializes the library. This must be called before any other SDK calls
  * but after the frame is loaded successfully.
  */
-export function initialize(hostWindow: any = window): void {
+export function initialize(callback?: () => void): void {
```
sachingupta: Do we want to support passing output values of initialize call to callback (context: string, clientType: string) ?

ydogandjiev: I don’t think so. The frameContext is an internal detail of the SDK used to determine what APIs are supported. We’ve never exposed that to developers. If we ever need to, we would do so via the getContext API.

The hostClientType is already available via the getContext API and I think that adding a parallel option for fetching it would be confusing.

ydogandjiev:
@JSteunou, the Teams Client SDK needs to work in browsers like IE 11 that don't yet have native support for ES6 promises so we do not use these in our SDK. We are looking at a few options for exposing a promise-based version of our SDK as part of #200 but this is currently not very high up on our priority list so it might take some time before we finish the changes.

PR #260, opened 2019-08-19
[Do Not Merge] Pr/CI build setup

rohithjidagam:
Added yml file to setup the PR/CI pipeline


deploy-to-cdn.js
```diff
@@ -0,0 +1,64 @@
...
+  })
+  const logger = console.log;
+
+  getConnectionString().then(connectionString => {
+    const opts = {
+      serviceOptions: [connectionString], // custom arguments to azure.createBlobService
+      containerName: 'sdk', // container name in blob
+      containerOptions: { publicAccessLevel: 'blob' }, // container options
+      folder: 'v' + version + '/js', // path within container
+      deleteExistingBlobs: true, // true means recursively deleting anything under folder
```
sachingupta: I think we don't want to delete any published blob/files

rohithjidagam: it deletes anything under the folder before uploading.

sachingupta: Lets not delete any published files/folder to avoid republishing files.

rohithjidagam: Done.

deploy-to-cdn.js
```diff
@@ -0,0 +1,64 @@
...
+      zip: true, // gzip files if they become smaller after zipping, content-encoding header will change if file is zipped
+      metadata: { cacheControl: 'public, max-age=31556926' }, // metadata for each uploaded file
+      testRun: false, // test run - means no blobs will be actually deleted or uploaded, see log messages for details
+    };
+
+    deploy(opts, filePaths, logger, function(err) {
+      if (err) {
+        console.log('Error deploying', err);
+      }
+      console.log('Deployment Successful.');
```
ydogandjiev: Should't this be inside of an else statement?

deploy-to-cdn.js
```diff
@@ -0,0 +1,64 @@
...
+      deleteExistingBlobs: false, // true means recursively deleting anything under folder
+      concurrentUploadThreads: 2, // number of concurrent uploads, choose best for your network condition
+      zip: true, // gzip files if they become smaller after zipping, content-encoding header will change if file is zipped
+      metadata: { cacheControl: 'public, max-age=31556926' }, // metadata for each uploaded file
+      testRun: false, // test run - means no blobs will be actually deleted or uploaded, see log messages for details
+    };
+
+    deploy(opts, filePaths, logger, function(err) {
+      if (err) {
+        console.log('Error deploying', err);
```
ydogandjiev: Why not throw the err here as you are doing in getConnectionString above? What's the correct way to handle errors inside of build tooling so that they get correctly picked up by the pipeline?

PR #288, opened 2019-11-08
Add the new setFrameContext and initializeWithFrameContext() API - Will be used in Go to website or Reload tab

debajyoti251088:
There are currently a few problems with our tab navigation inside the iframe. There is no good way for us to know if a navigation has happened inside the iframe (unless the user calls initialize() on every navigation - which is currently not always the case. eg. opening a task). 
 
The problem is when the user clicks on 'Go to website' or 'Reloads' the tab we launch it with the initial content URL even if a navigation has happened inside the iframe. 
 
We will have 2 new SDK APIs.

The new API will be called initializeWithContext(). The developer will need to pass in the content URL (which will be used for the tab reload) and the website URL (which will be used for the go to website workflow). 
This is being done to remove the need for calling a new SDK API along with initialize() every time a tab navigation happens within the SDK. 
This should preferably be used by the developers of tabs. 

Not all tabs call initialize() eveytime a navigation is done inside the iframe. Eg. Tasks. We wouldn’t want to force them to start calling initialize(). So we will provide a new API called setFrameContext() for them. This is just to set the content and the website URL.

We will respect the ensureInitialized() condition for the new initialize API as well

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -466,3 +466,19 @@ export interface LoadContext {
    */
   contentUrl: string;
 }
+
+/**
+ * @private
+ * Hide from docs.
+ */
+export interface FrameContext {
+  /**
+   * The current URL that needs to be used in the iframe if the tab is reloaded
+   */
+  contentURL: string;
+
+  /**
+   * The current URL that needs to be used for opening the website when the user clicks on 'Go to website'
+   */
+  websiteURL: string;
+}
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -467,18 +467,14 @@ export interface LoadContext {
   contentUrl: string;
 }
 
-/**
- * @private
- * Hide from docs.
- */
 export interface FrameContext {
   /**
    * The current URL that needs to be used in the iframe if the tab is reloaded
    */
-  contentURL: string;
+  contentUrl: string;
 
   /**
    * The current URL that needs to be used for opening the website when the user clicks on 'Go to website'
    */
-  websiteURL: string;
+  websiteUrl: string;
 }
--- a/src/public/publicAPIs.ts
+++ b/src/public/publicAPIs.ts
@@ -1,7 +1,13 @@
-import { processMessage, ensureInitialized, sendMessageRequest, handleParentMessage } from '../internal/internalAPIs';
+import {
+  processMessage,
+  ensureInitialized,
+  sendMessageRequestToParent,
+  handleParentMessage,
+  processAdditionalValidOrigins,
+} from '../internal/internalAPIs';
 import { GlobalVars } from '../internal/globalVars';
 import { version, frameContexts } from '../internal/constants';
-import { ExtendedWindow, MessageEvent } from '../internal/interfaces';
+import { ExtendedWindow, DOMMessageEvent } from '../internal/interfaces';
 import { settings } from './settings';
 import {
   TabInformation,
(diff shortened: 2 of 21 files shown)


src/public/interfaces.ts
```diff
@@ -466,3 +466,19 @@ export interface LoadContext {
    */
   contentUrl: string;
 }
+
+/**
+ * @private
```
ydogandjiev: Why mark this private?

src/public/interfaces.ts
```diff
@@ -466,3 +466,19 @@ export interface LoadContext {
...
+export interface FrameContext {
+  /**
+   * The current URL that needs to be used in the iframe if the tab is reloaded
+   */
+  contentURL: string;
+
+  /**
+   * The current URL that needs to be used for opening the website when the user clicks on 'Go to website'
+   */
+  websiteURL: string;
```
ydogandjiev: NIT: Rename `websiteURL` to `websiteUrl` to match the naming we use in the `Settings` interface

src/public/interfaces.ts
```diff
@@ -466,3 +466,19 @@ export interface LoadContext {
...
+
+/**
+ * @private
+ * Hide from docs.
+ */
+export interface FrameContext {
+  /**
+   * The current URL that needs to be used in the iframe if the tab is reloaded
+   */
+  contentURL: string;
```
ydogandjiev: NIT: Rename `contentURL` to `contentUrl` to match the naming we use in the `Settings` interface

src/public/publicAPIs.ts
```diff
@@ -350,3 +351,17 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
   const errorMessage = 'Invalid internalTabInstanceId and/or channelId were/was provided';
   GlobalVars.callbacks[messageId] = onComplete ? onComplete : getGenericOnCompleteHandler(errorMessage);
 }
+
+export function setFrameContext(url: FrameContext): void {
```
ydogandjiev: Rename `url` to `frameContext` to better reflect its type.

src/public/publicAPIs.ts
```diff
@@ -350,3 +351,17 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
...
+  ensureInitialized(frameContexts.content);
+  sendMessageRequestToParent('setFrameContext', [url]);
+}
+
+export function initializeWithFrameContext(
+  url: FrameContext,
```
ydogandjiev: Same as above - please rename `url` to `frameContext`.

test/public/publicAPIs.spec.ts
```diff
@@ -697,4 +698,19 @@ describe('MicrosoftTeams-publicAPIs', () => {
       expect(readyToUnloadMessage).not.toBeNull();
     });
   });
+
+  it('should successfully frame context', () => {
+    utils.initializeWithContext('content');
+
+    let url: FrameContext = {
```
ydogandjiev: Rename `url` to `frameContext`

src/public/publicAPIs.ts
```diff
@@ -350,3 +351,17 @@ export function navigateToTab(tabInstance: TabInstance, onComplete?: (status: bo
   const errorMessage = 'Invalid internalTabInstanceId and/or channelId were/was provided';
   GlobalVars.callbacks[messageId] = onComplete ? onComplete : getGenericOnCompleteHandler(errorMessage);
 }
+
+export function setFrameContext(url: FrameContext): void {
+  ensureInitialized(frameContexts.content);
+  sendMessageRequestToParent('setFrameContext', [url]);
+}
+
+export function initializeWithFrameContext(
```
ydogandjiev: Consider adding a UT for this API too.

PR #290, opened 2019-11-22
Add defaultOneNoteSectionId attribute to the Context interface

andreipatrascoiu:
Add the defaultOneNoteSectionId attribute to the Context interface. This is a new attribute that will be published by the MiddleTier

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -322,6 +322,12 @@ export interface Context {
    * Represents whether PSTN calling is allowed for the current logged in User
    */
   isPSTNCallingAllowed?: boolean;
+
+  /**
+   * The OneNote section ID that is linked to the channel.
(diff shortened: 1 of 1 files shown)


src/public/interfaces.ts
```diff
@@ -322,6 +322,12 @@ export interface Context {
    * Represents whether PSTN calling is allowed for the current logged in User
    */
   isPSTNCallingAllowed?: boolean;
+
+  /**
+   * The OneNote section ID that is linked to the channel.
+   * Used in Class/Staff/PLC teams.
```
ydogandjiev: NIT: You can remove the "Used in Class/Staff/PLC teams." part of the comment since this property will actually be set even in old channels that aren't part of Class/Staff/PLC teams.

PR #295, opened 2019-12-09
Fix the setframe context and initialize with frame context APIs by explicitly exporting them

debajyoti251088:


src/public/index.ts
```diff
@@ -29,6 +29,8 @@ export {
   registerOnLoadHandler,
   registerOnThemeChangeHandler,
   shareDeepLink,
+  setFrameContext,
```
ydogandjiev: NIT: Consider alphabetizing these as with the rest of the methods in the list

PR #307, opened 2020-01-27
Add the subtext to the list of parameters in the conversational subentities request

debajyoti251088:
We received feedback that certain apps want to use a subtext that will be displayed in the deeplink chiclet for the conversation. Adding that parameter in here.

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -419,6 +419,11 @@ export interface OpenConversationRequest {
   entityId: string;
 
   /**
+   * The subtext of the tab
+   */
+  subtext?: string;
+
+  /**
    * A function that is called once the conversation Id has been created
    */
   onStartConversation?: (conversationResponse: ConversationResponse) => void;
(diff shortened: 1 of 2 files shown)


sachingupta:
merge latest master to your feature branch.

sachingupta (dismissed):

debajyoti251088:
> merge latest master to your feature branch.

src/public/interfaces.ts
```diff
@@ -418,6 +418,11 @@ export interface OpenConversationRequest {
    */
   entityId: string;
 
+  /**
+   * The subtext of the tab
```
ydogandjiev: Can we provide a better description here? I think this should say something like "Description of the item being discussed" or something along those lines.

src/public/interfaces.ts
```diff
@@ -418,6 +418,11 @@ export interface OpenConversationRequest {
...
+   */
+  subtext?: string;
```
ydogandjiev: How did we come up with the `subtext` name? I think it might be better to call this `subtitle` or `description`.

debajyoti251088: Loki provided that in the work item. Will sync with him.

TrevorJoelHarris:
This PR is stale and will be closed, please feel free to re-open it if your intention is still to check into main.

PR #316, opened 2020-03-20
Add localeInfo to the Context interface

ydogandjiev:
This change adds the localeInfo property to the Context interface in preparation for enabling the Teams Client to provide this information to our hosted apps so that they can respect the OS date & time format settings when displaying this kind of data. Formatting can be handled via the @microsoft/globe library found here:
https://www.npmjs.com/package/@microsoft/globe

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -142,7 +142,6 @@ export interface LocaleInfo {
   longTime: string;
 }
 
-
 export interface Context {
   /**
    * The Office 365 group ID for the team with which the content is associated.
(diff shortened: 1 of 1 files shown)


src/public/interfaces.ts
```diff
@@ -129,6 +129,19 @@ export interface TeamInformation {
    */
   userTeamRole?: UserTeamRole;
 }
+
+/**
+ * Represents OS locale info used for formatting date and time data
+ */
+export interface LocaleInfo {
+  platform: string;
```
sachingupta, line 137: I think we should support predefined enum list for platform

cloudtx: changed the type to 'windows' | 'macos' to match globe library.

PR #320, opened 2020-04-08
Adding office host domains to valid Origins

sachingupta:
Update Client SDK to support Office.com

### DIFF
--- a/src/internal/constants.ts
+++ b/src/internal/constants.ts
@@ -1,7 +1,9 @@
 import { generateRegExpFromUrls } from './utils';
 
 export const version = '1.6.0';
-
+/**
+ * List of supported Host origins
+ */
 export const validOrigins = [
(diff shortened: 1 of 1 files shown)


src/internal/constants.ts
```diff
@@ -19,6 +19,13 @@ export const validOrigins = [
   'https://outlook.office.com',
   'https://outlook-sdf.office.com',
   'https://*.teams.microsoft.com',
+  'https://www.office.com',
+  'https://word.office.com',
+  'https://excel.office.com',
+  'https://powerpoint.office.com',
+  'https://www.officeppe.com',
+  'https://*.www.office.com',
+  'http://127.0.0.1:5000', // local Development
```
ydogandjiev, line 28: This is a little wacky. I feel like they should eventually move to something like our "https://local.teams.office.com" domain for local development.

PR #331, opened 2020-05-04
Media APIs for webapps

SonalJha:
We are adding a new API select attachment which will enable apps to take image and select image from gallery from the platform


ydogandjiev:
@SonalJha, can you please setup some time to review your spec for this feature with @sachingupta and myself?

SonalJha:
Description

This PR implements three new JS SDK APIs

**Select Media** Through this API a developer can get users to capture/select media from device and share the same with the web-app. In response to selectMedia API, web-app will receive uri of the media and may optionally receive a thumbnail of the selected media. Note: In Phase 1 (Q2 work) only image media type will be supported and support for other media types such as video and audio will come in the future. Use cases:

1. Multi image support
2. Support for additional configuration such image count, scan mode, editing/inking, etc.
3. Generic API definition that's scalable to support other media types in future

**Get Media** To fetch full base64 data for the given uri, web-app will invoke getMedia API. Through this API media files will be broken in smaller chunks  to facilitate transfer. Use cases:

1. Large media support Transfer original file
2. Lossless transfer

**View Images** - Using content uri webapp can view multiple images in native image viewer full screen mode
[Spec: here ](https://microsoft.sharepoint.com/%3Aw%3A/t/ActionPlatform/EdpbC4eEk3VBpheGUyKjdPMBIY5tNEbK0Ke7xY54KWh-jg?e=tiuhLr)
[Design Document ](https://microsoft-my.sharepoint.com/%3Aw%3A/r/personal/sojh_microsoft_com/_layouts/15/guestaccess.aspx?e=SODfiw&share=EY0ZYz8jxxFNhkwepOxdh9oBDdE4ED3i5hNSB0djZ0DjcQ)

Tests :
Did manual testing using a custom app , tested the following :

Verified that selectMedia API is returning back media selected by the user
Verified getMedia API is returning back media using content uri successfully
Verified that viewImages API is opening a multi native image viewer successfully
Verified that different status codes are communicated appropriately for different error cases for all three APIs
Verified all the three APIs have permissions handling according to device permission manager
Verified that the lock is released appropriately and the flow never ends up in a limbo state both in success/failure
Verified that for the scenario in case of these APIs : User leaves the app and comes back, the user is able to resume the flow from where it left off
Verified that for the scenario : User interacting with media APIs, leaves the app, comes back to the app in a different activity (say through deep link), in that case the previously opened activity, in our case the one hosting the JS app is destroyed. And the app has to then be opened and the subsequent flows work as expected.

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+            GlobalVars.getMediaHandler(
+              { errorCode: ErrorCode.INTERNAL_ERROR, message: 'data receieved is null' },
+              null,
+            );
+          }
+        }
+      }
+    }
+
+    GlobalVars.handlers['getMedia' + actionName] = handleGetMediaRequest;
```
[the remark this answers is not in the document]
SonalJha: Yes. We need to send the actionName to the platform in order to distinguish between 2 parallel requests. Here the callback comes back multiple times unlike the other APIs, hence we need a unique identifier. We cannot use the one returned by sendMessageRequestToParent because we need before that to be sent in its params.

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+  const messageId = sendMessageRequestToParent('selectMedia', params);
+  GlobalVars.callbacks[messageId] = callback;
+}
+
+/**
+ * View images using native image viewer
+ * @param uriList urilist of images to be viewed - can be content uri or server url
+ * @param result returns back error if encountered
+ */
+export function viewImages(uriList: ImageUri[], callback: (error?: SdkError) => void): void {
```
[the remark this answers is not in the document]
SonalJha: There will be no callback in case of success

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+   * Optional; indicate if image filtering mode is enabled on the selected image
+   * Default value is false
+   */
+  enableFilter?: boolean;
+}
+
+/**
+ * The modes in which camera can be launched in select Media API
+ */
+export const enum Mode {
```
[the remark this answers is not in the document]
SonalJha: It will be called as Imageprops.Mode, for other media types we can have audioProps.audioModes

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+  type: ImageUriType;
+}
+
+/**
+ * ID contains a mapping for content uri on platform's side, URL is generic
+ */
+export const enum ImageUriType {
+  ID = 1,
+  URL = 2,
+}
```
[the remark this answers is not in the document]
SonalJha: adding URL in fileFormat wouldn't be proper because URL is not a file format. Also we will be making the File class open-ended. We need URL only in case if viewImages so keeping it separately makes more sense to me.

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+   * A preview of the file which is a lightweight representation.
+   * In case of images this will be a thumbnail in base64 encoding.
+   */
+  public preview: string;
+
+  /**
+   * Gets the media in chunks irrespecitve of size, these chunks are assembled and sent back to the webapp as file/blob
+   * @param callback returns blob of media
+   */
+  public getMedia(callback: (error: SdkError, blob: Blob) => void): void {
```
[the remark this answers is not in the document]
SonalJha: media.getMedia() sounds better than getContent() or getBlob(). getContent() is not good because we have a member named content in Media object and getBlob() is kind of evident as the return param is blob itself. Also I am kind of pushing back on changing names unless must have, at this point of time because the name changes will have to be propagated in all the three platforms, in all design docs and wiki pages and test apps.

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
   const messageId = sendMessageRequestToParent('captureImage');
   GlobalVars.callbacks[messageId] = callback;
 }
+
+/**
+ * Media object returned by the select Media API
+ */
+export class Media extends File {
```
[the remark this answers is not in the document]
SonalJha: Media class is the superset of File class. File owners can anytime onboard to media class. Didn't want to add a lot of testing effort by changing existing File class hence building on top of that. There is no duplication so keeping two is also okay with me.

src/public/index.ts
```diff
@@ -37,4 +37,4 @@ export {
 export { settings } from './settings';
 export { tasks } from './tasks';
 export { ChildAppWindow, IAppWindow, ParentAppWindow } from './appWindow';
-export { captureImage, File, FileFormat } from './media';
+export { captureImage, File, FileFormat, selectMedia, viewImages, Media } from './media';
```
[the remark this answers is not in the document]
SonalJha: Media class is the superset of File class. File owners can anytime onboard to media class. Didn't want to add a lot of testing effort by changing existing File class hence building on top of that. There is no duplication so keeping two is also okay with me.

src/internal/globalVars.ts
```diff
@@ -34,4 +34,5 @@ export class GlobalVars {
   public static onStartConversationHandler: (conversationResponse: ConversationResponse) => void;
   public static onCloseConversationHandler: (conversationResponse: ConversationResponse) => void;
   public static getLogHandler: () => string;
+  public static getMediaHandler: (error: SdkError, blob: Blob) => void;
```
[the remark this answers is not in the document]
SonalJha: removed. Good catch.

src/public/media.ts
```diff
@@ -77,3 +81,233 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+   */
+  public getMedia(callback: (error: SdkError, blob: Blob) => void): void {
+    ensureInitialized(frameContexts.content, frameContexts.task);
+    let actionName = generateGUID();
+    let helper: MediaHelper = {
+      mediaMimeType: this.mimeType,
+      assembleAttachment: [],
+    };
+    const params = [actionName, this.content];
+    GlobalVars.getMediaHandler = callback;
```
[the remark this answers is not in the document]
SonalJha: removed. Good catch.

src/internal/mediaUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+  return file;
+}
+
+/**
+ * Helper function to convert Media chunks into another object type which can be later assemebled
+ * Converts base 64 encoded string to byte array and then into an array of blobs
+ */
+export function decodeAttachment(attachment: MediaChunk, mimeType: string): AssembleAttachment {
+  let decoded = atob(attachment.chunk);
```
ydogandjiev, line 28: NIT: `let` --> `const`

ydogandjiev: Same for the rest of the variables in this method

src/internal/mediaUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+    sequence: attachment.chunkSequence,
+    file: blob,
+  };
+  return assemble;
+}
+
+/**
+ * Returns true if the mediaInput params are valid and false otherwise
+ */
+export function validSelectMediaInputs(mediaInputs: MediaInputs): boolean {
```
ydogandjiev, line 45: NIT: `validSelectMediaInputs ` --> `validateSelectMediaInputs `

ydogandjiev: Same for the rest of the methods below

src/public/media.ts
```diff
@@ -77,3 +92,265 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+    if (!isAPISupportedByPlatform(mediaAPIVersion)) {
+      const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
+      callback(oldPlatformError, null);
+      return;
+    } else if (!validGetMediaInputs(this.mimeType, this.format, this.content)) {
+      const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
+      callback(invalidInput, null);
+      return;
+    }
+    let actionName = generateGUID();
```
ydogandjiev, line 124: NIT: `let` --> `const`

ydogandjiev: Same for `helper` and `mediaResult`

src/public/media.ts
```diff
@@ -77,3 +92,265 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+  Gallery = 2,
+}
+
+/**
+ * Specifies the type of Media
+ */
+export const enum MediaType {
+  Image = 1,
+  Video = 2,
+  ImageAndVideo = 3,
```
ydogandjiev, line 241: NIT: Shouldn't this be `ImageOrVideo`?

test/public/media.spec.ts
```diff
@@ -162,4 +164,409 @@ describe('media', () => {
...
+    media.mimeType = "image/jpeg";
+    media.format = FileFormat.ID;
+    media.getMedia((error: SdkError, blob: Blob) => {
+      mediaError = error;
+    });
+    expect(mediaError).not.toBeNull();
+    expect(mediaError.errorCode).toBe(ErrorCode.OLD_PLATFORM);
+  });
+
+  it('getmedia call in task frameContext works', () => {
```
ydogandjiev, line 391: NIT: `getmedia` --> `getMedia`

src/public/media.ts
```diff
@@ -77,3 +92,265 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+    const params = [actionName, this.content];
+    this.content && callback && sendMessageRequestToParent('getMedia', params);
+    function handleGetMediaRequest(response: string): void {
+      if (callback) {
+        let mediaResult: MediaResult = JSON.parse(response);
+        if (mediaResult.error) {
+          callback(mediaResult.error, null);
+        } else {
+          if (mediaResult.mediaChunk) {
+            if (mediaResult.mediaChunk.chunkSequence <= 0) {
```
ydogandjiev, line 138: Why are there two cases here? Do we have two different ways of transmitting the files across? Please add a comment explaining that.

src/internal/mediaUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  });
+  return file;
+}
+
+/**
+ * Helper function to convert Media chunks into another object type which can be later assemebled
+ * Converts base 64 encoded string to byte array and then into an array of blobs
+ */
+export function decodeAttachment(attachment: MediaChunk, mimeType: string): AssembleAttachment {
```
ydogandjiev, line 27: I don't see any UTs for the logic in this API. Can you please add some?

src/internal/mediaUtil.ts
```diff
@@ -0,0 +1,70 @@
+import { AssembleAttachment, MediaChunk, MediaInputs, FileFormat, ImageUri } from '../public/media';
+
+/**
+ * Helper function to create a blob from media chunks based on their sequence
+ */
+export function createFile(assembleAttachment: AssembleAttachment[], mimeType: string): Blob {
```
ydogandjiev, line 6: I don't see any UTs for the logic in this API. Can you please add some?

src/public/media.ts
SonalJha: added

src/public/media.ts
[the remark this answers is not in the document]
anandvnath: Agreed. But since this is defined as an independent enum, how do you propose we add the corresponding ones for audio or video? What would those enum be called?

src/public/media.ts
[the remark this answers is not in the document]
anandvnath: When will these be unregistered? I do not see the code that removes these.

src/public/media.ts
```diff
@@ -2,48 +2,63 @@ import { GlobalVars } from '../internal/globalVars';
...
 
 /**
  * This is the SDK version when captureImage API is supported on mobile.
  */
 const captureImageMobileSupportVersion = '1.7.0';
 
+/**
+ * This is the SDK version when media APIs is supported on all three platforms ios, android and web.
+ */
+const mediaAPIVersion = '1.8.0';
```
anandvnath, line 22: Can you also please verify that the version number in this PR in constants.ts is upgraded to '1.8.0-beta'?

sachingupta, line 22: rename to mediaAPISupportVersion

src/public/media.ts
```diff
@@ -77,3 +92,268 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+   * A preview of the file which is a lightweight representation.
+   * In case of images this will be a thumbnail/compressed image in base64 encoding.
+   */
+  public preview: string;
+
+  /**
+   * Gets the media in chunks irrespecitve of size, these chunks are assembled and sent back to the webapp as file/blob
+   * @param callback returns blob of media
+   */
+  public getMedia(callback: (error: SdkError, blob: Blob) => void): void {
```
sachingupta, line 110: is this callback use only for failure case. We are thinking to support promise based sdk API in future. It will be better to update this api public getMedia(successCallback: (blob: Blob) => void, failureCallback: (error: SdkError) => void):

src/public/media.ts
```diff
@@ -77,3 +92,268 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+  Photo = 1,
+  Document = 2,
+  Whiteboard = 3,
+  BusinessCard = 4,
+}
+
+/**
+ * Specifies the image source
+ */
+export const enum Source {
```
sachingupta, line 233: MediaSources ?

src/public/media.ts
```diff
@@ -77,3 +92,268 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+  mediaMimeType: string;
+  assembleAttachment: AssembleAttachment[];
+}
+
+/**
+ * Select an attachment using camera/gallery
+ * @param mediaInputs The input params to customize the media to be selected
+ * @param callback The callback to invoke after fetching the media
+ */
+export function selectMedia(mediaInputs: MediaInputs, callback: (error: SdkError, attachments: Media[]) => void): void {
```
sachingupta, line 315: same here support success and failure separate callback.

src/public/media.ts
```diff
@@ -77,3 +92,268 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
...
+  const params = [mediaInputs];
+  const messageId = sendMessageRequestToParent('selectMedia', params);
+  GlobalVars.callbacks[messageId] = callback;
+}
+
+/**
+ * View images using native image viewer
+ * @param uriList urilist of images to be viewed - can be content uri or server url. supports upto 10 Images in one go
+ * @param result returns back error if encountered, there will be no callback in case of success
+ */
```
sachingupta, line 339: same here support success and failure in callback explicit.

src/public/media.ts
SonalJha: After we introduced SdkError we had decided we will keep error as the 1st param of callback and remaining subsequently. When we move to promises we can do something like: Promise.resolve([error, data]). Right now in case of error we return error and in case of success we return blob.

src/public/media.ts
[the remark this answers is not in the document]
SonalJha: For video, Mode will be used as it is. All image props are applicable for vidoes. For audio we can use AudioMode maybe? I am sorry I don't want to rename variables at this point unless we see it as urgent as the changes will need to be propagated in all the three platforms.

src/public/media.ts
SonalJha: This is more like imageSource because for other medias we won't have same set of sources. However as explained above not taking up renaming changes unless must have. At the time of calling it will be called up as ImageProps.sources

src/public/media.ts
SonalJha: After we introduced SdkError we had decided we will keep error as the 1st param of callback and remaining subsequently. When we move to promises we can do something like: Promise.resolve([error, data]). Right now in case of error we return error and in case of success we return array of attachments.

src/public/media.ts
SonalJha: We don't need a success callback here. Default case is assumed as success.

src/public/media.ts
SonalJha: done

src/public/media.ts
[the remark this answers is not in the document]
SonalJha: done

PR #335, opened 2020-05-14
Introducing device API for image

anandvnath:
Added device namespace for hosting Android and iOS specific API. 
Added getImages API, related interfaces, documentation and tests.


src/public/device.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Launch camera, capture image or choose image from gallery and return the images as a File[] object to the callback
+   * File object returned will have status codes to call out error scenarios
+   * App can check against the StatusCode enum to find out the exact cause and present
+   * the user with appropriate error message. App should also do a `statusCode == Success` check before accessing the content
+   * Note: Currently we support getting one File through this API, i.e. the file arrays size will be one
+   * @see File
+   */
+  export function getImages(callback: (files: File[]) => void): void {
```
ydogandjiev, line 85: How are we planning to handle errors here?

src/public/device.ts
```diff
@@ -0,0 +1,93 @@
...
+   * The object will have "statusCode" as defined in StatusCode enum
+   * The user of this API should check "statusCode" to be "Success" before reading the "content"
+   * In case of other status codes, app can decide to show an error message to the user
+   * @see StatusCode
+   */
+  export interface File {
+    /**
+     * Status code
+     */
+    statusCode: StatusCode;
```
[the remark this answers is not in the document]
anandvnath: The error is associated with a File because if user selected more than one file and if we are able to return one file and we encountered problem while compressing / processing other file, it will be error per File.Returning a top level error and not responding with partial response is another option, which will move this error out to top level.

src/public/device.ts
```diff
@@ -0,0 +1,93 @@
+import { GlobalVars } from '../internal/globalVars';
+import { ensureInitialized, sendMessageRequestToParent } from '../internal/internalAPIs';
+import { frameContexts } from '../internal/constants';
+
+/**
+ * Namespace for device related APIs. These are implemented only for mobile (Android and iOS)
+ * On desktop, these APIs will be a noop.
```
pankajmi2, line 7: Shouldn't JS SDK return error if webapp invokes it for Desktop platform ?

src/public/device.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export namespace device {
+  /**
+   * Status codes for device APIs
+   * @see File
+   */
+  export enum StatusCode {
```
pankajmi2, line 14: Can we keep an error object in top level response object ? Like we discussed in API versioning review meeting, also in review here - https://github.com/OfficeDev/microsoft-teams-library-js/pull/326
It also tries to define some standard error codes which can be used across SDK APIs. Would recommend we use same codes here.

test/framelessPostUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+    let domEvent = {
+      data: {
+        id: message.id,
+        args: args,
+      } as MessageResponse,
+    } as DOMMessageEvent;
+    (window as ExtendedWindow).onNativeMessage(domEvent);
+  };
+
+  public sendMessageFromNativeToParent = (event: DOMMessageEvent): void => {
```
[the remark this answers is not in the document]
anandvnath: Only parameter is different. Will fix this.

src/public/device.ts
```diff
@@ -0,0 +1,93 @@
...
+   * App can check against the StatusCode enum to find out the exact cause and present
+   * the user with appropriate error message. App should also do a `statusCode == Success` check before accessing the content
+   * Note: Currently we support getting one File through this API, i.e. the file arrays size will be one
+   * @see File
+   */
+  export function getImages(callback: (files: File[]) => void): void {
+    if (!callback) {
+      throw new Error('[device.getImages] Callback cannot be null');
+    }
+    ensureInitialized(frameContexts.content);
```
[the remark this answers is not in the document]
anandvnath: Done

src/public/device.ts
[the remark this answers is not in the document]
anandvnath: Modified signature to have the callback as (error, file[])

src/public/device.ts
anandvnath: Added error to callback signature.

src/public/device.ts
anandvnath: I changed this to ErrorCode. Yes this is similar to what we are trying to do in versioning. Once that is checked in, we can migrate the ErrorCode to single place. As of now keeping it here. Also I think we won't need an Error object, rather an ErrorCode should suffice to convey the error. App can show appropriate text to the user. Please see my comment in PR #326

src/public/device.ts
anandvnath: Yes. I will check the behavior and do the needful.

src/internal/constants.ts
```diff
@@ -6,6 +6,11 @@ export const version = '1.6.0';
  */
 export const defaultSDKVersionForCompatCheck = '1.6.0';
 
+/**
+ * This is the SDK version when captureImage API is supported on mobile.
+ */
+export const captureImageMobileSupportVersion = '1.7.0';
```
pankajmi2, line 12: Can we directly consume '1.7.0' in captureImage API without declaring a constant for this. Doesn't look like it will be used anywhere else and going forward when we introduce new APIs, these constants will also grow in number.

src/public/interfaces.ts
```diff
@@ -527,12 +527,21 @@ export interface SdkError {
   errorCode: ErrorCode;
   /**
   Optional description for the error. This may contain useful information for web-app developers.
-  This string will not be localized and is not for end-user consumption.
+  This string will not be localized and is not for end-user consumption. 
+  App should not dependend on the string content. The exact value may change. This is only for debugging purposes.
```
pankajmi2, line 531: typo - "App should not be **dependent** on"

src/public/interfaces.ts
```diff
@@ -527,12 +527,21 @@ export interface SdkError {
...
   message?: string;
 }
 
 export const enum ErrorCode {
+  /**
+   * API not supported in the current platform.
+   */
+  NOT_SUPPORTED_ON_PLATFORM = 100,
+  /**
+   * Internal error encountered while performing the required operation.
```
pankajmi2, line 542: Do we have a scenario where we are passing it from the platform code ? 

If the scenario/error is not captured by existing codes, could we define more specific error instead of INTERNAL_ERROR as this gives very little information to developer regarding what went wrong in the handling of the API. Also, dev wouldn't know if the error can be rectified by the user and whether he can retry the operation etc.

src/public/media.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Callback will be called with an error, if there are any. App should first check the error.
+ * If it is present the user can be updated with appropriate error message.
+ * If error is null or undefined, then files will have the required result.
+ * Note: Currently we support getting one File through this API, i.e. the file arrays size will be one.
+ * Note: For desktop, this API is not supported. Callback will be resolved with ErrorCode.NotSupported.
+ * @see File
+ * @see SdkError
+ */
+export function captureImage(callback: (error: SdkError, files: File[]) => void): void {
+  if (!callback) {
```
pankajmi2, line 55: I think we concluded that this API will always support only one image. For multi-image and more advanced support, we would introduce another API. So, should we keep single File object in response instead of an array. Since, we never plan to enhance it beyond single image.

src/public/media.ts
```diff
@@ -0,0 +1,74 @@
...
+    throw new Error('[captureImage] Callback cannot be null');
+  }
+  ensureInitialized(frameContexts.content, frameContexts.task);
+
+  if (!GlobalVars.isFramelessWindow) {
+    let notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
```
pankajmi2, line 61: nit: const

src/public/media.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * File object that can be used to represent image or video or audio
+ */
+export interface File {
+  /**
+   * Content of the file
+   * App needs to convert this to dataUrl, if this must be used directly in HTML tags
+   */
+  content?: string;
```
pankajmi2, line 21: Shouldn't these be mandatory fields in File ?

src/public/interfaces.ts
SonalJha: We should keep either internal error or generic error because the exact cause of all errors are not likely to be known

ydogandjiev:
@sachingupta, can you please take a look at this PR when you get a chance?

src/public/interfaces.ts
```diff
@@ -527,12 +527,21 @@ export interface SdkError {
...
+  App should not dependend on the string content. The exact value may change. This is only for debugging purposes.
   */
   message?: string;
 }
 
 export const enum ErrorCode {
+  /**
+   * API not supported in the current platform.
+   */
+  NOT_SUPPORTED_ON_PLATFORM = 100,
```
samyuktha-tamvada, line 540: There is already an existing error code OLD_PLATFORM = 9000 which checks if platform supports the API or not. This seems to be redundant. What is the difference between this new error code and existing error code 9000?

src/internal/constants.ts
anandvnath: Done. Moved this to media.ts

src/public/interfaces.ts
anandvnath: Yes. The intention here is to let end user know something went wrong, knowing details of which may not make sense to him. Error can contain message which will contain debug information for dev. Keeping this as INTERNAL_ERROR

src/public/media.ts
anandvnath: Yes. Initially we had error with every file, at which point everything was non-mandatory. Now that we moved error out, few of these are mandatory. Thanks for calling this out.

src/public/media.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export function captureImage(callback: (error: SdkError, files: File[]) => void): void {
+  if (!callback) {
+    throw new Error('[captureImage] Callback cannot be null');
+  }
+  ensureInitialized(frameContexts.content, frameContexts.task);
+
+  if (!GlobalVars.isFramelessWindow) {
+    let notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
+    callback(notSupportedError, undefined);
```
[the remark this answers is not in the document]
anandvnath: For Javascript yes, but looks like typescript needs all parameters passed.

src/public/media.ts
```diff
@@ -0,0 +1,74 @@
...
+    return;
+  }
+
+  if (!isAPISupportedByPlatform(captureImageMobileSupportVersion)) {
+    let oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
+    callback(oldPlatformError, undefined);
```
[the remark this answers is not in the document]
anandvnath: For Javascript yes, but looks like typescript needs all parameters passed.

src/public/media.ts
anandvnath: Keeping this as array.

PR #351, opened 2020-06-17
Introducing Location APIs

pankajmi2:
**Scenario:** In Teams web-apps, if users need functionality to choose location from map, currently Teams SDK doesn't expose any APIs to facilitate that user need. Through this feature, we are adding two location APIs to address this need.

**Changes done:** We are introducing below two location APIs

**getLocation:** Through this API, based on input location props, web-app will be able to capture either current user location or will show location picker to the user where user can choose a location

**showLocation:** Through this API, web-app will ask the native Teams app to show location on map.

Both these APIs will have minimum required version of 1.9.0


pankajmi2:
@ydogandjiev, @anandvnath, @sachingupta - could you please take a look at this PR.

sachingupta (approved):

src/public/location.ts
```diff
@@ -0,0 +1,93 @@
+import { GlobalVars } from '../internal/globalVars';
+import { SdkError, ErrorCode } from './interfaces';
+import { ensureInitialized, sendMessageRequestToParent, isAPISupportedByPlatform } from '../internal/internalAPIs';
+import { FrameContexts } from './constants';
+
+/**
+ * This is the SDK version when location APIs (getLocation and showLocation) are supported.
+ */
```
sachingupta, line 8: Generally we follow nameSpace approach to club similar APIs together. Can you please move Location related API under Location namespace. please take a look at this https://github.com/OfficeDev/microsoft-teams-library-js/blob/master/src/public/settings.ts

sachingupta (changes requested):

src/public/location.ts
pankajmi2: @sachingupta, moved the APIs under location namespace.

src/public/location.ts
```diff
@@ -0,0 +1,95 @@
...
+      const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
+      callback(oldPlatformError, undefined);
+      return;
+    }
+    if (!props) {
+      const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
+      callback(invalidInput, undefined);
+      return;
+    }
+    const messageId = sendMessageRequestToParent('getLocation', [props]);
```
sachingupta, line 67: Please update API to locaton.getLocation

src/public/location.ts
```diff
@@ -0,0 +1,95 @@
...
+      const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
+      callback(oldPlatformError, undefined);
+      return;
+    }
+    if (!location) {
+      const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
+      callback(invalidInput, undefined);
+      return;
+    }
+    const messageId = sendMessageRequestToParent('showLocation', [location]);
```
sachingupta, line 92: same here. "location.showLocation"

sachingupta (changes requested):

src/public/location.ts
```diff
@@ -0,0 +1,93 @@
...
+    const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
+    callback(oldPlatformError, undefined);
+    return;
+  }
+  if (!location) {
+    const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
+    callback(invalidInput, undefined);
+    return;
+  }
+  const messageId = sendMessageRequestToParent('showLocation', [location]);
```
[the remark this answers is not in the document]
pankajmi2: I was trying to avoid changes on platform side by just changing developer facing name of the API, but now, I have changed the name of API with which invocation happens on the platform side as well.

src/public/location.ts
pankajmi2: done

src/public/location.ts
pankajmi2: done

src/public/location.ts
```diff
@@ -0,0 +1,93 @@
...
+    const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
+    callback(oldPlatformError, undefined);
+    return;
+  }
+  if (!props) {
+    const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
+    callback(invalidInput, undefined);
+    return;
+  }
+  const messageId = sendMessageRequestToParent('getLocation', [props]);
```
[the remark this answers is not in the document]
pankajmi2: done

sachingupta (approved):

PR #362, opened 2020-07-14
Sopurand/add stage view const to frame context

soorajcodehub:
Adding a new const for the stageViee to the FrameContext.

### DIFF
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -14,6 +14,7 @@ export const enum FrameContexts {
   remove = 'remove',
   task = 'task',
   sidePanel = 'sidePanel',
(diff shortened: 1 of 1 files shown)


src/public/constants.ts
```diff
@@ -14,6 +14,7 @@ export const enum FrameContexts {
   remove = 'remove',
   task = 'task',
   sidePanel = 'sidePanel',
+  stageView = 'stageView',
```
debajyoti251088, line 17: NIT: should we just make it 'stage'?

anindara: If we change the const then we will need to make a corresponding change in the consumer. Not a problem, just keep us updated.

PR #363, opened 2020-07-14
Allow the executeDeepLink API to be called from the tab configuration dialog

ydogandjiev:
This change allows the `microsoftTeams.executeDeepLink` API to be called from the tab configuration dialog (i.e. `settings` frame context)

### DIFF
--- a/src/public/publicAPIs.ts
+++ b/src/public/publicAPIs.ts
@@ -352,7 +352,7 @@ export function shareDeepLink(deepLinkParameters: DeepLinkParameters): void {
  * @param deepLink deep link.
  */
 export function executeDeepLink(deepLink: string, onComplete?: (status: boolean, reason?: string) => void): void {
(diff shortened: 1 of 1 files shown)


src/public/publicAPIs.ts
```diff
@@ -352,7 +352,7 @@ export function shareDeepLink(deepLinkParameters: DeepLinkParameters): void {
  * @param deepLink deep link.
  */
 export function executeDeepLink(deepLink: string, onComplete?: (status: boolean, reason?: string) => void): void {
-  ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task);
+  ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.settings, FrameContexts.task);
```
sachingupta, line 355: what will happen if deepLink change the route ? will it close setting dialog ?

ydogandjiev: Yes, I'm making the necessary change on the client to close the config dialog when executeDeepLink is called.

PR #372, opened 2020-08-05
Fix the type in Media[] returned in selectMedia callback

anandvnath:
This is a minor fix on top of the selectMedia API. When objects are sent over from native layer (Android or iOS) they are just objects and do not have a type. In selectMedia callback, we promise to deliver Media objects. Sending the objects as received from native layer does not really send Media objects, they do ducktype Media, but will be missing getMedia method.

The fix is to have one level of indirection, receive the array from native in our callback and create new Media objects which confirms with the promise and return those.

### DIFF
--- a/src/public/media.ts
+++ b/src/public/media.ts
@@ -97,6 +97,18 @@ export function captureImage(callback: (error: SdkError, files: File[]) => void)
  * Media object returned by the select Media API
  */
 export class Media extends File {
+  constructor(that: Media = null) {
+    super();
+    if (that) {
+      this.content = that.content;
+      this.format = that.format;
+      this.mimeType = that.mimeType;
+      this.name = that.name;
+      this.preview = that.preview;
+      this.size = that.size;
+    }
+  }
+
   /**
    * A preview of the file which is a lightweight representation.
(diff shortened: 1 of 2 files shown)


src/public/media.ts
```diff
@@ -350,7 +350,25 @@ export function selectMedia(mediaInputs: MediaInputs, callback: (error: SdkError
...
   const messageId = sendMessageRequestToParent('selectMedia', params);
-  GlobalVars.callbacks[messageId] = callback;
+
+  // What comes back from native at attachments would just be objects and will be missing getMedia method on them.
+  GlobalVars.callbacks[messageId] = (err: SdkError, attachs: Media[]) => {
+    let mediaArray: Media[] = null;
+    if (attachs) {
+      mediaArray = [];
+      for (let attachment of attachs) {
+        const media = new Media();
```
ydogandjiev, line 360: Why not add a constructor that just takes an attachment? You could even store that internally rather than copying each prop one by one.

anandvnath: Yes. It makes sense. I moved it into a constructor. Unfortunately I cannot use Object.assign as we are targeting es5. So had to copy properties one by one. Please let me know if there is a better way to do this.

PR #379, opened 2020-08-13
comment updated for viewImages callback and size check added for getMedia API

SonalJha:
1. viewImages API returns null error callback in case of success - updated comment
2. getMedia API is capped at 50 MB

### DIFF
--- a/src/internal/mediaUtil.ts
+++ b/src/internal/mediaUtil.ts
@@ -1,5 +1,7 @@
 import { AssembleAttachment, MediaChunk, MediaInputs, FileFormat, ImageUri } from '../public/media';
 
+export const fiftyMbInKb = 50 * 1024;
+
 /**
  * Helper function to create a blob from media chunks based on their sequence
  */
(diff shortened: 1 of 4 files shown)


src/internal/mediaUtil.ts
```diff
@@ -1,5 +1,7 @@
 import { AssembleAttachment, MediaChunk, MediaInputs, FileFormat, ImageUri } from '../public/media';
 
+export const fiftyMbInKb = 50 * 1024;
```
anandvnath, line 3: Isn't this an ECS controlled item? If so, the ECS will not have any effect, if we gate this at JS SDK level by hard coding. I

SonalJha: We are planning to introduce another function getMediaInChunks in JS library which will not assemble the media. The benefit will be we will be able to upload big videos from low end devices also. Developer doesn't need to assemble them, she can keeps uploading them to azure in chunks and get the url for viewing. getMedia and getMediaInChunks will use the exact same implementation underneath.

PR #380, opened 2020-08-14
updating comment for view images

SonalJha:
updating comment for view images


src/public/media.ts
```diff
@@ -380,7 +380,7 @@ export function selectMedia(mediaInputs: MediaInputs, callback: (error: SdkError
 /**
  * View images using native image viewer
  * @param uriList urilist of images to be viewed - can be content uri or server url. supports upto 10 Images in one go
- * @param result returns back error if encountered, there will be no callback in case of success
+ * @param callback returns back error if encountered, returns null in case of success
```
anandvnath, line 383: Just think its worth mentioning that the success case will be after user is done with viewing images, so that apps are clear on what to expect.

PR #407, opened 2020-09-09
App button hover enter and leave APIs

cloudtx:
Previous single app button hover api was not sufficient. Both a hover enter and hover leave API are now added to allow the sdk consumer to engage functionality both when entering hover and leaving hover.

### DIFF
--- a/src/public/publicAPIs.ts
+++ b/src/public/publicAPIs.ts
@@ -230,15 +230,27 @@ export function registerAppButtonClickHandler(handler: () => void): void {
 }
 
 /**
- * Registers a handler for hovering the app button.
+ * Registers a handler for entering hover of the app button.
  * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
- * @param handler The handler to invoke when the personal app button is hovered in the app bar.
+ * @param handler The handler to invoke when entering hover of the personal app button in the app bar.
  */
-export function registerAppButtonHoverHandler(handler: () => void): void {
+export function registerAppButtonHoverEnterHandler(handler: () => void): void {
   ensureInitialized();
 
-  GlobalVars.appButtonHoverHandler = handler;
-  handler && sendMessageRequestToParent('registerHandler', ['appButtonHover']);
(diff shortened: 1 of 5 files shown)


src/public/publicAPIs.ts
```diff
@@ -230,15 +230,27 @@ export function registerAppButtonClickHandler(handler: () => void): void {
...
+  handler && sendMessageRequestToParent('registerHandler', ['appButtonHoverEnter']);
+}
+
+/**
+ * Registers a handler for exiting hover of the app button.
+ * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
+ * @param handler The handler to invoke when exiting hover of the personal app button in the app bar.
+ */
+export function registerAppButtonHoverLeaveHandler(handler: () => void): void {
+  ensureInitialized();
```
ydogandjiev, line 250: Same here; pass `FrameContext.content` to the `ensureInitialized();` call so that developers get notified when their trying to use the API incorrectly.

src/public/publicAPIs.ts
```diff
@@ -230,15 +230,27 @@ export function registerAppButtonClickHandler(handler: () => void): void {
...
 /**
- * Registers a handler for hovering the app button.
+ * Registers a handler for entering hover of the app button.
  * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
- * @param handler The handler to invoke when the personal app button is hovered in the app bar.
+ * @param handler The handler to invoke when entering hover of the personal app button in the app bar.
  */
-export function registerAppButtonHoverHandler(handler: () => void): void {
+export function registerAppButtonHoverEnterHandler(handler: () => void): void {
   ensureInitialized();
```
ydogandjiev, line 238: I think we should limit the frame contexts from which this API can be called since it probably won't work as expected everywhere? I feel like this should only be supported in the `FrameContexts.content` for starters.

ydogandjiev (approved):

src/public/publicAPIs.ts
cloudtx: This should be applied to the click handler as well correct?

PR #411, opened 2020-09-15
Adding namespace to Media APIs

SonalJha:


ydogandjiev (approved):

src/public/media.ts
```diff
@@ -11,395 +11,400 @@ import {
...
-        } else {
-          if (mediaResult.mediaChunk) {
-            // If the chunksequence number is less than equal to 0 implies EOF
-            // create file/blob when all chunks have arrived and we get 0/-1 as chunksequence number
-            if (mediaResult.mediaChunk.chunkSequence <= 0) {
-              const file = createFile(helper.assembleAttachment, helper.mediaMimeType);
-              callback(mediaResult.error, file);
-              GlobalVars.handlers['getMedia' + actionName] = null;
+
+    const messageId = sendMessageRequestToParent('captureImage');
```
sachingupta, line 93: We need to update this name also with prefix as media "media.captureImage" . Please do it for all APIs in this file. On mobile side to make it backward compatible, We have to support both "captureImage" and "media.captureImage" for some time.

sachingupta (changes requested):

src/public/media.ts
ydogandjiev: If we do that it will be a breaking change for the SDK though and the new version wouldn't be usable on the old mobile clients that haven't been updated with this change. I think it might be safest to leave it as is if any partners have already taken a dependency on these APIs.

sachingupta: sounds good. In that case we can keep existing APIs name as it is and we can put media prefix for all new APIs that we add in future.

sachingupta (approved):

PR #425, opened 2020-10-05
ReturnFocus API 

cloudtx:
ReturnFocus api added. This api will move focus to the Teams search bar or appbar depending on the direction specified.


src/public/publicAPIs.ts
```diff
@@ -393,6 +393,16 @@ export function executeDeepLink(deepLink: string, onComplete?: (status: boolean,
   GlobalVars.callbacks[messageId] = onComplete ? onComplete : getGenericOnCompleteHandler();
 }
 
+/**
+ * Return focus to the main Teams app. Will focus search bar if navigating foward and app bar if navigating back.
+ * @param navigateForward Determines the direction to focus in teams app.
+ */
+export function returnFocus(navigateForward?: boolean): void {
```
sachingupta, line 400: Should we put under navigation namespace.

sachingupta (approved):

src/public/publicAPIs.ts
cloudtx: Added navigation namespace

src/public/navigation.ts
```diff
@@ -0,0 +1,16 @@
+import { ensureInitialized, sendMessageRequestToParent } from '../internal/internalAPIs';
+
+/**
+ * Namespace to interact with the navigation specific part of the SDK.
+ */
+export namespace navigation {
```
ydogandjiev, line 6: @sachingupta, how do you feel about this namespace? Today we have navigation APIs at the top level (e.g. navigateBack) so is the idea that we would move them into here at some point? Or should we just put this API at the top level since it's part of the "core" functionality?

ydogandjiev (approved):

src/public/navigation.ts
sachingupta: I spend some time thinking about place for returnFocus API. Lets go with putting return focus at top level so developer call it using microsoftTeams.returnFocus because

	
	Many other navigation APIs are already at top level - we can't move those under namespace without breaking changes.
	
	
	As Yuri also mentioned we can treat all navigation APIs as core APIs.
	
For better code organization Lets move navigation related APIs in a separate file - called navigationAPIs.ts without using namespace.

sachingupta (approved):

PR #442, opened 2020-10-26
Add API for toggle speaker audio

ronakbakshi:
This is the initial work on the disable audio API. This still has pending unit tests to add. Please provide initial feedback.

### DIFF
--- a/src/private/meetingAudio.ts
+++ b/src/private/meetingAudio.ts
@@ -19,7 +19,7 @@ export namespace meetingAudio {
    * Hide from docs
    * ------
    * Allows an app to toggle the incoming audio speaker setting for the meeting user from mute to unmute or vice-versa
-   * @param callback Callback contains 1 parameter of type IToggleClientAudio. This parameter is an object which contains error or type SdkError or null and result of type boolean or null 
+   * @param callback Callback contains 1 parameter of type IToggleClientAudio. This parameter is an object which contains error or type SdkError or null and result of type boolean or null
    */
   export function toggleIncomingClientAudio(callback: (response: IToggleClientAudio) => void): void {
     if (!callback) {
--- a/src/private/meetingAudio.ts
+++ b/src/private/meetingAudio.ts
@@ -3,8 +3,15 @@ import { GlobalVars } from '../internal/globalVars';
 import { SdkError } from '../public/interfaces';
 
 export namespace meetingAudio {
-  interface ToggleClientAudioReturnType {
+  export interface IToggleClientAudio {
+    /**
+     * error can either contain an error of type SdkError, incase of an error, or null when toggle is successful
+     */
     error: SdkError | null;
+    /**
+     * result can either contain the true/false value, incase of a successful toggle or null when the toggling fails
+     * result: True means incoming audio is muted and false means incoming audio is unmuted
+     */
     result: boolean | null;
   }
   /**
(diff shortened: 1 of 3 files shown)


ronakbakshi:
This PR is ready for review. I have added an API that provides 3P developers to be able to toggle the incoming client audio.

src/private/meetingAudio.ts
```diff
@@ -0,0 +1,45 @@
+import { ensureInitialized, sendMessageRequestToParent } from '../internal/internalAPIs';
+import { GlobalVars } from '../internal/globalVars';
+
+export namespace meetingAudio {
+  export interface IToggleResponse {
```
ydogandjiev, line 5: Instead of doing this can you please follow the pattern of the new media APIs like captureImage, getMedia, etc. You can see them here:
https://github.com/OfficeDev/microsoft-teams-library-js/blob/master/src/public/media.ts

Instead of the developer passing an object with a successCallback and failureCallback we now pass a single callback which takes two parameters - error and result. If we hit any issues then an error is returned and null result. Otherwise, a null error and valid result are returned.

ydogandjiev (changes requested):

src/private/meetingAudio.ts
```diff
@@ -0,0 +1,34 @@
...
+  export function toggleIncomingClientAudio(callback: (error: SdkError, result: boolean | null) => void): void {
+    if (!callback) {
+      throw new Error('[toggle incoming client audio] Callback cannot be null');
+    }
+    ensureInitialized();
+    const messageId = sendMessageRequestToParent('toggleIncomingClientAudio');
+    GlobalVars.callbacks[messageId] = (error: SdkError | null, result: boolean | null) => {
+      if (callback) {
+        if (error && error.errorCode) {
+          const internalError: SdkError = { errorCode: ErrorCode.INTERNAL_ERROR };
```
ydogandjiev, line 26: Why not just return the error that the client sends back? Seems weird to always map this to INTERNAL_ERROR.

src/private/meetingAudio.ts
```diff
@@ -0,0 +1,34 @@
...
+   * error: Incase of an error, errorCode will be 500 - internal error
+   */
+  export function toggleIncomingClientAudio(callback: (error: SdkError, result: boolean | null) => void): void {
+    if (!callback) {
+      throw new Error('[toggle incoming client audio] Callback cannot be null');
+    }
+    ensureInitialized();
+    const messageId = sendMessageRequestToParent('toggleIncomingClientAudio');
+    GlobalVars.callbacks[messageId] = (error: SdkError | null, result: boolean | null) => {
+      if (callback) {
```
ydogandjiev, line 24: I don't think this check is needed since you already ensure that the callback is defined above.

src/private/meetingAudio.ts
```diff
@@ -0,0 +1,34 @@
...
+   * Hide from docs
+   * ------
+   * Allows an app to toggle the incoming audio speaker setting for the meeting user from mute to unmute or vice-versa
+   * @param callback Callback contains 2 parameters, error of type SdkError and result of type boolen or null.
+   * result can either contain the isSpeakerMuted value, incase of a successful toggle or null when the toggling fails
+   * result: True means incoming audio is muted and false means incoming audio is unmuted
+   * error can either contain an error, incase of an error or null when toggle is successful
+   * error: Incase of an error, errorCode will be 500 - internal error
+   */
+  export function toggleIncomingClientAudio(callback: (error: SdkError, result: boolean | null) => void): void {
```
ydogandjiev, line 17: I think the correct type for the error parameter of the callback should be `SdkError | null`.

src/private/meetingAudio.ts
ronakbakshi: Thank you for the details. @ydogandjiev I've made updates.

ydogandjiev (changes requested):

src/private/meetingAudio.ts
```diff
@@ -0,0 +1,31 @@
...
+import { SdkError, ErrorCode } from '../public/interfaces';
+
+export namespace meetingAudio {
+  /**
+   * @private
+   * Hide from docs
+   * ------
+   * Allows an app to toggle the incoming audio speaker setting for the meeting user from mute to unmute or vice-versa
+   * @param callback Callback contains 2 parameters, error of type SdkError and result of type boolen or null.
+   * result can either contain the isSpeakerMuted value, incase of a successful toggle or null when the toggling fails
```
amansardana012, line 12: isSpeakerMuted value has no significance here. "result can either contain the isSpeakerMuted value" this part can be removed

src/private/meetingAudio.ts
ronakbakshi: Updated to have the error from the client. Thanks

src/private/meetingAudio.ts
ronakbakshi: Removed the check

src/private/meetingAudio.ts
ronakbakshi: Updated the type. Thanks

src/private/meetingAudio.ts
ronakbakshi: Removed it. Thanks

src/private/meetingAudio.ts
```diff
@@ -0,0 +1,31 @@
...
+   * error can either contain an error, incase of an error or null when toggle is successful
+   * error: Incase of an error, error will not be null and have an object
+   */
+  export function toggleIncomingClientAudio(callback: (error: SdkError | null, result: boolean | null) => void): void {
+    if (!callback) {
+      throw new Error('[toggle incoming client audio] Callback cannot be null');
+    }
+    ensureInitialized();
+    const messageId = sendMessageRequestToParent('toggleIncomingClientAudio');
+    GlobalVars.callbacks[messageId] = (error: SdkError | null, result: boolean | null) => {
```
ydogandjiev, line 23: I think you could just change this to:
```
GlobalVars.callbacks[messageId] = callback;
```

ydogandjiev (approved):

src/private/meetingAudio.ts
ronakbakshi: Updated it. Thanks

PR #447, opened 2020-11-10
Scan Barcode API for webapps

samyuktha-tamvada:
This PR implements a new JS SDK API (**scanBarCode**). Through **scanBarCode** API, developer will be able to scan barcode/QRcode from mobile device and share the result with the webapp. In response to scanBarCode API, the webapp will receive the decoded text scanned by the barcode scanner in the string format. 

[Dev design document](https://microsoft.sharepoint.com/:w:/t/ActionPlatform/EU4j09UujQlBmYYvHNM7RbkBdmXZNBChbHmwb9kP9d3zYA?e=fu8zmK)
[PM spec](https://nam06.safelinks.protection.outlook.com/ap/w-59584e83/?url=https%3A%2F%2Fmicrosoft-my.sharepoint.com%2F%3Aw%3A%2Fp%2Fshuchauh%2FEWrVlbxxa85CnutgJcKTmpcBE0_vzhe_dxNEu0RVPYneDw%3Fe%3DakbLmy&data=04%7C01%7Ctamsam%40microsoft.com%7C417f0b5e6bba4117641c08d86fc1bfe8%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637382225287906816%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=S7hEmsZgFo9iQFQDZFJ%2FJMbLuqXQwk6p3SNG%2BIncn24%3D&reserved=0):

Tests:
Did manual testing using a custom app, tested the following:

- Validated scan bar code is returning correct result back to the webapp

- Validated scan bar code works with null input

- Validated scan bar code works with no inputs

- Verified that default values are set as inputs in case of null or no inputs

- Validated scan bar code throws proper error with invalid inputs

- Verified that for the scenario in case of this API : User leaves the app and comes back, the user is able to resume the flow from where it left of

- Verified that for the scenario : User interacting with scan bar code API, leaves the app, comes back to the app in a different activity (say through deep link), in that case the previously opened activity, in our case the one hosting the JS app is destroyed. And the app has to then be opened and the subsequent flows work as expected.

- Verified that the flow never ends up in a limbo state both in success/failure

Sample test app to test out the new API:
Hosted on Azure for internal testing: https://testscanbarcodeapi.azurewebsites.net


src/internal/mediaUtil.ts
```diff
@@ -74,3 +74,25 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
...
   return true;
 }
+
+/**
+ * Returns true if the scan barcode param is valid and false otherwise
+ */
+export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): boolean {
+  if (barCodeConfig != null) {
+    if (
+      barCodeConfig.barCodeType != null &&
```
SonalJha, line 84: null and undefined check comment is applicable here and for timeOutIntervalInSec as well

src/internal/mediaUtil.ts
```diff
@@ -74,3 +74,25 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
...
+      (barCodeConfig.barCodeType < media.BarCodeType.All ||
+        barCodeConfig.barCodeType > media.BarCodeType.TwoDimensional)
+    ) {
+      return false;
+    }
+    if (
```
SonalJha, line 90: The opening and closing braces both ( ) and { } have been formatted as per c/c++ style. Javascript's ECMA script specification recommends that it be formatted like this: https://stackoverflow.com/questions/3960518/javascript-formatting-must-braces-be-on-the-same-line-as-the-if-function-etc-ke

src/public/interfaces.ts
```diff
@@ -595,6 +595,10 @@ export enum ErrorCode {
    * User aborted the operation
    */
   USER_ABORT = 8000,
+  /**
+   * Could not complete the operation in a given time interval
```
SonalJha, line 599: the*

src/public/media.ts
```diff
@@ -407,4 +413,66 @@ export namespace media {
...
   }
+
+  /**
+   * Optional barcode configuration supplied to scanBarCode API to customize barcode scanning experience in mobile
+   * All properties in BarCodeConfig are optional and have default values in the platform
+   */
+  export interface BarCodeConfig {
+    /**
+     * Optional; Lets the developer specify the type of barcode to be scanned
+     * Default value is All i.e one dimensional and two dimensional(only QRcode) barcodes can be scanned
```
SonalJha, line 424: i.e. both one dimensional ...

src/public/media.ts
```diff
@@ -407,4 +413,66 @@ export namespace media {
...
+     */
+    barCodeType?: BarCodeType;
+    /**
+     * Optional; Lets the developer specify the scan timeout interval in seconds
+     * Default value is 30 seconds and max value is 60 seconds
```
SonalJha, line 429: max allowed value is

src/public/media.ts
```diff
@@ -407,4 +413,66 @@ export namespace media {
...
+   * All properties in BarCodeConfig are optional and have default values in the platform
+   */
+  export interface BarCodeConfig {
+    /**
+     * Optional; Lets the developer specify the type of barcode to be scanned
+     * Default value is All i.e one dimensional and two dimensional(only QRcode) barcodes can be scanned
+     */
+    barCodeType?: BarCodeType;
+    /**
+     * Optional; Lets the developer specify the scan timeout interval in seconds
```
SonalJha, line 428: * Optional; Is this a standard way of specifying optional params? If it will just appear as a comment in documentation better use it in complete sentence.

SonalJha (changes requested):

src/internal/mediaUtil.ts
```diff
@@ -74,3 +74,25 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
   }
   return true;
 }
+
+/**
+ * Returns true if the scan barcode param is valid and false otherwise
+ */
+export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): boolean {
+  if (barCodeConfig != null) {
```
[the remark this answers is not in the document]
samyuktha-tamvada: done

src/internal/mediaUtil.ts
samyuktha-tamvada: done

src/internal/mediaUtil.ts
samyuktha-tamvada: This is formatted as per lint rules specified in Teams JS SDK repo

src/public/media.ts
samyuktha-tamvada: done

src/public/media.ts
samyuktha-tamvada: done

src/public/media.ts
samyuktha-tamvada: yes this is the way optional params are specified in the repo. Will keep it as is.

src/public/interfaces.ts
samyuktha-tamvada: done

src/public/media.ts
```diff
@@ -407,4 +413,66 @@ export namespace media {
...
+  }
+
+  /**
+   * Specifies the types of barcode which are supported by scanBarCode API
+   */
+  export enum BarCodeType {
+    /**
+     * One dimensional barcode types and QRcode are supported
+     */
+    All = 1,
```
ydogandjiev, line 441: How would this work if we add more types in the future? For example, what if a ThreeDimensional option was added and the developer wanted to specify only TwoDimensional and ThreeDimensional scanning? Seems like these should either be modeled as flags or we should just change `barCodeType` to a `barCodeTypes` array in the BarCodeConfig interface allowing the caller to provide multiple options.

src/internal/mediaUtil.ts
```diff
@@ -74,3 +74,27 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
...
+ */
+export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): boolean {
+  if (barCodeConfig) {
+    if (
+      barCodeConfig.barCodeType !== null &&
+      barCodeConfig.barCodeType !== undefined &&
+      (barCodeConfig.barCodeType < media.BarCodeType.All ||
+        barCodeConfig.barCodeType > media.BarCodeType.TwoDimensional)
+    ) {
+      return false;
```
ydogandjiev, line 89: Shouldn't we also log an error to the console or in some other way communicate to the developer exactly why we are rejecting their call?

ydogandjiev (approved):

src/internal/mediaUtil.ts
```diff
@@ -74,3 +74,27 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
...
+    }
+    if (
+      barCodeConfig.timeOutIntervalInSec !== null &&
+      barCodeConfig.timeOutIntervalInSec !== undefined &&
+      (barCodeConfig.timeOutIntervalInSec <= 0 || barCodeConfig.timeOutIntervalInSec > 60)
+    ) {
+      return false;
+    }
+  }
+  return true;
```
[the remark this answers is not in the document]
samyuktha-tamvada: Yes, this is intentional and skipping the validations if barCodeConfig is null/undefined as barCodeConfig is an optional param. True or false is not returned to the end user.  Platform will launch barcode scanner in default mode if barCodeConfig is null or undefined.

src/internal/mediaUtil.ts
samyuktha-tamvada: We will be returning INVALID_ARGUMENTS error code in the error response in case the user has passed invalid values in the barcodeConfig param.

SonalJha (approved):

src/public/media.ts
samyuktha-tamvada: Good catch but we have internally discussed in the team on keeping the optional barcodeType field and finally decided to remove it from BarCodeConfig. Currently, we don't see much value in keeping the optional barCodeType field in the barcodeConfig as Office Lens barcode SDK is only supporting 1 D barcode types and QRcode from 2 Dimension barcode types. We can introduce barCodeType as an array in future if more barcode types are supported by Office Lens and there is an ask from customers to support custom barcode type scanning.

amrutk-ms (approved):

PR #448, opened 2020-11-11
Update filePreviewParameters for openFilePreview

vivek3003:
Previously teams used to support opening a file only in view/edit mode. Hence we were using a boolean `editFile`.

Teams, can now support multiple modes.(`"view" | "edit" | "editNew"`). Adding a new filePreviewParameter to enable this.

### DIFF
--- a/src/private/interfaces.ts
+++ b/src/private/interfaces.ts
@@ -84,6 +84,7 @@ export interface FilePreviewParameters {
   baseUrl?: string;
 
   /**
+   * Deprecated; prefer using viewerAction instead
    * Optional; indicates whether the file should be opened in edit mode
    */
   editFile?: boolean;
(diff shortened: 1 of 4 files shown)


src/private/interfaces.ts
```diff
@@ -93,6 +94,11 @@ export interface FilePreviewParameters {
    * This field should be used to restore to a specific state within an entity, such as scrolling to or activating a specific piece of content.
    */
   subEntityId?: string;
+
+  /**
+   * Optional; indicates the mode in which file should be opened. Takes precedence over edit mode.
+   */
+  viewerAction?: 'view' | 'edit' | 'editNew';
```
ydogandjiev, line 101: Can you please add a new enum for these options - view, edit, editNew?

ydogandjiev (changes requested):

src/private/interfaces.ts
vivek3003: Have added the enum.

ydogandjiev (approved):

ydogandjiev (approved):

PR #459, opened 2020-11-30
Adding partial response infra to postIDResponse callback and Change getMedia API's postResponse from handler to callback with backward compatibility

SonalJha:
**Scenario**
Currently callbacks were supported for request/response but only one response could be sent for a callback. We have modified the pattern by introducing a new field isPartialResponse. If isPartialResponse is set to true callback will not be deleted on teams js sdk library and multiple callbacks can be sent back using the same id. When the platform doesn't need to send anymore callback, in the final one it can remove the isPartialResponse field and use the existing postIDResponse() or use postCompleteResponse() function to send the final callback. (which will get deleted on the platform)

getMedia needs to send more than one response. It was using handlers before but now we have onboarded getMedia() to callbacks with isPartialResponse.

[comment]: # "What is the user scenario this change is trying to fix? Try to bring in some context. Screenshots may be helpful but not always necessary."

**Validation**
- Verified that getMedia() works via the new callback method
- Verified backward compatibility of getMedia() that it continues to work on the older client versions
- Verified that isPartialResponse field is successfully received on teams js sdk library when postPartialResponse function is used

### DIFF
--- a/src/internal/internalAPIs.ts
+++ b/src/internal/internalAPIs.ts
@@ -232,6 +232,8 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
     const callback = GlobalVars.callbacks[message.id];
     if (callback) {
       callback.apply(null, message.args);
+
+      // Remove the callback to ensure that the callback is called only once and to free up memory if response is a complete response
       if (!isPartialResponse(evt)) {
         delete GlobalVars.callbacks[message.id];
       }
--- a/src/internal/internalAPIs.ts
+++ b/src/internal/internalAPIs.ts
@@ -232,9 +232,13 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
     const callback = GlobalVars.callbacks[message.id];
     if (callback) {
       callback.apply(null, message.args);
-
-      // Remove the callback to ensure that the callback is called only once and to free up memory.
-      delete GlobalVars.callbacks[message.id];
+      if (
+        !('isPartialResponse' in evt.data) ||
+        !(typeof evt.data.isPartialResponse === 'boolean') ||
+        !(evt.data.isPartialResponse === true)
+      ) {
+        delete GlobalVars.callbacks[message.id];
+      }
     }
   } else if ('func' in evt.data && typeof evt.data.func === 'string') {
     // Delegate the request to the proper handler
--- a/src/internal/internalAPIs.ts
+++ b/src/internal/internalAPIs.ts
@@ -232,11 +232,7 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
     const callback = GlobalVars.callbacks[message.id];
     if (callback) {
       callback.apply(null, message.args);
-      if (
-        !('isPartialResponse' in evt.data) ||
-        !(typeof evt.data.isPartialResponse === 'boolean') ||
-        !(evt.data.isPartialResponse === true)
-      ) {
+      if (!isPartialResponse(evt)) {
         delete GlobalVars.callbacks[message.id];
       }
     }
@@ -249,6 +245,14 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
       handler.apply(this, message.args);
     }
   }
+}
+
+function isPartialResponse(evt: DOMMessageEvent): boolean {
+  return (
+    'isPartialResponse' in evt.data &&
+    typeof evt.data.isPartialResponse === 'boolean' &&
+    evt.data.isPartialResponse === true
+  );
 }
 
 function handleChildMessage(evt: DOMMessageEvent): void {
--- a/src/internal/internalAPIs.ts
+++ b/src/internal/internalAPIs.ts
@@ -250,11 +250,7 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
 }
 
 function isPartialResponse(evt: DOMMessageEvent): boolean {
-  return (
-    'isPartialResponse' in evt.data &&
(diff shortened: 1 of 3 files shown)


src/internal/internalAPIs.ts
```diff
@@ -232,9 +232,13 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
     const callback = GlobalVars.callbacks[message.id];
     if (callback) {
       callback.apply(null, message.args);
-
-      // Remove the callback to ensure that the callback is called only once and to free up memory.
-      delete GlobalVars.callbacks[message.id];
+      if (
+        !('isPartialResponse' in evt.data) ||
+        !(typeof evt.data.isPartialResponse === 'boolean') ||
```
ydogandjiev, line 237: I think it would be clearer if you simply used `!==` instead of `===` and avoided negating the whole expression

src/internal/internalAPIs.ts
```diff
@@ -232,9 +232,13 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
...
+        !(evt.data.isPartialResponse === true)
```
ydogandjiev, line 238: Same here; just use `!==`

src/public/media.ts
```diff
@@ -141,6 +142,47 @@ export namespace media {
...
+            } else {
+              callback({ errorCode: ErrorCode.INTERNAL_ERROR, message: 'data receieved is null' }, null);
+            }
+          }
+        }
+      }
+      GlobalVars.callbacks[actionName] = handleGetMediaCallbackRequest;
+    }
+
+    private getMediaViaHandler(callback: (error: SdkError, blob: Blob) => void): void {
```
ydogandjiev, line 185: Keeping the old code path around totally negates the purpose of this change; we are just adding more complexity and confusion to the code. Can we remove the old code path? If not, then what's the plan for cleaning this up?

src/public/media.ts
```diff
@@ -141,6 +142,47 @@ export namespace media {
...
+      }
+    }
+
+    private getMediaViaCallback(callback: (error: SdkError, blob: Blob) => void): void {
+      const helper: MediaHelper = {
+        mediaMimeType: this.mimeType,
+        assembleAttachment: [],
+      };
+      const params = [this.content];
+      const actionName = sendMessageRequestToParent('getMedia', params);
```
ydogandjiev, line 159: Please rename `actionName` to `messageId` since that's what the `sendMessageRequestToParent` method is actually returning here.

src/public/media.ts
```diff
@@ -141,6 +142,47 @@ export namespace media {
...
+        this.getMediaViaHandler(callback);
+      }
+    }
+
+    private getMediaViaCallback(callback: (error: SdkError, blob: Blob) => void): void {
+      const helper: MediaHelper = {
+        mediaMimeType: this.mimeType,
+        assembleAttachment: [],
+      };
+      const params = [this.content];
```
ydogandjiev, line 158: Why are we sending `this.content` to the host? What is `this.content`? Could we give it a more clear name?

src/internal/internalAPIs.ts
```diff
@@ -232,9 +232,13 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
...
+      ) {
+        delete GlobalVars.callbacks[message.id];
```
ydogandjiev, line 240: Who would be responsible for cleaning up the callback in this case? Would the host be expected to send the final message with `isPartialResponse` set to false?

ydogandjiev (changes requested):

src/internal/internalAPIs.ts
SonalJha: That's right. Host is expected to send the final callback with either isPartialResponse set to false or without isPartialResponse param

src/public/media.ts
SonalJha: We will have to keep the old code for now to ensure backward compatibility. Ideally we can clean it up once we are sure every client's client version is greater than or equal to 2.0.0. But I don't know when that would be or how to track that.

src/internal/internalAPIs.ts
```diff
@@ -232,9 +232,13 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
     const callback = GlobalVars.callbacks[message.id];
     if (callback) {
       callback.apply(null, message.args);
-
-      // Remove the callback to ensure that the callback is called only once and to free up memory.
-      delete GlobalVars.callbacks[message.id];
+      if (
```
amrutk-ms, line 235: if (!isPartialResponse(evt.data) { ... }
isParitalResponse(evt) { return evt.isPartailResponse === true; }

src/public/media.ts
amrutk-ms: Create a work item to track telemetry and delete the old code once we determine that it is not being used anymore by anyone.

amrutk-ms: also: call delete GlobalVars.handlers['getMedia' + actionName] instead of setting it to null.

amrutk-ms (approved):

src/public/media.ts
SonalJha: done

src/public/media.ts
SonalJha: done

src/public/media.ts
SonalJha: done. https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/1376437

ydogandjiev (approved):

PR #462, opened 2020-12-02
Scan BarCode API to return NotSupported Error on desktop and web platform

samyuktha24:
ScanBarCode API is only available for mobile platform and not supported on Desktop and web platform
ErrorCode.NOT_SUPPORTED_ON_PLATFORM (i.e 100 errorcode) will be returned in the callback on desktop and web platform.

Tests:
Did manual testing using a custom app, tested the following:

- ScanBarCode API works on mobile platform

- ScanBarCode API returns ErrorCode.NOT_SUPPORTED_ON_PLATFORM (i.e 100 errorcode) on desktop and web platform

### DIFF
--- a/src/public/media.ts
+++ b/src/public/media.ts
@@ -440,7 +440,7 @@ export namespace media {
 
     if (!GlobalVars.isFramelessWindow) {
       const notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
-      callback(notSupportedError, undefined);
+      callback(notSupportedError, null);
       return;
     }
 
--- a/src/public/media.ts
+++ b/src/public/media.ts
@@ -428,7 +428,7 @@ export namespace media {
 
   /**
    * Scan Barcode/QRcode using camera
-   * Note: For desktop, this API is not supported. Callback will be resolved with ErrorCode.NotSupported.
+   * Note: For desktop and web, this API is not supported. Callback will be resolved with ErrorCode.NotSupported.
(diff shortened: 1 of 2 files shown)


src/public/media.ts
```diff
@@ -437,6 +438,12 @@ export namespace media {
     }
     ensureInitialized(FrameContexts.content, FrameContexts.task);
 
+    if (!GlobalVars.isFramelessWindow) {
+      const notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
+      callback(notSupportedError, undefined);
```
SonalJha, line 443: Please make it null instead of undefined

samyuktha24: done.

src/public/media.ts
```diff
@@ -437,6 +438,12 @@ export namespace media {
     }
     ensureInitialized(FrameContexts.content, FrameContexts.task);
 
+    if (!GlobalVars.isFramelessWindow) {
```
ydogandjiev, line 441: This is not a good way to check for web/desktop. Please look at the `GlobalVars.hostClientType` instead or better yet implement a handler in the web/desktop client code itself to return the error.

ydogandjiev (changes requested):

src/public/media.ts
samyuktha24: Sure. I have modified the code to use GlobalVars.hostClientType.

ydogandjiev (approved):

SonalJha (approved):

PR #464, opened 2020-12-07
Listen to child frame messages on frameless windows

Kanac:
In the initialize method, the event listener for frame messages seems to only be registered if the current window has a parent window. Otherwise, the code considers the window as a "frameless window" (i.e mobile app), and does not register the event listener.

For our app setup on mobile, we have nested frames (an iframe at the top level using Teams SDK, and another iframe inside it that also uses the Teams SDK). Currently, the inner frame's Teams SDK cannot communicate with the top level, because the top frame's Teams SDK doesn't register to listen to messages from other iframes, other than with the mobile app itself. 

This PR changes the initialization code so that the window listens to messages from other iframes regardless of whether its a frameless window or not. It also updates the `updateRelationship` method to properly parse frame messages as coming from the children for frameless windows.


src/internal/internalAPIs.ts
```diff
@@ -202,10 +202,17 @@ function shouldProcessMessage(messageSource: Window, messageOrigin: string): boo
 function updateRelationships(messageSource: Window, messageOrigin: string): void {
   // Determine whether the source of the message is our parent or child and update our
   // window and origin pointer accordingly
-  if (!GlobalVars.parentWindow || GlobalVars.parentWindow.closed || messageSource === GlobalVars.parentWindow) {
+  // For frameless windows (i.e mobile), there is no parent frame, so the message must be from the child.
+  const noParentOrMessageIsFromParent =
+    !GlobalVars.parentWindow || GlobalVars.parentWindow.closed || messageSource === GlobalVars.parentWindow;
+  const isParentMessage = !GlobalVars.isFramelessWindow && noParentOrMessageIsFromParent;
```
ydogandjiev, line 208: I find this new logic very confusing and difficult to understand. Neither the "noParentOrMessageIsFromParent" nor the "isParentMessage" and "isChildMessage" names seem accurate for what they are trying to capture. Please consider coming up with more accurate + descriptive names for these or simply inline the logic so that a reader just focuses on that instead of trying to decipher what the variable names mean.

ydogandjiev (changes requested):

src/internal/internalAPIs.ts
ydogandjiev: Also, please add UTs to exercise the newly added code paths here.

Kanac: That was just my interpretation of the existing logic here to make it less confusing. I was thinking of inlining the logic as well but I thought the logic would get too long, but if you're fine with that, then I'll update it to be inline.

I'll add unit tests as well for this 👍

Kanac: Unit tests added.

test/private/privateAPIs.spec.ts
```diff
@@ -323,6 +324,26 @@ describe('MicrosoftTeams-privateAPIs', () => {
...
+        id: 0,
+        func: 'themeChange',
+        args: ['testTheme'],
+      } as MessageResponse,
+    } as MessageEvent);
+
+    // The frameless window should send a response back to the child window
+    expect(utils.childMessages.length).toBe(1);
+
+    utils.mockWindow.parent = parent;
```
ydogandjiev, line 344: I understand why this is necessary but it's a little confusing for this cleanup code to live here. Can we move this into either the `initializeWithContext` helper or at least into a `beforeEach` function that resets this before every test case?

test/utils.ts
```diff
@@ -63,6 +63,12 @@ export class Utils {
           that.messages.push(message);
         },
       } as Window,
+      nativeInterface: {
+        framelessPostMessage: function(message: string): void {
+          let msg = JSON.parse(message);
```
ydogandjiev, line 68: NIT: `let` --> `const`

ydogandjiev (approved):

test/private/privateAPIs.spec.ts
Kanac: Done 👍

Kanac:
@prannayb Could you please take a look as well? Thanks!

src/public/publicAPIs.ts
```diff
@@ -46,12 +46,13 @@ export function initialize(callback?: () => void, validMessageOrigins?: string[]
         ? GlobalVars.currentWindow.parent
         : GlobalVars.currentWindow.opener;
 
+    // Listen to messages from the parent or child frame.
+    // Frameless windows will only receive this event from child frames.
+    GlobalVars.currentWindow.addEventListener('message', messageListener, false);
```
ydogandjiev, line 51: @Kanac, just to be safe would it be possible to put this call behind an explicit check. Either:
1. We can register it anytime `validMessageOrigins` is specified:
    ```
    if (validMessageOrigins) {
      // Register message handler
    }
    ```
2. Or we can add a new parameter to the `initialize` API - something like `enableChildFrameCommunication?: boolean`.

What do you think? Would either of these approaches work better for your use case?

ydogandjiev (changes requested):

src/public/publicAPIs.ts
Kanac: For #1, I thought the message wouldn't get pass the check here https://github.com/OfficeDev/microsoft-teams-library-js/blob/f3d43bc7f1953e010e340e39d21133bf34479bbc/src/internal/internalAPIs.ts#L193 if `validMessageOrigins` isn't defined. But then I noticed we have a bunch of default `validMessageOrigins` here https://github.com/OfficeDev/microsoft-teams-library-js/blob/69962f9344710a7a1f5a7b80b3f6f48712854f08/src/internal/constants.ts#L12 that would still be allowed to communicate with the frame. I'm guessing those won't cause a problem for the frameless case, but I guess we can add this check here.

for #2, I think that would be a breaking change for any existing users that already have child frame communication that would need to now remember to pass in that parameter.

ydogandjiev (approved):

prannayb (approved):
I went through the code flows, and all the logic checks out for the existing scenarios:
1. An app sets up an iframe inside itself that uses Teams SDK. (Existing unit test pass)
2. In same session, app removes a iframe and adds a new iframe (No existing unit test for this!)

And with the new info that mobile top level app frames are parentless (i.e. frameless) this will also serve that scenario. Could you add to the existing unit tests to also account for scenario #2 above for normal parent and frameless parent window cases? When top level frame is normal,frameless: Existing registered child iframe is superseded if a new child iframe is added by the app, and it must get registered as the new child (this assumes the old iframe was removed and replaced). The test would just post a new message posing as a new child frame.

Kanac:
Thanks @prannayb. Yes, the unit tests in the Teams SDK are missing some situations like you pointed out. Prior to my changes, there weren't any tests for frameless windows at all until now. For your 2nd point, I'm not aware of any usecases where an iframe is removed by app and another one is added. Or at least, our app doesn't do this. Perhaps we can have a follow-up PR for that test if its a common usecase? 

@ydogandjiev Are you supposed to merge this PR? I don't see a button for me to do it after getting approvals.

Kanac:
@ydogandjiev Do you know when this will be released to npm (i.e would it be in 1.9 beta 4?)

PR #470, opened 2020-12-11
Update meetingAudio namespace to meeting

ronakbakshi:
@ydogandjiev as discussed, I am updating the name of the namespace to "meeting" from "meetingAudio"


src/private/meeting.ts
```diff
@@ -2,7 +2,7 @@ import { ensureInitialized, sendMessageRequestToParent } from '../internal/inter
 import { GlobalVars } from '../internal/globalVars';
 import { SdkError } from '../public/interfaces';
 
-export namespace meetingAudio {
+export namespace meeting {
```
ydogandjiev, line 5: So we have a few other namespaces like this which use the plural version - e.g. `conversations` and `settings`. Do you think it would make sense to make this plural as well - i.e. `meetings`?

src/private/index.ts
```diff
@@ -26,4 +26,4 @@ export {
 } from './privateAPIs';
 export { conversations } from './conversations';
 export { meetingRoom } from './meetingRoom';
-export { meetingAudio } from './meetingAudio';
+export { meeting } from './meeting';
```
ydogandjiev, line 29: It seems like these APIs will be exposed publicly so you might want to move them under `src/public` folder instead of `src/private` where they are today.

ydogandjiev (changes requested):

src/private/meeting.ts
ronakbakshi: I saw a few namespaces like "media" and "meetingRoom" which are singular. I think it makes more sense for it to be singular.

src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './private/index';
 export * from './public/index';
+export { meeting } from './public/meeting';
```
ydogandjiev, line 3: I think this needs to go into the './public/index.ts' file instead. Can you please move it there?

test/public/meeting.spec.ts
```diff
@@ -1,10 +1,10 @@
-import { meetingAudio } from "../../src/private/meetingAudio";
+import { meeting } from "../../src/public/meeting";
 import { SdkError, ErrorCode } from "../../src/public/interfaces";
-import { DOMMessageEvent } from "../../src/internal/interfaces";
+import { DOMMessageEvent } from '../../src/internal/interfaces';
```
ydogandjiev, line 3: NIT: Please stick with `"` instead of `'` for consistency

ydogandjiev (changes requested):

src/index.ts
ydogandjiev: You can delete this line now

ydogandjiev (changes requested):

src/index.ts
ronakbakshi: Deleted

ydogandjiev (approved):

PR #476, opened 2020-12-23
Add get meeting details api

ronakbakshi:


src/public/meeting.ts
```diff
@@ -3,6 +3,37 @@ import { GlobalVars } from '../internal/globalVars';
...
+     * Scheduled start time of the meeting
+     */
+    scheduledStartTime?: string;
+    /**
+     * Scheduled end time of the meeting
+     */
+    scheduledEndTime?: string;
+    /**
+     * meeting title name of the meeting
+     */
```
sachingupta, line 21: are all fields optional for a meeting? This is public interface expose to deplorers. We should not make all fields optional. I think if there is meeting id then it will have start, endtime and may be other fields as required.

sachingupta (changes requested):

src/public/meeting.ts
ronakbakshi: @sachingupta I have made start time and end time as required. Please let me know if you have any other questions.

sachingupta (approved):

PR #483, opened 2021-01-09
Fix partial response messages for nested frame setup

Kanac:
Related PR https://github.com/OfficeDev/microsoft-teams-library-js/pull/459

The above PR introduces the notion of partial responses. This is for Teams SDK calls that require more than 1 response from Teams to fully complete the request (i.e `getMedia`). However, there is a problem when there are nested frames that are both using the Teams SDK. If the inner frame uses an API like `getMedia` which uses partial responses, it doesn't quite work. 

When the outer frame passes the partial response from Teams back to the inner frame, the response is mistakenly not formatted as a partial response for the inner frame. This causes APIs like `getMedia` to terminate early before the whole response is given back to the inner frame. This PR fixes this issue by passing `isPartialResponse` to the end of the argument list to the response proxy callbacks so that it could be used for formatting the proxy response as partial.

Method Response Chain **Before** (for calls like `getMedia`): 
Teams -> (Sends Partial Response) -> Outer Frame using Teams SDK -> (Sends **Non-Partial** Response) -> Inner Frame using Teams SDK

Method Response Chain **After** (for calls like `getMedia`): 
Teams -> (Sends Partial Response) -> Outer Frame using Teams SDK -> (Sends **Partial** Response) -> Inner Frame using Teams SDK

### DIFF
--- a/src/internal/interfaces.ts
+++ b/src/internal/interfaces.ts
@@ -31,12 +31,13 @@ export interface ExtendedWindow extends Window {
 export interface MessageRequest {
   id?: number;
   func: string;
-  args?: any[]; // tslint:disable-line:no-any The args here are a passthrough to postMessage where we do allow any[]
+  args?: any[]; // tslint:disable-line:no-any The args here are a passthrough to postMessage where wes do allow any[]
 }
 
 export interface MessageResponse {
   id: number;
   args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
+  isPartialResponse?: boolean;
 }
 
 /**
--- a/src/internal/interfaces.ts
+++ b/src/internal/interfaces.ts
@@ -31,7 +31,7 @@ export interface ExtendedWindow extends Window {
 export interface MessageRequest {
   id?: number;
   func: string;
-  args?: any[]; // tslint:disable-line:no-any The args here are a passthrough to postMessage where wes do allow any[]
+  args?: any[]; // tslint:disable-line:no-any The args here are a passthrough to postMessage where we do allow any[]
 }
 
 export interface MessageResponse {
--- a/src/internal/interfaces.ts
+++ b/src/internal/interfaces.ts
@@ -37,7 +37,7 @@ export interface MessageRequest {
 export interface MessageResponse {
   id: number;
   args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
-  isPartialResponse?: boolean;
+  isPartialResponse?: boolean; // If the message is partial, then there will be more future responses for the given message ID.
 }
 
 /**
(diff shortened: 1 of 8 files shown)


test/utils.ts
```diff
@@ -49,20 +51,20 @@ export class Utils {
       outerHeight: 768,
       screenLeft: 0,
       screenTop: 0,
-      addEventListener: function (type: string, listener: (ev: MessageEvent) => void, useCapture?: boolean): void {
```
Kanac, line 52: Most of the changes here are formatting fixes from my IDE, but they seem to make the code cleaner. I could remove these if there's any issues though.

Kanac:
@ydogandjiev

ydogandjiev (approved):

src/internal/internalAPIs.ts
```diff
@@ -235,7 +235,7 @@ export function handleParentMessage(evt: DOMMessageEvent): void {
     const message = evt.data as MessageResponse;
     const callback = GlobalVars.callbacks[message.id];
     if (callback) {
-      callback.apply(null, message.args);
+      callback.apply(null, [...message.args, message.isPartialResponse]);
```
ydogandjiev, line 238: Why is this change necessary? Seems weird that we would be passing this property to the callback? I don't think this is something a callback should ever have to handle itself.

ydogandjiev (changes requested):

src/internal/internalAPIs.ts
Kanac: Its used specifically for this part here where we proxy the response from Teams to the current frame to the child frame that called the API https://github.com/OfficeDev/microsoft-teams-library-js/pull/483/files#diff-fb785af0e32520a03c689ec3c20c481d2c9b5d362ef181402123f2b50be7a418R277

I don't really see any other better way to pass `isPartialResponse` to reconstructing the proxy partial response to the child frame so that's why this change is necessary.

SonalJha:
Changes look good to me. Please test with mobile (android and ios) before merging the PR.

SonalJha (approved):

src/internal/internalAPIs.ts
Kanac: One thing we can do to be safer is to only pass `isPartialResponse` to the callback args if the callback is specificially the proxy callback defined here https://github.com/OfficeDev/microsoft-teams-library-js/pull/483/files#diff-fb785af0e32520a03c689ec3c20c481d2c9b5d362ef181402123f2b50be7a418L274 
(i.e we can change it from an anonymous function to a named one, and check the callbacks name).
But either way, given how the current code works, passing this property as an arg to the callback seems to be the only way to solve this issue.

ydogandjiev (approved):

src/internal/internalAPIs.ts
ydogandjiev: I'm fine with either option; I would say do what you think is best. FYI @sachingupta

src/internal/interfaces.ts
```diff
@@ -37,6 +37,7 @@ export interface MessageRequest {
 export interface MessageResponse {
   id: number;
   args?: any[]; // tslint:disable-line:no-any The args here are a passthrough from OnMessage where we do receive any[]
+  isPartialResponse?: boolean;
```
sachingupta, line 40: add some detail about this field.

sachingupta (approved):

PR #485, opened 2021-01-12
Add the sourceOrigin and userClickTime in the getContext call

debajyoti251088:
This will be used by the WPX apps mostly to understand where the tab was opened from and the timestamp when the tab was opened

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -384,6 +384,16 @@ export interface Context {
    * Personal app icon y coordinate position
    */
   appIconPosition?: number;
+
+  /**
+   * Source origin from where the tab is opened
+   */
+  sourceOrigin?: string;
+
+  /**
+   * Time when the user clicked on the tab
+   */
+  userClickTime?: string;
 }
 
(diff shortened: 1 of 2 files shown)


src/public/interfaces.ts
```diff
@@ -384,6 +384,16 @@ export interface Context {
    * Personal app icon y coordinate position
    */
   appIconPosition?: number;
+
+  /**
+   * Source origin from where the tab is opened
```
sachingupta, line 389: is it host like teams.microsoft.com origin?

sachingupta (approved):

src/public/interfaces.ts
debajyoti251088: @sachingupta Its mostly from where the tab is opened. Like say 'files' in the case of WPX tabs

ydogandjiev: We can document what values the developer can expect here later. For now lets just keep it vague since we don't actually know what all the possible sources are yet.

ydogandjiev (approved):

PR #486, opened 2021-01-14
Adding namespace and APIs for PTZ Extension

nimeickl:
Adding nampespace "ptzExtension" to hold 8 new APIs needed for the PTZ Extension app being built as a 3P app under the N* experience extensibility-apps. PTZ extension will be used in a Teams meeting side panel and these APIs are required to send messages from the UI (in extensibility-apps experience) to the meeting window (in multi-window experience) such that it can then interact with TSW and finally slimcore.


src/private/ptzExtension.ts
```diff
@@ -0,0 +1,296 @@
...
+   * error can either contain an error of type SdkError, incase of an error, or null when fetch is successful
+   * result can either contain the true/false value, incase of a successful request or null when it fails
+   * result: True means request was accepted and false means request was denied
+   */
+  export function requestRemoteCameraControl(
+    participant: PtzParticipantInput,
+    callback: (error: SdkError, result: boolean) => void,
+  ): void {
+    if (!participant) {
+      throw new Error('[ptzExtension.requestRemotePtzCameraControl] Participant cannot be null or empty');
```
mrgarlap, line 198: should this be '[ptzExtension.requestRemoteCameraControl] .....'?

src/private/ptzExtension.ts
```diff
@@ -0,0 +1,296 @@
...
+    }
+    if (!callback) {
+      throw new Error('[ptzExtension.requestRemotePtzCameraControl] Callback cannot be null');
```
mrgarlap, line 201: should this be '[ptzExtension.requestRemoteCameraControl] .....'?

src/private/ptzExtension.ts
```diff
@@ -0,0 +1,296 @@
...
+   */
+  export function sendRemoteCustomCommand(
+    ptzControlCommand: PtzControlCommand,
+    callback: (error: SdkError) => void,
+  ): void {
+    if (!ptzControlCommand) {
+      throw new Error('[ptzExtension.sendRemotePtzCustomCommand] PtzControlCommand cannot be null');
+    }
+    if (!callback) {
+      throw new Error('[ptzExtension.sendRemotePtzCustomCommand] Callback cannot be null');
```
mrgarlap, line 224: should this be '[ptzExtension.sendRemoteCustomCommand] .....'?

mrgarlap: Make sure to put the right name, in all other places aswell, after you confirm that, it needs tobe changed!!

src/private/ptzExtension.ts
```diff
@@ -0,0 +1,296 @@
...
+  ): void {
+    ensureInitialized();
+    GlobalVars.ptzControlDeviceStateChangeHandler = handler;
+    handler && sendMessageRequestToParent('registerHandler', ['ptzControlDeviceStateChange']);
+  }
+
+  /**
+   * Registers a handler for PTZ session status change.
+   * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
+   * @param handler The handler to invoke when the current PTZ status changes.
```
mrgarlap, line 287: @param handler The handler to invoke when the current PTZ **session** status changes. ?

mrgarlap (approved):

src/private/ptzExtension.ts
```diff
@@ -0,0 +1,336 @@
+import { ensureInitialized, sendMessageRequestToParent } from '../internal/internalAPIs';
+import { GlobalVars } from '../internal/globalVars';
+import { SdkError } from '../public/interfaces';
+export namespace ptzExtension {
```
sachingupta, line 4: we can have namespace as "remoteCamera"

sachingupta (changes requested):

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+import { GlobalVars } from '../internal/globalVars';
+import { SdkError } from '../public/interfaces';
+export namespace remoteCamera {
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Data structure to represent patricipant details needed to request control of camera.
+   */
+  export interface ParticipantInput {
```
sachingupta, line 11: ParticipantBaseInfo

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+   * @private
+   * Hide from docs
+   *
+   * Fetch a list of the participants with controllable-cameras in a meeting.
+   * @param callback Callback contains 2 parameters, error and result.
+   * error can either contain an error of type SdkError, incase of an error, or null when fetch is successful
+   * result can either contain the CapableParticipants object, incase of a successful fetch or null when it fails
+   * result: object that contains an array of participants with controllable-cameras
+   */
+  export function getCapableParticipants(callback: (error: SdkError, result: CapableParticipants) => void): void {
```
sachingupta, line 200: why not just return Participant[] ? what the use of class that has just one field.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+   * Request control of a participant's camera.
+   * @param participant Participant specifies the participant to send the request for camera control.
+   * @param callback Callback contains 2 parameters, error and result.
+   * error can either contain an error of type SdkError, incase of an error, or null when fetch is successful
+   * result can either contain the true/false value, incase of a successful request or null when it fails
+   * result: True means request was accepted and false means request was denied
+   */
+  export function requestControl(
+    participant: ParticipantInput,
+    callback: (error: SdkError, result: boolean) => void,
```
sachingupta, line 222: we can define better var name for result ?

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Send control command to the participant's camera.
+   * @param ControlCommand ControlCommand specifies the command for controling the camera.
+   * @param callback Callback to invoke when the command response returns.
+   */
+  export function sendControlCommand(ControlCommand: ControlCommand, callback: (error: SdkError) => void): void {
```
sachingupta, line 243: do we need to specify which participant we are sending Control to or it depend on current selected participant?

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+  export function terminateSession(callback: (error: SdkError) => void): void {
+    if (!callback) {
+      throw new Error('[remoteCamera.terminateSession] Callback cannot be null');
+    }
+    ensureInitialized();
+    const messageId = sendMessageRequestToParent('remoteCamera.terminateSession');
+    GlobalVars.callbacks[messageId] = callback;
+  }
+
+  GlobalVars.handlers['remoteCamera.capableParticipantsChange'] = handleCapableParticipantsChange;
```
sachingupta, line 271: these can move to before other SDK APIs.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Data structure to return error reason when a handler error occurs.
+   */
+  export interface HandlerFailed {
+    handlerError: ErrorReason;
+  }
```
sachingupta, line 66: can we directly return ErrorReason. No need to have class for one field.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Data structure to track the controlled device's state.
+   */
+  export interface DeviceStateChanged {
```
sachingupta, line 74: can we directly return DeviceState. No need to have class for one field.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Data structure to track the control session status.
+   */
+  export interface SessionStatusChanged {
+    sessionStatus: SessionStatus;
+  }
```
sachingupta, line 86: can we directly return SessionStatus. No need to have class for one field.

src/internal/globalVars.ts
```diff
@@ -41,4 +42,8 @@ export class GlobalVars {
   public static appButtonHoverLeaveHandler: () => void;
   public static meetingRoomCapabilitiesUpdateHandler: (capabilities: meetingRoom.MeetingRoomCapability) => void;
   public static meetingRoomStatesUpdateHandler: (states: meetingRoom.MeetingRoomState) => void;
+  public static capableParticipantsChangeHandler: (participantChange: remoteCamera.CapableParticipants) => void;
```
sachingupta, line 45: These function looks more generic since they are not under namespace, can we prefix with remoteCamera for these functions?

sachingupta (changes requested):

src/private/remoteCamera.ts
nimeickl: This is type is taken from the CDL / schema code; I was remaining consistent with the data returned by those queries/mutations/subscriptions

src/private/remoteCamera.ts
nimeickl: This is type is taken from the CDL / schema code; I was remaining consistent with the data returned by those queries/mutations/subscriptions

src/private/remoteCamera.ts
nimeickl: This is type is taken from the CDL / schema code; I was remaining consistent with the data returned by those queries/mutations/subscriptions

src/private/remoteCamera.ts
nimeickl: You can only be controlling one participant's camera at a time for our use case; perhaps if the need arises, it could be added later

sachingupta (approved):

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,305 @@
...
+   * participants can either contain an array of Participant objects, incase of a successful fetch or null when it fails
+   * participants: object that contains an array of participants with controllable-cameras
+   */
+  export function getCapableParticipants(
+    callback: (error: SdkError | null, participants: Participant[] | null) => void,
+  ): void {
+    if (!callback) {
+      throw new Error('[remoteCamera.getCapableParticipants] Callback cannot be null');
+    }
+    ensureInitialized();
```
ydogandjiev, line 157: It's probably best to pass `FrameContexts.sidePanel` to the `ensureInitialized();` call since these APIs won't work from any other context.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,305 @@
...
+    participant: Participant,
+    callback: (error: SdkError | null, requestResponse: boolean | null) => void,
+  ): void {
+    if (!participant) {
+      throw new Error('[remoteCamera.requestControl] Participant cannot be null');
+    }
+    if (!callback) {
+      throw new Error('[remoteCamera.requestControl] Callback cannot be null');
+    }
+    ensureInitialized();
```
ydogandjiev, line 183: It's probably best to pass `FrameContexts.sidePanel` to the `ensureInitialized();` call since these APIs won't work from any other context.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,305 @@
...
+   * @param callback Callback to invoke when the command response returns.
+   */
+  export function sendControlCommand(ControlCommand: ControlCommand, callback: (error: SdkError | null) => void): void {
+    if (!ControlCommand) {
+      throw new Error('[remoteCamera.sendControlCommand] ControlCommand cannot be null');
+    }
+    if (!callback) {
+      throw new Error('[remoteCamera.sendControlCommand] Callback cannot be null');
+    }
+    ensureInitialized();
```
ydogandjiev, line 203: It's probably best to pass `FrameContexts.sidePanel` to the `ensureInitialized();` call since these APIs won't work from any other context.

src/private/remoteCamera.ts
```diff
@@ -0,0 +1,305 @@
...
+   * Hide from docs
+   *
+   * Terminate the remote  session
+   * @param callback Callback to invoke when the command response returns.
+   */
+  export function terminateSession(callback: (error: SdkError | null) => void): void {
+    if (!callback) {
+      throw new Error('[remoteCamera.terminateSession] Callback cannot be null');
+    }
+    ensureInitialized();
```
ydogandjiev, line 219: It's probably best to pass `FrameContexts.sidePanel` to the `ensureInitialized();` call since these APIs won't work from any other context.

ydogandjiev: Same for the remaining APIs below.

ydogandjiev (approved):

PR #488, opened 2021-01-19
Chharpe/globalvars

charpercyr:
## Removing GlobalVars Part 1/3
This PR moves all the parent/child message handling to a file called communication.ts.
- Move all message handling functions to the Communication file
- Changed sendMessageRequestToParent to sendMessageToParent, which optionally accepts a callback. This new function does the work of both sendMessageRequestToParent and GlobalVars.callback[id] = ...
- Some housekeeping in functions in Communication
- Removed file level handler registration, replaced with calls in initialization functions


src/internal/communication.ts
```diff
@@ -0,0 +1,472 @@
...
+   */
+  public static sendMessageToParent(actionName: string): void;
+  /**
+   * Send a message to parent. Uses nativeInterface on mobile to communicate with parent context
+   */
+  public static sendMessageToParent(actionName: string, args: any[]): void;
+  /**
+   * Send a message to parent. Uses nativeInterface on mobile to communicate with parent context
+   */
+  public static sendMessageToParent(actionName: string, callback: Function): void;
```
sachingupta, line 95: is there any better way to minimum no of declaration for sendMessageToParent function?

src/internal/communication.ts
```diff
@@ -0,0 +1,472 @@
...
+      GlobalVars.onStartConversationHandler({
+        subEntityId: subEntityId,
+        conversationId: conversationId,
+        channelId: channelId,
+        entityId: entityId,
+      });
+    }
+  }
+
+  private static handleCloseConversation(
```
sachingupta, line 155: There are still using GlobalVars.onStartConversationHandler. May be we can move these back to InternalAPIs file.

src/internal/internalAPIs.ts
```diff
@@ -1,121 +1,7 @@
...
-GlobalVars.handlers['backButtonPress'] = handleBackButtonPress;
-GlobalVars.handlers['load'] = handleLoad;
-GlobalVars.handlers['beforeUnload'] = handleBeforeUnload;
-GlobalVars.handlers['changeSettings'] = handleChangeSettings;
-GlobalVars.handlers['startConversation'] = handleStartConversation;
-GlobalVars.handlers['closeConversation'] = handleCloseConversation;
-GlobalVars.handlers['appButtonClick'] = handleAppButtonClick;
-GlobalVars.handlers['appButtonHoverEnter'] = handleAppButtonHoverEnter;
-GlobalVars.handlers['appButtonHoverLeave'] = handleAppButtonHoverLeave;
-
```
sachingupta, line 20: we can keep all handlers to its own nameSpace? may be "EventHandlers" instead of putting to communication. We should try to make communication nameSpace independent of handler so minimize future changes in that file.

sachingupta (changes requested):
Few comments
1. Merge latest master - there were few APIs added recently
2.  we should not put handers inside communication nameSpace, either keep these in internalAPIs or to separate nameSpace EventHandlers.

src/internal/globalVars.ts
```diff
@@ -1,29 +1,15 @@
...
-  public static nextMessageId: number = 0;
-  public static handlers: {
-    [func: string]: Function;
-  } = {};
-  public static callbacks: {
-    [id: number]: Function;
-  } = {};
   public static hostClientType: string;
   public static clientSupportedSDKVersion: string;
   public static printCapabilityEnabled: boolean = false;
```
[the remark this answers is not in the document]
charpercyr: Yes! I will do the work over 3 PRs in total

src/internal/communication.ts
charpercyr: See PR #503. It takes care of handlers

src/internal/internalAPIs.ts
charpercyr: See PR #503. It takes care of handlers

src/internal/communication.ts
charpercyr: Good point! I have reduced it to 2

charpercyr:
End-to-End scenario testing passed. Can you approve @sachingupta @ydogandjiev?

ydogandjiev (approved):

charpercyr:
Changes included in #518

PR #492, opened 2021-01-21
Change type of userClickTime to number

ydogandjiev:


src/public/interfaces.ts
```diff
@@ -393,7 +393,7 @@ export interface Context {
   /**
    * Time when the user clicked on the tab
```
sachingupta, line 394: is it timespan/tick ? should we update comment

sachingupta (approved):

src/public/interfaces.ts
debajyoti251088: its the particular time Sachin. not really a span

debajyoti251088 (approved):

PR #500, opened 2021-02-02
Adding people picker capability to teams client library

SonalJha:


sachingupta (changes requested):
We just merge one of major refactoring PR which will impact the way we register handler. https://github.com/OfficeDev/microsoft-teams-library-js/pull/518

Please merge latest master and update your changes.

SonalJha:
> We just merge one of major refactoring PR which will impact the way we register handler. #518
> 
> Please merge latest master and update your changes.

I have updated my changes with the latest master

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
+import { ensureInitialized, isAPISupportedByPlatform } from '../internal/internalAPIs';
+import { FrameContexts } from './constants';
+import { ErrorCode, SdkError } from './interfaces';
+import { validatePeoplePickerInput } from '../internal/mediaUtil';
+import { sendMessageToParent } from '../internal/communication';
+
+export namespace peoplePicker {
```
ydogandjiev, line 7: Can we just use "people" as the namespace here? The namespace should really be more generic so it can encompass a wider variety of related APIs.

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * This is the SDK version when people picker API is supported on mobile.
+   */
+  export const peoplePickerRequiredVersion = '2.0.0';
```
ydogandjiev, line 11: We are actually going to v1.10.0 next. I noticed that we were using 2.0.0 here and also for some of the new media APIs. Any chance we can update that? Have we taken any dependencies on the 2.0.0 number for any in-production code yet?

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @param peoplePickerInputs Input parameters to launch customized people picker
+   */
+  export function selectPeople(
+    callback: (error: SdkError, people: PeoplePickerResult[]) => void,
+    peoplePickerInputs?: PeoplePickerInputs,
+  ): void {
+    if (!callback) {
+      throw new Error('[people picker] Callback cannot be null');
+    }
+    ensureInitialized(FrameContexts.content, FrameContexts.task);
```
ydogandjiev, line 26: Why only these two contexts? What about settings?

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+     * Default value is "Select people" for multiselect and "Select a person" for single select
+     */
+    title?: string;
+
+    /**
+     * Optional; AAD ids of the users to be pre-populated in the search box of people picker control
+     * If single select is enabled this value, only the first user in the list will be pre-populated
+     * Default value is null
+     */
+
```
ydogandjiev, line 58: NIT: Delete extra empty line between comment and parameter.

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+    setSelected?: string[];
+
+    /**
+     * Optional; launches the people picker in org wide scope even if the app is added to a chat or channel
+     * Default value is false
+     */
+
+    openOrgWideSearchInChatOrChannel?: boolean;
+
```
ydogandjiev, line 67: NIT: Delete extra empty line between comment and parameter.

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Optional; launches the people picker for which only 1 person can be selected
+     * Default value is false
+     */
+
```
ydogandjiev, line 72: NIT: Delete extra empty line between comment and parameter.

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+  export interface PeoplePickerResult {
+    /**
+     * aad Id of the selected user
+     */
+    aadId: string;
+
+    /**
+     * Optional; display name of the selected user
+     */
+
```
ydogandjiev, line 88: NIT: Delete extra empty line between comment and parameter.

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+    displayName?: string;
+
+    /**
+     * Optional; email of the selected user
+     */
+
```
ydogandjiev, line 94: NIT: Delete extra empty line between comment and parameter.

src/public/peoplePicker.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Output user object of people picker API
+   */
+  export interface PeoplePickerResult {
+    /**
+     * aad Id of the selected user
+     */
+    aadId: string;
```
ydogandjiev, line 83: AAD ID is very confusing. Is this the user's UPN? Or their Object ID? Please be more specific here and do not call this aadId. This should probably be objectId or just oid.

ydogandjiev (changes requested):

src/public/peoplePicker.ts
SonalJha: Media APIs, location APIs etc and just initializing these two hence followed suit. Please let me know if I should add settings for those as well.

src/public/peoplePicker.ts
SonalJha: done

src/public/peoplePicker.ts
SonalJha: done

src/public/peoplePicker.ts
sachingupta: Can not we fix compareSDKVersions function and use 1.10.0. I am hesitating to go with 2.0.0 which indicate major version changes means breaking changes.

@SonalJha has any previous version of prod sdk release used 2.0.0 ? if not then definitely we should fix compareSDKVersions  and use 1.10.0 here.

sachingupta (changes requested):
We need to fix version comparison here.

src/public/peoplePicker.ts
SonalJha: even if we fix compareSDKVersion() the older versions e.g. 1.9.0 will still break

SonalJha:
@sachingupta Let's take version comparisons fix as a different PR as these two are totally unrelated

ydogandjiev (approved):

sachingupta (approved):

PR #501, opened 2021-02-03
Add requestPreferredFileViewModeChange helper function

vivek3003:
With `requestPreferredFileViewModeChange` apps can allow users to update heir preferred file view mode. File opens will then respect this preference.

### DIFF
--- a/src/private/interfaces.ts
+++ b/src/private/interfaces.ts
@@ -37,6 +37,17 @@ export interface ShowNotificationParameters {
  * Hide from docs.
  * ------
  */
+export enum ViewerActionTypes {
+  view = 'view',
+  edit = 'edit',
+  editNew = 'editNew',
+}
+
+/**
+ * @private
+ * Hide from docs.
+ * ------
+ */
 export interface FilePreviewParameters {
   /**
    * The developer-defined unique ID for the file.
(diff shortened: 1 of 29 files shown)


src/private/interfaces.ts
```diff
@@ -110,6 +121,11 @@ export interface FilePreviewParameters {
    * Optional; indicates the mode in which file should be opened. Takes precedence over edit mode.
    */
   viewerAction?: ViewerActionTypes;
+
+  /**
+   * Optional; indicates where the file should be viewed when opened.
+   */
+  fileViewMode?: FileViewModes;
```
ydogandjiev, line 128: Why are we giving this control to the app? Should Teams be the one that decides how to open the file based on the user's preference?

ydogandjiev (changes requested):

src/private/interfaces.ts
vivek3003: So currently, a user can explicitly choose where to view a particular file while opening it. By adding this field apps can provide this ability.
In case this is not provided, teams will decide based on the set the preference. @ydogandjiev

vivek3003: Actually undoing this. For now the underlying app can provide this functionality. We can support this when we see more use cases.
For now teams will decide based on user preference. Will be undoing the changes.

vivek3003:
Ignoring this. Will breakdown and share the new PRs

PR #503, opened 2021-02-09
Chharpe/globalvars handlers

charpercyr:
Removing GlobalVars Part 2/3
This PR moves handler logic to the Handlers class

- Moved handler logic to Handlers
- Removed most handler variables, since they can be simply be called directly from Communication
- Made other variables private
- Removed some boilerplate code

This PR is dependent on #488


src/internal/communication.ts
```diff
@@ -0,0 +1,343 @@
+import { validOriginRegExp, version } from './constants';
+import { GlobalVars } from './globalVars';
+import { Handlers } from './handlers';
+import { MessageResponse, MessageRequest, ExtendedWindow, DOMMessageEvent } from './interfaces';
+
+export class Communication {
```
coshaw, line 6: I'm curious why you opted to create a static class here. This means that anywhere we want to use the members of this class, you need to import the class and call `Communication.foo`, which seems to be a pretty common pattern in this PR. It might be cleaner if these were free functions and variables on the module itself, so that in the places where it's used you could choose to do either of the following:
`import * as Communication from './communication'`
`import { sendMessageToParent } from './communication'`

The former would import everything from this file, and you'd still reference the members like `Communication.sendMessageToParent` as you're already doing - but the latter would mean that you could just call `sendMessageToParent`. As it stands now there isn't that flexibility, and you must import the entire Communication class from this module anywhere it's used.

charpercyr: My idea was to copy the "style" of GlobalVars. There would be no problem having free functions, but I wonder what @ydogandjiev and @sachingupta think about this

src/private/remoteCamera.ts
```diff
@@ -134,11 +136,6 @@ export namespace remoteCamera {
     ControlDisabled,
   }
 
-  GlobalVars.handlers['remoteCamera.capableParticipantsChange'] = handleCapableParticipantsChange;
-  GlobalVars.handlers['remoteCamera.handlerError'] = handleHandlerError;
```
sachingupta, line 138: How are we handling these registration ?

src/private/meetingRoom.ts
```diff
@@ -126,10 +127,6 @@ export namespace meetingRoom {
      */
     leaveMeeting: boolean;
   }
-
-  GlobalVars.handlers['meetingRoom.meetingRoomCapabilitiesUpdate'] = handleMeetingRoomCapabilitiesUpdate;
```
sachingupta, line 130: We also need initialize for these registration

src/private/logs.ts
```diff
@@ -9,15 +10,6 @@ import { GlobalVars } from '../internal/globalVars';
  * Hide from docs
  */
 export namespace logs {
-  GlobalVars.handlers['log.request'] = handleGetLogRequest;
-
-  function handleGetLogRequest(): void {
-    if (GlobalVars.getLogHandler) {
-      const log: string = GlobalVars.getLogHandler();
```
sachingupta, line 16: should this also be part of initialize

src/internal/globalVars.ts
```diff
@@ -1,49 +1,14 @@
...
-  public static backButtonPressHandler: () => boolean;
-  public static loadHandler: (context: LoadContext) => void;
-  public static beforeUnloadHandler: (readyToUnload: () => void) => boolean;
-  public static changeSettingsHandler: () => void;
-  public static onStartConversationHandler: (conversationResponse: ConversationResponse) => void;
-  public static onCloseConversationHandler: (conversationResponse: ConversationResponse) => void;
-  public static getLogHandler: () => string;
-  public static appButtonClickHandler: () => void;
-  public static appButtonHoverEnterHandler: () => void;
-  public static appButtonHoverLeaveHandler: () => void;
```
sachingupta, line 42: How these handlers are handle now ?

src/internal/communication.ts
sachingupta: I also like using free functions, similar to ensureInitialized(). GlobalVars were use to keep non functional object. I think we should update static class to free functions for communications and handlers files

sachingupta (changes requested):

src/private/remoteCamera.ts
charpercyr: For all handlers that don't have a default / always needs to run behavior (does nothing if there is no handlers), the registration is done in the "registerXXXHandler" call. The rest will use initialize to register the default handler and use a handler function variable to register the handler.

src/internal/globalVars.ts
charpercyr: See previous comment

charpercyr:
@sachingupta, I will do refactor to free functions as part of this PR for both Communication and Handlers. Is that okay for you?

charpercyr:
Changes included in #518

PR #505, opened 2021-02-09
Add user file open preference in context & send user's file open pref while opening files

vivek3003:
Apps can read where user's prefer to open files. The current values are:
Teams, Desktop (Desktop app of the file such as excel/powerpoint), Web (browser)

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -404,6 +404,11 @@ export interface Context {
    */
   userClickTime?: number;
 
+  /*
+   * Team Template ID if there was a Team Template associated with the creation of the team.
+   */
+  teamTemplateId?: string;
+
   /**
    * Where the user prefers the file to be opened from by default during file open
    */
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -384,6 +384,16 @@ export interface Context {
    * Personal app icon y coordinate position
    */
   appIconPosition?: number;
+
+  /**
+   * Source origin from where the tab is opened
+   */
+  sourceOrigin?: string;
+
+  /**
+   * Time when the user clicked on the tab
+   */
+  userClickTime?: number;
 }
 
 export interface DeepLinkParameters {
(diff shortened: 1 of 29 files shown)


ydogandjiev:
@vivek3003, were you planning to also add a query API for the setting that apps can use to figure out what the current setting is?

ydogandjiev:
> @vivek3003, were you planning to also add a query API for the setting that apps can use to figure out what the current setting is?

src/public/interfaces.ts
```diff
@@ -394,6 +403,11 @@ export interface Context {
    * Time when the user clicked on the tab
    */
   userClickTime?: number;
+
+  /**
+   * Where the user prefers the file to be opened from by default during file open
+   */
+  userFileOpenPreference?: FileOpenPreference;
```
ydogandjiev, line 410: Where is this information stored on the client? Does it ever require making a network call or is the information always available locally?

src/public/interfaces.ts
```diff
@@ -158,6 +158,15 @@ export interface LocaleInfo {
   longTime: string;
 }
 
+/**
+ * Allowed user file open preferences
+ */
+export enum FileOpenPreference {
+  Teams = 'teams',
```
ydogandjiev, line 165: Any chance we can make this a little more generic - e.g. Hub, InApp, or something like that? We're soon going to be replacing the Teams Client SDK with the MetaOS App SDK which will used to host apps in a number of different Office hubs like Outlook, Office.com, etc.

ydogandjiev (commented):
As discussed offline, lets please review these changes with @sachingupta and myself before committing this PR.

src/public/interfaces.ts
vivek3003: This is stored and read from the preferences store. It's read synchronously while passing to the context. 
No network called involved.

src/public/interfaces.ts
vivek3003: Yes, we should be able to. But we will need to write a mapping function in teams to read&pass this values as inApp.

vivek3003: Have updated to 'inApp'

ydogandjiev (approved):

PR #507, opened 2021-02-13
Add surfaceHub platform to HostClientType

kpyapali:
Add surfaceHub2 platform to HostClientType with value='surfaceHub'.
Use this value for SurfaceHub 2s specific scenarios/workflows.

### DIFF
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -4,6 +4,7 @@ export enum HostClientType {
   android = 'android',
   ios = 'ios',
   rigel = 'rigel',
+  surfaceHub2 = 'surfaceHub2',
 }
 
 // Ensure these declarations stay in sync with the framework.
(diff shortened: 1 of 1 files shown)


msudankapoor (approved):

src/public/constants.ts
```diff
@@ -4,6 +4,7 @@ export enum HostClientType {
   android = 'android',
   ios = 'ios',
   rigel = 'rigel',
+  surfaceHub2 = 'surfaceHub2',
```
sachingupta, line 7: as discussed please rename it to surfaceHub

ydogandjiev, line 7: As discussed in our sync please change this to the more generic `surfaceHub` instead of `surfaceHub2`.

ydogandjiev (changes requested):

sachingupta (changes requested):

src/public/constants.ts
kpyapali: Changed.

kpyapali: changed.

sachingupta (approved):

ydogandjiev (approved):

PR #517, opened 2021-02-25
Add registerUserSettingsChange handler

vivek3003:
The following PR adds a new registerHandler.

This handler can be used to register for changes in user's settings. The callback will be triggered whenever the settings subscribed to change.

Currently, adding support for file open preference.

Sample usage:

```
microsoftTeams.registerUserSettingsChangeHandler([microsoftTeams.userSettingKeys.fileOpenPreference], (settingKey, newValue) => {
        // The setting (identified by the key has been changed by the user
    });
```


src/internal/globalVars.ts
```diff
@@ -46,4 +46,5 @@ export class GlobalVars {
   public static remoteCameraErrorHandler: (error: remoteCamera.ErrorReason) => void;
   public static remoteCameraDeviceStateChangeHandler: (deviceStateChange: remoteCamera.DeviceState) => void;
   public static remoteCameraSessionStatusChangeHandler: (sessionStatusChange: remoteCamera.SessionStatus) => void;
+  public static userSettingsChangeHandler: (updatedSettingKey: userSettingKeys, updatedValue: any) => void;
```
shishir-srivastava, line 49:
```suggestion
  public static userSettingsChangeHandler: (settingKey: userSettingKeys, value: any) => void;
```
The handler is already called a "change handler". I feel that adding the "updated" prefix to each param may be redundant.

src/internal/internalAPIs.ts
```diff
@@ -116,6 +117,12 @@ function handleAppButtonHoverLeave(): void {
   }
 }
 
+function handleUserSettingsChange(udatedSettingKey, updatedSettingValue): void {
+  if (GlobalVars.userSettingsChangeHandler) {
+    GlobalVars.userSettingsChangeHandler(udatedSettingKey, updatedSettingValue);
+  }
+}
```
shishir-srivastava, line 124: Same as above (param naming).

src/public/interfaces.ts
```diff
@@ -167,6 +167,15 @@ export enum FileOpenPreference {
   Web = 'web',
 }
 
+/**
+ * User setting changes that can be subscribed to,
+ */
+export enum userSettingKeys {
```
shishir-srivastava, line 173: This should be in Pascal case.

shishir-srivastava: Also, add provision to subscribe to theme changes.

sachingupta (changes requested):
We just merge one of major refactoring PR which will impact the way we register handler. https://github.com/OfficeDev/microsoft-teams-library-js/pull/518

Please merge latest master and update your changes.

vivek3003:
> We just merge one of major refactoring PR which will impact the way we register handler. #518
> 
> Please merge latest master and update your changes.

@sachingupta Updated the PR.

src/internal/handlers.ts
```diff
@@ -88,6 +90,14 @@ export function registerBeforeUnloadHandler(handler: (readyToUnload: () => void)
   handler && sendMessageToParent('registerHandler', ['beforeUnload']);
 }
 
+export function registerUserSettingsChangeHandler(
+  settingKeys: UserSettingKeys[],
+  handler: (settingKey: UserSettingKeys, value: any) => void,
```
sachingupta, line 95: We should use registerHandler, which ca be extended to take params/args as part of register handler.

src/public/publicAPIs.ts
```diff
@@ -328,3 +329,17 @@ export function initializeWithFrameContext(
...
   setFrameContext(frameContext);
 }
+
+/**
+ * register a handler to be called when a user setting changes. The changed setting key & value is provided in the callback.
+ * @param settingKeys List of user setting changes to subscribe
+ * @param handler When a subscribed setting is updated this handler is called
+ */
+export function registerUserSettingsChangeHandler(
+  settingKeys: UserSettingKeys[],
```
sachingupta, line 339: We just need to extend registerHandler to support args in this case settingKeys and use it here.

sachingupta (changes requested):

src/private/interfaces.ts
```diff
@@ -43,6 +43,23 @@ export enum ViewerActionTypes {
   editNew = 'editNew',
 }
 
+/**
+ * * @private
+ * Hide from docs.
+ * ------
+ * User setting changes that can be subscribed to,
+ */
+export enum UserSettingKeys {
```
sachingupta, line 52: May be we just say UserSettingTypes ? Keys looks very technical here.

src/private/privateAPIs.ts
```diff
@@ -203,3 +204,18 @@ export function getConfigSetting(callback: (value: string) => void, key: string)
...
+
+/**
+ * @private
+ * register a handler to be called when a user setting changes. The changed setting key & value is provided in the callback.
+ * @param settingKeys List of user setting changes to subscribe
+ * @param handler When a subscribed setting is updated this handler is called
+ */
+export function registerUserSettingsChangeHandler(
+  settingKeys: UserSettingKeys[],
+  handler: (settingKey: UserSettingKeys, value: any) => void,
```
sachingupta, line 216: with this design we are expecting handler to execute for each setting change. Which make sense.

sachingupta (approved):

charpercyr (approved):

src/private/privateAPIs.ts
```diff
@@ -203,3 +204,18 @@ export function getConfigSetting(callback: (value: string) => void, key: string)
...
+): void {
+  ensureInitialized(FrameContexts.content);
```
ydogandjiev, line 218: This API should work in any frame context so you probably shouldn't pass a frame context to the `ensureInitialized` call.

ydogandjiev (approved):

src/private/privateAPIs.ts
vivek3003: This is done

src/private/interfaces.ts
vivek3003: Changed to UserSettingTypes.

PR #518, opened 2021-02-25
GlobalVars changes part 1/3 & 2/3

charpercyr:
This is a merge of PRs #488 and #503.

> ## Removing GlobalVars Part 1/3
> This PR moves all the parent/child message handling to a file called communication.ts.
> 
> * Move all message handling functions to the Communication file
> * Changed sendMessageRequestToParent to sendMessageToParent, which optionally accepts a callback. This new function does the work of both sendMessageRequestToParent and GlobalVars.callback[id] = ...
> * Some housekeeping in functions in Communication
> * Removed file level handler registration, replaced with calls in initialization functions

> ## Removing GlobalVars Part 2/3
> This PR moves handler logic to the Handlers class
> 
> * Moved handler logic to Handlers
> * Removed most handler variables, since they can be simply be called directly from Communication
> * Made other variables private
> * Removed some boilerplate code
> 
> This PR is dependent on #488

This PR also includes the changes to move all the static functions in Communication & Handlers to free functions


src/internal/handlers.ts
```diff
@@ -0,0 +1,99 @@
+import { LoadContext, navigateBack } from '../public';
+import { Communication, sendMessageEventToChild, sendMessageToParent } from './communication';
+
+class Handlers {
```
charpercyr, line 4: I could change this to HandlersPrivate to be more consistent @sachingupta what do you think?

charpercyr:
End-to-End just passed for this change

coshaw (approved):
LGTM

src/internal/communication.ts
```diff
@@ -0,0 +1,337 @@
...
+
+/**
+ * Send a message to parent. Uses nativeInterface on mobile to communicate with parent context
+ */
+export function sendMessageToParent(actionName: string, callback?: Function): void;
+/**
+ * Send a message to parent. Uses nativeInterface on mobile to communicate with parent context
+ */
+export function sendMessageToParent(actionName: string, args: any[], callback?: Function): void;
+export function sendMessageToParent(actionName: string, argsOrCallback?: any[] | Function, callback?: Function): void {
```
sachingupta, line 79: why do we define this function 3 times?

src/private/logs.ts
```diff
@@ -28,7 +20,13 @@ export namespace logs {
...
     ensureInitialized();
 
-    GlobalVars.getLogHandler = handler;
-    handler && sendMessageRequestToParent('registerHandler', ['log.request']);
+    if (handler) {
+      registerHandler('log.request', () => {
+        const log: string = handler();
+        sendMessageToParent('log.receive', [log]);
+      });
+    } else {
```
sachingupta, line 28: why remove handler is not passed. Is this for handling 2nd call to override it?

src/public/settings.ts
```diff
@@ -10,8 +11,11 @@ import { getGenericOnCompleteHandler } from '../internal/utils';
 export namespace settings {
   let saveHandler: (evt: SaveEvent) => void;
   let removeHandler: (evt: RemoveEvent) => void;
-  GlobalVars.handlers['settings.save'] = handleSave;
-  GlobalVars.handlers['settings.remove'] = handleRemove;
+
+  export function initialize(): void {
+    registerHandler('settings.save', handleSave, false);
+    registerHandler('settings.remove', handleRemove, false);
```
sachingupta, line 17: why not do it inside this function registerOnRemoveHandler?

src/internal/communication.ts
```diff
@@ -0,0 +1,337 @@
...
+import { MessageResponse, MessageRequest, ExtendedWindow, DOMMessageEvent } from './interfaces';
+
+export class Communication {
+  public static currentWindow: Window | any;
+  public static parentOrigin: string;
+  public static parentWindow: Window | any;
+  public static childWindow: Window;
+  public static childOrigin: string;
+}
+
```
sachingupta, line 13: May be CommunicationInternal ?

src/public/publicAPIs.ts
```diff
@@ -212,9 +182,7 @@ export function registerOnThemeChangeHandler(handler: (theme: string) => void):
  */
 export function registerFullScreenHandler(handler: (isFullScreen: boolean) => void): void {
   ensureInitialized();
-
-  GlobalVars.fullScreenChangeHandler = handler;
-  handler && sendMessageRequestToParent('registerHandler', ['fullScreen']);
+  Handlers.registerHandler('fullScreenChange', handler);
```
sachingupta, line 185: Loved this approach, will simplify this kind of APIs in future.

sachingupta (approved):

src/public/settings.ts
charpercyr: Because they both have default behavior, I kept the old pattern for when there is more work to do than just calling a handler (see also Handlers for more examples)

src/internal/communication.ts
charpercyr: This will create better documentation. The first two are declarations of possible interfaces (with or without args), the last one the implementation.
https://www.typescriptlang.org/docs/handbook/functions.html#overloads

src/private/logs.ts
charpercyr: I am not sure what you mean by this. I was simply copying the existing logic for logs.

PR #520, opened 2021-03-02
Support new frame context for tab rendered in meeting stage

ronakbakshi:


sachingupta (approved):
Changes looks good to me. One thing to consider that, All SDK APIs supported in content context should also be supported for meetingStage.

PR #531, opened 2021-03-22
Add cloud storage folders api

vivek3003:
The following PR introduces the following new private APIs

- files.getCloudStorageFolders
- files.addCloudStorageFolder
- files.deleteCloudStorageFolder

This will be privateAPIs that will be nested under files namespace. They will be allowed only for 1P apps.


src/private/files.ts
```diff
@@ -0,0 +1,160 @@
...
+    callback: (error: SdkError, isFolderAdded: boolean, folders: CloudStorageFolder[]) => void,
+  ): void {
+    if (!channelId || channelId.length == 0) {
+      throw new Error('[files.addCloudStorageFolder] channelId name cannot be null or empty');
+    }
+    if (!callback) {
+      throw new Error('[files.addCloudStorageFolder] Callback cannot be null');
+    }
+
+    ensureInitialized();
```
ydogandjiev, line 129: Please changes this to `ensureInitialized(FrameContexts.content)` since this API should never be called from any other frame context.

src/private/files.ts
```diff
@@ -0,0 +1,160 @@
...
+      throw new Error('[files.deleteCloudStorageFolder] channelId name cannot be null or empty');
+    }
+    if (!folderToDelete) {
+      throw new Error('[files.deleteCloudStorageFolder] folderToDelete cannot be null or empty');
+    }
+    if (!callback) {
+      throw new Error('[files.deleteCloudStorageFolder] Callback cannot be null');
+    }
+
+    ensureInitialized();
```
ydogandjiev, line 157: Please changes this to `ensureInitialized(FrameContexts.content)` since this API should never be called from any other frame context.

src/private/files.ts
ydogandjiev: Also, it's probably best to make this the first call in the method before you check any of the parameters.

src/private/files.ts
```diff
@@ -0,0 +1,160 @@
...
+    callback: (error: SdkError, folders: CloudStorageFolder[]) => void,
+  ): void {
+    if (!channelId || channelId.length == 0) {
+      throw new Error('[files.getCloudStorageFolders] channelId name cannot be null or empty');
+    }
+    if (!callback) {
+      throw new Error('[files.getCloudStorageFolders] Callback cannot be null');
+    }
+
+    ensureInitialized();
```
ydogandjiev, line 106: Please changes this to `ensureInitialized(FrameContexts.content)` since this API should never be called from any other frame context.

ydogandjiev (approved):

src/private/files.ts
```diff
@@ -0,0 +1,163 @@
...
+ */
+export namespace files {
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Cloud storage providers registered with Microsoft Teams
+   */
+  export enum CloudStorageProviderCode {
+    Dropbox = 'DROPBOX',
```
sachingupta, line 19: better name could be "CloudStorageProvider"

sachingupta (approved):

PR #536, opened 2021-03-26
Added check for clientSupportedSdkVersion and hostclient for `getUserJoinedTeams` execution on android

shbhadra:
Explain Defect

- The android client released support for a private JS SDK API userJoinedTeams(), support for the same was released at around August 2020.
- However, the released native implementation had a bug that did not return the correct user role. It returned Admin as the role for all TeamInformation.userTeamRole returned. This impaired critical workflows in the Assignments app, wherein they show different UI based on whether a user is a teacher or a student, leading to a student seeing UI meant for a Teacher.
- The bug was fixed for the android client release in January 2021, for the build version number 1416/1.0.0.2021022202.
- However even with the fix, there are customers who are on an older version of the android client, due to the critical nature of the impact, assignments is not able to roll out the feature.
- Along with the fix, Assignments app also requires a way to identify if they are running on an older android client that does not have the required fix, so that they can turn off the functionality only on that client and roll it out for rest.

Explain fix
- Please refer to the following [one-pager](https://microsoft-my.sharepoint.com/:w:/p/shbhadra/EXgJP9sLeKpIq_TmPAqtjOEBV4PGei_9KSeNodqO80jYag?e=qi3tOr) for details of the proposed solution and the associated pros and cons.
- The associated changes on the android client can be seen in this [PR](https://domoreexp.visualstudio.com/Teamspace/_git/SkypeSpaces-Android/pullrequest/331425)

Validation
Validated by adding the JS SDK changes to a local version of microsoft-teams-test-tab, validated the following flows:

With newer JS SDK :
- Throws an error on older android clients (i.e. clients that set the CLIENT_SUPPORTED_JS_SDK_VERSION to 2.0.0 and lower).
- Returns the correct result on the newer android clients (setting the above value to 2.0.1)
- For non-full trust apps on the newer android client, sends a null response.

Did sanity test for selectMedia, getMedia and location APIs.

### DIFF
--- a/src/internal/constants.ts
+++ b/src/internal/constants.ts
@@ -7,6 +7,41 @@ export const version = '1.9.0';
 export const defaultSDKVersionForCompatCheck = '1.6.0';
 
 /**
+ * Minimum required client supported version for {@link getUserJoinedTeams} to be supported on {@link HostClientType.android}
+ */
+export const getUserJoinedTeamsSupportedAndroidClientVersion = '2.0.1';
+
+/**
+ * This is the SDK version when location APIs (getLocation and showLocation) are supported.
+ */
+export const locationAPIsRequiredVersion = '1.9.0';
+
+/**
+ * This is the SDK version when people picker API is supported on mobile.
+ */
+export const peoplePickerRequiredVersion = '2.0.0';
+
+/**
+ * This is the SDK version when captureImage API is supported on mobile.
+ */
+export const captureImageMobileSupportVersion = '1.7.0';
+
+/**
+ * This is the SDK version when media APIs is supported on all three platforms ios, android and web.
+ */
+export const mediaAPISupportVersion = '1.8.0';
+
+/**
+ * This is the SDK version when getMedia API is supported via Callbacks on all three platforms ios, android and web.
+ */
+export const getMediaCallbackSupportVersion = '2.0.0';
+
+/**
+ * This is the SDK version when scanBarCode API is supported on mobile.
+ */
+export const scanBarCodeAPIMobileSupportVersion = '1.9.0';
+
+/**
  * List of supported Host origins
  */
 export const validOrigins = [
--- a/src/internal/constants.ts
+++ b/src/internal/constants.ts
@@ -27,7 +27,7 @@ export const peoplePickerRequiredVersion = '2.0.0';
 export const captureImageMobileSupportVersion = '1.7.0';
 
 /**
- * This is the SDK version when media APIs is supported on all three platforms ios, android and web.
+ * This is the SDK version when media APIs are supported on all three platforms ios, android and web.
  */
 export const mediaAPISupportVersion = '1.8.0';
 
--- a/src/private/privateAPIs.ts
+++ b/src/private/privateAPIs.ts
@@ -14,8 +14,7 @@ import { menus } from './menus';
 import { registerHandler } from '../internal/handlers';
 import { GlobalVars } from '../internal/globalVars';
 import { ErrorCode, SdkError } from '../public/interfaces';
-
-const getUserJoinedTeamsSupportedAndroidClientVersion = '2.0.1';
+import { getUserJoinedTeamsSupportedAndroidClientVersion } from '../internal/constants';
 
 export function initializePrivateApis(): void {
   menus.initialize();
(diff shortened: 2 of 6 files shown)


src/private/privateAPIs.ts
```diff
@@ -12,6 +12,10 @@ import { getGenericOnCompleteHandler } from '../internal/utils';
 import { Communication, sendMessageToParent, sendMessageEventToChild } from '../internal/communication';
 import { menus } from './menus';
 import { registerHandler } from '../internal/handlers';
+import { GlobalVars } from '../internal/globalVars';
+import { ErrorCode, SdkError } from '../public/interfaces';
+
+const getUserJoinedTeamsSupportedAndroidClientVersion = '2.0.1';
```
sachingupta, line 18: Could you please move all min required versions to internal const file. Wanted to keep all hardcoded version at one place.

sachingupta (changes requested):

src/private/privateAPIs.ts
shbhadra: Have moved the hardcoded versions to the internal constants file. Please take a look.

src/private/privateAPIs.ts
```diff
@@ -31,6 +34,14 @@ export function getUserJoinedTeams(
 ): void {
   ensureInitialized();
```
pankajmi2, line 36: I see this API was introduced after we introduced versioning change in Teams SDK. Then, why is there no versioning check in this API ? You are just placing a check now and that too for android. What about iOS ?

shbhadra: Not sure why the API didn't have the versioning parameter as a part of it's signature, maybe @sachingupta can help with that. As highlighted in the document attached in the description. iOS doesn't have the mentioned issue and hence doesn't need any special checks in place

src/private/privateAPIs.ts
shbhadra: @sachingupta could you please review the changes.

src/internal/constants.ts
```diff
@@ -6,6 +6,41 @@ export const version = '1.9.0';
...
+ */
+export const peoplePickerRequiredVersion = '2.0.0';
+
+/**
+ * This is the SDK version when captureImage API is supported on mobile.
+ */
+export const captureImageMobileSupportVersion = '1.7.0';
+
+/**
+ * This is the SDK version when media APIs is supported on all three platforms ios, android and web.
```
SonalJha, line 30: There are 3 media APIs so please change is to are.

src/private/privateAPIs.ts
SonalJha: If the user is on ios and the ios app is old and doesn't have implementation of this API the web app developers will not get an error message(OLD_PLATFORM). They will simply find the API no-op. To ensure passing proper error message back to the user Pankaj had introduced versioning. It will be good to keep it consistent with all APIs.

shbhadra: That is correct. I completely understand the rationale behind the previously introduced versioning strategy and the repercussions of not having a version check in the implementation of the API. However I am not the primary contributor for the API, and the API has already been implemented without a version check, with a method signature that is not compatible with the versioning strategy.

The scope of this PR is to only address the issues faced on android, the existing behaviour on any of the other clients is not to be taken up in this PR. Making the API compliant with the versioning strategy requires a larger discussion and should be taken up as a separate item. 
@ydogandjiev  @sachingupta to chime in on how to take this on further

src/internal/constants.ts
shbhadra: changed.

pankajmi2 (approved):

SonalJha (approved):

src/private/privateAPIs.ts
```diff
@@ -31,6 +34,14 @@ export function getUserJoinedTeams(
...
 
+  if (
+    GlobalVars.hostClientType === HostClientType.android &&
```
ydogandjiev, line 38: Why are we just checking for Anroid here? This doesn't seem like a good idea.

ydogandjiev (changes requested):

ydogandjiev (approved):

sachingupta (approved):

PR #546, opened 2021-04-15
Kpolnitz/timestamp post msg

kpolnitz:
PR for adding timestamp to MessageRequests sent through PostMessage

### DIFF
--- a/src/internal/communication.ts
+++ b/src/internal/communication.ts
@@ -86,6 +86,7 @@ export function sendMessageToParent(actionName: string, argsOrCallback?: any[] |
 
   const targetWindow = Communication.parentWindow;
   const request = createMessageRequest(actionName, args);
+  //put in delay to mimic behavior, setTimeout 5 secs 
   if (GlobalVars.isFramelessWindow) {
     if (Communication.currentWindow && Communication.currentWindow.nativeInterface) {
       (Communication.currentWindow as ExtendedWindow).nativeInterface.framelessPostMessage(JSON.stringify(request));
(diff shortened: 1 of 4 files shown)


src/internal/communication.ts
```diff
@@ -312,6 +312,7 @@ function createMessageRequest(func: string, args: any[]): MessageRequest {
   return {
     id: CommunicationPrivate.nextMessageId++,
     func: func,
+    time: new Date().toISOString(),
```
ydogandjiev, line 315: Could we just use a simple milliseconds timestamp here instead of a fully formatted date? Formatting the date like this seems to incur a performance hit which even though small would just add latency to our API calls:
image

ydogandjiev (changes requested):

src/internal/communication.ts
```diff
@@ -312,6 +312,7 @@ function createMessageRequest(func: string, args: any[]): MessageRequest {
   return {
     id: CommunicationPrivate.nextMessageId++,
     func: func,
+    creationTimestamp: new Date().getTime(),
```
sachingupta, line 315: Lets use Date.Now()

sachingupta (changes requested):

src/internal/communication.ts
```diff
@@ -312,6 +312,7 @@ function createMessageRequest(func: string, args: any[]): MessageRequest {
   return {
     id: CommunicationPrivate.nextMessageId++,
     func: func,
+    creationTimestamp: Date.now(),
```
ydogandjiev, line 315: Why not just call this `timestamp`? The `creationTimestamp` seems a little wordy. @sachingupta

ydogandjiev (approved):

sachingupta (approved):

PR #551, opened 2021-04-20
Add selectAppEntity private API

andreipatrascoiu:
Add a new private API.

This API will be called from Assignments and it will open the Tab Gallery in Teams. After the user configures their app, the configuration will be returned by this API.

### DIFF
--- a/src/private/appEntity.ts
+++ b/src/private/appEntity.ts
@@ -1,7 +1,6 @@
-import { SdkError } from '@microsoft/teams-js';
 import { sendMessageToParent } from '../internal/communication';
 import { ensureInitialized } from '../internal/internalAPIs';
-import { FrameContexts } from '../public';
+import { FrameContexts, SdkError } from '../public';
 
 /**
  * Namespace to interact with the application entities specific part of the SDK.
--- a/src/private/appEntity.ts
+++ b/src/private/appEntity.ts
@@ -1,7 +1,7 @@
+import { SdkError } from '@microsoft/teams-js';
 import { sendMessageToParent } from '../internal/communication';
 import { ensureInitialized } from '../internal/internalAPIs';
 import { FrameContexts } from '../public';
-import { AppEntityConfiguration } from '../private';
 
 /**
  * Namespace to interact with the application entities specific part of the SDK.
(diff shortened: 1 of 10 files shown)


src/private/appEntity.ts
```diff
@@ -0,0 +1,31 @@
...
+   *
+   * Open the Tab Gallery and retrieve the app entity configuration
+   * @param channelId ID of the channel where the app entity will be created
+   * @param categories A list of app categories that will be displayed in the open tab gallery
+   * @param callback Callback that will be triggered once the app entity information is available
+   */
+  export function selectAppEntity(
+    channelId: string,
+    categories: string[],
+    callback: (appEntity: AppEntityConfiguration) => void,
```
sachingupta, line 25: how are we handling error?

src/private/interfaces.ts
```diff
@@ -159,3 +159,36 @@ export interface UserJoinedTeamsInformation {
...
   userJoinedTeams: TeamInformation[];
 }
+
+/**
+ * @private
+ * Hide from docs
+ * --------
+ * Information on an app entity
+ */
+export interface AppEntityConfiguration {
```
sachingupta, line 169: We can keep this interface inside appEntity namespace or in same file.

sachingupta (changes requested):

src/private/appEntity.ts
andreipatrascoiu: I have added an SdkErrro parameter to the callback. We will use that to communicate any error conditions to the main application.

src/private/appEntity.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Open the Tab Gallery and retrieve the app entity configuration
+   * @param channelId ID of the channel where the app entity will be created
+   * @param categories A list of app categories that will be displayed in the open tab gallery
+   * @param callback Callback that will be triggered once the app entity information is available.
+   *                 The callback takes two arguments: the app entity configuration, if available and
+   *                 an optional SdkError in case something happened (i.e. the window was closed)
+   */
+  export function selectAppEntity(
+    channelId: string,
+    categories: string[],
```
ydogandjiev, line 58: I wonder if we should use an enum for the categories so they are not arbitrary strings that the developer has to decipher...

src/private/appEntity.ts
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * Open the Tab Gallery and retrieve the app entity configuration
+   * @param channelId ID of the channel where the app entity will be created
+   * @param categories A list of app categories that will be displayed in the open tab gallery
+   * @param callback Callback that will be triggered once the app entity information is available.
+   *                 The callback takes two arguments: the app entity configuration, if available and
+   *                 an optional SdkError in case something happened (i.e. the window was closed)
+   */
+  export function selectAppEntity(
+    channelId: string,
```
ydogandjiev, line 57: I wonder if we should just call this `threadId` since we might want to expand this functionality to chats/meetings at some point...

ydogandjiev (approved):
@sachingupta, can you please take a look at this PR and some of the comments I left when you get a chance?

sachingupta (approved):

PR #554, opened 2021-04-23
Add notifyInitializedWithErrors API

ht4963-ms:
Add notifyInitializedWithErrors along with new data contract.
Updated API comments for clarity.

### DIFF
--- a/src/public/appInitialization.ts
+++ b/src/public/appInitialization.ts
@@ -7,7 +7,7 @@ export namespace appInitialization {
     appLoaded: 'appInitialization.appLoaded',
     success: 'appInitialization.success',
     failure: 'appInitialization.failure',
-    initializedWithErrors: 'appInitialization.initializedWithErrors',
+    expectedFailure: 'appInitialization.expectedFailure',
   };
 
   /**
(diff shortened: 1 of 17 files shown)


src/public/appInitialization.ts
```diff
@@ -3,41 +3,70 @@ import { version } from '../internal/constants';
...
+    sendMessageToParent(notifyMessages.failure, [
       appInitializationFailedRequest.reason,
       appInitializationFailedRequest.message,
     ]);
   }
 
+  /**
+   * Notifies the frame that app initialized with some expected errors.
+   */
+  export function notifyInitializedWithErrors(appInitializedWithErrorsRequest: IInitWithErrorsRequest): void {
```
ydogandjiev, line 43: Consider adding a UT to validate the API sends the expected message and passes along the expected parameters

ydogandjiev (approved):

src/public/appInitialization.ts
```diff
@@ -3,41 +3,70 @@ import { version } from '../internal/constants';
 import { sendMessageToParent } from '../internal/communication';
 
 export namespace appInitialization {
+  export const notifyMessages = {
+    appLoaded: 'appInitialization.appLoaded',
+    success: 'appInitialization.success',
+    failure: 'appInitialization.failure',
```
sachingupta, line 9: [NIT] this is something we should do for all sdk APIs. Probably rename it to export SDKAPIs = {
appInitialization: {
   appLoaded: 'appInitialization.appLoaded',
    success: 'appInitialization.success',
    failure: 'appInitialization.failure',
 initializedWithErrors: 'appInitialization.initializedWithErrors',
}

}
}

sachingupta (approved):

src/public/appInitialization.ts
ht4963-ms: Since this is exported under the appInitialization namespace, it is already available as a constant when accessing the package.  You can reference it like this:
I've noticed that none of the other message strings are exported. This is something we should fix in a separate PR. All the message string constants should be exposed so we can reference them directly. We can create a new object for SDKMessages.

src/public/appInitialization.ts
ht4963-ms: Just added a new test suite for appInitialization

PR #556, opened 2021-04-26
process message with response timestamp

kpolnitz:
Changes to read response from teams client response object

### DIFF
--- a/src/internal/interfaces.ts
+++ b/src/internal/interfaces.ts
@@ -31,6 +31,7 @@ export interface ExtendedWindow extends Window {
 export interface MessageRequest {
   id?: number;
   func: string;
+  timestamp?: number;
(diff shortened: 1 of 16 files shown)


src/internal/interfaces.ts
```diff
@@ -37,6 +37,7 @@ export interface MessageRequest {
 
 export interface MessageResponse {
   id: number;
+  timestamp?: number;
```
ydogandjiev, line 40: How would the developer get this value so they can log it?

ydogandjiev (approved):

sachingupta (changes requested):
We had a discussion offline. Sending timestamp from client to app require some more thinking and changes in sdk. Which will take more time. for now we are only doing timestamp one way

PR #578, opened 2021-05-13
Add new hostClientTypes

jmsy:
Add new hostClientType for Teams Android devices and MTR on Windows devices.


src/public/interfaces.ts
```diff
@@ -334,7 +334,8 @@ export interface Context {
   isTeamArchived?: boolean;
 
   /**
-   * The type of the host client. Possible values are : android, ios, web, desktop, rigel
+   * The type of the host client. Possible values are : android, ios, web, desktop, rigel,
```
steffenloesch, line 337: I think it would be good to indicate that rigel is being deprecated.

jmsy: Thanks, added @deprecated annotation.

ydogandjiev (approved):

src/private/privateAPIs.ts
```diff
@@ -35,7 +35,10 @@ export function getUserJoinedTeams(
   ensureInitialized();
 
   if (
-    GlobalVars.hostClientType === HostClientType.android &&
+    (GlobalVars.hostClientType === HostClientType.android ||
+      GlobalVars.hostClientType === HostClientType.teamsRoomsAndroid ||
```
ydogandjiev, line 39: Why add these here? This change was made to address a very specific issue in Android Mobile Client. I don't think we should be adding more entries here.

ydogandjiev (changes requested):

src/private/privateAPIs.ts
jmsy: Originally all the Android platforms, including mobile and devices were sharing the same HostClientType - `android`, so I assume here the we should apply the logic to all the Android platforms to keep the same behavior as before. Do you think it's unnecessary? How about other changes in `authentication.ts` and `media.ts` ?

ydogandjiev: Hmm... Are you saying that all of these are just variations of our Android client? Do you know who owns the Client SDK implementation for these different client types? @shbhadra, can you please chime in here.

shbhadra: Hey Edward, is the corresponding change for this checked into the android client? 
The value for `GlobalVars.hostClientType` is set during the initialize() call, and is a value that is sent by the client. 
Android code seems to be still setting this as 'android' always.

jmsy: I have another PR to make the hostClientType change on Android side, it's not merged to develop now. All the new types here are subtype of "android". 
https://domoreexp.visualstudio.com/Teamspace/_git/SkypeSpaces-Android/pullrequest/352658

ydogandjiev (approved):

shbhadra (approved):

PR #585, opened 2021-06-01
Get a team's channels by providing teamId API

yumikohey:
The following PR introduces the following new private API

`teams.getTeamsChannels`

This will be privateAPI that will be nested under `teams` namespace. It will be allowed only for 1P apps.

### DIFF
--- a/src/private/teams.ts
+++ b/src/private/teams.ts
@@ -30,10 +30,7 @@ export namespace teams {
    * Get a list of channels belong to a Team
    * @param groupId a team's objectId
    */
-  export function getTeamChannels(
-    groupId: string,
-    callback: (error: SdkError, channels: ChannelInfo[]) => void,
-  ): void {
+  export function getTeamChannels(groupId: string, callback: (error: SdkError, channels: ChannelInfo[]) => void): void {
     ensureInitialized(FrameContexts.content);
 
     if (!groupId) {
--- a/src/private/teams.ts
+++ b/src/private/teams.ts
@@ -1,0 +1,50 @@
+import { sendMessageToParent } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { FrameContexts, SdkError } from '../public';
+
+/**
+ * Namespace to interact with the files specific part of the SDK.
+ *
+ * @private
+ * Hide from docs
+ */
+export namespace teams {
+
+  export enum ChannelType {
+    Regular = 0,
+    Private = 1,
+    Shared = 2
+  }
+
+  export interface TeamsChannelInfo {
+    siteUrl: string;
+    objectId: string;
+    folderRelativeUrl: string;
+    displayName: string;
+    channelType: ChannelType;
+  }
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Get a list of channels belong to a Team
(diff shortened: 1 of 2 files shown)


src/private/teams.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Get a list of channels belong to a Team
+   * @param teamId a team's objectId
+   */
+  export function getTeamsChannels(
```
ydogandjiev, line 33: NIT: Seems like `getTeamChannels` would be more accurate than `getTeamsChannels` since the API only retrieves the channels for one team at a time

src/private/teams.ts
```diff
@@ -0,0 +1,49 @@
...
+    Private = 1,
+    Shared = 2,
+  }
+
+  export interface TeamsChannelInfo {
+    siteUrl: string;
+    objectId: string;
+    folderRelativeUrl: string;
+    displayName: string;
+    channelType: ChannelType;
```
ydogandjiev, line 23: Do you need all of these properties?

src/private/teams.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Get a list of channels belong to a Team
+   * @param teamId a team's objectId
+   */
+  export function getTeamsChannels(
+    teamId: string,
```
ydogandjiev, line 34: If `teamId` is just the ID of the AAD group then it might be better to call this `groupId`; that's what we call it in the SDK context.

ydogandjiev (changes requested):

src/private/teams.ts
yumikohey: Yes, these are all the properties we currently need for our feature :)

src/private/teams.ts
yumikohey: Thanks Yuri. Good to know. I will make the changes.

src/private/teams.ts
```diff
@@ -0,0 +1,49 @@
...
+   * Hide from docs
+   *
+   * Get a list of channels belong to a Team
+   * @param groupId a team's objectId
+   */
+  export function getTeamChannels(
+    groupId: string,
+    callback: (error: SdkError, channels: TeamsChannelInfo[]) => void,
+  ): void {
+    ensureInitialized(FrameContexts.content);
```
sachingupta, line 37: will this be also supported in channel meeting?

sachingupta (approved):

src/private/teams.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Hide from docs
+ */
+export namespace teams {
+  export enum ChannelType {
+    Regular = 0,
+    Private = 1,
+    Shared = 2,
+  }
+
+  export interface TeamsChannelInfo {
```
ydogandjiev, line 18: NIT: Consider renaming this to just `ChannelInfo`; this is the Teams SDK so it's generally not necessary to prefix type/class/interface/method names with "Teams".

ydogandjiev (approved):

ydogandjiev (approved):

PR #596, opened 2021-06-24
Remove unnecessary handler variable

charpercyr:
In a previous PR, a private variable named `focusEnterHandler` was added to internal/handler.ts. This handler does not have a default behavior and can simply be registered when a user calls the register function.

### DIFF
--- a/src/internal/handlers.ts
+++ b/src/internal/handlers.ts
@@ -7,7 +7,6 @@ class HandlersPrivate {
   } = {};
   public static themeChangeHandler: (theme: string) => void;
   public static backButtonPressHandler: () => boolean;
-  public static focusEnterHandler: (navigateForward: boolean) => void;
   public static loadHandler: (context: LoadContext) => void;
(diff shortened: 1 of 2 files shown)


MarcinJuraszek:
@charpercyr is this something we'd like to get in to 2.0-preview branch?

erinha:
> @charpercyr is this something we'd like to get in to 2.0-preview branch?

@MarcinJuraszek Do you mean instead of fixing in master? If it's fixed in master, we'll get it in one of our weekly merges from master to 2.0-preview.

MarcinJuraszek:
Either way, in general, this has been open since Jun and am trying to understand if it's still something we want to fix (in either branch) or it's not worth it.

aeunms:
Checking in on old PRs- it seems like we haven't had any issues with registerFocusEnterHandler and I can't seem to find a ticket on ADO regarding this. Perhaps we can close this PR without merging?

PR #598, opened 2021-06-30
shbeniwa/js sdk add video and image mobile support

ShivangiBeniwal:
**To summarise the changes :**

- Introduced VideoProps & VideoAndImageProps
- As some of the params were common between VideoProps and ImageProps - created a base interface called MediaProps
- VideoProps is not exported as of now to JS developers because of LENS BUG. 
- Introduced videoAndImageMediaAPISupportVersion as 2.0.2 for mobile.
- Added 2 UTs for VideoAndImage mediaType :
  - SelectMedia calls with successful result for mediaType = 3
  - SelectMedia call for mediaType = 3 in mediaAPISupportVersion of platform support fails (below 2.0.2 client support version it will fail)


src/internal/constants.ts
```diff
@@ -6,6 +6,11 @@ export const version = '1.10.0';
  */
 export const defaultSDKVersionForCompatCheck = '1.6.0';
 
+/**
+ * This is the SDK version when selectMedia API - VideoAndImage is supported on mobile.
```
sachingupta, line 10: could you please sync with @SonalJha I think we are moving to plan where we don't need to use API versioning.

sachingupta (changes requested):

src/internal/constants.ts
SonalJha: Video capability is not planned for web and desktop. For mobiles we are keeping the versioning changes as it is for now. As this is an existing API (SelectMedia) a new constant is needed.

src/public/media.ts
```diff
@@ -391,7 +425,10 @@ export namespace media {
       throw new Error('[select Media] Callback cannot be null');
     }
     ensureInitialized(FrameContexts.content, FrameContexts.task);
-    if (!isAPISupportedByPlatform(mediaAPISupportVersion)) {
+    if (
+      !isAPISupportedByPlatform(mediaAPISupportVersion) ||
```
SonalJha, line 429: As video is not being planned for web and desktop, please return appropriate error if user is trying selectMedia Video capability on web and desktop

src/public/media.ts
```diff
@@ -226,22 +228,30 @@ export namespace media {
...
+    videoAndImageProps?: VideoAndImageProps;
+
     /**
      * Additional properties for audio capture flows.
      */
     audioProps?: AudioProps;
   }
 
   /**
-   *  All properties in ImageProps are optional and have default values in the platform
```
SonalJha, line 240: why is this comment removed? It is still applicable I believe

src/public/media.ts
ShivangiBeniwal: Done

src/public/media.ts
ShivangiBeniwal: The comment is still on top of ImageProps (This is MediaProps where you are referring)

src/public/media.ts
```diff
@@ -226,22 +228,30 @@ export namespace media {
...
+   * @private
```
ydogandjiev, line 247: Why are we making this private? Aren't these APIs supposed to be public?

src/public/media.ts
```diff
@@ -277,6 +292,25 @@ export namespace media {
     enableFilter?: boolean;
   }
 
+  /**
+   * @private
```
ydogandjiev, line 296: Same question; why is this being marked as private?

src/public/media.ts
```diff
@@ -390,12 +424,30 @@ export namespace media {
...
       const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
       callback(oldPlatformError, null);
       return;
     }
+
+    if (isMediaCallForVideoAndImageInputs(mediaInputs)) {
+      if (
+        GlobalVars.hostClientType === HostClientType.desktop ||
+        GlobalVars.hostClientType === HostClientType.web ||
+        GlobalVars.hostClientType === HostClientType.rigel
```
ydogandjiev, line 439: This approach for determining API availability is really messy and fragile. @sachingupta, can you please take a look at this when you get a chance and advise on the best path forward? I am really worried that we are sprinkling a bunch of this kind of code throughout our SDK and this is creating an extremely fragmented platform. I don't want us to add any more functionality here until we can sort this out.

ydogandjiev (changes requested):

src/public/media.ts
ShivangiBeniwal: There are some common parameters between VideoProps and ImageProps, so MediaProps is a base interface to them, especially needed for mediaType = 3 (where can select/capture both Video and Image).

- For example, a case where JS developer mentions CAMER SWITCHER as FALSE in Image mode and TRUE in Video mode. 
- This functionality is not supported by our platform, i.e. to have different configurations of camera in different modes.
- Therefore to avoid confusion for JS developers, we have created a base interface encapsulating common parameters that will restrict the developers from specifying two different values. 
- Hence there is no direct usage of MediaProps that's why marked this as private to hide it from public docs. Feel free to refer to the Contract document attached in the channel post.

src/public/media.ts
ShivangiBeniwal: Launching lens camera in only Video mode is not supported as of now. Once the lens enables this support we will expose it to developers for use. 

For now, VideoProps is required for VideoAndImageProps (mediatype = 3) and hence kept it private to hide it from public docs.

ydogandjiev (approved):

src/public/media.ts
ShivangiBeniwal: Resolving this conversation as discussed on the Teams **Client SDK API Review** channel post.

src/internal/constants.ts
```diff
@@ -8,6 +8,11 @@ export const version = '1.10.0';
  */
 export const defaultSDKVersionForCompatCheck = '2.0.1';
 
+/**
+ * This is the SDK version when selectMedia API - VideoAndImage is supported on mobile.
+ */
```
sachingupta, line 13: Shivangi - As part of this PR https://github.com/OfficeDev/microsoft-teams-library-js/pull/591 Sonal has done some changes to move out of sdk version. Now in your PR we are adding bigger version which doesn't align to model where we are going. Could you please sync with Sonal and update your PR without adding new version.

sachingupta (changes requested):

src/public/media.ts
```diff
@@ -390,12 +424,30 @@ export namespace media {
...
+      ) {
```
sachingupta, line 440: can we check if hostClientType is mobile since in future we can have more non mobile host type.

sachingupta (approved):

src/public/media.ts
ShivangiBeniwal: Makes sense, updated the check.

PR #605, opened 2021-07-06
added refreshSiteUrl as a private API

yumikohey:
The following PR introduces the following new private API

`refreshSiteUrl`

This will be a privateAPI. It will be allowed only for 1P apps.


sachingupta:
Have we reviewed this API with extensibility team. can you please setup some time with Yuri and me.

sachingupta (approved):

src/private/teams.ts
```diff
@@ -43,4 +43,21 @@ export namespace teams {
...
+   * provided, the threadId from route params will be used.
+   */
+  export function refreshSiteUrl(callback: (error: SdkError) => void, threadId?: string): void {
+    ensureInitialized();
+
+    if (!callback) {
+      throw new Error('[teams.refreshSiteUrl] Callback cannot be null');
+    }
+
+    sendMessageToParent(threadId);
```
ydogandjiev, line 61: The `sendMessageToParent` API call takes an action name first and then an array of parameters. Have you tested this out?

test/private/teams.spec.ts
```diff
@@ -73,4 +75,11 @@ describe('teams', () => {
       expect(callback).toHaveBeenCalled();
     });
   });
-});
\ No newline at end of file
+
+  describe('refreshSiteUrl', () => {
+    it('should not allow calls before initialization', () => {
```
ydogandjiev, line 80: Please add a test case that validates the non-error scenario

PR #606, opened 2021-07-06
share in teams api

ht4963-ms:
Add shareInTeams API


src/public/sharing.ts
```diff
@@ -77,19 +77,19 @@ export namespace messaging {
...
   /**
    * @private
    * Feature is under development
    *
    * Opens a share dialog for creating a class asignment
-   * @param shareWebContentRequest web content info
+   * @param createAssignmentRequest assignment info
    * @param callback optional callback
    */
   export function createAssignment(
```
ydogandjiev, line 91: Is this API needed for v1 of this feature? Is this covered in Pankaj's PM spec? I personally don't think we should force the apps to choose how they want to share the web content because they often won't know what's appropriate. Apps don't even have a way to know whether Assignments is even available (which is only in EDU tenants).

ydogandjiev (approved):

src/public/sharing.ts
ht4963-ms: I've removed create assignment, deferring it to a future add-on.

src/public/sharing.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    if (!validateContentForSupportedTypes(shareWebContentRequest, callback)) {
+      return;
+    }
+
+    ensureInitialized(
+      FrameContexts.content,
+      FrameContexts.sidePanel,
+      FrameContexts.task,
```
ydogandjiev, line 71: Is task a valid frame context here? Are we going to support opening a share from from within a task module? That can be tricky since it would require opening one dialog from inside of another. Same comment applies to the stage.

ydogandjiev (approved):

src/public/sharing.ts
ht4963-ms: I think it's safe to ensure that these contexts are available just in case and let the Teams framework decide how to handle dialog on top of dialog scenarios.

PR #607, opened 2021-07-07
export teams namespace

yumikohey:
Export `teams` namespace


sachingupta:
Looks like we need to update comment in teams.ts file 
image

sachingupta (approved):

PR #612, opened 2021-07-09
Release 1.11.0 beta.4

JulianDeng:
Request beta release for this branch for testing purpose.

Building my folder in Yuri's test app in following PR. 
https://github.com/ydogandjiev/microsoft-teams-test-tab/pull/47/files

Expecting this release can generate a static script like this that I can import to my test tab index.html.
    <script src="https://statics.teams.microsoft.com/sdk/v1.4.1/js/MicrosoftTeams.min.js"></script>
Expect the v1.4.1 would be replaced for 1.11.0-beta.4, or please advise me what the version string would be like.

Without this script, the microsoftTeams function is simply not working.

### DIFF
--- a/src/public/monetization.ts
+++ b/src/public/monetization.ts
@@ -1,0 +1,37 @@
+import { sendMessageToParent } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { SdkError } from './interfaces';
+
+export namespace monetization {
+  /**
+   * @private
+   * Hide from docs
+   * Data structure to represent a subscription plan.
+   */
+  export interface PlanInfo {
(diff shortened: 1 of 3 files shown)


src/public/monetization.ts
```diff
@@ -0,0 +1,37 @@
+import { sendMessageToParent } from '../internal/communication';
```
sachingupta, line 1: is this file already merged to master ?

sachingupta (changes requested):

src/public/monetization.ts
JulianDeng: Yes. ../internal/communication is an existing file.

PR #613, opened 2021-07-09
add new public api openPurchaseExperience()

JulianDeng:


JulianDeng:
Hi @ydogandjiev , @sachingupta , I have the following questions.

1. When will next sdk be released? Can we include this change in next release (1.11.0)?
2. After changes are pushed, may I request a 1.11.0 beta.4 for testing?
3. Is there any guidelines to edit Microsoft Docs (e.g. https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/microsoftteams.meeting?view=msteams-client-js-latest) for the new api?

src/public/monetization.ts
```diff
@@ -0,0 +1,37 @@
...
+   * Open dialog to start user's purchase experience
+   * @param callback Callback contains 1 parameters, error.
+   * @param planInfo optional parameter. It contains info of the subscription plan pushed to users.
+   * error can either contain an error of type SdkError, incase of an error, or null when get is successful
+   */
+  export function openPurchaseExperience(callback: (error: SdkError | null) => void, planInfo?: PlanInfo): void {
+    if (!callback) {
+      throw new Error('[open purchase experience] Callback cannot be null');
+    }
+    ensureInitialized();
```
ydogandjiev, line 34: Please add a list of frame contexts from which you expect this API to be called (unless it works everywhere). Starting off with just the content frame context is probably a safe option.

ydogandjiev (approved):

src/public/monetization.ts
JulianDeng: This API is expected to be called within only Tab or Chat. 

added "FrameContexts.content" inside the initialization. Please check new changes.

PR #614, opened 2021-07-13
Change Date.now() to performance.now() for timestamp

kpolnitz:
This change removes date.now() and uses performance.now() for the message request


MarcinJuraszek:
This would be a breaking change. `performance.now()` returns time elapsed from a different 'point in time' than `Date.now()` does. It would break whoever expects the value to be consistent with how `Date.now()` works and because of that I don't think we can take it in.

PR #651, opened 2021-09-09
Add subEntityId param to selectAppEntity function

andreipatrascoiu:


src/private/appEntity.ts
```diff
@@ -48,14 +48,17 @@ export namespace appEntity {
...
+   * @param launchContext An object that will be made available to the application being configured
+   *                      through the Teams Context's subEntityId field.
    * @param callback Callback that will be triggered once the app entity information is available.
    *                 The callback takes two arguments: the app entity configuration, if available and
    *                 an optional SdkError in case something happened (i.e. the window was closed)
    */
   export function selectAppEntity(
     threadId: string,
     categories: string[],
+    launchContext: string,
```
ydogandjiev, line 61: Shouldn't we also update the Context interface that apps will receive this launchContext through?

ydogandjiev (changes requested):

src/private/appEntity.ts
andreipatrascoiu: There is no new field added to the Context after the discussions with the MetaOS team. This launchContext param will show up in the subEntityId field.

ydogandjiev (approved):

ydogandjiev (approved):

PR #652, opened 2021-09-09
Add people.openCard API

grhewitt:
Adds an API for opening the [profile card](https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501) inside Teams hosted apps. Teams will open the already existing profile card in the correct position over the top of the app according to the positional information provided in the API call.

### DIFF
--- a/src/internal/peopleUtil.ts
+++ b/src/internal/peopleUtil.ts
@@ -37,17 +37,17 @@ export function validatePeoplePickerInput(peoplePickerInputs: people.PeoplePicke
  * @returns true if the parameters are valid, false otherwise
  */
 export function validateOpenCardRequest(openCardRequest: people.OpenCardRequest): boolean {
-  if (!openCardRequest || !openCardRequest.parameters || !openCardRequest.parameters.personaInfo) {
+  if (!openCardRequest || !openCardRequest.cardParameters || !openCardRequest.cardParameters.personaInfo) {
     return false;
   }
 
-  if (!validateHostAppProvidedPersonaIdentifiers(openCardRequest.parameters.personaInfo.identifiers)) {
+  if (!validateHostAppProvidedPersonaIdentifiers(openCardRequest.cardParameters.personaInfo.identifiers)) {
     return false;
   }
 
   if (
-    !openCardRequest.parameters.openCardTriggerType ||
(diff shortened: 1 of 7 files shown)


test/public/people.spec.ts
```diff
@@ -4,166 +4,224 @@ import { FrameContexts } from '../../src/public/constants';
...
+      _initialize(mobilePlatformMock.mockWindow);
+    });
 
-  afterEach(() => {
-    // Reset the object since it's a singleton
-    if (_uninitialize) {
-      _uninitialize();
-    }
-  });
+    afterEach(() => {
```
mtennoe, line 25: Looks like a lot of whitespace diffs - Intentional or something we can revert to reduce noise?

grhewitt: It was intentional since I added a new describe scope to the suite. I wasn't expecting the diff to look like this though (I have a gut feeling VSO wouldn't show the diff this way). 

I'd recommend looking at the file without the diff, and focusing on the tests under "describe('openCard', ..." as that's what I added.

grhewitt: I just discovered you can toggle hide whitespace in GitHub, the diff looks a lot better with it on.
image

src/internal/peopleUtil.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  if (openCardRequest.cardParameters.behavior && openCardRequest.cardParameters.behavior !== 'object') {
+    return false;
+  }
+
+  return true;
+}
+
+function validateHostAppProvidedPersonaIdentifiers(identifiers: people.IHostAppProvidedPersonaIdentifiers): boolean {
```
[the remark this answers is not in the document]
grhewitt: Yeah the interface was taken directly from LPC since I thought initially there was value in making them equal for us to maintain. Thinking more about it now, I think I'd prefer the interface names to be clearly named for consumption and we can keep it in sync via the property names matching.

mtennoe (approved):

grhewitt:
We're starting again with a fresh PR as this project was paused for a while and now we're targeting the 2.0 SDK patterns.

PR #657, opened 2021-09-20
Release 1.11.0 beta.8

yumikohey:


src/private/teams.ts
```diff
@@ -51,13 +51,17 @@ export namespace teams {
...
+    if (!threadId) {
+      throw new Error('[teams.refreshSiteUrl] threadId cannot be null or empty');
+    }
+
     if (!callback) {
       throw new Error('[teams.refreshSiteUrl] Callback cannot be null');
     }
 
-    sendMessageToParent(threadId);
+    sendMessageToParent('teams.refreshSiteUrl', [threadId], callback);
```
ydogandjiev, line 65: There shouldn't be any actual code changes in the release branch. Please check those into master separately.

ydogandjiev (changes requested):

src/private/teams.ts
yumikohey: This line of code is in master code base :)

sachingupta:
why are we update beta7 from beta 8?

yumikohey:
> why are we update beta7 from beta 8?

As 1.11.0-beta.7 branch was already created

PR #666, opened 2021-09-24
Add scripts for dev releasing

aeunms:
Part of carrying out the addition of automatic dev release publishing. This script is meant to be run only by the pipeline task that will be created next, and it will edit not the package.json in the code base but the temporary file in the build artifacts.


prepDevRelease.js
```diff
@@ -0,0 +1,126 @@
...
+  console.log('FATAL ERROR: package.json path could not be found in the file system.');
+  process.exitCode = EXIT_CODE_FATAL_ERROR;
+  return;
+}
+
+function savePackageJson(packageJson) {
+  fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson));
+}
+
+function getFileVersion(packageJson) {
```
ydogandjiev, line 23: NIT: Add a comment describing what this function does like you did for the ones below

prepDevRelease.js
```diff
@@ -0,0 +1,126 @@
...
+
+  // append the suffix to form a new version
+  const newVersion = newDevPrefix + '-dev.' + newDevSuffix;
+
+  console.log('new version: ' + newVersion);
+
+  // update package.json with the new version
+  packageJson.version = newVersion;
+
+  savePackageJson(packageJson);
```
ydogandjiev, line 125: Might be out of scope for this PR but it might actually make sense to have UTs for all the logic in this file to avoid regressing it accidentally when we need to evolve it in the future.

ydogandjiev (approved):

prepDevRelease.js
aeunms: Do you think it would be good to have this for all of the functions including getPackageJson and savePackageJson? Or do you think those two are clearer on what they're doing there?

prepDevRelease.js
erinha: @aeunms I think we should open a work item to track adding tests separately for this particular file. We can work on scheduling it after our current milestone.

prepDevRelease.js
erinha: If it's not expensive, I will also vote in favor of more comments. :)

erinha (approved):
We need to figure out what to do with the src/internal/constants.ts file. Other than that, this all looks good!

prepDevRelease.js
aeunms: Created and tracked through this work item now: https://office.visualstudio.com/MetaOS/_workitems/edit/5472937

ydogandjiev (approved):

erinha (approved):
Thanks for doing this!

PR #684, opened 2021-10-22
Added third-party cloud storages support

yumikohey:


test/private/teams.spec.ts
```diff
@@ -20,65 +20,65 @@ describe('teams', () => {
   });
 
   afterEach(() => {
-    // Reset the object since it's a singleton
+    // Reset the object since it"s a singleton
```
aeunms, line 23: I think this got replaced during a "replace all."

aeunms (approved):
This looks good to me! I think we're still waiting on a review from Yuri or Sachin though.

src/private/files.ts
```diff
@@ -258,4 +274,63 @@ export namespace files {
...
+
+    sendMessageToParent('files.getExternalProviders', [excludeAddedProviders], callback);
+  }
+
+  /**
+   * @private
+   * Allow 1st party apps to call this function to move files
+   * among SharePoint and third party cloud storages.
+   */
+  export function copyMoveFiles(
```
vivek3003, line 303: nit: `copyMoveFiles` for an API name might be confusing if we chose to make this public.
Can we have `copyFiles` and `moveFiles` as separate functions? Then no need to ask for `isMove` param.

vivek3003 (approved):

ydogandjiev (changes requested):

src/private/files.ts
erinha: Then can we at least rename it to be `copyOrMoveFiles`?

src/private/files.ts
```diff
@@ -21,18 +21,35 @@ export namespace files {
...
+    thumbnails: IWopiThumbnail[];
+  }
+
+  export interface IExternalProvider extends IWopiService {
+    providerType: CloudStorageProviderType;
+    providerCode: CloudStorageProvider;
   }
 
-  /**
-   * @private
```
ydogandjiev, line 27: Did you intentionally remove the `@private` comment? This will cause the interfaces/enums to be added to our public docs here which might not be what we want since these APIs won't work for 3P apps:
https://docs.microsoft.com/en-us/javascript/api/overview/msteams-client?view=msteams-client-js-latest

src/private/files.ts
```diff
@@ -124,6 +141,67 @@ export namespace files {
...
+     */
+    email: string;
+
+    /**
+     * User MRI.
+     */
+    mri: string;
+  }
+
+  export enum SpecialDocumentLibraryType {
```
ydogandjiev, line 167: Add `@private`

src/private/files.ts
```diff
@@ -124,6 +141,67 @@ export namespace files {
...
+    ClassMaterials = 'classMaterials',
+  }
+
+  export enum DocumentLibraryAccessType {
```
ydogandjiev, line 171: Add `@private`

src/private/files.ts
```diff
@@ -258,4 +336,63 @@ export namespace files {
...
+  export function copyMoveFiles(
+    selectedFiles: CloudStorageFolderItem[] | ISharePointFile[],
+    providerCode: CloudStorageProvider,
+    destinationFolder: CloudStorageFolderItem | ISharePointFile,
+    destinationProviderCode: CloudStorageProvider,
+    isMove = false,
+    callback: (error?: SdkError) => void,
+  ): void {
+    ensureInitialized(FrameContexts.content);
+    if (isMove === undefined) {
```
ydogandjiev, line 374: NIT: This check can never be false because you are providing a default value for `isMove` in the method params.

ydogandjiev (approved):
Please make sure all new interfaces and enums that you are exporting are marked as `@private` before checking in your changes.

PR #693, opened 2021-11-08
Rename @privateRemarks TSDoc tag to @hidden

erinha:
Turns out that the @privateRemarks tag is not currently supported by TypeDoc generator used by MS Docs for TypeScript documentation. To suppress comments from publish on docs.microsoft.com, the recommendation is to use @hidden.

### DIFF
--- a/packages/teams-js/src/private/interfaces.ts
+++ b/packages/teams-js/src/private/interfaces.ts
@@ -1,6 +1,6 @@
 import { FileOpenPreference, TeamInformation } from '../public/interfaces';
 /**
- * @privateRemarks
+ * @hidden
  * Hide from docs
  * --------
  * Information about all members in a chat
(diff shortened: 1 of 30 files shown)


packages/teams-js/src/private/interfaces.ts
```diff
@@ -56,108 +56,108 @@ export enum ViewerActionTypes {
 }
 
 /**
- * * @privateRemarks
+ * * @hidden
```
erikadoyle, line 59: There's an extraneous asterisk before @hidden (line 59) that's causing the "Hide from docs" text to generate documentation

erikadoyle (commented):
Just that one issue--otherwise the docs generate great when I run typedoc locally.

packages/teams-js/src/private/interfaces.ts
erinha: Great catch!

aeunms:
Looks like there's some @private tags in pages.ts that didn't get updated to @privateRemarks before, I think those need to be @hidden as well.

aeunms (approved):
LGTM!

PR #697, opened 2021-11-10
Update e2etests pipeline to work with 2.0-preview branch

erinha:
Update the branch triggers for the e2e pipeline.

### DIFF
--- a/azure-pipelines-with-e2etests.yml
+++ b/azure-pipelines-with-e2etests.yml
@@ -3,8 +3,8 @@
 # And then running E2E tests and publishing them.
 
 trigger:
-  - main
-  - develop
+  - 2.0-preview
+  - 'release*'
 
 pool:
   vmImage: 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


azure-pipelines-with-e2etests.yml
```diff
@@ -3,8 +3,8 @@
 # And then running E2E tests and publishing them.
 
 trigger:
-  - main
-  - develop
+  - 2.0-preview
+  - 'release*'
```
erinha, line 7: I think we might want to be more specific than just "release*". I think we might want to use the prefix "release-2*" so that v1 releases aren't triggering these pipelines. Unless I'm overthinking this.

erinha:
The e2e pipeline is failing currently because a change in the hosting repo needs to be merged first. Once that's been merged, I'll rerun the e2e pipeline.

azure-pipelines-with-e2etests.yml
erinha: Since these yml files are only in 2.0-preview branch, there's no need to further filter the trigger here. It doesn't exist in master, so it won't be triggered for v1 releases.

murthys (approved):

aeunms (approved):

PR #705, opened 2021-11-11
Publish bundle analysis artifact on CI to destination branch

anmaur:


azure-pipelines.yml
```diff
@@ -86,8 +86,7 @@ steps:
   - task: PublishBuildArtifacts@1
     condition: and(
       in(variables['Agent.JobStatus'], 'Succeeded', 'SucceededWithIssues'),
-      eq(variables['Build.Reason'], 'PullRequest'),
-      eq(variables['System.PullRequest.TargetBranch'], '2.0-preview'))
+      ne(variables['Build.Reason'], 'PullRequest')
```
erinha, line 89: Do we not need the `eq(variables['System.PullRequest.TargetBranch'], '2.0-preview')`?

anmaur: No, Because that would be conflicting with Build.Reason to be PullRequest. Like why do you want to publish in PR? If you are referring to running comparison, that is in different task where we call compareBundleAnalysis

erinha: Right, thanks for clarifying.

erinha (approved):
Thanks for fixing this!

PR #710, opened 2021-11-12
Lakhveer/meeting room capabilities/state restriction 

lakhveerk:
In MeetingRoomCapability and MeetingRoomState interfaces, fields are restricted to only some string values. However, we are not enforcing this in code (only written in comments). This PR restricts these fileds to take only values they suppose to take. 

I used `type` to limit a variable to only some certain strings. Please let me know if there is a better way to achieve the same purpose (e.g. using eums)

### DIFF
--- a/packages/teams-js/src/private/meetingRoom.ts
+++ b/packages/teams-js/src/private/meetingRoom.ts
@@ -40,60 +40,42 @@ export namespace meetingRoom {
    * @hidden
    * Hide from docs
    * ------
-   * Enum used to indicate meeting room capabilities.
+   * Type of Media control capabilities of a meeting room.
    */
-  export enum Capability {
-    /**
-     * @hidden
-     * Media control capability: toggle mute.
-     */
-    toggleMute = 'toggleMute',
-    /**
-     * @hidden
-     * Media control capability: toggle camera.
-     */
-    toggleCamera = 'toggleCamera',
-    /**
-     * @hidden
-     * Media control capability: toggle captions.
-     */
-    toggleCaptions = 'toggleCaptions',
-    /**
-     * @hidden
-     * Media control capability: volume ajustion.
-     */
-    volume = 'volume',
-    /**
-     * @hidden
-     * Stage layout control capability: show gallery mode.
-     */
-    showVideoGallery = 'showVideoGallery',
-    /**
-     * @hidden
-     * Stage layout control capability: show content mode.
-     */
-    showContent = 'showContent',
-    /**
-     * @hidden
-     * Stage layout control capability: show content + gallery mode.
-     */
-    showVideoGalleryAndContent = 'showVideoGalleryAndContent',
-    /**
-     * @hidden
-     * Stage layout control capability: show laryge gallery mode.
-     */
-    showLargeGallery = 'showLargeGallery',
-    /**
-     * @hidden
-     * Stage layout control capability: show together mode.
-     */
-    showTogether = 'showTogether',
-    /**
-     * @hidden
-     * Meeting control capability: leave meeting.
-     */
-    leaveMeeting = 'leaveMeeting',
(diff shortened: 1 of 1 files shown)


erinha:
<p>Analyzed commit id : 87069358d3579671347d28defdccaae07e82fb47</p><hr><details><summary><b><span style="color: green">⯆</span> @microsoft/bundle-analysis-app:</b> -377 Bytes</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>57.64 KB</td><td>57.27 KB</td><td><span style="color: green">⯆</span> -377 Bytes</td></tr><tr><td>Total Size</td><td>57.64 KB</td><td>57.27 KB</td><td><span style="color: green">⯆</span> -377 Bytes</td></tr></table></details><hr><p>Baseline commit: 45c39a825f731bcecdbde98a65a9e7a4f395b690</p>

MarcinJuraszek:
@ydogandjiev, @sachingupta - this is a breaking change, but I think it's worth making for v2. Can you share your opinion?

@lakhveerk - can you make sure to validate that the "over-the-wire" (aka what's sent through to the Hub/Host as post message) is the same before and after your change?

ydogandjiev:
> @ydogandjiev, @sachingupta - this is a breaking change, but I think it's worth making for v2. Can you share your opinion?
> 
> @lakhveerk - can you make sure to validate that the "over-the-wire" (aka what's sent through to the Hub/Host as post message) is the same before and after your change?

@lakhveerk, @MarcinJuraszek, is this just a breaking change for the apps that use the media APIs or also for the hubs? From the hub perspective does the underlying contract stay the same? If so, then I'm OK with this change. Otherwise, if the underlying format is changing we would need to update the Hub SDK and existing Teams Clients to understand both the old and new formats which is a non-starter.

MarcinJuraszek:
These APIs are not yet part of Hub SDK. Lakhveer is working on adding them there, so we'll add them to match whatever decision we make here.

We do need to make sure the post message structure stays the say. Otherwise it would break it on all existing Teams clients. So at the end of the day, this should only affect the apps using these APIs when they switch to 2.0.

lakhveerk:
> > @ydogandjiev, @sachingupta - this is a breaking change, but I think it's worth making for v2. Can you share your opinion?
> > @lakhveerk - can you make sure to validate that the "over-the-wire" (aka what's sent through to the Hub/Host as post message) is the same before and after your change?
> 
> @lakhveerk, @MarcinJuraszek, is this just a breaking change for the apps that use the media APIs or also for the hubs? From the hub perspective does the underlying contract stay the same? If so, then I'm OK with this change. Otherwise, if the underlying format is changing we would need to update the Hub SDK and existing Teams Clients to understand both the old and new formats which is a non-starter.

@ydogandjiev Marcin helped me check that over the wire is the same data type (strings). So, we are good on that part!

ydogandjiev (approved):

erinha:
<p>Analyzed commit id : fb9db8ad7a6de81c954f01f6245268513362989e</p><hr><details><summary><b><span style="color: green">⯆</span> @microsoft/bundle-analysis-app:</b> -377 Bytes</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>57.64 KB</td><td>57.27 KB</td><td><span style="color: green">⯆</span> -377 Bytes</td></tr><tr><td>Total Size</td><td>57.64 KB</td><td>57.27 KB</td><td><span style="color: green">⯆</span> -377 Bytes</td></tr></table></details><hr><p>Baseline commit: b14911ddde1c1f8437e603a5ec1264c3f8eb5461</p>

PR #718, opened 2021-11-16
Update each @deprecated tag to point to the replaced API

Ella-ly:
Update each @deprecated tag to point to the replaced API and use the @link tag inside the @deprecated comments with full signature.
For example:
` @deprecated As of 2.0.0-beta.1. Please use {@link app.initialize app.initialize(validMessageOrigins?: string[]): Promise\<void\>} instead.`

### DIFF
--- a/packages/teams-js/src/public/appInitialization.ts
+++ b/packages/teams-js/src/public/appInitialization.ts
@@ -1,35 +1,42 @@
 import { app } from './app';
 
 /**
- * @deprecated As of 2.0.0-beta.1. Please use {@link app} namespace instead.
+ * @deprecated
+ * As of 2.0.0-beta.1, please use {@link app} namespace instead.
  */
 export namespace appInitialization {
   /**
-   * @deprecated As of 2.0.0-beta.1. Please use {@link app.Messages} instead.
+   * @deprecated
+   * As of 2.0.0-beta.1, please use {@link app.Messages} instead.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   export import Messages = app.Messages;
   /**
-   * @deprecated As of 2.0.0-beta.1. Please use {@link app.FailedReason} instead.
+   * @deprecated
+   * As of 2.0.0-beta.1, please use {@link app.FailedReason} instead.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   export import FailedReason = app.FailedReason;
   /**
-   * @deprecated As of 2.0.0-beta.1. Please use {@link app.ExpectedFailureReason} instead.
+   * @deprecated
+   * As of 2.0.0-beta.1, please use {@link app.ExpectedFailureReason} instead.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   export import ExpectedFailureReason = app.ExpectedFailureReason;
   /**
-   * @deprecated As of 2.0.0-beta.1. Please use {@link app.IFailedRequest} instead.
+   * @deprecated
+   * As of 2.0.0-beta.1, please use {@link app.IFailedRequest} instead.
    */
   export import IFailedRequest = app.IFailedRequest;
   /**
-   * @deprecated As of 2.0.0-beta.1. Please use {@link app.IExpectedFailureRequest} instead.
+   * @deprecated
+   * As of 2.0.0-beta.1, please use {@link app.IExpectedFailureRequest} instead.
    */
   export import IExpectedFailureRequest = app.IExpectedFailureRequest;
 
   /**
-   * @deprecated As of 2.0.0-beta.1. Please use {@link app.notifyAppLoaded app.notifyAppLoaded(): void} instead.
+   * @deprecated
+   * As of 2.0.0-beta.1, please use {@link app.notifyAppLoaded app.notifyAppLoaded(): void} instead.
    * Notifies the frame that app has loaded and to hide the loading indicator if one is shown.
    */
   export function notifyAppLoaded(): void {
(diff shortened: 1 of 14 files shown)


erinha:
<p>Analyzed commit id : f9fd63718c017b07e3bc29c4bb6b93e2cd05c25f</p><hr><details><summary><b><span style="color: green">■</span> @microsoft/bundle-analysis-app:</b> No change</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>57.27 KB</td><td>57.27 KB</td><td><span style="color: green">■</span> No change</td></tr><tr><td>Total Size</td><td>57.27 KB</td><td>57.27 KB</td><td><span style="color: green">■</span> No change</td></tr></table></details><hr><p>Baseline commit: 1eb3b020461fc39c8808f1f8153097e002cbc4bd</p>

AE-MS:
I'm not in the best position to review the accuracy of the content in the `@deprecated` comments, but I'm wondering if it would be possible to use the [`@link`](https://tsdoc.org/pages/tags/link/) tag inside the `@deprecated` comments to make it easer for developers to navigate to the docs for the new versions? The [`@deprecated`](https://tsdoc.org/pages/tags/deprecated/) doc page gives an example of how it could be used.

erinha:
<p>Analyzed commit id : cb323d68b37d5600a0419d1c61d0e3942625b208</p><hr><details><summary><b><span style="color: green">■</span> @microsoft/bundle-analysis-app:</b> No change</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>57.32 KB</td><td>57.32 KB</td><td><span style="color: green">■</span> No change</td></tr><tr><td>Total Size</td><td>57.32 KB</td><td>57.32 KB</td><td><span style="color: green">■</span> No change</td></tr></table></details><hr><p>Baseline commit: 1165acabda8f7cea16e921a1ccbeccfa080e85e3</p>

packages/teams-js/src/public/appInitialization.ts
```diff
@@ -1,59 +1,59 @@
...
+   * @deprecated As of 2.0.0-beta.1. Please use {@link app.notifyAppLoaded(): void} instead.
    * Notifies the frame that app has loaded and to hide the loading indicator if one is shown.
    */
   export function notifyAppLoaded(): void {
     app.notifyAppLoaded();
   }
 
   /**
-   * @deprecated with TeamsJS v2 upgrades
+   * @deprecated As of 2.0.0-beta.1. Please use {@link app.notifySuccess(): void} instead.
```
MarcinJuraszek, line 40: I'm not sure what the right way is, but this doesn't seem to produce the results you'd expect:

image

MarcinJuraszek: That probably applies to a lot of the comments you have here.

MarcinJuraszek: I think what we actually want is simples:

```
@deprecated As of 2.0.0-beta.1. Please use {@link app.notifySuccess} instead.
```

That shows the correct comment when somebody tries to use the deprecated function.

MarcinJuraszek: So yeah, I believe you'll need to update these not to use the full signature. After that you should validate in VSCode that the warnings show up with the right documentation.

MarcinJuraszek (changes requested):

packages/teams-js/src/public/appInitialization.ts
Ella-ly: > I think what we actually want is simpler:
> 
> ```
> @deprecated As of 2.0.0-beta.1. Please use {@link app.notifySuccess} instead.
> ```
> 
> That shows the correct comment when somebody tries to use the deprecated function.

For some APIs, for example: `viewImages(uriList: ImageUri[]): Promise<void>`,  the way to make this API call is the same as the old version, but it cuts one param and returns a Promise now. I thought using the full signature would be more accurate and avoid confusion to developers in this situation. Do you think we should apply to other APIs for consistency?
Now I use:
`@deprecated As of 2.0.0-beta.1. Please use {@link media.viewImages media.viewImages(uriList: ImageUri[]): Promise<void>} instead.`
viewImage

erinha:
<p>Analyzed commit id : 33eb9b375bbea2f46e417c6c7c329fffd303ba80</p><hr><details><summary><b><span style="color: green">⯆</span> @microsoft/bundle-analysis-app:</b> -6 Bytes</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>57.34 KB</td><td>57.34 KB</td><td><span style="color: green">⯆</span> -6 Bytes</td></tr><tr><td>Total Size</td><td>57.34 KB</td><td>57.34 KB</td><td><span style="color: green">⯆</span> -6 Bytes</td></tr></table></details><hr><p>Baseline commit: 9835bb8c2b9e89de9b3eb4ef4b11985e945916c3</p>

erikadoyle (commented):
Looks like the MS Docs build system expands the @deprecated to a full sentence ("This API is now deprecated"), and any text after that is parsed as a new paragraph ([example](https://review.docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/@microsoft.teams-js.public.appinitialization.appinitialization?view=msteams-client-js-beta&branch=AddBeta2_2#notifyAppLoaded__)).

I think it might render most cleanly if we could:
- Declare @deprecated tag without any trailing text on the same line
- Rework our custom deprecation text on a separate line as a single sentence. For example: *As of 2.0.0-beta.1, please use {@link app.ExpectedFailureReason} instead.*

packages/teams-js/src/public/appInitialization.ts
```diff
@@ -1,59 +1,59 @@
 import { app } from './app';
 
 /**
- * @deprecated with TeamsJS v2 upgrades
+ * @deprecated As of 2.0.0-beta.1. Please use {@link app} namespace instead.
```
erinha, line 4: Can you please update these as per Erika's suggestion: e.g.,

`As of 2.0.0-beta.1, please use {@link app} namespace instead.`

Note the "... , please ..." instead of two separate sentences.

erinha (changes requested):
Please update tags to be one sentence separated with "," instead of two sentences. Thanks!

erinha (approved):
LGTM!

PR #720, opened 2021-11-16
Trharris/readmefix

TrevorJoelHarris:
we don't mention yarn install in the readme file, which is super minor but hey look at me I'm helping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,9 @@ This JavaScript library is part of the [Microsoft Teams developer platform](http
 ## Getting Started
 
 1. Clone the repo
-2. `yarn build` from repo root
-3. to run Unit test `yarn test`
+2. `yarn install` from repo root
+3. `yarn build` from repo root
+4. to run Unit test `yarn test`
 
   TIP: whenever building or testing the Teams client SDK, you can run `yarn build` or `yarn test` from the packages/teams-js directory.
 


MarcinJuraszek (approved):

erinha (approved):

README.md
```diff
@@ -10,8 +10,9 @@ This JavaScript library is part of the [Microsoft Teams developer platform](http
 ## Getting Started
 
 1. Clone the repo
-2. `yarn build` from repo root
-3. to run Unit test `yarn test`
+2. `yarn install` from repo root
```
erinha, line 13: Can you please make this change in our other packages/teams-js/README.md too?

erinha (changes requested):
Whoops, forgot to mention we have two README files

erinha (approved):
Double-whoops. That instruction exists in the other README file.

README.md
TrevorJoelHarris: The change is already in that README file, but it was missing in this one. Would probably make sense to only have one README file and then copy it around during the build. Not sure how you all track issues but I guess this would be a good one to throw on the pile?

TrevorJoelHarris:
OK, this coming from a fork is still causing problems so abandoning again and trying to actually do things correctly

PR #732, opened 2021-11-18
Stop generating blank documentation files with weird titles

TrevorJoelHarris:
Our MicrosoftTeams.d.ts file was getting generated with module declarations around each item (and those modules had names like @microsoft/teams-js...). These modules weren't really doing anything, and after talking with coshaw added this line to our webpack.config.js file:

image

This stops generating the outer module declarations which means that generating documentation using MicrosoftTeams.d.ts works correctly again because the typedoc decorations are at the correct level again.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,6 +16,13 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "Run current file",
+      "program": "${file}",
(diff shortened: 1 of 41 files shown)


TrevorJoelHarris:
Going to branch off of this instead of the main 2.0 branch to bring in fewer changes

PR #733, opened 2021-11-18
Documentation fixes for release

TrevorJoelHarris:
Update webpack-dev-server types to match webpack 5 versions and stop generating module wrappers in MicrosoftTeams.d.ts

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1974,7 +1974,7 @@
     write-file-atomic "^2.3.0"
 
 "@microsoft/teams-js@file:packages/teams-js":
-  version "2.0.0-beta.0"
+  version "2.0.0-beta.1"
 
 "@mixer/webpack-bundle-compare@^0.1.0":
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -1974,7 +1974,7 @@
     write-file-atomic "^2.3.0"
 
 "@microsoft/teams-js@file:packages/teams-js":
-  version "2.0.0-beta.0"
+  version "2.0.0-beta.1"
```
aeunms, line 1977: This is interesting. Should we be changing the package.json to be beta.2 and targeting that instead? Or is it as designed to trail one beta version behind? (I feel like this is also more of a question for folks who's been here longer)

erinha (approved):
Looks good, thanks!

yarn.lock
erinha: I'll be bumping the version in a separate change for this next release.

PR #735, opened 2021-11-19
Remove unnecessary deprecated tags from index

TrevorJoelHarris:
These tags caused a warning during doc generation because in the MicrosoftTeams.d.ts file the things they were referencing were removed, which meant we just had 5 deprecated comments in a row at the top of the file. I checked, and each function/module marked as deprecated in index are ALSO marked as deprecated in their actual files, and removing the deprecated tags from index appears to have no impact on the actual documentation (other than fixing the warning)

If the @deprecated tags in index are doing something else that I don't know about let me know!

### DIFF
--- a/packages/teams-js/src/public/index.ts
+++ b/packages/teams-js/src/public/index.ts
@@ -41,13 +41,7 @@ export { people } from './people';
 export { video } from './video';
 export { sharing } from './sharing';
 export { call } from './call';
-/**
- * @deprecated with TeamsJS v2 upgrades
- */
 export { appInitialization } from './appInitialization';
-/**
- * @deprecated with TeamsJS v2 upgrades
- */
 export {
   enablePrintCapability,
   executeDeepLink,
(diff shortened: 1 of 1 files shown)


erinha:
@ndangudubiyyam I believe you added these tags originally. Could you please take a look here and double-check whether there is some other reason they are required? Thanks!

erinha:
<p>Analyzed commit id : be86ff7dc20153eb1af9f45780e3e4d015333bc5</p><hr><details><summary><b><span style="color: green">⯆</span> @microsoft/bundle-analysis-app:</b> -6 Bytes</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>57.34 KB</td><td>57.34 KB</td><td><span style="color: green">⯆</span> -6 Bytes</td></tr><tr><td>Total Size</td><td>57.34 KB</td><td>57.34 KB</td><td><span style="color: green">⯆</span> -6 Bytes</td></tr></table></details><hr><p>Baseline commit: d4fca89a93d4ee7dc26ada0b2613893953712853</p>

ndangudubiyyam:
> @ndangudubiyyam I believe you added these tags originally. Could you please take a look here and double-check whether there is some other reason they are required? Thanks!

I added the tag at all the places referencing the deprecated classes. If removing them is not affecting the documentation then go ahead and remove them.

MarcinJuraszek (approved):

erinha (approved):

PR #750, opened 2021-11-26
Support for outputFileFormats in imageProps in selectMedia API

adityakamath90:
**Introduced selectMedia to support imageOutputFormats for Image Media Type**

* This feature is natively supported by both iOS and Android platforms as is part of the Lens flow
* Requirement of the feature for MVP is 
    * Let the developer specify the output format (Image and PDF). Based on the format in JS SDK, mobile platform to convert the Image via Lens SDK to the respective output format
    * If no imageOutputFormat specified then mobile platform would default to Image format
    * Throwing JS error 4000 when output format is not recognised in mobile platform
* Changes made in JS SDK MEDIA FILE (MVP)  :
    * Added field imageOutputFormats in ImageProps
    * Introduced enum ImageOutputFormat consisting of option for Image and PDF.. 
    * Added check for older client version and clientSupportedSdkVersion updated to 2.4.0 
* For supporting future scalability of this feature in other Media Types like Video, Audio and ImageAndVideo following steps would be taken (Non MVP) 
    *  For MediaType Video -  A separate enum for to support video output formats can similarly be introduced to VideoProps
    *  For MediaType ImageAndVideo - ImageOutputFormats and VideoOutputFormats would be included in ImageAndVideo props 
    *  For MediaType Audio - AudioOutputFormats would be included in AudioProps

Note - This is a mobile feature only and won't be implemented on desktop. 

 [SDK Document](https://microsoftapc-my.sharepoint.com/:w:/g/personal/adityakamath_microsoft_com/ERjYz2qXFGFLuYy1Kbkp47wBsX54AUqv26ZOaahAJoXagw?e=ZBKvob)
[Changes Required](https://microsoftapc-my.sharepoint.com/:p:/g/personal/adityakamath_microsoft_com/EfB6-kqxARdJsbUpYaaSX-YBrmbTF18d1WF6tXRjPEpM1g?e=0CMrQw)
[Design Document](https://microsoftapc-my.sharepoint.com/:w:/g/personal/adityakamath_microsoft_com/EUBHOx5NeWdEuEm9ojKV92EBGI0ZQdu3uP0m-p0EeYnfgQ?e=nXcjae)
 [Feature Spec](https://microsoft.sharepoint.com/:w:/t/skypespacesteamnew/ERE3HaitVRxJlWEcSnU-bdgBH4lp0Q3eGZjwJLdtWe04kA?e=7jQgKc)


sachingupta (approved):

MarcinJuraszek:
Are these APIs public and released now? If so, isn't this a breaking change? E.g. the developer specifies the output types using the new prop, but the app is running in an old version of Teams that does not support these (e.g. some old version of the mobile client). In this case, the experience would be broken and the data returned might not match the output type requested.

adityakamath90:
@MarcinJuraszek 

That's a fair comment on the behavior that people will get if they don't update. However, that does not make it a breaking change. Every feature we deliver requires users to update their mobile clients

Breaking change would mean that users with older builds will not be able to get any output at all when they try and use the selectMedia

Assignments, who contributes to 95% of our Select Media usage is aware of this possible limitation. Since we get to a 80% update in about 30 days timeframe, this is acceptable limitation.

If we keep this as a blocking limitation, then we would not be able to ship any client side changes w.r.t. JS SDK functions after a V1.

src/public/media.ts
```diff
@@ -22,6 +22,7 @@ import {
   scanBarCodeAPIMobileSupportVersion,
   nonFullScreenVideoModeAPISupportVersion,
 } from '../internal/constants';
+import { userOriginUrlValidationRegExp } from '@microsoft/teams-js';
```
ydogandjiev, line 25: Why are you importing this but then not using it anywhere?

ydogandjiev (approved):
Please address my comment before committing the change.

src/public/media.ts
adityakamath90: @ydogandjiev This wasn't intentional, Will remove this import..

adityakamath90:
> Are these APIs public and released now? If so, isn't this a breaking change? E.g. the developer specifies the output types using the new prop, but the app is running in an old version of Teams that does not support these (e.g. some old version of the mobile client). In this case, the experience would be broken and the data returned might not match the output type requested.
> 
> @sachingupta this is a pattern I've seen multiple times now, where the change is technically a breaking change but it's being approved and merged anyway. Is that something we should address more broadly? Have some guidance when it is or isn't allowed to do these kinds of changes?

@MarcinJuraszek Updated the PR with the check for older client platforn

src/internal/mediaUtil.ts
```diff
@@ -85,6 +91,16 @@ export function validateSelectMediaInputs(mediaInputs: media.MediaInputs): boole
   return true;
 }
 
+/**
+ * Returns true if the mediaInput params are called for mediatype Image and contains Image outputs formats false otherwise
```
ShivangiBeniwal, line 95: nit: formats `and` false otherwise OR formats`,` false otherwise

adityakamath90: @ShivangiBeniwal Review comments addressed

ShivangiBeniwal (approved):

PR #759, opened 2021-12-03
Changes for audio-powered avatar

shilcare:


src/public/video.ts
```diff
@@ -78,7 +109,7 @@ export namespace video {
   /**
    *  Video effect change call back function definition
    */
-  type VideoEffectCallBack = (effectId: string | undefined) => void;
+  type VideoEffectCallBack = (effect: string | undefined, effectParam?: string) => void;
```
zhml530, line 112: Is there any reason to change 'effectId' to 'effect? For video app, the parameter's value is still effect Id, if changed to effect may need 3rd party developer update its code when using new sdk.

src/public/video.ts
```diff
@@ -99,13 +131,15 @@ export namespace video {
    * in-meeting scenario, we will call videoEffectCallback when apply button clicked.
    * @param effectChangeType the effect change type.
    * @param effectId Newly selected effect id.
+   * @param effectParam param for the selected effect.
    */
   export function notifySelectedVideoEffectChanged(
     effectChangeType: EffectChangeType,
     effectId: string | undefined,
+    effectParam: string | undefined,
```
zhml530, line 139: Shall we make the effectParam to be optional? Since for videoApp not used this parameter.

src/public/video.ts
```diff
@@ -114,6 +148,15 @@ export namespace video {
   export function registerForVideoEffect(callback: VideoEffectCallBack): void {
     ensureInitialized(FrameContexts.sidePanel);
     registerHandler('video.effectParameterChange', callback);
+    sendMessageToParent('video.registerForVideoEffect');
```
ydogandjiev, line 151: Why do you need to send two messages to Teams here? The registerHandler call is already sending a message that you can handle on the Teams side.

src/public/video.ts
```diff
@@ -66,6 +78,25 @@ export namespace video {
...
+    /**
+     * display name
+     */
+    name: string;
+    /**
+     * effect type defined by app
+     */
+    type: string;
+    /**
+     * URL of the thumbnail
```
ydogandjiev, line 95: How are these URLs used on the Teams side? Are the protected by any kind of authentication?

src/public/video.ts
```diff
@@ -50,6 +54,14 @@ export namespace video {
      * video format
      */
     format: VideoFrameFormat;
+    /**
+     * Synthesizing video frame based on raw frame or not
+     */
+    requireCameraStream?: boolean;
+    /**
+     * onnx model for audio inference
```
ydogandjiev, line 62: Please be consistent with capitalization and also use language that is easier to understand. What is the onnx model? Is this a publicly defined concept?

src/public/video.ts
```diff
@@ -66,6 +78,25 @@ export namespace video {
     EffectDisabled,
   }
 
+  export interface PersonalizedEffect {
+    /**
+     * personalized effect id
```
ydogandjiev, line 83: NIT: Capitalize

ydogandjiev: Same below; please remember that these comments become part of our public docs once you commit your changes

ydogandjiev (changes requested):

src/public/video.ts
shilcare: Changed to optional

src/public/video.ts
shilcare: Updated

src/public/video.ts
shilcare: Fixed

src/public/video.ts
zhml530: the registerHandler has called the sendMessageToParent(), we don't need call SendMessageToParent() again. Just change the handler of 'video.registerForVideoEffect' to 'video.effectParameterChange'

TrevorJoelHarris:
This PR is stale and will be closed, please feel free to re-open it if your intention is still to check into main.

PR #765, opened 2021-12-08
Reject Initialization promise when no parent or nativeInterface found

MarcinJuraszek:
This modifies the initialization logic to fail early in no parent window, no opener window and no nativeInterface are available to send the `initialize` message to.

Based on my conversation with @ydogandjiev, there was some weird bug with `authentication.authenticate` flow on IE11. As we don't really know if we want to continue officially support that browser, plus the scenario is quite niche I'm potentially breaking it for now. If we decide to explicitly support IE11 before GA we should consider reverting this change.

Validation performed:
- `yarn test` for UTs
- `yarn test:e2e` in the host sdk repo for scenario tests
- manually tested the `authentication.authenticate` scenario using Orange


erinha:

MarcinJuraszek:
@ydogandjiev I'm a bit nervous about this change and would appreciate your review!

erinha:

packages/teams-js/src/internal/communication.ts
```diff
@@ -60,10 +60,14 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
   }
 
   if (!Communication.parentWindow) {
-    GlobalVars.isFramelessWindow = true;
-    /* eslint-disable  @typescript-eslint/ban-ts-comment */
-    // @ts-ignore: window as ExtendedWindow
-    (window as ExtendedWindow).onNativeMessage = handleParentMessage;
+    const extendedWindow = (Communication.currentWindow as unknown) as ExtendedWindow;
```
anmaur, line 63: I am not sure if I am stating it right but: Wouldn't this mean that for a user loading/testing an app in a browser which doesnot have any parent or NativeInterface could break their JS unless they use try catch?

anmaur, line 63: Can be directly casted to ExtendedWindow

MarcinJuraszek: It can't. TypeScript complains. Which is why there was a `@ts-ignore` comment here before. I think going through `as unknown` and not trying to disable type checks is better, which is why I changed this. The behavior should be the same.

MarcinJuraszek: Correct. Feels like a valid breaking change to take in v2.0. But I'm curious if that's just my opinion and other people see it differently.

erinha:

ydogandjiev (approved):

PR #769, opened 2021-12-08
oauth external browser

lnnl:
add the isExternal parameter
replace oauthRedirectMethod with 'web' in the url for the web flow

### DIFF
--- a/src/public/authentication.ts
+++ b/src/public/authentication.ts
@@ -60,7 +60,7 @@ export namespace authentication {
       // Ask the parent window to open an authentication window with the parameters provided by the caller.
       sendMessageToParent(
         'authentication.authenticate',
-        [link.href, authenticateParams.width, authenticateParams.height],
+        [link.href, authenticateParams.width, authenticateParams.height, authenticateParams.isExternal],
         (success: boolean, response: string) => {
           if (success) {
             authenticateParams.successCallback(response);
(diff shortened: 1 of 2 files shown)


src/public/authentication.ts
```diff
@@ -326,6 +326,10 @@ export namespace authentication {
      * The preferred height for the pop-up. This value can be ignored if outside the acceptable bounds.
      */
     height?: number;
+    /**
+     * The flag which indicates whether the auth page should be opened in an external browser.
```
ydogandjiev, line 330: Consider adding a note here that this does flag has no effect in web browsers.

src/public/authentication.ts
```diff
@@ -138,7 +138,7 @@ export namespace authentication {
     height = Math.min(height, Communication.currentWindow.outerHeight - 200);
     // Convert any relative URLs into absolute URLs before sending them over to the parent window
     const link = document.createElement('a');
-    link.href = authParams.url;
+    link.href = authParams.url.replace('{oauthRedirectMethod}', 'web');
```
ydogandjiev, line 141: Should we just do this when `isExternal` is set to true?

ydogandjiev (approved):

src/public/authentication.ts
lnnl: has no effect when isExternal = false

ht4963-ms (approved):

PR #771, opened 2021-12-08
Ramourt/readme generator

ramourt:
Adding a script that will update the existing Readme.md file in the packages\teams-js folder and updates it with the integrity hash value provided from the MicrosoftTeams-manifest.json generated from webpack builds.


erinha:

packages/teams-js/webpack.config.js
```diff
@@ -71,12 +71,12 @@ module.exports = {
     new WebpackAssetsManifest({
       integrity: true,
       integrityHashes: ['sha384'],
-      output: 'MicrosftTeams-manifest.json',
+      output: 'MicrosoftTeams-manifest.json',
```
erinha, line 74: Nice catch!

azure-pipelines.yml
```diff
@@ -68,6 +68,16 @@ steps:
       Arguments: 'bundleAnalyze:collect'
     displayName: 'Run bundle analysis and collect'
 
+  - script: 'node prepNewReadme.js'
+    workingDirectory: '$(System.DefaultWorkingDirectory)\packages\teams-js'
+    condition: and(
+      eq(variables['Build.SourceBranch'], 'refs/heads/2.0-preview'),
```
erinha, line 74: I think we only want to update the README.md when the integrity hash is being generated for the MicrosoftTeams.min.js file that is going to be pushed to prod CDN. We always want the HTML script tag that has a prod CDN URL to include the integrity hash that is valid for it. 

If I understand the trigger correctly, I think this will execute on every CI build to 2.0-preview. 

This pipeline still reflects all the steps that are done to publish as we haven't yet onboarded the release pipeline that will take care of those steps.

The matching condition that I think we want here (given that it's still reflecting our current/soon-to-be-old release process) is:

`startsWith(variables['Build.SourceBranch'], 'refs/tags/v'),`

Does this make sense?

packages/teams-js/.npmignore
```diff
@@ -9,5 +9,6 @@ coverage/**
 CONTRIBUTING.md
 gulpfile.js
 karma.conf.js
+prepNewReadme.js
```
erinha, line 12: Sorry if this is a really naive question, but what is the effect of adding this here? What happens if you don't?

erinha (changes requested):
Appreciate the quick turnaround here! Left a comment about the trigger we're using to run the prepNewReadme.js script.

azure-pipelines.yml
ramourt: Yes, this makes perfect sense. Since we publish next-dev versions after merges (2.0.0-beta.x-dev.x), do we care about those dev versions Readmes on npm being correct?

If so, I think we might want both triggers, like the triggers set for the publishing to npm step?
```
or(startsWith(variables['Build.SourceBranch'], 'refs/tags/v'), eq(variables['Build.SourceBranch'], 'refs/heads/2.0-preview')),
```

packages/teams-js/.npmignore
ramourt: Not naive at all. All matching names set in this file will be skipped when npm publish is called on this package.
Once a dev adds teams-js package to their repository (with npm/yarn add teams-js), the files they will see under that node module will exclude those set in the npmignore file.

azure-pipelines.yml
erinha: The README file on @next-dev packages published to npm will still just refer to prod CDN, so there's no need to update the integrity check. I think. Unless we want to update those with the SDF CDN URL rather than the prod CDN URL. I don't think we've considered modifying the README to point to SDF CDN for those @next-dev releases. We should probably discuss this further offline.

packages/teams-js/.npmignore
erinha: Thanks, that's super helpful!

azure-pipelines.yml
ramourt: Fixed

erinha:

erinha:

erinha (approved):
Looks great, thanks!

erinha:

PR #774, opened 2021-12-11
Update v2 with getGenericOnCompleteHandler when the user didn't provide a callback.

ndangudubiyyam:
There are few places in v1(master) where the return of the method getGenericOnCompleteHandler used when the user doesn't provide a callback.
appWindow.ts
 - ChildAppWindow.postMessage
 - ParentAppWindow.postMessage
navigation.ts (deprecated)
- navigateToTab
- navigateCrossDomain
- navigateBack
publicAPIs.ts (deprecated)
- executeDeepLink
settings.ts (deprecated)
- setSettings
I have updated the deprecated code with the generic handler. 
For appWindow.ts functions, they throw an error when the promise fails, so no need to update it.


erinha:

MarcinJuraszek (approved):

PR #784, opened 2021-12-16
Update package.json

lnnl:


ht4963-ms (approved):

ydogandjiev (changes requested):
These PRs don't get checked in. They are only used to deploy beta versions of the SDK.

PR #785, opened 2021-12-16
Update constants.ts

lnnl:


src/internal/constants.ts
```diff
@@ -1,6 +1,6 @@
 import { generateRegExpFromUrls } from './utils';
 
-export const version = '1.11.0';
+export const version = '1.12.0-beta.2';
```
ht4963-ms, line 3: Why the jump from 1.11.0 to 1.12.0-beta.2? Did we skip some versions?

PR #797, opened 2021-12-27
Added DisplayMode Property to MenuItem as part of setNavBarMenu API

meethjain:
Introduced Display Mode Enum & corresponding property to `MenuItem` part of `setNavBarMenu` 

- This feature is required by both iOS and Android platforms as part of **setNavBarMenu** API
- Adding capability for Web Apps to define how a menu items should appear in the NavBar, either on the NavBar or inside the Overflow Menu.​
- JS SDK Changes
  - Added DisplayMode enum, consisting of the following options.
    - ifRoom – Only place this item in the NavBar if there's room for it. If there's no room, item is shown in the overflow menu.​
    - never - Never place this item in the NavBar. The item would always be shown in NavBar's overflow menu. ​
  - Adding new property displayMode as part of MenuItem.​
  - If null, the behavior would default to ifRoom.​
- API is modelled on similar lines as showAsAction for TitleBar on Android platform

@reenudeswal @ydogandjiev @aeunms

### DIFF
--- a/src/private/menus.ts
+++ b/src/private/menus.ts
@@ -26,6 +26,23 @@ export namespace menus {
      */
     contentDescription?: string;
   }
+
+  /**
+   * Defines how a menu item should appear in the NavBar.
+   */
+  export enum DisplayMode {
+    /**
+     * Only place this item in the NavBar if there's room for it.
+     * If there's no room, item is shown in the overflow menu.
+     */
+    ifRoom = 0,
+    /**
+     * Never place this item in the NavBar.
+     * The item would always be shown in NavBar's overflow menu.
+     */
+    never = 1,
+  }
+
   /**
    * Represents information about menu item for Action Menu and Navigation Bar Menu.
    */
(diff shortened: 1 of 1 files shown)


src/private/menus.ts
```diff
@@ -26,6 +26,23 @@ export namespace menus {
      */
     contentDescription?: string;
   }
+
+  /**
+   * Defines how a menu item should appear in the NavBar.
+   */
+  export enum DisplayMode {
```
ht4963-ms, line 33: The naming here is a bit confusing. DisplayMode.never sounds like it should not be displayed at all. The fact that overflow is part of the explanation but not in the name is, I think, part of the confusion. How about something like:
default = 0, // show unless there's no room, in which case push it into overflow
overflowOnly = 1, // show in overflow alway

meethjain: Agree, however `never` to `overflowOnly`
However, instead of `default`, I feel it should be still `ifRoom` to indicate that intended behaviour.

meethjain: @ht4963-ms Let me know what you think.

ht4963-ms: I think ifRoom **is** the default behavior because you will never force something to be displayed if there's no room, so this distinction is unnecessary. By specifying "ifRoom", it begs the question: does it not get displayed at all when there's no room, then you'll have add the additional explanation that it does get displayed, but only in the overflow.

aeunms (approved):
Thanks for the changes! I believe these will not conflict with the vision for v2. I would wait to hear from the v1 team for their go-ahead still.

ydogandjiev (approved):

PR #798, opened 2021-12-27
Move getConfig to pages capability

MarcinJuraszek:


erinha:
<p>Analyzed commit id : 96ba4c5dfdf70ee355c0fc060c9d0bc227f485cf</p><hr><details><summary><b><span style="color: green">⯆</span> @microsoft/bundle-analysis-app:</b> -14 Bytes</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>59.43 KB</td><td>59.42 KB</td><td><span style="color: green">⯆</span> -14 Bytes</td></tr><tr><td>Total Size</td><td>59.43 KB</td><td>59.42 KB</td><td><span style="color: green">⯆</span> -14 Bytes</td></tr></table></details><hr><p>Baseline commit: bd112e44bd54dff4b1b231e9709ce6490a1b5076</p>

packages/teams-js/src/public/pages.ts
```diff
@@ -216,27 +251,16 @@ export namespace pages {
...
-
     /**
      * Sets the config for the current instance.
      * This is an asynchronous operation; calls to getConfig are not guaranteed to reflect the changed state.
-     * @param Config The desired config for this instance.
+     * @param instanceConfig The desired config for this instance.
      * @returns Promise that resolves when the operation has completed.
      */
-    export function setConfig(instanceSettings: Config): Promise<void> {
+    export function setConfig(instanceConfig: InstanceConfig): Promise<void> {
```
TrevorJoelHarris, line 260: >setConfig [](http://example.com/codeflow?start=20&length=9)

Why does this one stay here when getConfig moved into pages?

TrevorJoelHarris (changes requested):
:clock1:

MarcinJuraszek:
Sharing some context I shared internally, for future reference:

> Having a capability (pages.config in this example) with optional methods makes the isSupported check pretty much useless. So unless the Host SDK can provide the default implementation for them too (which I don't think is true), I don't think that's a good idea. In addition to that, getConfig (based on my research on how it works in Teams) is available to all the different flavors of Tab apps (personal, configurable, task modules, etc). So it makes sense for that API to be in a top level pages capability (Tabs === Pages). Whereas the other methods in pages.config only make sense in settings or remove contexts.
> 
> Now, that last part is a bit of an assumption for setConfig as the code uses ensureInitialized(FrameContexts.content, FrameContexts.settings, FrameContexts.sidePanel) there. I'm suspecting this is similar to how submitTask used to allow for all sorts of FrameContexts for no reason and only recently was updated to only work in task context (Restrict updateTask and submitTask to FrameContext.task by MarcinJuraszek · Pull Request #658 · OfficeDev/microsoft-teams-library-js (github.com)). We should confirmthis assumption with Yuri.
> 
> If we were to keep getConfig in pages.config with Host SDK providing the default implementation, without doing anything else, it would be quite weird because pages.config.isSupported() would return false, but pages.config.getConfig() would be OK to call.

@TrevorJoelHarris let me know if there's still something you want me to change in this PR.

packages/teams-js/src/public/pages.ts
MarcinJuraszek: `setConfig` is only (afaik) applicable in `FrameContext.settings` wheras `getConfig` is available in all the FrameContexts. If we were to move `setConfig` to top-level `pages` it would require the developer to do both `pages.isSupported` and `context.pages.FrameContext === FrameContexts.settings` (or whatever the property on context is that has frameContext on it). I feel like leaving it in `pages.config` (which after moving `getConfig` out of only has methods available in `frameContexts.settings` and `frameContexts.remove` makes it easier for the developer to reason about what is and isn't available.

MarcinJuraszek: I would expect us to have to have a deeper conversation on how `isSupported` on capabilities play with `frameContext` at some point, but with how things work right now I think leaving `setConfig` in `pages.config` gives a better dev experience to the app developer.

TrevorJoelHarris (approved):
:shipit:

PR #800, opened 2021-12-27
Add trace logging to TeamsJs

MarcinJuraszek:
This change is a first stab at adding trace logging to TeamsJs v2. It integrates [debug ](https://www.npmjs.com/package/debug) package and uses it to add trace logging around app->parent communication: sending, queuing, receiving and handling messages between the application and the parent host.

The logging is disabled by default and can be turned on by setting a logging pattern in local storage:

```
localStorage.debug = "teamsJs:*"
```

Sample output for a test app loaded successfully, followed by `getContext` call, `registerThemeChanged` and host notifying the application about the theme changing:

image

In subsequent changes, assuming there's an agreement around this change, I'll add similar trace logging around app->child communication and other important aspects of the SDK.

Open questions:

- [ ] is there any sort of attribution we need to do for debug package usage?
- [ ] where should I add documentation around this new functionality?
- [ ] how do I let the support folks know that this will be a thing, so they can ask the customers reporting issues to enable trace logging and share the logs, if appropriate?

### DIFF
--- a/packages/teams-js/package.json
+++ b/packages/teams-js/package.json
@@ -18,6 +18,12 @@
     "prettier": "prettier --write '**/*.{ts,js,css,html}'",
     "test": "jest"
   },
+  "dependencies": {
+    "debug": "4.3.3"
+  },
+  "devDependencies": {
+    "@types/debug": "4.1.7"
+  },
   "license": "MIT",
   "files": [
     "dist/**",
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -7,7 +7,10 @@ import { version } from './constants';
 import { GlobalVars } from './globalVars';
 import { callHandler } from './handlers';
 import { DOMMessageEvent, ExtendedWindow, MessageRequest, MessageResponse } from './interfaces';
+import { getLogger } from './telemetry';
 import { validateOrigin } from './utils';
+
+const communicationLogger = getLogger('communication');
 
 /**@internal */
 export class Communication {
@@ -180,12 +183,20 @@ export function sendMessageToParent(actionName: string, argsOrCallback?: any[] |
   }
 }
 
+const sendMessageToParentHelperLogger = communicationLogger.extend('sendMessageToParentHelper');
+
 /**@internal */
 function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequest {
+  const logger = sendMessageToParentHelperLogger;
+
   const targetWindow = Communication.parentWindow;
   const request = createMessageRequest(actionName, args);
+
+  logger('Message %i information: %o', request.id, { actionName, args });
+
   if (GlobalVars.isFramelessWindow) {
     if (Communication.currentWindow && Communication.currentWindow.nativeInterface) {
+      logger('Sending message %i to parent via framelessPostMessage interface', request.id);
       (Communication.currentWindow as ExtendedWindow).nativeInterface.framelessPostMessage(JSON.stringify(request));
     }
   } else {
@@ -194,8 +205,10 @@ function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequ
     // If the target window isn't closed and we already know its origin, send the message right away; otherwise,
     // queue the message and send it after the origin is established
     if (targetWindow && targetOrigin) {
+      logger('Sending message %i to parent via postMessage', request.id);
       targetWindow.postMessage(request, targetOrigin);
     } else {
+      logger('Adding message %i to parent message queue', request.id);
       getTargetMessageQueue(targetWindow).push(request);
     }
   }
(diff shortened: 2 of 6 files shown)


erinha:
<p>Analyzed commit id : 5c8f0f90d63c64b8f7ef1de95ba08ca8a1cd530a</p><hr><details><summary><b><span style="color: coral">⯅</span> @microsoft/bundle-analysis-app:</b> +7.05 KB</summary><table><tr><th>Metric Name</th><th>Baseline Size</th><th>Compare Size</th><th>Size Diff</th></tr><tr><td>main.js</td><td>59.42 KB</td><td>66.48 KB</td><td><span style="color: coral">⯅</span> +7.05 KB</td></tr><tr><td>Total Size</td><td>59.42 KB</td><td>66.48 KB</td><td><span style="color: coral">⯅</span> +7.05 KB</td></tr></table></details><hr><p>Baseline commit: 247aa802c407b894cd515b6a838bf15db5b279a0</p>

packages/teams-js/src/internal/communication.ts
```diff
@@ -180,12 +183,20 @@ export function sendMessageToParent(actionName: string, argsOrCallback?: any[] |
...
+const sendMessageToParentHelperLogger = communicationLogger.extend('sendMessageToParentHelper');
+
 /**@internal */
 function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequest {
+  const logger = sendMessageToParentHelperLogger;
+
   const targetWindow = Communication.parentWindow;
   const request = createMessageRequest(actionName, args);
+
+  logger('Message %i information: %o', request.id, { actionName, args });
```
ht4963-ms, line 195: How do we ensure the args do not contain EUII data?

packages/teams-js/src/internal/communication.ts
```diff
@@ -286,29 +299,42 @@ function updateRelationships(messageSource: Window, messageOrigin: string): void
...
 function handleParentMessage(evt: DOMMessageEvent): void {
+  const logger = handleParentMessageLogger;
+
   if ('id' in evt.data && typeof evt.data.id === 'number') {
     // Call any associated Communication.callbacks
     const message = evt.data as MessageResponse;
     const callback = CommunicationPrivate.callbacks[message.id];
+    logger('Received a response from parent for message %i', message.id);
     if (callback) {
+      logger('Invoking the registered callback for message %i with arguments %o', message.id, message.args);
```
ht4963-ms, line 314: Same here. For location API, e.g., the callback would consist of a user's location, correct?

packages/teams-js/src/internal/communication.ts
MarcinJuraszek: This is only ever logged to brower's developer console. I'm not sure having EUII here is really a concern. It might be, if we ask users to share these with us when reporting issues (e.g. as part of issues here on GitHub), but other than that it should be ok.

packages/teams-js/src/internal/communication.ts
```diff
@@ -180,12 +183,20 @@ export function sendMessageToParent(actionName: string, argsOrCallback?: any[] |
   }
 }
 
+const sendMessageToParentHelperLogger = communicationLogger.extend('sendMessageToParentHelper');
+
 /**@internal */
 function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequest {
+  const logger = sendMessageToParentHelperLogger;
```
ht4963-ms, line 190: Can we implement logging levels? This looks to be a lot of logging, maybe these could be divided into a couple of levels to avoid excessive logging? Also, how will this be turned on? Is it by the user, the app developer, or a feature flag?

MarcinJuraszek: The `debug` package, which is, from what I understand, the gold standard for this kind of logging, does not provide log levels. Instead, it leverages namespaces/categories to allow users to enable/disable certain logs. Nothing will be turned on by default, and the user can either enable all the logging (by setting `localStorage.debug = 'teamsJs.*'`) or e.g. only the communication functions (`localStorage.debug = 'teamsJs:communication:*`) or even just one of them (`localStorage.debug = 'teamsJs:communication:flushMessageQueue`). 

Because this is not getting uploaded anywhere, is off by default, I don't want to engineer a new layer on top of that, unless strictly necessary.

packages/teams-js/src/internal/communication.ts
MarcinJuraszek: Yes, but because even if enabled it'll only log that to the browser's developer console, this should not be a concern, imo. Would you agree?

packages/teams-js/src/internal/communication.ts
ht4963-ms: That makes sense. It would be good to instrument some ways for the user to control these settings through the Teams app (even if it's just through the url query), rather than having to muck with the console.

packages/teams-js/src/internal/communication.ts
ht4963-ms: Sounds good. I assume this log will be retrievable and collectable as part of the client logs?

packages/teams-js/src/internal/communication.ts
ht4963-ms: Yes, I agree.

packages/teams-js/package.json
```diff
@@ -18,6 +18,12 @@
     "prettier": "prettier --write '**/*.{ts,js,css,html}'",
     "test": "jest"
   },
+  "dependencies": {
+    "debug": "4.3.3"
+  },
```
AE-MS, line 23: IIRC, there was some trouble when we added the dependency on es6-promise here originally and so we had to do something different? Does that apply here?

I think maybe it means we have to document other libraries that developers have to reference if they pull TeamsJS from the CDN?

Does debug have a minimum supported browser version?

AE-MS (changes requested):
:clock1:

packages/teams-js/package.json
MarcinJuraszek: afaik, that was not because of the dependency itself, but because we called into it through `require('es6-promise')` which doesn't work. 

This bundles `debug` into our bundle (the one we deploy to CDN) and there's no need for the developer to reference anything extra for it to work.

The documentation does not specify their browser support matrix (not surprising, tbh). It is used widely across a lot of projects, including things like Fluid at Microsoft, so I would be very surprised if we had to support a browser it doesn't. Also, the project itself isn't very complicated and, from my skim through the code, does not use any modern potentially limiting browser features.

AE-MS (approved):

PR #817, opened 2022-01-06
Made getAuthToken's param optional

TrevorJoelHarris:
getAuthToken took in a parameter that was an array that contained only optional values. Made the array an optional parameter as well and added two new unit tests to verify that passing in no array still works as expected.

This is a fix for bug #5377609


AE-MS (approved):
:shipit:

PR #820, opened 2022-01-06
Add FrameContext to public\index.ts exports

TrevorJoelHarris:
Erin checked with Yuri and it sounds like not exporting this was an oversight in the main branch.


ydogandjiev (approved):

AE-MS (approved):
:shipit:

PR #824, opened 2022-01-08
Make the clean command work on non-Windows platform

erinha:
I think this is the only script command we have that isn't happy when you run it on a Mac. Tested it on Mac and Windows and it works on both.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "build-test-app-local": "yarn workspace teams-test-app build:local",
     "clean": "lerna run clean --stream",
     "clean:full": "yarn clean && yarn clean:modules",
-    "clean:modules": "lerna clean -y && if exist node_modules (rd /s /q node_modules)",
+    "clean:modules": "lerna clean -y && rm -rf node_modules",
     "efb": "lerna run eslint-fix-build",
     "publish-sdk": "yarn publish packages/teams-js",
     "start-perf-app": "yarn workspace teams-perf-test-app start",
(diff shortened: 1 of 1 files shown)


TrevorJoelHarris (dismissed):
:shipit:

package.json
```diff
@@ -18,7 +18,7 @@
     "build-test-app-local": "yarn workspace teams-test-app build:local",
     "clean": "lerna run clean --stream",
     "clean:full": "yarn clean && yarn clean:modules",
-    "clean:modules": "lerna clean -y && if exist node_modules (rd /s /q node_modules)",
+    "clean:modules": "lerna clean -y && rm -rf node_modules",
```
erinha, line 21: Sadly I was too optimistic about it working on Windows. My test case was atypical and it looks like it won't work for all Windows developers. Going to use `rimraf` instead, as per a Stack Overflow article (thanks Aaron!). Stay tuned for the update.

AE-MS:
For posterity: per our discussion and testing, this doesn't work on Windows generally. There is no `rm` command on Windows. There is one in the git/bin directory, but that directory isn't in the PATH by default. We could perhaps doing something about that for engineers on our team but I suggested using rimraf instead

AE-MS (changes requested):
:clock1:

package.json
erinha: Fixed.

AE-MS (approved):
:shipit:

PR #826, opened 2022-01-10
Add localhost to allowed cross frame message origins for testing

jekloudaMSFT:


AE-MS (approved):

MarcinJuraszek (approved):

src/internal/constants.ts
```diff
@@ -74,6 +74,7 @@ export const validOrigins = [
   'https://local.teams.live.com:8080', // local development
   'https://local.teams.office.com', // local development
   'https://local.teams.office.com:8080', // local development
+  'https://localhost:8080', // local development
```
ydogandjiev, line 77: This is not a good idea. We used to have localhost in this list and removed it for security reasons.

ydogandjiev (changes requested):

src/internal/constants.ts
jekloudaMSFT: What was the security concern?

PR #832, opened 2022-01-11
Intentional unit test failure

TrevorJoelHarris:
Want to see if this causes the cloud build to notice the failed test. When I run the test locally it prints errors to the console but jest reports that 763/763 tests passed

This should NOT be merged.


TrevorJoelHarris:
Learned what I wanted, closing.

PR #833, opened 2022-01-13
Backwards compatibility for tasks API

Ella-ly:
Added in callback capability for the Tasks API depending on if we are in Teams JS v1 mode or TeamsJSv2 mode.


apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -1,13 +1,13 @@
...
+import { dialog, DialogInfo, IAppWindow, ParentAppWindow, TaskInfo, tasks } from '@microsoft/teams-js';
 import React, { ReactElement } from 'react';
 
 import { ApiWithoutInput, ApiWithTextInput } from './utils';
 
 const DialogAPIs = (): ReactElement => {
   const childWindowRef = React.useRef<IAppWindow | null>(null);
   const OpenDialog = (): ReactElement =>
-    ApiWithTextInput<DialogInfo>({
+    ApiWithTextInput<DialogInfo | TaskInfo>({
```
MarcinJuraszek, line 10: Are these two different? I think they should be exactly the same?

apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -16,29 +16,56 @@ const DialogAPIs = (): ReactElement => {
...
+            const onComplete = (err: string, result: string | object): void => {
+              setResult('Error: ' + err + '\nResult: ' + result);
+            };
+            // Store the reference of child window in React
+            const childWindow = dialog.open(dialogInfo, onComplete);
+            childWindow.addEventListener('message', (message: string) => {
+              // Message from parent
+              setResult(message);
+            });
+            childWindowRef.current = childWindow;
```
MarcinJuraszek, line 30: can this be put into some sort of a helper function to avoid duplication?

apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -16,29 +16,56 @@ const DialogAPIs = (): ReactElement => {
...
+          }
+        },
+        submit: {
+          withPromise: async dialogInfo => {
+            dialog.resize(dialogInfo);
+            return 'Teams client SDK call dialog.resize was called';
+          },
+          withCallback: (taskInfo, setResult) => {
+            tasks.updateTask(taskInfo);
+            setResult('Teams client SDK call tasks.updateTask was called');
```
MarcinJuraszek, line 66: would the difference in the result string here mean we'll need separate tests for both promise and callback?

apps/teams-test-app/src/components/DialogAPIs.tsx
Ella-ly: Yes. I am thinking of using teams-v1 script situation in which there is no `DialogInfo`.

apps/teams-test-app/src/components/DialogAPIs.tsx
Ella-ly: Discussed offline.

apps/teams-test-app/src/components/DialogAPIs.tsx
MarcinJuraszek: Can you use the helper for promise version too?

Ella-ly: Yes. I am still doing the test to check the e2e test failure. I should put that in the PR description. Sorry about that.

jadahiya-MSFT (approved):

PR #838, opened 2022-01-13
Unit tests for mail.ts

TrevorJoelHarris:
An initial set of tests for openMailItem for now, curious to see if anyone has comments or suggestions


packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    it('should not allow calls from settings context', async () => {
+      await utils.initializeWithContext('settings');
+
+      await mail
+        .openMailItem(openMailItemParams)
+        .catch(e =>
+          expect(e).toMatchObject(
+            new Error('This call is only allowed in following contexts: ["content"]. Current context: "settings".'),
+          ),
```
AE-MS, line 47: I know that this is not a new pattern that you're implementing here but I've always thought it was kind of weird that we matched against the entire string. E.g. if we update the mail capability to be allowed in a new context later, we have to come back and update these tests. It seems like the only thing we should care about matching against is that it is throwing an error that says "settings" is an invalid context. But I don't really know if you can practically do that. In theory it seems like we should be throwing a more specific error like `InvalidContextError` and just checking for that error and ignoring the string. But I'm sure that's a much bigger change.

You can ignore me I guess. I mean more than usual.

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+      await utils.initializeWithContext('meetingStage');
+
+      await mail
+        .openMailItem(openMailItemParams)
+        .catch(e =>
+          expect(e).toMatchObject(
+            new Error('This call is only allowed in following contexts: ["content"]. Current context: "meetingStage".'),
+          ),
+        );
+    });
```
AE-MS, line 123: It feels like all these tests should be a loop over the list of possible contexts or something. Or at least a one-line call to a helper method that takes the context as a parameter?

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+  afterEach(() => {
+    // Reset the object since it's a singleton
+    if (app._uninitialize) {
+      app._uninitialize();
+    }
+  });
+
+  describe('openMailItem', () => {
+    const openMailItemParams: mail.OpenMailItemParams = {
+      itemId: '',
```
AE-MS, line 30: It seems like the `openMailItem` function should throw if it gets an empty item id?

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+        success: true,
+        error: 'Something went wrong...',
+      };
+
+      utils.respondToMessage(openMailItemMessage, data);
+      await promise;
+
+      expect(openMailItemMessage).not.toBeNull();
+      expect(openMailItemMessage.args.length).toEqual(1);
+      expect(openMailItemMessage.args[0]).toStrictEqual(openMailItemParams);
```
AE-MS, line 167: Since the only assertions here are on the message object received by the infrastructure, do you need to create a data object, respond to the message, and await the result? It seems like the thing you are testing don't require those actions to happen.

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+    });
+
+    it('should not allow calls if runtime does not support mail', async () => {
+      await utils.initializeWithContext('content');
+      utils.setRuntimeConfig({ apiVersion: 1, supports: {} });
+
+      await mail.composeMail(composeMailParams).catch(e => expect(e).toBe('Not Supported'));
+    });
+
+    it('should successfully throw if the openMailItem message sends and fails', async () => {
```
AE-MS, line 275: >openMailItem [](http://example.com/codeflow?start=41&length=12)

Copy pasta

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+      const data = {
+        success: false,
+        error: 'Something went wrong...',
+      };
+
+      utils.respondToMessage(composeMail, data);
+      await composeMailPromise.catch(e => expect(e).toMatchObject(new Error('Something went wrong...')));
+    });
+
+    it('should successfully send the openMailItem message', async () => {
```
AE-MS, line 292: >openMailItem [](http://example.com/codeflow?start=37&length=12)

Same

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      const composeMail = utils.findMessageByFunc('mail.composeMail');
+
+      const data = {
+        success: false,
+        error: 'Something went wrong...',
+      };
+
+      utils.respondToMessage(composeMail, data);
+      await composeMailPromise.catch(e => expect(e).toMatchObject(new Error('Something went wrong...')));
```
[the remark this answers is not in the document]
AE-MS: Same on lines 142 & 146

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+        success: true,
+        error: 'Something went wrong...',
+      };
+
+      utils.respondToMessage(composeMailMessage, data);
+      await promise;
+
+      expect(composeMailMessage).not.toBeNull();
+      expect(composeMailMessage.args.length).toEqual(1);
+      expect(composeMailMessage.args[0]).toStrictEqual(composeMailParams);
```
AE-MS, line 310: Same comment as on line 162 about responding to the message being unnecessary?

AE-MS (changes requested):
:clock1:

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,325 @@
+import { GlobalVars } from '../../src/internal/globalVars';
+import { app } from '../../src/public/app';
+import { mail } from '../../src/public/mail';
+import { Utils } from '../utils';
+
+describe('mail', () => {
```
AE-MS, line 6: >describe [](http://example.com/codeflow?start=0&length=8)

Love seeing these new unit tests! Thank you!

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: I agree with everything you wrote here, including the part about it being too big of a change for this. I think in general (in the future) we're going to want to be more deliberate about how we're handling errors and error types. The functions being tested here are occasionally throwing strings instead of Errors which is not great (as an example).

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: No problem

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: This is a good question. I think the answer is that without responding to the message the promise from line 153 never finishes and the test times out.

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: Repeating here just in case: believe you have to respond to the message or the promise never resolves and the test times out after 5 seconds.

erinha (dismissed):
:shipit:

packages/teams-js/src/public/mail.ts
```diff
@@ -14,6 +14,10 @@ export namespace mail {
         throw 'Not Supported';
       }
 
+      if (!openMailItemParams.itemId || !openMailItemParams.itemId.trim()) {
```
MarcinJuraszek, line 17: How do we decide where it's OK to just let the TypeScript type system enforce stuff, and when we need to do our own validation on top of it? E.g. should you check if `openMailItemParams.itemId` is a `string` too before calling `trim()` on it?

packages/teams-js/src/public/mail.ts
```diff
@@ -14,6 +14,10 @@ export namespace mail {
...
+        throw new Error('Must supply an item id to openMailItem');
```
erinha, line 18: This message doesn't match the test expectation and is failing the tests (yay!)

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+      utils.setRuntimeConfig({ apiVersion: 1, supports: { mail: {} } });
+
+      const promise = mail.openMailItem(openMailItemParams);
+
+      const openMailItemMessage = utils.findMessageByFunc('mail.openMailItem');
+
+      const data = {
+        success: true,
+        error: dataError,
+      };
```
AE-MS, line 147: Not sure if this is something to be fixed in this PR or if it can even be fixed but it's notable that some part of the system allows `success` to be `true` but also allows there to be a non-empty `error` string. It seems like it should either be `true` or an error message, but in theory it shouldn't be allowable to have both.

AE-MS (dismissed):
:shipit:

packages/teams-js/src/public/mail.ts
TrevorJoelHarris: I am excited to announce that this is because I forgot to save the other half of the change

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: Okay, I had to look into this more and it caused me to fix a bug with the test which is good!

It turns out that "respond to message" doesn't actually expect a single data object with values, it wants separate "success" and "error message" values. The test looked like it was passing because the function was just not actually throwing (it was testing truthiness on the first argument and since it was an object it was true).

I find the the respondToMessage/communication code a little hard to follow because of just how async it is and how much it relies on passing around args objects that don't use types, but I think that the use of mysterious args is the thing I might look into more closely if I were to try to fix this. That being said, definitely outside the bounds of what this PR is doing.

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+      utils.respondToMessage(openMailItemMessage, data.success, data.error);
+
+      let error;
+      try {
+        await openMailItemPromise;
+      } catch (e) {
+        error = e;
+      }
+
+      expect(error).toMatchObject(new Error(dataError));
```
erinha, line 120: I was curious about why this changed and appreciate the offline explanation (there was nothing to fail if no exception was thrown). I am curious if we could use expect.assertions here: https://stackoverflow.com/questions/50816254/necessary-to-use-expect-assertions-if-youre-awaiting-any-async-function-calls

(My curiosity drove me to try it locally and the test code is actually throwing 5 assertions, not sure why)

packages/teams-js/src/public/mail.ts
MarcinJuraszek: OK. That was a bit of a leading question I guess, and I hoped to get the very answer you gave. A follow up question - why do we then check for `!openMailItemParams.itemId`? It's already covered by the type system as `itemId` is a required property:

```
  export interface ComposeReplyOrForwardParams<T extends ComposeMailType> extends ComposeMailBase<T> {
    itemid: string;
  }
```

The `...trim()` check is a different story, as that's not something you can push to type system. But the undefined/null check is already covered.

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: More good questions which led me to find more problems with our unit tests.

It turned out that the utils.ts file in the tests directory contained 4 expects statements that were running in over half of our tests and just silently passing all the time. I've converted those to be throws and then added expect.assertions(n) to any function that is checking expects of an async call in a .then or .catch. This will verify that we are hitting the right number of expect checks without making us jump through hoops like this.

packages/teams-js/test/utils.ts
```diff
@@ -39,10 +39,10 @@ export class Utils {
 
     this.parentWindow = {
       postMessage: function(message: MessageRequest, targetOrigin: string): void {
-        if (message.func === 'initialize') {
-          expect(targetOrigin).toEqual('*');
-        } else {
-          expect(targetOrigin).toEqual(that.validOrigin);
```
TrevorJoelHarris, line 45: I really don't think this is a good pattern to use. Putting expect checks in shared functions used by all tests means that we end up with a bunch of hidden expects that people don't know about and make it very hard to use expect.assertions() anywhere.

packages/teams-js/test/public/mail.spec.ts
TrevorJoelHarris: (To be clear, you could do what I had originally and jump through hoops to validate that you were actually testing what you meant but hey now you can do it easily)

packages/teams-js/test/utils.ts
erinha: Agree!

erinha (approved):
Looks great, thanks for drilling down into (and fixing) some of these tangent-y issues!

packages/teams-js/src/public/mail.ts
TrevorJoelHarris: I think we could leave out the !openMailItemParams.itemId check if it makes us feel bad. Fwiw it is enforcing that someone can't create an OpenMailItemParams object that looks like this:

`{itemId: undefined}`

I think it's unlikely someone would do that, but having this check in place will prevent it explicitly (and I have a test that checks that)

PR #839, opened 2022-01-13
Backwards compatibility for sharing API

Ella-ly:


apps/teams-test-app/src/components/SharingAPIs.tsx
```diff
@@ -28,15 +28,25 @@ const ShareWebContent = (): React.ReactElement =>
...
-      submit: async input => {
-        return new Promise<string>((res, rej) => {
-          sharing.shareWebContent(input, error => {
-            if (error) {
-              rej(JSON.stringify(error));
-            }
-            res('Success');
+      submit: {
+        withPromise: async (input, setResult) => {
+          return new Promise<string>((res, rej) => {
```
MarcinJuraszek, line 33: I know it's what it was before, but I'm a bit confused by why this is not something like:
```
await sharing.shareWebContent(input);
return 'Success';
```

wouldn't the error be handled and printed out automatically by the `ApiWithTextInput` component?

Ella-ly: You are right, the `await` codes were what I wrote in the first commit. I have over-think about this part and I will change it back.

MarcinJuraszek (approved):

PR #840, opened 2022-01-13
Merging changes from v1 into v2 - Git magic January 13, 2022

aeunms:
For your convenience, here are [the changes from v1](https://github.com/OfficeDev/microsoft-teams-library-js/compare/aa80fbdcba0afa8da57c7d940f03e26282e6f4da...2005bdbc55f2f702c4dfd06c80787bb7299bfcda) that have been altered to fit in our repo. Please compare those changes with the file changes in this PR to ensure they have been altered correctly and nothing has been dropped on the way here.

### DIFF
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -125,6 +125,7 @@ export namespace authentication {
             link.href,
             authenticateParameters.width,
             authenticateParameters.height,
+            authenticateParameters.isExternal,
           ]).then(([success, response]: [boolean, string]) => {
             if (success) {
               return response;
@@ -281,7 +282,7 @@ export namespace authentication {
     height = Math.min(height, Communication.currentWindow.outerHeight - 200);
     // Convert any relative URLs into absolute URLs before sending them over to the parent window
     const link = document.createElement('a');
-    link.href = authenticateParameters.url;
+    link.href = authenticateParameters.url.replace('{oauthRedirectMethod}', 'web');
     // We are running in the browser, so we need to center the new window ourselves
     let left: number =
       typeof Communication.currentWindow.screenLeft !== 'undefined'
@@ -494,6 +495,10 @@ export namespace authentication {
      * The preferred height for the pop-up. This value can be ignored if outside the acceptable bounds.
      */
     height?: number;
+    /**
+     * The flag which indicates whether the auth page should be opened in an external browser. This flag has no effect on the web client.
+     */
+    isExternal?: boolean;
   }
 
   /**
--- a/packages/teams-js/test/public/authentication.spec.ts
+++ b/packages/teams-js/test/public/authentication.spec.ts
@@ -336,15 +336,17 @@ describe('authentication', () => {
           url: 'https://someUrl',
           width: 100,
           height: 200,
+          isExternal: true,
         };
         authentication.authenticate(authenticationParams);
 
         const message = utils.findMessageByFunc('authentication.authenticate');
         expect(message).not.toBeNull();
-        expect(message.args.length).toBe(3);
+        expect(message.args.length).toBe(4);
         expect(message.args[0]).toBe(authenticationParams.url.toLowerCase() + '/');
         expect(message.args[1]).toBe(authenticationParams.width);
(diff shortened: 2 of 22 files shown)


packages/teams-js/src/public/authentication.ts
```diff
@@ -281,7 +282,7 @@ export namespace authentication {
     height = Math.min(height, Communication.currentWindow.outerHeight - 200);
     // Convert any relative URLs into absolute URLs before sending them over to the parent window
     const link = document.createElement('a');
-    link.href = authenticateParameters.url;
+    link.href = authenticateParameters.url.replace('{oauthRedirectMethod}', 'web');
```
TrevorJoelHarris, line 285: >link.href = authenticateParameters.url.replace('{oauthRedirectMethod}', 'web'); [](http://example.com/codeflow?start=4&length=79)

Not something that should hold up the merge, but @AE-MS have you seen something like this before? Came from this change: [https://github.com/OfficeDev/microsoft-teams-library-js/pull/750](https://github.com/OfficeDev/microsoft-teams-library-js/pull/750)

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -396,62 +398,6 @@ describe('authentication', () => {
...
-        height: 200,
-      };
-      const promise = authentication.authenticate(authenticationParams);
-
-      const message = utils.findMessageByFunc('authentication.authenticate');
-      expect(message).not.toBeNull();
-
-      utils.respondToMessage(message, false, 'someReason');
-
-      return expect(promise).rejects.toThrowError('someReason');
```
TrevorJoelHarris, line 451: I don't see these tests removed in the original change, why do we need to remove them?

packages/teams-js/test/public/media.spec.ts
```diff
@@ -874,25 +944,14 @@ describe('media', () => {
         return blobReadingPromise;
       }
 
-      it('getMedia calls with successful result via the handler', done => {
-        mobilePlatformMock.initializeWithContext(FrameContexts.content).then(() => {
+      it('getMedia calls with successful result via the handler', () => {
```
TrevorJoelHarris, line 947: I don't see the changes from here down in the original set of changes you linked. What caused them to get introduced here?

TrevorJoelHarris (changes requested):
:clock1:

packages/teams-js/test/public/authentication.spec.ts
aeunms: These were discovered as duplicates. I knew I fixed these tests and thought it was weird they were failing, so I looked and there were two of each.

aeunms: The ones kept can be seen at line 332

packages/teams-js/test/public/media.spec.ts
aeunms: This was me tinkering around to try and get them to pass the tests. I might leave them like this because I understood this as what should be done anyways from some of Niharika's feedback from the last git magic PR.

test-results/unit/unit-tests-report.xml
```diff
@@ -0,0 +1,283 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
aeunms, line 1: I staged this file by accident. I'll be adding this to gitignore.

packages/teams-js/test/public/media.spec.ts
TrevorJoelHarris: If possible, I think it would be better to merge this without extra tinkering and then add this right away after the merge. It keeps the merge checkins limited to "just the merge" which can be helpful. Was this change necessary to make the test pass?

packages/teams-js/src/public/authentication.ts
AE-MS: @MarcinJuraszek want to look at this?

packages/teams-js/test/public/media.spec.ts
aeunms: Great clarification, they're not needed for making the tests pass so I've just reverted these changes.

TrevorJoelHarris (approved):
:shipit:

PR #841, opened 2022-01-13
Add unit tests for calendar

TrevorJoelHarris:
Very similar to mail unit tests

### DIFF
--- a/packages/teams-js/test/public/calendar.spec.ts
+++ b/packages/teams-js/test/public/calendar.spec.ts
@@ -1,0 +1,323 @@
+import { GlobalVars } from '../../src/internal/globalVars';
+import { app } from '../../src/public/app';
+import { calendar } from '../../src/public/calendar';
+import { Utils } from '../utils';
+
+describe('calendar', () => {
+  // Use to send a mock message from the app.
+  const utils = new Utils();
+
+  beforeEach(() => {
+    utils.processMessage = null;
+    utils.messages = [];
+    utils.childMessages = [];
+    utils.childWindow.closed = false;
+    GlobalVars.frameContext = undefined;
+
+    // Set a mock window for testing
+    app._initialize(utils.mockWindow);
+  });
+
+  afterEach(() => {
+    // Reset the object since it's a singleton
+    if (app._uninitialize) {
+      app._uninitialize();
+    }
+  });
+
+  describe('openCalendarItem', () => {
+    const openCalendarItemParams: calendar.OpenCalendarItemParams = {
+      itemId: '',
+    };
+
+    it('should not allow calls before initialization', async () => {
+      await calendar
+        .openCalendarItem(openCalendarItemParams)
+        .catch(e => expect(e).toMatchObject(new Error('The library has not yet been initialized')));
+    });
+
+    it('should not allow calls from settings context', async () => {
+      await utils.initializeWithContext('settings');
+
+      await calendar
+        .openCalendarItem(openCalendarItemParams)
+        .catch(e =>
+          expect(e).toMatchObject(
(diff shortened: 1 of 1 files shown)


packages/teams-js/test/public/calendar.spec.ts
```diff
@@ -0,0 +1,275 @@
+import { GlobalVars } from '../../src/internal/globalVars';
+import { FrameContexts } from '../../src/public';
+import { app } from '../../src/public/app';
+import { calendar } from '../../src/public/calendar';
+import { Utils } from '../utils';
+
+describe('calendar', () => {
+  // Use to send a mock message from the app.
```
AE-MS, line 8: Not really sure you should do anything about this here but it's notable how many times the comments on lines 8, 18, and 23 appear in the codebase.

packages/teams-js/test/public/calendar.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+    });
+
+    it('should throw if a null itemId is supplied', async () => {
+      expect.assertions(1);
+
+      await utils.initializeWithContext('content');
+      utils.setRuntimeConfig({ apiVersion: 1, supports: { calendar: {} } });
+
+      await calendar
+        .openCalendarItem({ itemId: null })
```
AE-MS, line 82: Why doesn't this trigger compiler error 2322? I'm guessing the answer is because we don't have it turned on and we should probably have a bug to do that (or figure out what set of warnings it is in that we don't have enabled)

image

packages/teams-js/test/public/calendar.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+            .composeMeeting(composeMeetingParams)
+            .catch(e =>
+              expect(e).toMatchObject(
+                new Error(
+                  `This call is only allowed in following contexts: ["content"]. Current context: "${frameContext}".`,
+                ),
+              ),
+            );
+        });
+      });
```
AE-MS, line 198: I bet this whole thing could be made a helper function that takes in three parameters: the allowed frame context, a function it should call and expect an error, and the string it should test against.

I'm guessing this code only occurs in three places now (two in this file and one in mail?) but even three copies is probably too many and I'm guessing the test should probably be in many other files.

Up to you if you want to fix it as part of this change though; I've just become very familiar with the code which triggered my spidey sense.

TrevorJoelHarris: Yeah, I went back and forth on this. I can definitely make a shared function that does this but I also kind of like having expect statements in actual test files and not hidden in a helper function somewhere cough cough utils.ts

That being said, there might be some way to make a new file for this sort of thing and name it in such a way that it is not mysterious. I will think about it (but probably not tackle it as part of this change)

packages/teams-js/test/public/calendar.spec.ts
TrevorJoelHarris: @erinha Where should I parent a bug like this? Do you have a likely candidate in mind?

packages/teams-js/test/public/calendar.spec.ts
TrevorJoelHarris: I think in the interest of putting a bow on this PR I am going to leave this as is for now. I think this is a good idea, but I'm not entirely sure where I would want to put these sort of "cross file tests".

packages/teams-js/test/public/calendar.spec.ts
erinha: Answered in chat

TrevorJoelHarris: bug has been opened

PR #842, opened 2022-01-14
Lakhveer/teams core

lakhveerk:
1. Adding isSupported function in teams Core test app
2. Moving RegisterFocusEnterHandler  from teamscore to pagesAPI


aeunms (approved):
LGTM!

apps/teams-test-app/src/components/TeamsCoreAPIs.tsx
```diff
@@ -82,6 +88,7 @@ const TeamsCoreAPIs = (): ReactElement => (
     <RegisterOnLoadHandler />
     <RegisterBeforeUnloadHandler />
     <RegisterFocusEnterHandler />
```
MarcinJuraszek, line 90: This should move to `pages`, where we already have `returnFocus`.

MarcinJuraszek (changes requested):

apps/teams-test-app/src/components/PagesAPIs.tsx
```diff
@@ -121,6 +122,28 @@ const ReturnFocus = (): React.ReactElement =>
     },
   });
 
+const RegisterFocusEnterHandler = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'registerFocusEnterHandler',
+    title: 'Register On Focus Enter Handler',
+    onClick: {
+      withPromise: async setResult => {
```
MarcinJuraszek, line 130: These look exactly the same. you don't really need both.

MarcinJuraszek (changes requested):

apps/teams-test-app/src/components/PagesAPIs.tsx
lakhveerk: Sure, I can keep only one. btw, I copied it from teamsCoreAPI, where it was added by PR#854 (https://github.com/OfficeDev/microsoft-teams-library-js/pull/854/files)

MarcinJuraszek (approved):

PR #845, opened 2022-01-14
Fixing build badges and moving them to top

ramourt:
Still need to fix the coverage badge. 

Build badges look like this: 
image


aeunms (approved):
LGTM!

PR #849, opened 2022-01-18
Backwards compatibility for AppWindow API

jadahiya-MSFT:


apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -46,29 +46,71 @@ const DialogAPIs = (): ReactElement => {
...
+                if (reason && !status) {
+                  setResult(JSON.stringify(reason));
+                } else {
+                  setResult('Message sent to child');
+                }
+              };
+              childWindow.postMessage(message, onComplete);
+            } else {
+              setResult("childWindow doesn't exist");
+            }
```
Ella-ly, line 75: Can we please make the string consistent with `withPromise` here? Also `SendMessageToParent` below.

jadahiya-MSFT: This was more of a lint error, the output string would be the same. The difference is we needed to use \' to get the apostraphe into the first string due to using single quotations. The second one we can't pass single quotations in. I could change it to `childWindow doesnt exist` to keep it consistent though

Ella-ly (approved):

PR #853, opened 2022-01-19
Adding try and catch in test-app: ApiWithoutTextInput

lakhveerk:
If some error occurs on host side, we need to catch that. (Just like we do it in ApiWithTextInput).


apps/teams-test-app/src/components/utils/ApiWithoutInput.tsx
```diff
@@ -18,26 +18,25 @@ export interface ApiWithoutInputProps {
 export const ApiWithoutInput = (props: ApiWithoutInputProps): React.ReactElement => {
   const { name, onClick, title } = props;
   const [result, setResult] = React.useState('');
-
+  const onClickCallback = React.useCallback(async () => {
```
jadahiya-MSFT, line 21: Not 100% certain on this, but I don't think it is recommended to do `async () => { ...` inside a react hook like this. It might be better to do this instead.
`const onClickCallback = React.useCallback( () => { 
           const asyncHandler = async () => { 
           ...
           };
           
           asyncHandler();
     }
 `
 
 Additionally, I'm not entirely sure why we are using a React hook here in the first place, sorry if that's a dumb question, just unsure.

MarcinJuraszek: Interesting! Do you have any resources you could share on why `async () => {}` in `React.useCallback` isn't recommended?

jadahiya-MSFT: I believe it's because `React.useCallback` gets morphed into `React.useMemo` and if I recall correctly performing a `async` call would lead to side effects. I believe that according to the [ReactJS Hooks FAQ](https://reactjs.org/docs/hooks-faq.html). Side effects aren't allowed in `React.useMemo`. It actually appears that the recommended approach would be to use `React.useEffect`. But I could just be completely misunderstanding and I'm sorry if that's the case.

lakhveerk: Closing this comment per our convo. useCallback returns its function uncalled so you can call it later, while useMemo calls its function and returns the result. That means async is not a prob when we are using usecallback

MarcinJuraszek (approved):

PR #854, opened 2022-01-20
Backwards compatibility for APIs in publicAPIs.ts

Ella-ly:


apps/teams-test-app/src/components/PagesBackStackAPIs.tsx
```diff
@@ -61,16 +62,26 @@ const PagesBackStackAPIs = (): ReactElement => {
...
-    pages.backStack.registerBackButtonHandler((): boolean => {
-      if (totalStates > 0) {
-        const newNumStates = totalStates - 1;
-        setTotalStates(newNumStates);
-        setRegisterBackButtonHandlerValue('back button clicked. total remaining state: ' + newNumStates);
-        return true;
-      }
-      return false;
-    });
+    if (getTestBackCompat()) {
```
jadahiya-MSFT, line 76: This breaks our current way of doing things, but I'm not sure how to go around it.

jadahiya-MSFT (approved):

apps/teams-test-app/src/components/PagesBackStackAPIs.tsx
Ella-ly: @MarcinJuraszek Can you please take a look here?

apps/teams-test-app/src/components/PagesAPIs.tsx
```diff
@@ -115,9 +131,15 @@ const SetCurrentFrame = (): React.ReactElement =>
...
-        pages.setCurrentFrame(input);
-        return 'called';
+      submit: {
+        withPromise: async input => {
+          pages.setCurrentFrame(input);
+          return 'called';
+        },
+        withCallback: (input, setResult) => {
+          setFrameContext(input);
+          setResult('called');
```
MarcinJuraszek, line 141: you're probably going to hate me for this, but now seeing how many of these call `setResult` anyway, maybe we should add the option to return a string from here back :D Let's chat about this when we meet later today.

apps/teams-test-app/src/components/PagesBackStackAPIs.tsx
MarcinJuraszek: I think this is fine. It's a non-standard API which is why it's not using the `ApiWithTextInput` component. I don't think we should spend any more time than necessary to try to get it look/feel 'better'. This seems totally appropriate to me.

MarcinJuraszek (approved):

apps/teams-test-app/src/components/PagesAPIs.tsx
jadahiya-MSFT: It's interesting because there are a lot of calls that would also lead to importing `noHostSdk` if we add this back in, do you think we should modify `withCallback` to allow for both scenarios? One where we would want to return a string and the one where we don't?

PR #856, opened 2022-01-20
Organize tools into separate directory

aeunms:
bundle-analysis-tools and bundle-analysis-app are not meant to test the individual APIs but instead are used by the codeowners to gauge package size changes. As such, they are being moved into tools to better reflect their usage.


.gitignore
```diff
@@ -21,6 +21,7 @@ yarn-error.log
 /packages/*/bundleAnalysis
 /apps/*/bundleAnalysis
 /common/*
+/tools/bundle-size-tools/bundle-analysis-app/bundleAnalysis
```
AE-MS, line 24: Perhaps these could be alphabetized?

AE-MS (dismissed):
:shipit:

.gitignore
```diff
@@ -21,6 +21,7 @@ yarn-error.log
 /packages/*/bundleAnalysis
```
ramourt, line 21: nit: Don't think these 2 lines for bundleAnalysis are still necessary

tools/bundle-size-tools/bundle-analysis-app/tsconfig.json
```diff
@@ -1,11 +1,7 @@
 {
```
ramourt, line 1: I don't know if it matters but just double checking if this is intentional: 
bundle-analysis-app package is now in a subfolder within the package bundle-size-tools?

ramourt (dismissed):
lgtm

tools/bundle-size-tools/bundle-analysis-app/tsconfig.json
aeunms: It was intentional, I thought it would be good to put the app along with the tools, but I could move it elsewhere!

ramourt (approved):

PR #857, opened 2022-01-21
app.spec.ts unit tests

jekloudaMSFT:
Error validation in app.spec.ts unit tests


erinha:
(Sorry GitHub doesn't let me highlight code that isn't close to lines that have been changed. 👎)
I think test case on line 150 ('should throw an error if the given runtime config causes a non parsing related error') should have the error be explicitly verified. In general for our unit tests, if the SDK is throwing an error, I think we should be verifying the specific error that is thrown whenever possible. The unit test currently just has `expect(utils.respondToMessage(initMessage, FrameContexts.content, HostClientType.web, null)).toThrowError` and I would like to see the exact error message validated.

erinha:
>     expect(utils.respondToMessage(initMessage, FrameContexts.content, HostClientType.web, null)).toThrowError;

(Oooh CodeFlow _does_ let me add comments in other places!)
This is the place my other comment refers to - can you please validate the specific error thrown?

---
Refers to: packages/teams-js/test/public/app.spec.ts:154 in 7afa03e. [](commit_id = 7afa03ef5b494c332b2715d01ed0158d59dbc717, deletion_comment = False)

erinha:
>     expect(utils.respondToMessage(initMessage, FrameContexts.content, HostClientType.web, null)).toThrowError;

Okay, I looked at the code more closely and I see that `initialize` just rethrows whatever exception is thrown by either JSON.parse or applyRuntimeConfig. 

It may still be worth it to verify a specific exception. One reason for that is if something changes in test or product code and if a different exception is being thrown, it might be covering a real issue.

---
In reply to: [1018093989](https://github.com/OfficeDev/microsoft-teams-library-js/pull/857#issuecomment-1018093989) [](http://example.com/codeflow?ancestors=1018093989)

---
Refers to: packages/teams-js/test/public/app.spec.ts:154 in 7afa03e. [](commit_id = 7afa03ef5b494c332b2715d01ed0158d59dbc717, deletion_comment = False)

erinha:
>   it('should initialized with clientSupportedSDKVersion and runtimeConfig arguments flipped', async () => {

nit: since you're touching this file... :sunglasses: can you please change this to "initialize"? And the next test too.

---
Refers to: packages/teams-js/test/public/app.spec.ts:190 in 7afa03e. [](commit_id = 7afa03ef5b494c332b2715d01ed0158d59dbc717, deletion_comment = False)

packages/teams-js/test/public/app.spec.ts
```diff
@@ -578,7 +604,7 @@ describe('AppSDK-app', () => {
         error: 'Something went wrong...',
```
erinha, line 604: >'Something went wrong...' [](http://example.com/codeflow?start=15&length=25)

I know you didn't introduce this, but can you please use a variable for this message? We use it several times in this file.

erinha:
>     });

These tests look awfully similar. Can you please use a function to iterate through the FrameContexts? Trevor recently did this in mail and calendar tests.

---
Refers to: packages/teams-js/test/public/app.spec.ts:519 in 7afa03e. [](commit_id = 7afa03ef5b494c332b2715d01ed0158d59dbc717, deletion_comment = False)

erinha (changes requested):
:clock1:

jekloudaMSFT:
> > ```
> > expect(utils.respondToMessage(initMessage, FrameContexts.content, HostClientType.web, null)).toThrowError;
> > ```
> 
> Okay, I looked at the code more closely and I see that `initialize` just rethrows whatever exception is thrown by either JSON.parse or applyRuntimeConfig.
> 
> It may still be worth it to verify a specific exception. One reason for that is if something changes in test or product code and if a different exception is being thrown, it might be covering a real issue.
> 
> In reply to: [1018093989](https://github.com/OfficeDev/microsoft-teams-library-js/pull/857#issuecomment-1018093989)
> 
> Refers to: packages/teams-js/test/public/app.spec.ts:154 in [7afa03e](https://github.com/OfficeDev/microsoft-teams-library-js/commit/7afa03ef5b494c332b2715d01ed0158d59dbc717). [](commit_id = [7afa03e](https://github.com/OfficeDev/microsoft-teams-library-js/commit/7afa03ef5b494c332b2715d01ed0158d59dbc717), deletion_comment = False)

Not sure how to fix this test. After rewriting it, I'm seeing that the initialize handler isn't throwing on the null runtime config, and I don't know what value to put in that wouldn't just throw a parsing error, any ideas?

jekloudaMSFT:
> > ```
> > expect(utils.respondToMessage(initMessage, FrameContexts.content, HostClientType.web, null)).toThrowError;
> > ```
> 
> Okay, I looked at the code more closely and I see that `initialize` just rethrows whatever exception is thrown by either JSON.parse or applyRuntimeConfig.
> 
> It may still be worth it to verify a specific exception. One reason for that is if something changes in test or product code and if a different exception is being thrown, it might be covering a real issue.
> 
> In reply to: [1018093989](https://github.com/OfficeDev/microsoft-teams-library-js/pull/857#issuecomment-1018093989)
> 
> Refers to: packages/teams-js/test/public/app.spec.ts:154 in [7afa03e](https://github.com/OfficeDev/microsoft-teams-library-js/commit/7afa03ef5b494c332b2715d01ed0158d59dbc717). [](commit_id = [7afa03e](https://github.com/OfficeDev/microsoft-teams-library-js/commit/7afa03ef5b494c332b2715d01ed0158d59dbc717), deletion_comment = False)

I'm having trouble coming up with a proper test here. I can't come up with any value that would throw any error other than a SyntaxError. Any ideas?

packages/teams-js/src/public/app.ts
```diff
@@ -470,6 +470,14 @@ export namespace app {
...
             try {
               const givenRuntimeConfig: IRuntime = JSON.parse(runtimeConfig);
+              // Check that givenRuntimeConfig is a valid instance of IRuntimeConfig
+              try {
+                if (!givenRuntimeConfig.apiVersion) {
+                  throw new Error('Received runtime config is invalid');
+                }
+              } catch {
+                throw new Error('Received runtime config is invalid');
+              }
```
erinha, line 480: Why the catch and re-throw?

erinha: Won't it be caught in the existing catch?

erinha: If you change the check to `if (!givenRuntimeConfig || !givenRuntimeConfig.apiVersion)` then you don't need to put the check in its own try-catch.

packages/teams-js/test/public/app.spec.ts
```diff
@@ -417,91 +423,96 @@ describe('AppSDK-app', () => {
   });
 
   describe('navigateCrossDomain', () => {
```
erinha, line 425: I noticed that this test, which you did not add, seems like it should actually belong in the pages.spec.ts file, not the app.spec.ts. I suspect this was an oversight when the APIs were organized into the pages capability. Will follow up offline about whether these should be moved as part of this change.

erinha (approved):
Looks great, thanks!

PR #858, opened 2022-01-21
Added file download API : To be used in Files app 2.0

HarshJain27:
Design Document for File Download API: [File Download API Design Document](https://microsoft.sharepoint.com/:w:/t/whatsupdoc/EdrTxFIZmRBMpk4cNAFSfJYBkqp2zt8r9SEBnXW6juay1Q?e=z1tkk2)

Design Document for Open Download Preference Folder API: [Open Download Preference Folder API Design Document](https://microsoft.sharepoint.com/:w:/t/whatsupdoc/EZ1FqiNnys9JjedcAYs_VYkBumHjioXMsgAY0k0GzNVmXA?e=pK5HVq)

Adding APIs for Files 2.0 App -> Download pivot: 1. File download API which will be used to list download files  2. To Open download preference folder
These will be private **files** function and will work only for 1P apps.

Relevant Task : [Workitem](https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2106088)


src/private/files.ts
```diff
@@ -226,6 +226,43 @@ export namespace files {
...
+     */
+    objectId?: string;
+    /**
+     * Path of the file
+     */
+    path?: string;
+    /**
+     * Size of the file
+     */
+    size?: number;
```
TrevorJoelHarris, line 247: >size [](http://example.com/codeflow?start=4&length=4)

This should indicate what the unit is (bytes for example)

src/private/files.ts
```diff
@@ -226,6 +226,43 @@ export namespace files {
...
+    /**
+     * Download status
+     */
+    status?: string;
```
TrevorJoelHarris, line 251: >status?: string; [](http://example.com/codeflow?start=4&length=16)

If this is a limited set of values, please use a type or enum so that callers can see what the potential status values are. Just using the string type doesn't give callers much information

src/private/files.ts
```diff
@@ -226,6 +226,43 @@ export namespace files {
...
+    /**
+     * Download timestamp
+     */
+    timestamp: Date | string;
```
TrevorJoelHarris, line 255: >string [](http://example.com/codeflow?start=22&length=6)

Why string? And what format would the string be in?

src/private/files.ts
```diff
@@ -226,6 +226,43 @@ export namespace files {
...
+    /**
+     * File name
+     */
+    title: string;
+    /**
+     * Type of file
+     */
+    type: string;
```
TrevorJoelHarris, line 263: >type: string; [](http://example.com/codeflow?start=4&length=13)

What does type actually indicate or contain? Is this file extension? Could it be represented as a limited set of possible values in an enum or type? I realize this is a microsoft internal function but it will be very hard for other microsoft developers to use without more information.

src/private/files.ts
```diff
@@ -416,4 +453,21 @@ export namespace files {
...
   }
+
+  /**
+   * @private
+   * Hide from docs
+   *
+   * Gets list of downloads for current user
+   * @param callback Callback that will be triggered post downloads load
+   */
+  export function getFileDownloads(callback: (error: SdkError, files: IFileItem[]) => void): void {
```
MarcinJuraszek, line 464: The callback should have both error and files as optional, to better represent the data that'll actually be provided back when it's invoked.

src/private/files.ts
HarshJain27: Updated the doc comment.

src/private/files.ts
HarshJain27: timestamp is represented as Date or string to remain consistent with the underlying API output data structure format. The string will also be a **Date()** object representation.

src/private/files.ts
HarshJain27: As you rightly guessed, type indicates file extension here. Updated the doc comment for the same. The reason it is represented as string and not type is to avoid the complexity involved in doing so because of huge list of such extensions and/or avoid missing any extension from the type list which might be encountered dynamically. The respective client app code is handling it to infer required icon details etc. Note that the underlying client API output data structure represents this field as string too.

src/private/files.ts
HarshJain27: Comment incorporated.

src/private/files.ts
HarshJain27: Comment Incorporated. Converted string to enum.

src/private/files.ts
MarcinJuraszek: I'd rather have the property be called `sizeInBytes` instead of this being in a comment.

src/private/files.ts
MarcinJuraszek: I don't think we should leak the underlying implementation details of the system that's providing this data. If this is a string, how would anybody know what format this string is in?

src/private/files.ts
MarcinJuraszek: If this is an extension, the property should be called `extension`.

src/private/files.ts
MarcinJuraszek: Also, if it's a timestamp, shouldn't it be a `number` to begin with?

src/private/files.ts
HarshJain27: Since the property is named as **type** in Teams Web Client, hence retained the same name here, so that the conversion/switch can happen seamlessly between SDK and Teams Web Client code.
@ydogandjiev Please correct me if it can work otherwise, as suggested by Marcin.

src/private/files.ts
HarshJain27: Since the property is named as **size** in Teams Web Client, hence retained the same name here, so that the conversion/switch can happen seamlessly between SDK and Teams Web Client code.
@ydogandjiev Please correct me if it can work otherwise, as suggested by Marcin.

src/private/files.ts
HarshJain27: Name of the filed is **timestamp** but it is represented by Date object.

MarcinJuraszek: ok. In that case, it should be typed as `Date`, and not `Date | string`.

src/private/files.ts
MarcinJuraszek: As discussed offline, this is not good enough of a reason for me. We need to focus on designing the best possible developer experience for the developers using the SDK. Not guide the design based on implementation details of Teams (or any other host) trying to host these.

src/private/files.ts
MarcinJuraszek: As discussed offline, this is not good enough of a reason for me. We need to focus on designing the best possible developer experience for the developers using the SDK. Not guide the design based on implementation details of Teams (or any other host) trying to host these.

src/private/files.ts
HarshJain27: Suggestion incorporated.

src/private/files.ts
HarshJain27: Suggestion incorporated.

src/private/files.ts
HarshJain27: Suggestion incorporated.

MarcinJuraszek (approved):

ydogandjiev (approved):

TrevorJoelHarris (approved):

PR #861, opened 2022-01-24
Import TeamsJs v1 in the test App

Ella-ly:
Import TeamsJs v1 in the test App, set up webpack config and webpage, and update azure pipeline to run e2e tests on every PR.


azure-pipelines-with-e2etests-teamsV1.yml
```diff
@@ -0,0 +1,80 @@
+# This pipeline checks-out current client sdk repo and stable app hosting sdk repo (as specified in AppHostingSdkGitRef).
```
aeunms, line 1: Is there a reason we'd want to create a new yml file for this process? I'm currently working on combining all of our yaml files and it seems a lot of the content here is duplicated from our other actions. (It would be a lot more efficient not running them twice if we don't need to.)

Ella-ly: This pipeline is for running e2e tests running against TeamsJs V1. The reason we want it to run on every PR is to make sure changes made to hub SDK don’t break the backwards compatibility.

aeunms: We can still run it on every PR without creating another file. We could maybe add steps to an existing file like azure-pipelines-with-e2etests.yml?

Ella-ly: You are right, thanks. To avoid messing current pipeline, I will delete the file for now and make the change in azure-pipelines-with-e2etests.yml later or maybe in other PR.

aeunms: It looks like just this task needs to be included to run the test, maybe you can just put it in the file now? (You're free to do it whenever you see fit tho ofc)

` - bash: 'node tools/cli/runAppsWithE2ETests.js --envType=teamsV1'
    displayName: 'Run E2E integration tests'
    condition: succeeded()
    workingDirectory: '$(AppHostingSdkProjectDirectory)'`

MarcinJuraszek: This is a tradeoff. On one hand, if we create a new pipeline, it'll force us to clone, build and start the app before running the tests. If we were to add it to the existing pipeline we could reuse the fact that the app is already running. On the other hand, with a separate pipeline all this would happen in parallel to the existing pipeline. If we add it to the existing one, it'll be sequential.

MarcinJuraszek: If the reason to have them in a single pipeline is to avoid duplicating all the steps, we could accomplish it via stage templates: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example%2Cparameter-schema#stage-templates

MarcinJuraszek: Actually, looks like the build command is different between these (`build-test-app` vs `build-test-app-V1`, so yeah, I think we should have them as separate pipelines, and maybe consider using stage templates.

aeunms: Thanks for catching that I missed the build command for the test app here being different though ofc!

Ella-ly: As discussed offline I'll be going ahead with this file and this staging work can be tabled to another PR

aeunms (approved):
:shipit:

azure-pipelines-with-e2etests_teamsV1.yml
```diff
@@ -0,0 +1,80 @@
+# This pipeline checks-out current client sdk repo and stable app hosting sdk repo (as specified in AppHostingSdkGitRef).
```
MarcinJuraszek, line 1: Now that I'm thinking about it, I don't think we want to run these tests as part of TeamsJs v2 validation. No changes we make here should affect this. We definitely want to run it as part of validating the host part in the other repo in ADO.

MarcinJuraszek: On the other hand, we might want to run it to make sure the test app is not changed in a way that would make the tests fail. What do you think Ella?

Ella-ly: Discussed offline. We will keep the pipeline here.

MarcinJuraszek (approved):

aeunms:
As I'm doing my work to combing our yaml files, I noticed that currently we're not using this added e2e test file in our GitHub checks yet- was that intended? Do we not want to include it in our automatic runs yet?

Ella-ly:
> As I'm doing my work to combing our yaml files, I noticed that currently we're not using this added e2e test file in our GitHub checks yet- was that intended? Do we not want to include it in our automatic runs yet?

Yes. There are some jobs that need to be done in the other repo before we have this one running.

PR #865, opened 2022-01-26
Adding workflows to automate release

ramourt:
Adding workflows to automate all steps in release preparation and post release merging into 2.0-preview


.github/workflows/postrelease.yml
```diff
@@ -0,0 +1,55 @@
+name: Post Release
+on:
+  pull_request:
+    branches: [release/2.0.0-beta.*]
+    types: [closed]
```
erinha, line 5: I don't know much about this trigger, but does "closed" differentiate between PRs being merged vs closed? I would think we would not want this to execute if we closed the PR but didn't merge it. (Although I'm not 100% sure the workflow would allow the PR to be closed and not merged.)

tools/cli/utils.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable */
```
erinha, line 1: Do you anticipate adding more to this file in the future? I think the filename could be more indicative of what this file does.

azure-pipelines-with-e2etests.yml
```diff
@@ -3,8 +3,12 @@
...
 
 trigger:
-  - '2.0-preview'
-  - 'release*'
+  branches:
+    include:
+      - '2.0-preview'
+  tags:
+    include:
+      - v2.0.0-beta.*
```
erinha, line 11: So the steps are: 

1. prerelease (creates release branch, makes changes to integrity hash and version constants, creates PR with changelog contents against release branch), 
2. postrelease (creates GitHub (pre)Release, creates tag, creates PR against 2.0-preview), 
3. e2e & script tag tests trigger on tag creation from postrelease step, 
4. CI build triggers on tag creation from postrelease step, which triggers publishing internally. 

Did I summarize that correctly?

erinha (changes requested):
I'm so excited you're doing this! Mostly just questions so I know I understand this correctly.

.github/workflows/postrelease.yml
ramourt: Correct, there is a condition under the job itself that checks if the pr is merged. 
`if: github.event.pull_request.merged == true`
Unfortunately, there isn't a trigger for merged specifically, so even if they are closed this will run and just instantly end. I have this tested out on my fork.

azure-pipelines-with-e2etests.yml
ramourt: Yes!

tools/cli/utils.js
ramourt: I made this at some point because I had written more than one script and didn't like duplicate code, but since it's just the one prerelease.js script now, I think I can just move these into that file.

ramourt: I moved the functions to the prerelease.js script

.github/workflows/postrelease.yml
erinha: Thanks for the clarification!

erinha (approved):

PR #867, opened 2022-01-26
public/authentication.spec.ts Error Validation Tests

jekloudaMSFT:
public/authentication.spec.ts Error Validation Tests


packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -1,13 +1,20 @@
...
 
-  const utils = new Utils();
+  const errorMessage = 'mockError';
+  const mockResult = 'someResult';
+  const mockResource = 'https://someresource/';
+  const mockClaim = 'some_claim';
 
   beforeEach(() => {
+    // Use to send a mock message from the app.
+    utils = new Utils();
```
erinha, line 16: Beautiful, thank you for cleaning this up!

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -25,98 +32,74 @@ describe('authentication', () => {
...
   it('should not allow authentication.authenticate calls before initialization', () => {
+    expect.assertions(1);
     const authenticationParams: authentication.AuthenticatePopUpParameters = {
       url: 'https://someurl/',
       width: 100,
       height: 200,
     };
 
-    expect(() => authentication.authenticate(authenticationParams)).toThrowError(
+    return expect(() => authentication.authenticate(authenticationParams)).toThrowError(
```
erinha, line 42: I'd prefer that we use async/await rather than adding a return here for this test.

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -25,98 +32,74 @@ describe('authentication', () => {
...
+        it(`should not allow authentication.authenticate calls from ${context} context`, async () => {
+          expect.assertions(1);
+          await utils.initializeWithContext(context);
+          const authenticationParams: authentication.AuthenticatePopUpParameters = {
+            url: 'https://someurl/',
+            width: 100,
+            height: 200,
+          };
+
+          return expect(() => authentication.authenticate(authenticationParams)).toThrowError(
```
erinha, line 94: Can we use await instead of return here?

packages/teams-js/test/public/authentication.spec.ts
jekloudaMSFT: We don't need either here since `() => authentication.authenticate()` is not an async function. Reverted these to how they were before.

packages/teams-js/test/public/authentication.spec.ts
jekloudaMSFT: Just removed the return because this isn't an async function

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -581,19 +556,23 @@ describe('authentication', () => {
...
 
     // Wait 100ms for the message queue and 200ms for the close delay
-    setTimeout(() => {
-      expect(closeWindowSpy).toHaveBeenCalled();
-    }, 301);
+    await new Promise<void>(resolve =>
+      setTimeout(() => {
+        expect(closeWindowSpy).toHaveBeenCalled();
+        resolve();
+      }, 350),
```
TrevorJoelHarris, line 563: Do we really need 350ms timeouts? This kind of pattern can really add to the runtime of the unit tests (and if we are waiting a certain amount of time, are these tests actually reliable?)

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -604,28 +583,32 @@ describe('authentication', () => {
...
```
TrevorJoelHarris, line 590: Same question about timeouts

packages/teams-js/test/public/authentication.spec.ts
jekloudaMSFT: Hmm, this test didn't work with the 301ms timeout, so there's not a lot of room to lower that number, but I could see getting rid of this check altogether since the important part of this test is that closeWindow isn't called before the message comes through

TrevorJoelHarris: I would maybe leave these for now and open a bug and assign it to me to figure out why we need them the sleeps at all

jekloudaMSFT: Sure, opened bug 5790851 for this

TrevorJoelHarris (approved):

PR #870, opened 2022-01-27
update function to throw errors

Jessica-Koch:
update function so that it throws instead of returns an error and add a try/catch to functions that call it to handle that error call.  All tests are passing


packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -66,10 +66,10 @@ export function isHostClientMobile(): boolean {
 export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): SdkError {
```
erinha, line 66: Can we please change the signature to not return SdkError?

Jessica-Koch: yes, i think the only type i can add is a `never` type

packages/teams-js/src/public/media.ts
```diff
@@ -412,14 +412,17 @@ export namespace media {
      * @param mediaEvent indicates what the event that needs to be signaled to the host client
      * Optional; @param callback is used to send app if host client has successfully handled the notification event or not
      */
-    protected notifyEventToHost(mediaEvent: MediaControllerEvent, callback?: (err?: SdkError) => void): void {
+    protected notifyEventToHost(
```
erinha, line 415: I'm going to look at this more tomorrow, but I was expecting to see this function and the VideoController.stop function (so the MediaController.stop function as well) be converted to return Promises, like `selectMedia` was. With overloads to support the existing callback pattern that are marked deprecated and a new function that returns a Promise. Will provide more detail tomorrow!

packages/teams-js/src/internal/internalAPIs.ts
erinha: void?

packages/teams-js/src/public/media.ts
TrevorJoelHarris: I agree with Erin. The ultimate goal of this change should be to make the public, user facing function of MediaController.stop() return a Promise and throw an exception if an error occurs. In order to continue supporting back compat with v1 we need to have multiple signatures for stop (which is a pattern they follow in other places in this code base. As Erin pointed out, selectMedia is one example of this pattern.

I think the signature we'd like to add for stop is something like:

public stop(): Promise<void>

And then notifyEventToHost and isApiSupportedOnMobile should both throw SdkErrors instead of returning them

Happy to talk more if it would be helpful!

packages/teams-js/src/public/media.ts
```diff
@@ -412,14 +412,17 @@ export namespace media {
...
-          callback(err);
-        }
-        return;
+      let err;
+      try {
+        isApiSupportedOnMobile(nonFullScreenVideoModeAPISupportVersion);
+      } catch (e) {
+        err = e;
+        return callback(err);
       }
```
TrevorJoelHarris, line 426: After the other changes I mentioned, you shouldn't need to catch here. We want this exception to bubble up and be thrown to the app developer because they are trying to call something on a platform where it isn't supported

TrevorJoelHarris:
I would expect to see new or updated unit tests for the stop() function after this change

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -63,13 +63,13 @@ export function isHostClientMobile(): boolean {
  *
  * @internal
  */
-export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): SdkError {
+export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): never | null {
   if (!isHostClientMobile()) {
     const notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
-    return notSupportedError;
+    throw notSupportedError;
   } else if (!isAPISupportedByPlatform(requiredVersion)) {
```
TrevorJoelHarris, line 70: >isAPISupportedByPlatform [](http://example.com/codeflow?start=14&length=24)

This is not a request for you to change this here (unless it's easy) but writing this will help me hold in my head what this function is actually doing (because what this function is doing does not match the name of the function).

This function would be more correctly named something like "isCurrentSDKVersionAtLeast"

packages/teams-js/src/internal/internalAPIs.ts
TrevorJoelHarris: I've really been racking my brain. Wracking my brain? Trying to figure out how to fix this function. It's very small but it's also very misleading. Here is the best idea that I have come up with:

1) Change name to isMobileApiSupported
2) Change return value to boolean
3) Change line 74 to return true

I *think* the changed name would more closely match what the function is actually doing (validating that you are currently running on a mobile platform AND that the call you're trying to make will work on the current version  of the SDK the user is using), AND it makes the function named like it is returning a boolean return a boolean.

I still don't love that it only returns true OR throws, but I think this at least will make it a little better. Open to other suggestions though

TrevorJoelHarris (changes requested):
:clock1:

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -63,13 +63,13 @@ export function isHostClientMobile(): boolean {
  *
  * @internal
  */
-export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): SdkError {
+export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): never | null {
```
TrevorJoelHarris, line 66: I left a comment here in CodeFlow but I don't see it in git, so I will add it here again. if it's a double comment I'm sorry!

I've really been racking my brain. Wracking my brain? Trying to figure out how to fix this function. It's very small but it's also very misleading. Here is the best idea that I have come up with:

1) Change name to isMobileApiSupported
2) Change return value to boolean
3) Change line 74 to return true

I think the changed name would more closely match what the function is actually doing (validating that you are currently running on a mobile platform AND that the call you're trying to make will work on the current version  of the SDK the user is using), AND it makes the function named like it is returning a boolean return a boolean.

I still don't love that it only returns true OR throws, but I think this at least will make it a little better. Open to other suggestions though

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -65,14 +64,15 @@ export function decodeAttachment(attachment: media.MediaChunk, mimeType: string)
...
+    return isMobileApiSupported(videoAndImageMediaAPISupportVersion);
   } else if (isMediaCallForNonFullScreenVideoMode(mediaInputs)) {
-    return isApiSupportedOnMobile(nonFullScreenVideoModeAPISupportVersion);
+    return isMobileApiSupported(nonFullScreenVideoModeAPISupportVersion);
   } else if (isMediaCallForImageOutputFormats(mediaInputs)) {
-    return isApiSupportedOnMobile(imageOutputFormatsAPISupportVersion);
+    return isMobileApiSupported(imageOutputFormatsAPISupportVersion);
   }
+
   return null;
```
TrevorJoelHarris, line 76: >null [](http://example.com/codeflow?start=9&length=4)

this should be true

TrevorJoelHarris:
>  * Function returns null if the media call is supported on current mobile version, else SdkError.

Need to update the comment

---
Refers to: packages/teams-js/src/internal/mediaUtil.ts:63 in d1eba25. [](commit_id = d1eba25d848f2badcb7946c77daffd98b57414c8, deletion_comment = False)

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -37,7 +37,7 @@ export function ensureInitialized(...expectedFrameContexts: string[]): void {
  *
  * @internal
  */
-export function isAPISupportedByPlatform(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
+export function isCurrentSDKVersionAtLeast(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
```
TrevorJoelHarris, line 40: >isCurrentSDKVersionAtLeast [](http://example.com/codeflow?start=16&length=26)

thanks for renaming this! It makes the lower function easier to read

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +410,38 @@
      * @param mediaEvent indicates what the event that needs to be signaled to the host client
```
TrevorJoelHarris, line 410: >isSupported [](http://example.com/codeflow?start=10&length=11)

this should be !isSupported

TrevorJoelHarris: (and lines 415 and 416 could be combined into:

if (!isMobileApiSupported(nonFullScreenVideoModeAPISupportVersion))

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +410,38 @@
...
      * Optional; @param callback is used to send app if host client has successfully handled the notification event or not
      */
-    protected notifyEventToHost(mediaEvent: MediaControllerEvent, callback?: (err?: SdkError) => void): void {
```
TrevorJoelHarris, line 415: >callback(isSupported) [](http://example.com/codeflow?start=17&length=21)

What does it mean to return a boolean to this callback? The reason the callback existed was to pass the error back to the caller I think. Passing a false here won't do much (and won't ever get hit because isMobileApiSupported throws in the false case now).

TrevorJoelHarris: I think in order to maintain backcompatibility, we still want the callback to take in SdkErrors if the callback is present. In this function, if the callback is specified we would need to catch the exception from isMobileApiSupported and then pass it to the callback. If there was no callback, instead we just let isMobileApiSupported's exception keep bubbling up

TrevorJoelHarris: (Alternatively you could keep this function the way it is but with no callback at all, and then do the catch and pass to callback if callback is defined in the stop function. That might be cleaner?)

packages/teams-js/test/public/media.spec.ts
```diff
@@ -508,67 +508,56 @@ describe('media', () => {
...
+        expect(message).not.toBeNull();
+        expect(message.args.length).toBe(1);
+        const callbackId = message.id;
+        mobilePlatformMock.respondToMessage({
+          data: {
+            id: callbackId,
+            args: [err],
+          },
+        } as DOMMessageEvent);
+        await expect(mediaError).rejects.toEqual({ errorCode: ErrorCode.INTERNAL_ERROR });
```
TrevorJoelHarris, line 560: >rejects [](http://example.com/codeflow?start=33&length=7)

What does it mean to use rejects when evaluating a boolean? I thought rejects was for promises

TrevorJoelHarris: Actually I think this whole line doesn't work because you're expecting a boolean to equal an SdkError

packages/teams-js/test/public/media.spec.ts
```diff
@@ -508,67 +508,56 @@ describe('media', () => {
...
-          expect(message.args.length).toBe(1);
-
-          const callbackId = message.id;
-          mobilePlatformMock.respondToMessage({
-            data: {
-              id: callbackId,
-              args: [undefined],
-            },
-          } as DOMMessageEvent);
+      it('videoController notifyEventToHost should fail in default version of platform', async () => {
```
TrevorJoelHarris, line 511: >it('videoController notifyEventToHost should fail in default version of platform', async () => { [](http://example.com/codeflow?start=6&length=96)

I don't see any tests where you are testing the new stop() signature with no callback

packages/teams-js/src/public/media.ts
erinha: I agree. I think because this function is called internally within the `stop` function, we can implement this function the way we want it (no callback) and handle the callback compatibility in the caller (`stop`).

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +410,38 @@ export namespace media {
...
-      sendMessageToParent('media.controller', [params], (err?: SdkError) => {
+
+      sendMessageToParent('media.controller', [params], (err?: boolean) => {
         if (callback) {
           callback(err);
         }
       });
     }
 
+    public stop(): Promise<void>;
```
erinha, line 432: This version of `stop` needs documentation comments above it (I looked at `selectMedia` and if that was your example (since I suggested it as an example elsewhere ;)), it also needs documentation comments for it).

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +410,38 @@ export namespace media {
...
     /**
+     * @deprecated
+     * As of 2.0.0-beta.1, please use {@link MediaController.stop MediaController.stop(): Promise\<void\>} instead.
```
erinha, line 435: This will be 2.0.0-beta.3

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -63,15 +63,15 @@ export function isHostClientMobile(): boolean {
  *
  * @internal
  */
-export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): SdkError {
+export function isMobileApiSupported(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
```
erinha, line 66: Okay, don't kill me, but after reading through how this is used, I'm not sure this name is clearer. I think that the name `isApiSupportedOnMobile` is clearer than `isMobileApiSupported`. `isMobileApiSupported` makes me think oh, this API that is calling this is only for mobile, so check if that mobile-only API is supported. But if the function is supported on other platforms (and unfortunately it's not clear to me whether that is the case or not), then I don't think this is a better name.

packages/teams-js/src/public/media.ts
TrevorJoelHarris: I think adding comments to selectMedia might be a good example of something that is out of scope for Jess's PR. Totally agree that the function needs comments but at some point we need to stop pulling on the thread :)

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -58,20 +58,20 @@ export function isHostClientMobile(): boolean {
...
-    return notSupportedError;
-  } else if (!isAPISupportedByPlatform(requiredVersion)) {
+    throw notSupportedError;
+  } else if (!isCurrentSDKVersionAtLeast(requiredVersion)) {
     const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
-    return oldPlatformError;
+    throw oldPlatformError;
   }
-  return null;
+  return undefined;
```
TrevorJoelHarris, line 74: I think you can just delete this return statement entirely. Since it's a void function there's no need

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -58,20 +58,20 @@ export function isHostClientMobile(): boolean {
...
  * @hidden
  * Helper function which indicates if current API is supported on mobile or not.
- * @returns SdkError if host client is not android/ios or if the requiredVersion is not
+ * @throws SdkError if host client is not android/ios or if the requiredVersion is not
  *          supported by platform or not. Null is returned in case of success.
  *
  * @internal
  */
-export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): SdkError {
+export function throwExceptionIfApiIsNotSupported(requiredVersion: string = defaultSDKVersionForCompatCheck): void {
```
TrevorJoelHarris, line 66: I realize you've changed the name of this function a lot, and I'm sorry that we've been going back and forth on it a lot. I think this name is much closer to correct, but what if we did "throwExceptionIfMobileApiIsNotSupported". The function seems intended to verify both that you are on a mobile device AND that the SDK version matches what you say is required.

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -62,18 +61,19 @@ export function decodeAttachment(attachment: media.MediaChunk, mimeType: string)
 /**
  * @hidden
  * Function returns null if the media call is supported on current mobile version, else SdkError.
```
TrevorJoelHarris, line 63: This comment should be updated since it is no longer accurate

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -62,18 +61,19 @@ export function decodeAttachment(attachment: media.MediaChunk, mimeType: string)
...
   } else if (isMediaCallForNonFullScreenVideoMode(mediaInputs)) {
-    return isApiSupportedOnMobile(nonFullScreenVideoModeAPISupportVersion);
+    return throwExceptionIfApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
   } else if (isMediaCallForImageOutputFormats(mediaInputs)) {
-    return isApiSupportedOnMobile(imageOutputFormatsAPISupportVersion);
+    return throwExceptionIfApiIsNotSupported(imageOutputFormatsAPISupportVersion);
   }
-  return null;
+
+  return;
```
TrevorJoelHarris, line 76: Can get of this return, not necessary since this is a void function

TrevorJoelHarris:
>         }

Do we need to throw here if callback is not defined, and if so do we need or have a unit tests that tests for that case?

---
Refers to: packages/teams-js/src/public/media.ts:432 in e444bb0. [](commit_id = e444bb02b094d1d914d3cce1769c7830ec0b49e1, deletion_comment = False)

packages/teams-js/src/public/media.ts
```diff
@@ -414,15 +416,16 @@ export namespace media {
...
+          return callback(err);
         }
-        return;
       }
 
       const params: MediaControllerParam = { mediaType: this.getMediaType(), mediaControllerEvent: mediaEvent };
+
       sendMessageToParent('media.controller', [params], (err?: SdkError) => {
         if (callback) {
           callback(err);
```
TrevorJoelHarris, line 431: Do we need to throw here if callback is undefined? I assume if we don't the Promise version of the function won't fail correctly in this case. If we do need to make that change we should add corresponding unit tests

packages/teams-js/src/public/media.ts
```diff
@@ -414,15 +416,16 @@ export namespace media {
      */
     protected notifyEventToHost(mediaEvent: MediaControllerEvent, callback?: (err?: SdkError) => void): void {
       ensureInitialized(FrameContexts.content, FrameContexts.task);
-      const err = isApiSupportedOnMobile(nonFullScreenVideoModeAPISupportVersion);
-      if (err) {
+      try {
+        throwExceptionIfApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
+      } catch (err) {
         if (callback) {
```
TrevorJoelHarris, line 422: If callback isn't defined this will fail silently. I would think we would only want the try catch here if callback was defined, and otherwise just call the throw function directly and let it throw up to the caller

packages/teams-js/src/public/media.ts
```diff
@@ -431,11 +434,31 @@ export namespace media {
...
-      this.notifyEventToHost(MediaControllerEvent.StopRecording, callback);
+    public stop(callback: (err?: SdkError) => void): void;
+    public stop(callback?: (err?: SdkError) => void): Promise<void> {
+      if (callback) {
+        // if there is a callback we want to pass the callback to the notifyevent to host fn - if an exception is caught, it gets passed to the callback
+        const inputFn = (): Promise<void> =>
+          new Promise<void>(resolve => {
+            resolve(this.notifyEventToHost(MediaControllerEvent.StartRecording, callback));
+          });
+        return callCallbackWithErrorOrResultFromPromiseAndReturnPromise(inputFn, callback);
```
TrevorJoelHarris, line 457: Most of the other instances in our code base like this have one call to callCallbackWithErrorOrResultFromPromiseAndReturnPromise at the end of the function and then they pass the wrapped function and the callback to that. Is there a reason that that's not the pattern we're following here? I think callCallbackWithErrorOrResultFromPromiseAndReturnPromise has logic for handling the if (callback) scenario

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +414,42 @@
...
-      if (err) {
-        if (callback) {
-          callback(err);
-        }
-        return;
-      }
+
+      const promise = Promise.resolve().then(() =>
+        throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion),
+      );
```
TrevorJoelHarris, line 422: >return err; [](http://example.com/codeflow?start=10&length=11)

shouldn't we be throwing here?

packages/teams-js/src/public/media.ts
TrevorJoelHarris: Curious to hear why you decided not to go with this pattern? I don't think it's wrong necessarily, but does mean that we're not following the same pattern as a lot of the other backcompatible public functions.

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +414,42 @@
...
```
TrevorJoelHarris, line 423: Do we return Promises if we hit these cases? And do we need to throw on line 430?

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +414,42 @@ export namespace media {
...
+     * As of 2.0.0-beta.3, please use {@link MediaController.stop MediaController.stop(): Promise\<void\>} instead.
+     
      * Function to programatically stop the ongoing media event
      * Optional; @param callback is used to send app if host client has successfully stopped the event or not
      */
-    public stop(callback?: (err?: SdkError) => void): void {
-      this.notifyEventToHost(MediaControllerEvent.StopRecording, callback);
+    public stop(callback: (err?: SdkError) => void): void;
+    public stop(callback?: (err?: SdkError) => void): Promise<void> {
+      const inputFn = (): Promise<void> => this.notifyEventToHost(MediaControllerEvent.StartRecording, callback);
```
TrevorJoelHarris, line 451: Curious to hear why you decided not to go with the callCallbackWithErrorOrResultFromPromiseAndReturnPromise  pattern from other back compatible public functions? I don't think what you've done is wrong necessarily, but it does mean that we're not following the same pattern as other similar functions. How come?

TrevorJoelHarris: This is my bad, Jess is doing the right thing here but since inputFn wasn't named the same as the other similar variables I missed it.

packages/teams-js/test/public/media.spec.ts
```diff
@@ -838,11 +822,58 @@ describe('media', () => {
...
+        expect(message.args.length).toBe(1);
+        const callbackId = message.id;
+        mobilePlatformMock.respondToMessage({
+          data: {
+            id: callbackId,
+            args: [err],
+          },
+        } as DOMMessageEvent);
+        expect(stopPromise).resolves;
       });
```
TrevorJoelHarris, line 874: I think this test should be making sure the right exception comes out of stop when it resolves

packages/teams-js/test/public/media.spec.ts
```diff
@@ -838,11 +822,58 @@ describe('media', () => {
...
+        mobilePlatformMock.respondToMessage({
+          data: {
+            id: callbackId,
+            args: [undefined],
+          },
+        } as DOMMessageEvent);
+        expect(stopPromise).resolves;
+      });
+
+      it('videoController notifyEventToHost is not handled successfully but stop resolves', async () => {
```
TrevorJoelHarris, line 856: I think this test should be making sure the right exception comes out of stop when it resolves

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +414,52 @@ export namespace media {
...
+     
      * Function to programatically stop the ongoing media event
      * Optional; @param callback is used to send app if host client has successfully stopped the event or not
      */
-    public stop(callback?: (err?: SdkError) => void): void {
-      this.notifyEventToHost(MediaControllerEvent.StopRecording, callback);
+    public stop(callback: (err?: SdkError) => void): void;
+    public stop(callback?: (err?: SdkError) => void): Promise<void> {
+      const wrappedFunction = (): Promise<void> =>
+        this.notifyEventToHost(MediaControllerEvent.StartRecording, callback);
```
ndangudubiyyam, line 461: is it supposed to be MediaControllerEvent.StopRecording?
Also, since the stop is just calling the notifyEventToHost, you don't have to wrap it again and call the utility function.
```
public stop(callback?: (err?: SdkError) => void): Promise<void> {
return Promise.resolve(this.notifyEventToHost(MediaControllerEvent.StopRecording, callback)
}
```

packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +414,51 @@ export namespace media {
...
         }
       });
+
+      return promise
+        .then(() => {
+          if (mediaError) {
+            throw mediaError;
+          }
+        })
+        .catch(err => {
```
TrevorJoelHarris, line 440: nitpicky, but I don't think you need to catch and rethrow here, I believe if you just remove the catch the same error will be thrown to the caller

TrevorJoelHarris (approved):
:shipit:

packages/teams-js/test/public/media.spec.ts
```diff
@@ -488,131 +474,102 @@ describe('media', () => {
...
-      });
-
-      it('videoController notifyEventToHost should fail in default version of platform', () => {
-        return mobilePlatformMock.initializeWithContext(FrameContexts.content, HostClientType.android).then(() => {
-          mobilePlatformMock.setClientSupportedSDKVersion(originalDefaultPlatformVersion);
-          let error;
-          new media.VideoController().stop((e: SdkError) => {
-            error = e;
-          });
+        media.selectMedia(mediaInputs, (error: SdkError, attachments: media.Media[]) => {
```
ndangudubiyyam, line 485: nit: You are testing the callback.,so it should be placed user v1

PR #874, opened 2022-01-27
private/communication.spec.ts Error Validation Test

jekloudaMSFT:


packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -5,7 +5,7 @@ import * as utils from '../../src/internal/utils';
 describe('communication', () => {
   it('initializeCommunication should reject if no parent window and no native interface found', () => {
     const initPromise = communication.initializeCommunication(undefined);
-    return expect(initPromise).rejects.toThrowError('Initialization Failed. No Parent window found.');
+    return expect(initPromise).rejects.toMatchObject(new Error('Initialization Failed. No Parent window found.'));
```
lakhveerk, line 8: internal API is just throwing an error
`throw new Error('The library has not yet been initialized'); `

Why do we prefer toMatchObject over just toThrowError? 

Jest docs says that:
Use .toMatchObject to check that a JavaScript object matches a subset of the properties of an object. It will match received objects with properties that are not in the expected object.

ndangudubiyyam: I agree with @lakhveerk  on this
the unit test is written correctly. No need to change it.
https://jestjs.io/docs/expect#rejects

lakhveerk: I just realized that error is produced from 
`return Promise.reject(new Error('Initialization Failed. No Parent window found.'));` 
@ndangudubiyyam  can we use `.toThrowError` even when error was produced by `promise.reject` (and not by `throw new error`)?

jekloudaMSFT: Based on this doc:
https://jestjs.io/docs/asynchronous#resolves--rejects
I got the impression that `.rejects` returns the rejected value so I would expect to use something like `.toMatch` whereas `.toThrow` should be called on a function, but that doc does use `rejects.toThrow`. Both patterns work as far as I can tell, so we should just pick one we prefer.

TrevorJoelHarris (approved):

PR #875, opened 2022-01-27
Added open stage view API

richardle798:
Adding the openStageView API.  The design doc can be found here: https://microsoft-my.sharepoint-df.com/:w:/p/rile/ETE2H4uz1tFEpkM7PctZYZwBC2JxxHPkegcYi_vYaAAUVA?e=vTn4cc

### DIFF
--- a/src/public/stageView.ts
+++ b/src/public/stageView.ts
@@ -56,9 +56,9 @@ export namespace stageView {
     ensureInitialized(FrameContexts.content);
 
     if (!stageViewParams) {
-      throw new Error('[open] Stage view params cannot be null');
+      throw new Error('[stageView.open] Stage view params cannot be null');
     }
 
-    sendMessageToParent('openStageView', [stageViewParams], callback);
+    sendMessageToParent('stageView.open', [stageViewParams], callback);
   }
 }
--- a/src/public/stageView.ts
+++ b/src/public/stageView.ts
@@ -68,21 +68,13 @@ export namespace stageView {
    *                 The callback takes as an argument an SdkError in case something happened (i.e.
    *                 no permissions to execute the API)
    */
-  export function openStageView(
-    applicationId: string,
-    context: string,
-    callback?: (sdkError?: SdkError) => void,
-  ): void {
+  export function openStageView(stageViewParams: StageViewParams, callback?: (sdkError?: SdkError) => void): void {
     ensureInitialized(FrameContexts.content, FrameContexts.stage);
 
-    if (!applicationId) {
-      throw new Error('[openStageView] Application ID cannot be null');
+    if (!stageViewParams) {
+      throw new Error('[openStageView] Stage view params cannot be null');
     }
 
-    if (!context) {
-      throw new Error('[openStageView] Context cannot be null');
-    }
-
-    sendMessageToParent('openStageView', [applicationId, context], callback);
+    sendMessageToParent('openStageView', [stageViewParams], callback);
   }
 }
--- a/src/public/stageView.ts
+++ b/src/public/stageView.ts
@@ -8,6 +8,56 @@ import { FrameContexts } from './constants';
  */
 export namespace stageView {
   /**
+   * Parameters to open a stage view.
+   */
+  export interface StageViewParams {
+    /**
+     * The application ID of the Teams application to be opened.
+     */
+    appId: string;
+
+    /**
+     * The context passed into the stage view.
+     */
+    context: {
+      /**
+       * The URL of the content to display.
+       */
+      contentUrl?: string;
+
+      /**
+       * The Teams app website URL.
+       */
+      websiteUrl?: string;
+
+      /**
+       * The name of the stage view.
+       */
+      name?: string;
+
+      /**
+       * The entity ID.
+       */
+      entityId?: string;
+
+      /**
+       * The thread that initiated the request.
+       */
+      threadId?: string;
+
+      /**
+       * The initatiator of the stage view request.
+       */
+      source?: string;
+
+      /**
+       * The title to give the stage view.
+       */
+      title?: string;
+    };
+  }
+
+  /**
    * @private
    * Feature is under development
    *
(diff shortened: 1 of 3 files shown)


src/public/stageView.ts
```diff
@@ -0,0 +1,38 @@
...
+   * Feature is under development
+   *
+   * Opens a stage view to display a Teams app
+   * @param applicationId The ID of the Teams app
+   * @param context The context required to launch the Teams app
+   * @param callback Callback that will be triggered once the stage view is closed.
+   *                 The callback takes as an argument an SdkError in case something happened (i.e.
+   *                 no permissions to execute the API)
+   */
+  export function openStageView(
```
MarcinJuraszek, line 21: It seems very similar to what dialog/task module is, so maybe we should try to have the API be similar too? `stageView.open` to align with `dialog.open`?

src/public/stageView.ts
```diff
@@ -0,0 +1,38 @@
...
+    applicationId: string,
+    context: string,
```
MarcinJuraszek, line 23: What is this exactly representing? What kind of string should be provided here? From the doc, it seems to be a JSON-serialized object. That's not great. Let's try to keep this strongly typed instead.

src/public/stageView.ts
```diff
@@ -0,0 +1,82 @@
...
+   * Feature is under development
+   *
+   * Opens a stage view to display a Teams app
+   * @param stageViewParams The parameters to pass into the stage view.
+   * @param callback Callback that will be triggered once the stage view is closed.
+   *                 The callback takes as an argument an SdkError in case something happened (i.e.
+   *                 no permissions to execute the API)
+   */
+  export function openStageView(stageViewParams: StageViewParams, callback?: (sdkError?: SdkError) => void): void {
+    ensureInitialized(FrameContexts.content, FrameContexts.stage);
```
ydogandjiev, line 74: Please remove FrameContext.stage from the `ensureInitialized` call since this API shouldn't be called from within the stage.

src/public/stageView.ts
```diff
@@ -0,0 +1,82 @@
...
+   * @private
+   * Feature is under development
+   *
+   * Opens a stage view to display a Teams app
+   * @param stageViewParams The parameters to pass into the stage view.
+   * @param callback Callback that will be triggered once the stage view is closed.
+   *                 The callback takes as an argument an SdkError in case something happened (i.e.
+   *                 no permissions to execute the API)
+   */
+  export function openStageView(stageViewParams: StageViewParams, callback?: (sdkError?: SdkError) => void): void {
```
ydogandjiev, line 73: @TrevorJoelHarris, @MarcinJuraszek, @erinha, so what do we wanna call this API? Should it just be `open`? Also, do we want to leave the namespace as is or change it to something simpler - e.g. `modal` or `stage`?

src/public/stageView.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * The initatiator of the stage view request.
+     * Delete seems unused
+     */
+    source?: string;
+
+    /**
+     * The title to give the stage view.
+     * Used by the error dialog.  Should this be required?
```
ydogandjiev, line 58: What is the error dialog you're referring to here? Can we just use the `name` in that as well? Can you please share some examples of real world values for the `name` and `title`? When are these different?

src/public/stageView.ts
```diff
@@ -0,0 +1,82 @@
...
+     * The chat or channel ID.
+     * Not sure
+     */
+    threadId?: string;
+
+    /**
+     * The initatiator of the stage view request.
+     * Delete seems unused
+     */
+    source?: string;
```
ydogandjiev, line 54: Please delete as per our discussion

ydogandjiev (changes requested):

ydogandjiev (approved):

src/public/stageView.ts
```diff
@@ -0,0 +1,64 @@
...
+     */
+    title: string;
+
+    /**
+     * The Teams app website URL.
+     */
+    websiteUrl?: string;
+
+    /**
+     * The entity ID of the Teams app.
```
ydogandjiev, line 40: This should say something like "The entity ID of the Teams app content being opened."

ydogandjiev (approved):

src/public/stageView.ts
TrevorJoelHarris: Open is good, I agree with Marcin's comment above.

I suspect that eventually we will want to call this something other than stageView, but since we will still have the opportunity to change this up later as part of merging into 2.0 and I don't feel confident that modal is *better* than stageView, I am fine with leaving as is.

test/public/stageView.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    websiteUrl: 'websiteUrl',
+    entityId: 'entityId',
+  };
+
+  describe('open', () => {
+    it('should not allow calls before initialization', () => {
+      expect(() => stageView.open(stageViewParams)).toThrowError('The library has not yet been initialized');
+    });
+
+    it('should not allow calls from stage context', () => {
```
TrevorJoelHarris, line 39: Can you unit test the other frame contexts that are not supported as well? Should be pretty straightforward to write as a for loop, let me know if you have trouble.

test/public/stageView.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    websiteUrl: 'websiteUrl',
+    entityId: 'entityId',
+  };
+
+  describe('open', () => {
+    it('should not allow calls before initialization', () => {
+      expect(() => stageView.open(stageViewParams)).toThrowError('The library has not yet been initialized');
+    });
+
+    test.each([['settings'], ['sidePanel'], ['authentication'], ['remove'], ['task'], ['meetingStage'], ['stage']])(
```
TrevorJoelHarris, line 39: How would you feel about modifying this slightly to be more like:

 Object.keys(FrameContexts)
      .map(k => FrameContexts[k])
      .forEach(frameContext => {
        it(`should not allow calls from ${frameContext} context`, async () => {

The reason I like this better is that it will catch issues in the future in case FrameContexts change or new ones are added.

TrevorJoelHarris: (Or you might be able to replace your inner array with something like tests.each([Object.keys(FrameContexts)]))

TrevorJoelHarris (approved):

PR #877, opened 2022-01-27
public/call.spec.ts Error Validation Tests

jekloudaMSFT:


packages/teams-js/test/public/call.spec.ts
```diff
@@ -22,13 +22,17 @@ describe('call', () => {
...
+  it('should not allow calls before initialization', async () => {
+    expect.assertions(1);
+    await call
+      .startCall(mockStartCallParams)
+      .catch(e => expect(e).toMatchObject(new Error('The library has not yet been initialized')));
   });
 
-  it('shoud not allow calls if not supported', () => {
+  it('shoud not allow calls if not supported', async () => {
+    expect.assertions(1);
```
Jessica-Koch, line 33: Silly question, is there any reason why in one test you use `.toMatchObject` and another you use `.toEqual`

Jessica-Koch: I know from just having looked it up, that they are ever so slightly different

Jessica-Koch: i think they'd both be fine, i'm just curious

Jessica-Koch (dismissed):
:shipit:

packages/teams-js/test/public/call.spec.ts
```diff
@@ -22,13 +22,17 @@ describe('call', () => {
     }
   });
 
-  it('should not allow calls before initialization', () => {
-    return expect(call.startCall(mockStartCallParams)).rejects.toThrowError('The library has not yet been initialized');
```
lakhveerk, line 26: image

The above code looks correct to me. Attached is the pic from official jest doc. 
Are we changing it for code readability?

packages/teams-js/test/public/call.spec.ts
jekloudaMSFT: Good call, we should stick to one for consistency. Just taking a quick look, `.toEqual` seems more correct for what we're testing, but I think we could go either way

packages/teams-js/test/public/call.spec.ts
jekloudaMSFT: Mentioned this on the `communication.spec.ts` PR, but I thought that `.rejects` returned the rejected value based on https://jestjs.io/docs/asynchronous#resolves--rejects, so I thought it needed to be treated as an object rather than a function. These tests will work either way, so we should just pick an approach and be consistent with it. I don't have a strong preference here.

erinha: As mentioned offline, we're looking into getting consistent guidance here. In the meantime, can you confirm that if the test fails, it reports failure correctly (e.g., showing a red X in VS Code with the Jest plugin, reporting failed test in the command line summary, etc.)?

jekloudaMSFT: Sorry for the delay, but I have confirmed that if the test fails it is reported correctly in both implementations.

shrshindeMSFT: **We have two different ways to go for it but there is a difference as some functions don't return a promise.**

**Function which returns Promise the code we can use either as below:**

`try {
      await call.startCall(mockStartCallParams);
    } catch (e) {
      expect(e).toEqual(new Error('The library has not yet been initialized'));
    }`
or 

`await expect(call.startCall(mockStartCallParams)).rejects.toThrowError('The library has not yet been initialized');`

**Function which does not return Promise we need to use the try method one as using await will make no difference.**
This is the function which I've been working on it's in **_public/teamsAPI.spec.ts_**
`     try {
        teamsCore.registerOnLoadHandler(() => {
          return false;
        });
      } catch (e) {
        expect(e).toEqual(new Error('The library has not yet been initialized'));
      }`

For code consistency I'd recommend **try** method.

packages/teams-js/test/public/call.spec.ts
```diff
@@ -22,17 +22,16 @@ describe('call', () => {
...
+  it('should not allow calls before initialization', async () => {
+    await expect(call.startCall(mockStartCallParams)).rejects.toThrowError('The library has not yet been initialized');
   });
 
-  it('shoud not allow calls if not supported', () => {
+  it('shoud not allow calls if not supported', async () => {
```
TrevorJoelHarris, line 29: should

packages/teams-js/test/public/call.spec.ts
TrevorJoelHarris: This is good discussion! Erin recently published some internal docs about how we want to test this case and why, and it matches what we have here: we prefer async/await to return "because it more closely resembles how app developers would call the function".

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

PR #878, opened 2022-01-27
public/pages.spec.ts Error Validation Tests

jekloudaMSFT:


Jessica-Koch (dismissed):

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -89,52 +116,42 @@ describe('AppSDK-TeamsAPIs', () => {
...
+        'The library has not yet been initialized',
+      );
     });
 
-    it('should not allow calls with a bad origin', async () => {
-      expect.assertions(1);
-      await pages
-        .navigateCrossDomain('https://badorigin.com')
-        .catch(e => expect(e).toMatchObject(new Error('The library has not yet been initialized')));
+    // Skipping these tests as url validation is not implemented
```
TrevorJoelHarris, line 124: Isn't this test testing the same thing as the previous one?

TrevorJoelHarris: I see, so I think you converted tests that were already here and in place. In my opinion, might be better to comment out the tests for features that don't work yet and then open bugs for this work if there aren't already. We can then maybe add the bug number(s) in the comments here?

jekloudaMSFT: Sounds good. I commented out the tests and created bug 5790202 for this issue

TrevorJoelHarris (approved):

PR #880, opened 2022-01-28
Combine our yaml files

aeunms:
First step in taking our numerous yaml files and combining them into one. This PR will allow for us to not only avoid depending on several pipeline runs but also keep the jobs parallelized and efficient.

A few notable/potentially confusing points explained:
 - Newly introduced pipeline variables `$(AppHostingSdkGitPath2)`, `$(AppHostingSdkGitPath3)`, etc. (Example: Line 317)
   - These `$(AppHostingSdkGitPath#)` variables actually contain the same value as the original `$(AppHostingSdkGitPath)` variable. This is the best getaround I've found to YAML's issue which doesn't allow for checking out the same repository defined by the same variable more than once in the same file. 
 - If we're combining the yaml files, why aren't the old ones being deleted?
   - Before this PR is merged, all of this test content still needs to be run on all other PRs to 2.0-preview, so those GitHub checks have to stay. And since those GitHub checks have to stay, this PR won't pass the checks unless those files are present. After this PR is merged, I will turn off the GitHub checks that utilize those files and then submit a PR to delete the outdated files. This multi-step process will allow us to have no 'gap' where the test content isn't being covered.
 - Lots of tasks that are "duplicated" in the e2e test jobs
   - I tried to have them depend on a job that would `yarn install` and `build` the teamsJs library and the appHostSdk library to avoid the duplication, but all these things have to be done specifically by the job/agent running each test. 
 - Tons of commits
   - Please ignore them, I had to test a ton with our pipeline. Also, they'll be squashed :)

Update 1: 
 - Commented out the v1 test steps. Will uncomment after PR when hub sdk is ready and include it in new PR

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -291,13 +291,14 @@ jobs:
         inputs:
           Arguments: 'build'
           ProjectDirectory: 'microsoft-teams-library-js'
-          - task: CmdLine@2
-            displayName: 'Configure host machine'
-            inputs:
-              script: |
-                sudo chmod -R 755 ./ 
-                sudo yarn setup
-              workingDirectory: '$(AppHostingSdkProjectDirectory)'
+
+      - task: CmdLine@2
+        displayName: 'Configure host machine'
+        inputs:
+          script: |
+            sudo chmod -R 755 ./ 
+            sudo yarn setup
+          workingDirectory: '$(AppHostingSdkProjectDirectory)'
 
       - bash: 'node tools/cli/runAppsWithE2ETests.js'
         displayName: 'Run E2E integration tests'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,11 +14,6 @@ trigger:
     include:
       - 'v*'
 
-steps:
-  - checkout: self
-  - checkout: git://$(AppHostingSdkGitPath)@$(AppHostingSdkGitRef)
-    persistCredentials: true
-
 jobs:
   - job: Security
     displayName: 'Security Tasks'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -232,19 +248,253 @@ jobs:
...
+          testResultsFiles: '**/e2e-tests-report*.xml'
+        condition: succeededOrFailed()
+
+  - job: E2e_Performance_Test
+    displayName: 'E2e performance test'
+    pool:
+      vmImage: 'ubuntu-latest'
+    steps:
+      - checkout: self
+      - checkout: git://$(AppHostingSdkGitPath2)
```
ramourt, line 318: I think the duplicate variables are fine. 
Maybe worth opening a very low-pri bug for it. Will just save people effort to have to add a new variable if they ever need to add more e2e tests.

ramourt (dismissed):
:shipit:

azure-pipelines.yml
```diff
@@ -232,19 +248,241 @@ jobs:
...
+        inputs:
+          Arguments:
+          ProjectDirectory: '$(ClientSdkProjectDirectory)'
+
+      - task: Yarn@2
+        displayName: 'Build client sdk'
+        inputs:
+          Arguments: 'build'
+          ProjectDirectory: '$(ClientSdkProjectDirectory)'
+
```
Ella-ly, line 414: Missing one build here: 
`  - task: Yarn@2
    displayName: 'Build Test App Local'
    inputs:
      Arguments: 'build-test-app-local'
      ProjectDirectory: '$(ClientSdkProjectDirectory)'`

aeunms: Great catch!

Ella-ly (changes requested):
Please add `Build Test App Local build` before merging.

Ella-ly (approved):

azure-pipelines.yml
aeunms: Just created a bug under Untriaged for it! I've sent it to you in case you want to take a look. (But you don't have to.)

PR #882, opened 2022-01-28
add code comment to select media method

Jessica-Koch:


packages/teams-js/src/public/media.ts
```diff
@@ -597,6 +597,11 @@ export namespace media {
     assembleAttachment: AssembleAttachment[];
   }
 
+  /**
+   * Select an attachment using camera/gallery
+   *
+   * @returns  A promise resolved with an array of media data or rejected with an @see SdkError
```
erinha, line 603: Thanks for turning this out quickly! Can you add `@param` for mediaInputs please?

Jessica-Koch: omg of course, sorry i'm still not used to the double documentation

erinha (approved):
:shipit:

PR #883, opened 2022-01-31
Turn eslint warnings into errors

aeunms:
This work will uphold eslint rules to be errors instead of warnings so anyone merging in code has to ensure that their code changes don't introduce eslint warnings. This will keep our repo nice and tidy and eliminate the workload involved in upkeep. 

There were already some eslint warnings in the repo that needed fixing along with changes for the package.json scripts run by the GitHub checks that should be updated, so this PR is a bit big.


apps/teams-perf-test-app/webpack.config.js
```diff
@@ -43,5 +43,6 @@ module.exports = {
     https: true,
     disableHostCheck: true,
   },
+  performance: { hints: false },
```
AE-MS, line 46: Is this a new behavior we want?

apps/teams-perf-test-app/package.json
```diff
@@ -5,9 +5,9 @@
   "description": "Teams Perf Test App utilizing Teams JavaScript client SDK to test Teams Host Perf",
   "version": "0.0.1",
   "scripts": {
-    "build": "yarn eslint ./src/**/*.tsx && webpack",
+    "build": "yarn lint && webpack",
     "clean": "rimraf ./build",
-    "eslint-fix-build": "yarn eslint --fix --ext .tsx ./src && webpack",
+    "lint": "yarn eslint ./src --max-warnings 0 --fix --ext .tsx",
```
AE-MS, line 10: >tsx [](http://example.com/codeflow?start=61&length=3)

I know this isn't new (and this question applies to the other package.json files in which it appears) -- but does this cause us to not lint .TS files also?

apps/teams-test-app/src/components/LocationAPIs.tsx
```diff
@@ -55,6 +55,7 @@ const ShowLocation = (): React.ReactElement =>
           return 'Completed';
         },
         withCallback: (locationProps, setResult) => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
AE-MS, line 58: >  // eslint-disable-next-line @typescript-eslint/no-unused-vars [](http://example.com/codeflow?start=8&length=63)

Not saying you have to fix it as part of this change, but should we actually be using the status variable in this test app? E.g. putting it into the result somewhere?
(you can always file a bug to fix it separately if so)

AE-MS (changes requested):
:clock1:

apps/teams-perf-test-app/package.json
aeunms: I left it as .tsx for the app package.jsons since all they have are .tsx files and it throws an error saying no files matching the pattern ".ts" were found.

apps/teams-perf-test-app/webpack.config.js
aeunms: Yes. We had a thread going in our Engineering channel about how we expect the Teams package to be larger than webpack is used to, so we concluded to turn the hints off about performance. If this makes you feel too icky, we can also totally explore setting a higher number that webpack should warn us at, although my issue was that I couldn't decide on just what number we should be setting it at.

apps/teams-test-app/src/components/LocationAPIs.tsx
aeunms: Good idea. I initially thought no because this function is deprecated (I wasn't sure we wanted to allocate more resources towards something we're not maintaining) and I couldn't find any documentation on what exactly the status means here. (succeeded = true maybe?) But after thinking more on it thanks to your comment ( :smile: ) I realize all I would have to do is add to the string message displayed. I'll be making that change. As for understanding what exactly the status means, I imagine it's probably just true = good and false = something went wrong...?

apps/teams-perf-test-app/package.json
Jessica-Koch: does the script accept a conditional like `(ts|tsx)$`

aeunms: Hmm I think it might be possible, I tested a few ways to use it and that didn't work though. (`(.ts|.tsx)`, `(.ts|.tsx)$`, `$(.ts|.tsx)`)

AE-MS: Seems crazy that you can't allow for future expansion into a different file extension without causing the error you say but 🤷‍♂️

Does .ts* work?  :)

apps/teams-test-app/src/components/LocationAPIs.tsx
AE-MS: Is there some automated test that is going to fail because the string is different now? I'm assuming if there was it would have failed on your PR with the change... does that mean there's no automation that calls and tests this?

apps/teams-perf-test-app/package.json
aeunms: I tried variations using those ( moving the argument order around, etc.) but I can't seem to make it work. Maybe I can open a bug item so we're prepared for any .ts files coming in?

AE-MS: I don't think I would open a bug for it since it's not really actionable now and I don't think it's going to cause anyone who happens to add a TS file to remember to do something. I guess we just have to leave it like this for now.

tools/bundle-size-tools/src/utilities/getChunkParsedSize.ts
```diff
@@ -10,7 +10,7 @@ import { WebpackStatsJson } from '../BundleBuddyTypes';
  */
 export function getChunkParsedSize(stats: WebpackStatsJson, chunkId: string | number): number {
   if (stats.assets === undefined) {
-    throw new Error(`No assets property in the stats file, can't compute parsed sizes of chunks`);
+    throw new Error("No assets property in the stats file, can't compute parsed sizes of chunks");
```
erinha, line 13: This looks like the only place we use double quotes. Something we need to tweak somewhere to keep it consistently single quotes?

apps/teams-perf-test-app/package.json
aeunms: I was thinking maybe they could figure out how to add the .ts extension without there being ts files in the mo 😆

apps/teams-test-app/src/components/LocationAPIs.tsx
aeunms: This was a great question! (I missed thinking about e2e.) It seems it passes the tests, although I'm not sure that's intended. I looked at the tests in the app host sdk side and I'm not quite sure when it utilizes the callback version here vs the promises version. For now I'm thinking I'll actually revert my latest commit to use the eslint-disable comment and create a ticket as you mentioned we don't necessarily have to fix it in the PR- and then I can grab the ticket and work on that one next. What do you think?

tools/bundle-size-tools/src/utilities/getChunkParsedSize.ts
aeunms: That's a good point. I think I should be able to include that in this PR.

apps/teams-test-app/src/components/LocationAPIs.tsx
AE-MS: Sounds good.

AE-MS (approved):

tools/bundle-size-tools/src/utilities/getChunkParsedSize.ts
aeunms: Ok I finally found out what's happening here. We have avoidEscape set to true in the root eslintrc.js which makes it so strings can use single or double quotes so long as the string contains a quote that would have to be escaped otherwise. Do we want to change this behavior?

aeunms: [Official docs](https://eslint.org/docs/2.0.0/rules/quotes) for more details

aeunms: @erinha friendly bump on this question :)

erinha: Sorry I missed your reply! That makes sense, this is fine to leave as it is.

erinha (approved):

PR #884, opened 2022-02-01
Jekoch/is api supported on mobile update

Jessica-Koch:


packages/teams-js/src/public/media.ts
```diff
@@ -412,30 +414,47 @@ export namespace media {
...
+     
      * Function to programatically stop the ongoing media event
      * Optional; @param callback is used to send app if host client has successfully stopped the event or not
      */
-    public stop(callback?: (err?: SdkError) => void): void {
-      this.notifyEventToHost(MediaControllerEvent.StopRecording, callback);
+    public stop(callback: (err?: SdkError) => void): void;
+    public stop(callback?: (err?: SdkError) => void): Promise<void> {
+      const wrappedFunction = (): Promise<void> => this.notifyEventToHost(MediaControllerEvent.StopRecording, callback);
+      return callCallbackWithErrorOrResultFromPromiseAndReturnPromise(wrappedFunction, callback);
```
ndangudubiyyam, line 457: This implementation is wrong. you are calling the callback once in the notifyEventfunction, then again in the utility function.
Also why bother wrapping up with the utility function, which is btw wrong, it should be `callCallbackWithSdkErrorFromPromiseAndReturnPromise`

Just return the resolved Promise.
```
public stop(callback?: (err?: SdkError) => void): Promise<void> {
      return Promise.resolve(this.notifyEventToHost(MediaControllerEvent.StopRecording, callback));
    }
```

packages/teams-js/test/public/media.spec.ts
```diff
@@ -488,131 +474,102 @@ describe('media', () => {
...
+          expect(mediaError).toEqual({ errorCode: ErrorCode.OLD_PLATFORM });
         });
       });
-
-      it('videoController notifyEventToHost is not handled successfully', () => {
-        return mobilePlatformMock.initializeWithContext(FrameContexts.content, HostClientType.android).then(() => {
-          mobilePlatformMock.setClientSupportedSDKVersion(nonFullScreenVideoModeAPISupportVersion);
-          let mediaError: SdkError;
-          new media.VideoController().stop((e: SdkError) => {
+      it('videoController notifyEventToHost is handled successfully', async () => {
```
ndangudubiyyam, line 499: I will review the test cases again after you fix the notifyEvent function

PR #885, opened 2022-02-01
Adding prerelease yml to master

ramourt:
This will add the manually triggerable workflow to master. The workflow on master will not really do anything now. 
However for the same workflow in 2.0-preview to become available for invoking, we need to add this file to master.

For context, we are creating these workflows in 2.0-preview to facilitate the release process, and automate all manual steps that would need to be done by developers when a new version is to be released. More details in #865

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -1,0 +1,17 @@
+# This needs to be added in the master branch to show the workflow fir all branches on GitHub
+# Running this on master will not do anything
+name: Initiate Release Workflow
+on:
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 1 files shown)


.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,17 @@
+# This needs to be added in the master branch to show the workflow fir all branches on GitHub
```
erinha, line 1: typo: ...show the workflow **for** all...

ramourt: Fixed!

erinha (approved):

ydogandjiev (approved):

PR #886, opened 2022-02-01
Using yml templates for readability

ramourt:


azure-pipelines.yml
```diff
@@ -14,281 +16,19 @@ trigger:
...
-        condition: and(
-          eq(variables['Build.Reason'], 'PullRequest'),
-          eq(variables['System.PullRequest.TargetBranch'], '2.0-preview'))
-        inputs:
-          scanType: 'LogOnly'
-          verbosity: 'Verbose'
-          alertWarningLevel: 'High'
-          failOnAlert: true
-        displayName: 'Component Governance (Pull Request)'
+  - template: tools/yaml/security.yml
```
erinha, line 19: Do you think there is value in storing the yaml templates in a subdirectory called templates? Relatedly, is there a problem if any of the template yaml files (like security.yml) is used directly or in a certain way, and if so, would we be more likely to avoid using it in a bad way if it were more clearly obvious that it's a template? 

I'm new to templates, so trying to figure out if we should be more obvious about their use.

ramourt: That is a great question!

The main reason I'm doing this is simply syntactical (we had an earlier discussion about being able to merge into one yaml file without a cost of readability since our yaml file is like 500 lines now). Templates solve this, at the risk of then being reused incorrectly if we ever need to make more yaml pipelines and reuse templates. 

I don't know if we'll really ever need to make a new yaml pipeline, so it really depends on what another pipeline is trying to do if it can then reuse the steps here. I think if that ever happens, based on new pipeline requirements, we can edit the templates to then fit both pipelines. (ie security.yml and build-test-publish.yml are jobs files, whereas build-app-host is a steps file, and you can see how they are used different in the azure-pipelines.yml file now)

I doubt the 1st two templates I mentioned are ever going to be reused. The only reason I made them templates is to make the main file easier to read. I am ok with keeping them inside however.
The nice thing about the build-app-host.yml however is it shortens all the copied steps for building the app host before running e2e tests. So the template build-app-host does all the checking out and building both repos, and then we just add a script afterwards in each job to run the specific e2e test.

PR #890, opened 2022-02-02
Reorganizing yaml with templates

ramourt:
This change is mostly to allow the azure-pipelines.yml pipeline file become more readable. 

This is done with templates. All the templates are now in the tools folder under a directory called yaml-templates. They are written as a set of steps, allowing them to be injected into any pipeline job and extended if needed. 
- security.yml: should be used in any other new pipelines we have that are linked to this repo. This does all the Guardian Security Checks and any Component Governance checks.
- build-test-publish.yml: is not really expected to be reused in its current state as it is very specific to validating build/unit tests, and then publishing artifacts for release. If we want this to be reused later, we can make it such that it only does building and testing, and then explicitly add the PR comment & publishing steps into azure-pipelines.yml. If we ever need to add more pipelines, we can make that decision then.
- build-app-host.yml: This does all the steps for cloning both app sdk and app host sdk repos and building them, then configuring the app host. It is then reused in 3 (soon 4) e2e test jobs that will run this template and then adds a node script with different parameters for which type of e2e test that needs to be run. I have modified the commented out e2e test to also use templates so it can be uncommented once it is ready.

**Known issues**: There is an issue with the checkout step for the appHostSdk git path where using the same variable in the checkout causes the pipeline to break. We have a bug for this, and the only workaround at the time being is to create more variables in the pipeline containing the same value and use each one as a parameter for each e2e test.


azure-pipelines.yml
```diff
@@ -19,276 +19,26 @@ jobs:
...
-
-      - task: PublishBuildArtifacts@1
-        inputs:
-          PathtoPublish: '$(Build.ArtifactStagingDirectory)\NPMFeed'
-          ArtifactName: 'NPMFeed'
-        displayName: 'Publish NPM feed to Build Artifacts'
-
-  - job: E2e_Test
-    displayName: 'E2e test'
+  - job: E2ETest1
```
erinha, line 31: Does the order the jobs are listed in have any impact on which tests start (or finish) first?

azure-pipelines.yml
```diff
@@ -425,53 +87,14 @@ jobs:
           testResultsFormat: 'JUnit'
           testResultsFiles: '**/e2e-tests-report*.xml'
         condition: succeededOrFailed()
-  # - job: E2e_Test_TeamsV1
-  #   displayName: 'E2e test TeamsJS V1'
+  # - job: E2ETest4
```
erinha, line 90: I think this is now enabled in 2.0-preview (a merge should bring that change in).

azure-pipelines.yml
aeunms: I can answer this one: Unless there is a dependsOn/conditional property, the jobs are run in parallel.

erinha (changes requested):
The comments in your description about what each template yml file is for are super useful! Can you please put them in each template file in comments at the top so we don't lose this info?

azure-pipelines.yml
ramourt: Fixed

ramourt:
> The comments in your description about what each template yml file is for are super useful! Can you please put them in each template file in comments at the top so we don't lose this info?

Done

erinha (approved):
Looks good, thanks!

PR #891, opened 2022-02-02
Added in E2E test for Backwards Compatibility

jadahiya-MSFT:
Added in an E2E test for testing callback/backwards compatibility mode against latest version of the sdk


ramourt (approved):
lgtm

PR #898, opened 2022-02-03
Jekoch/throwing errors

Jessica-Koch:
* deprecated notifyEventToHost in media.ts
* deprecated stop in media.ts
* renamed utility functions to better describe their new behavior
* modified utility functions to throw sdk errors rather than return objects containing error codes

### DIFF
--- a/packages/teams-js/src/internal/internalAPIs.ts
+++ b/packages/teams-js/src/internal/internalAPIs.ts
@@ -37,7 +37,7 @@ export function ensureInitialized(...expectedFrameContexts: string[]): void {
  *
  * @internal
  */
-export function isAPISupportedByPlatform(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
+export function isCurrentSDKVersionAtLeast(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
   const value = compareSDKVersions(GlobalVars.clientSupportedSDKVersion, requiredVersion);
   if (isNaN(value)) {
     return false;
--- a/packages/teams-js/src/internal/internalAPIs.ts
+++ b/packages/teams-js/src/internal/internalAPIs.ts
@@ -58,20 +58,21 @@ export function isHostClientMobile(): boolean {
 /**
  * @hidden
  * Helper function which indicates if current API is supported on mobile or not.
- * @returns SdkError if host client is not android/ios or if the requiredVersion is not
+ * @throws SdkError if host client is not android/ios or if the requiredVersion is not
  *          supported by platform or not. Null is returned in case of success.
  *
  * @internal
  */
-export function isApiSupportedOnMobile(requiredVersion: string = defaultSDKVersionForCompatCheck): SdkError {
+export function throwExceptionIfMobileApiIsNotSupported(
+  requiredVersion: string = defaultSDKVersionForCompatCheck,
+): void {
   if (!isHostClientMobile()) {
     const notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
-    return notSupportedError;
-  } else if (!isAPISupportedByPlatform(requiredVersion)) {
+    throw notSupportedError;
+  } else if (!isCurrentSDKVersionAtLeast(requiredVersion)) {
     const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
-    return oldPlatformError;
+    throw oldPlatformError;
   }
-  return null;
 }
 
 /**
--- a/packages/teams-js/src/internal/mediaUtil.ts
+++ b/packages/teams-js/src/internal/mediaUtil.ts
@@ -1,4 +1,3 @@
-import { SdkError } from '../public/interfaces';
 import { media } from '../public/media';
 import { people } from '../public/people';
 import {
@@ -67,11 +66,11 @@ export function decodeAttachment(attachment: media.MediaChunk, mimeType: string)
  */
 export function throwExceptionIfMediaCallIsNotSupportedOnMobile(mediaInputs: media.MediaInputs): void {
   if (isMediaCallForVideoAndImageInputs(mediaInputs)) {
-    return throwExceptionIfMobileApiIsNotSupported(videoAndImageMediaAPISupportVersion);
+    throwExceptionIfMobileApiIsNotSupported(videoAndImageMediaAPISupportVersion);
   } else if (isMediaCallForNonFullScreenVideoMode(mediaInputs)) {
-    return throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
+    throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
   } else if (isMediaCallForImageOutputFormats(mediaInputs)) {
-    return throwExceptionIfMobileApiIsNotSupported(imageOutputFormatsAPISupportVersion);
+    throwExceptionIfMobileApiIsNotSupported(imageOutputFormatsAPISupportVersion);
   }
 }
 
--- a/packages/teams-js/src/internal/mediaUtil.ts
+++ b/packages/teams-js/src/internal/mediaUtil.ts
@@ -6,7 +6,7 @@ import {
   nonFullScreenVideoModeAPISupportVersion,
   videoAndImageMediaAPISupportVersion,
 } from './constants';
-import { isApiSupportedOnMobile } from './internalAPIs';
+import { throwExceptionIfMobileApiIsNotSupported } from './internalAPIs';
 
 /**
  * @hidden
@@ -61,19 +61,18 @@ export function decodeAttachment(attachment: media.MediaChunk, mimeType: string)
 
 /**
  * @hidden
- * Function returns null if the media call is supported on current mobile version, else SdkError.
- *
+ * Function throws an SdkError if the media call is not supported on current mobile version, else undefined.
+ * @throws an SdkError if the media call is not supported
  * @internal
  */
-export function isMediaCallSupportedOnMobile(mediaInputs: media.MediaInputs): SdkError {
+export function throwExceptionIfMediaCallIsNotSupportedOnMobile(mediaInputs: media.MediaInputs): void {
   if (isMediaCallForVideoAndImageInputs(mediaInputs)) {
-    return isApiSupportedOnMobile(videoAndImageMediaAPISupportVersion);
+    return throwExceptionIfMobileApiIsNotSupported(videoAndImageMediaAPISupportVersion);
   } else if (isMediaCallForNonFullScreenVideoMode(mediaInputs)) {
-    return isApiSupportedOnMobile(nonFullScreenVideoModeAPISupportVersion);
+    return throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
   } else if (isMediaCallForImageOutputFormats(mediaInputs)) {
-    return isApiSupportedOnMobile(imageOutputFormatsAPISupportVersion);
+    return throwExceptionIfMobileApiIsNotSupported(imageOutputFormatsAPISupportVersion);
   }
-  return null;
 }
 
 /**
--- a/packages/teams-js/src/public/media.ts
+++ b/packages/teams-js/src/public/media.ts
@@ -412,34 +412,69 @@ export namespace media {
      * @hidden
      * Hide from docs
      * --------
+     *
+     * Function to notify the host client to programatically control the experience
+     * @param mediaEvent indicates what the event that needs to be signaled to the host client
+     * @returns A promise resolved promise
+     */
+    protected notifyEventToHost(mediaEvent: MediaControllerEvent): Promise<void>;
+    /**
+     * @hidden
+     * Hide from docs
+     * --------
+     *
+     * @deprecated
+     * As of 2.0.0-beta.3, please use {@link media.MediaController.notifyEventToHost media.MediaController.notifyEventToHost(mediaEvent: MediaControllerEvent): Promise\<void\>} instead.
+     *
      * Function to notify the host client to programatically control the experience
      * @param mediaEvent indicates what the event that needs to be signaled to the host client
      * Optional; @param callback is used to send app if host client has successfully handled the notification event or not
      */
-    protected notifyEventToHost(mediaEvent: MediaControllerEvent, callback?: (err?: SdkError) => void): void {
+    protected notifyEventToHost(mediaEvent: MediaControllerEvent, callback?: (err?: SdkError) => void): void;
+    protected notifyEventToHost(mediaEvent: MediaControllerEvent, callback?: (err?: SdkError) => void): Promise<void> {
       ensureInitialized(FrameContexts.content, FrameContexts.task);
-      const err = throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
-      if (err) {
-        if (callback) {
-          callback(err);
-        }
-        return;
-      }
 
-      const params: MediaControllerParam = { mediaType: this.getMediaType(), mediaControllerEvent: mediaEvent };
-      sendMessageToParent('media.controller', [params], (err?: SdkError) => {
-        if (callback) {
-          callback(err);
-        }
-      });
+      return Promise.resolve()
+        .then(() => {
+          return throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
+        })
+        .then(() => {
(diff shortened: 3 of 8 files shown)


packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -37,7 +37,7 @@ export function ensureInitialized(...expectedFrameContexts: string[]): void {
  *
  * @internal
  */
-export function isAPISupportedByPlatform(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
+export function isCurrentSDKVersionAtLeast(requiredVersion: string = defaultSDKVersionForCompatCheck): boolean {
```
jadahiya-MSFT, line 40: Not sure about this naming convention but not sure what else to name it

packages/teams-js/src/public/media.ts
```diff
@@ -408,34 +412,69 @@ export namespace media {
...
+        .then(() => {
+          const params: MediaControllerParam = { mediaType: this.getMediaType(), mediaControllerEvent: mediaEvent };
+
+          sendMessageToParent('media.controller', [params], (err?: SdkError) => {
+            if (callback) {
+              callback(err);
+            } else if (err) {
+              throw err;
+            }
+          });
```
jadahiya-MSFT, line 450: I'm kind of confused by this, couldn't we do something like 
```
return Promise.resolve()
  .then(() => {
         const err = isMobileApiSupported(nonFullScreenVideoModeAPISupportVersion);
          if(err){
              throw err;
           }
         const params: MediaControllerParam = {...}
         ...  
    }
    .catch
```

jadahiya-MSFT: this way we don't have to use .then().then().catch()

packages/teams-js/src/internal/internalAPIs.ts
jadahiya-MSFT: what would happen if someone calls `isCurrentSDKVersionAtLeast`  with an older version of the SDK passed into `requiredVersion`, I think that would break the naming convention here

packages/teams-js/src/public/media.ts
Jessica-Koch: you can't assign the error to a variable because it is throwing errors now

Jessica-Koch: i can definitely put comments in there though because i was very confused by this function originally

packages/teams-js/src/internal/internalAPIs.ts
Jessica-Koch: these are internal functions, so this shouldn't break anything, but I am new to the backwards  compat stuff so i could be wrong

packages/teams-js/src/public/media.ts
Jessica-Koch: plus, this is handling the old use case but with promises, we need the function to exit after the first then if it throws an error, and go to the catch. This was something i was confused about in the original implementation

packages/teams-js/src/internal/internalAPIs.ts
jadahiya-MSFT: Oh I didn't mean actually breaking anything, I meant that the name wouldn't make sense in that situation

packages/teams-js/src/public/media.ts
jadahiya-MSFT: Hmm so it's not possible to propagate the Error up to this call and then throw it? Basically just switching the callback in the original implementation to a throw

packages/teams-js/src/public/media.ts
```diff
@@ -408,34 +412,69 @@ export namespace media {
...
 
-      const params: MediaControllerParam = { mediaType: this.getMediaType(), mediaControllerEvent: mediaEvent };
-      sendMessageToParent('media.controller', [params], (err?: SdkError) => {
-        if (callback) {
-          callback(err);
-        }
-      });
+      return Promise.resolve()
+        .then(() => {
+          return throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
```
ndangudubiyyam, line 439: This revision looks much better :) 
You don't need the `return` before the `throwExceptionIfMobileApiIsNotSupported`, it will go to catch block anyway.

Bonus Tip: As people have commented below, there the utility functions which hide the ugly code if callback stuff. I would suggest if you can use any existing utility functions (eg:`callCallbackWithSdkErrorFromPromiseAndReturnPromise`), your code will be much cleaner. Else you can write up your own utility function, so it will help incase same scenarios pop up.

packages/teams-js/test/public/media.spec.ts
```diff
@@ -669,6 +517,31 @@ describe('media', () => {
...
+        expect(message.args.length).toBe(1);
+        const callbackId = message.id;
+        mobilePlatformMock.respondToMessage({
+          data: {
+            id: callbackId,
+            args: [undefined, undefined, 2],
+          },
+        } as DOMMessageEvent);
+        expect(mediaError).toBeFalsy();
+        expect(jest.fn()).not.toHaveBeenCalled();
```
ndangudubiyyam, line 543: You haven't assigned the callback insde the selectMedia call to the jest.fn(). Correct it

packages/teams-js/test/public/media.spec.ts
```diff
@@ -815,9 +688,11 @@ describe('media', () => {
         expect(mediaAttachment.size).not.toBeNull();
         expect(typeof mediaAttachment.size === 'number').toBeTruthy();
         expect(mediaAttachment.getMedia).toBeDefined();
+        await expect(promise).resolves;
```
ndangudubiyyam, line 691: I think .resolves takes a matcher like .toBe or something the promise that is expected to resolve with some result.
So line 691 is redundant and doesn't test anything.

packages/teams-js/test/public/media.spec.ts
```diff
@@ -838,7 +713,146 @@ describe('media', () => {
           },
         } as DOMMessageEvent);
 
-        return expect(promise).rejects.toEqual({ errorCode: ErrorCode.SIZE_EXCEEDED });
+        await expect(promise).rejects.toMatchObject({ errorCode: ErrorCode.SIZE_EXCEEDED });
```
ndangudubiyyam, line 716: I think the new standard is to use .toEqual for comparing SdkError objects.

packages/teams-js/test/public/media.spec.ts
```diff
@@ -940,10 +954,54 @@ describe('media', () => {
         return blobReadingPromise;
       }
 
-      it('getMedia calls with successful result via the handler', done => {
+      it('getMedia calls with successful result via the handler', async () => {
```
ndangudubiyyam, line 957: This is a success test case, so the callback is being called. For the rest of the failure cases, we are checking with `done`(). I won't change the format just because it is a success case, I would want to keep it consistent for both success and failure cases.

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: oh, i think i just moved this code, i will fix it, but my next task is actually to make sure all the tests in this file are running as appropriate

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: you're right!

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: this is another piece of code that i just moved, but didn't change, I would be more than happy to discuss with you how you think these tests in this file should be structured.

packages/teams-js/src/public/media.ts
Jessica-Koch: it is, it's in the catch, you can't write something like if(err) because you can't assign it to a variable, it will always be void.  so i have to split that function into a couple thens because if the first then fails, it has to immediately exit the function, if it doesn't it continues, but the callback could have an error, and since we don't want to swallow that case, we need to call the callback with the error, but throw it in the case where there is no callback.  

i'd be more than happy to discuss this offline, i had to have it explained to me a couple of times :)

packages/teams-js/src/internal/internalAPIs.ts
TrevorJoelHarris: I think this name makes sense doesn't it? If GlobalVars.clientSupportedSDKVersion == 2.0 and requiredVersion == 1.0, the function would return true because the current version was at least 1.0. I suppose it's just a different way of saying "GreaterThanOrEqualTo".

I guess that's my way of saying "I think this name makes sense but if people want to pick an even clearer name it's okay with me".

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -61,19 +61,18 @@ export function decodeAttachment(attachment: media.MediaChunk, mimeType: string)
...
- *
+ * Function throws an SdkError if the media call is not supported on current mobile version, else undefined.
+ * @throws an SdkError if the media call is not supported
  * @internal
  */
-export function isMediaCallSupportedOnMobile(mediaInputs: media.MediaInputs): SdkError {
+export function throwExceptionIfMediaCallIsNotSupportedOnMobile(mediaInputs: media.MediaInputs): void {
   if (isMediaCallForVideoAndImageInputs(mediaInputs)) {
-    return isApiSupportedOnMobile(videoAndImageMediaAPISupportVersion);
+    return throwExceptionIfMobileApiIsNotSupported(videoAndImageMediaAPISupportVersion);
```
TrevorJoelHarris, line 70: These returns are unnecessary now that throwExceptionIfMobileApiIsNotSupported returns void, can just call that function.

(I don't think the return statements are hurting anything, just unnecessary.)

packages/teams-js/src/public/media.ts
TrevorJoelHarris: Agreed, this revision looks really good.

(this is another return that we don't technically need, but it isn't hurting anything)

packages/teams-js/test/public/media.spec.ts
```diff
@@ -655,11 +505,9 @@ describe('media', () => {
             expect(mediaAttachments).toBeFalsy();
             expect(mediaError.errorCode).toBe(ErrorCode.SIZE_EXCEEDED);
```
TrevorJoelHarris, line 506: @erinha This might be an example of testing for an SDK error thrown from synchronous code. We were trying to find examples of this yesterday for the unit testing guidelines you were writing

packages/teams-js/test/public/media.spec.ts
```diff
@@ -655,11 +505,9 @@ describe('media', () => {
             expect(mediaAttachments).toBeFalsy();
```
TrevorJoelHarris, line 505: Do we need to use expect.assertions in this function to make sure that the expect statements inside of the callback are hit? (I'm not sure off the top of my head, legit question)

packages/teams-js/test/public/media.spec.ts
TrevorJoelHarris, line 716: Based on the unit test guidelines Erin wrote up yesterday, we prefer to use toEqual instead of toMatchObject for validating SdkError objects.

"toMatchObject  will only check the object's properties that are passed to the toMatchObject function, so in general, prefer greater specificity with toEqual"

packages/teams-js/test/public/media.spec.ts
```diff
@@ -838,7 +713,146 @@ describe('media', () => {
...
+          data: {
+            id: callbackId,
+            args: [undefined],
+          },
+        } as DOMMessageEvent);
+
+        expect(sendMessageToParentSpy).toHaveBeenCalled();
+      });
+
+      it('videoController notifyEventToHost is not handled successfully', async () => {
```
TrevorJoelHarris, line 747: A slightly more descriptive name will help people understand what test is failing if it ever does. Maybe something like "stop returns SdkError to callback when parent rejects message"?

packages/teams-js/test/public/media.spec.ts
```diff
@@ -838,7 +713,146 @@ describe('media', () => {
...
-        return expect(promise).rejects.toEqual({ errorCode: ErrorCode.SIZE_EXCEEDED });
+        await expect(promise).rejects.toMatchObject({ errorCode: ErrorCode.SIZE_EXCEEDED });
+      });
+    });
+  });
+
+  describe('videoController', () => {
+    describe('v1', () => {
+      it('videoController notifyEventToHost is handled successfully', async () => {
+        expect.assertions(7); // initializeWithContext has 3 assertions + 4 in this test = 7
```
TrevorJoelHarris, line 724: I love these comments. It makes it a lot easier to understand where the hidden assertions are. (Not a huge fan of hidden assertions but not something you introduced here so no need to go fix it as part of this change)

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: not sure what happened here but you're right, if we do it one place, we should do it everywhere and the code hadn't even been completely converted

packages/teams-js/test/public/media.spec.ts
```diff
@@ -838,7 +713,146 @@ describe('media', () => {
...
+      it('videoController notifyEventToHost should fail in default version of platform and should exit early', async () => {
+        expect.assertions(5);
+
+        await mobilePlatformMock.initializeWithContext(FrameContexts.content, HostClientType.android);
+        mobilePlatformMock.setClientSupportedSDKVersion(originalDefaultPlatformVersion);
+
+        const sendMessageToParentSpy = jest.spyOn(communication, 'sendMessageToParent');
+
+        try {
+          await new media.VideoController().stop(e => {
```
TrevorJoelHarris, line 786: I'm curious how this works. By looking at your code, it seems like if we pass the callback in then we always return the error through the callback (which is correct), but in this test it seems to be expecting and catching the exception on 790. Given that you are expecting 5 assertions on 778 (3 in initializeWithContext and 2 in this test) clearly it's getting hit I'm just not sure I understand why

packages/teams-js/src/internal/mediaUtil.ts
Jessica-Koch: they are unnecessary and you're right they aren't hurting anything but to avoid confusion i will remove them

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: lol, glad i can help

packages/teams-js/src/internal/internalAPIs.ts
Jessica-Koch: i explained this to @jadahiya-MSFT in teams and not here, I think if you know the parameters being passed in are '2.0' and not 'some random string', it totally makes sense.

packages/teams-js/src/public/media.ts
```diff
@@ -408,34 +412,69 @@ export namespace media {
...
+              callback(err);
+            } else if (err) {
+              throw err;
+            }
+          });
+        })
+        .catch(err => {
+          if (callback) {
+            callback(err);
+          }
```
TrevorJoelHarris, line 455: Do we need an else here to make sure that in the case where someone is passing in a callback we don't change the contract of the old function?

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: i agree, this confused me a LOT when i was writing the tests, I'm not sure why we have assertions in a mock, but that's something we can discuss outside of this PR

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: image
 this is where i got the information from. If i checked in the callback, it wouldn't wait for the callback to execute and the assertions it would expect would be 4.  When i would write that I expected it to reject, it would say it threw an error, but the test wouldn't pass, i'm actually not sure why this happened.

Jessica-Koch: when i was referring to testing the thrown error  that i am confused by, what i wrote was 
```
await expect(new media.VideoController().stop(e=>e))).rejects.toEqual({old platform error})
```

packages/teams-js/src/public/media.ts
Jessica-Koch: we discussed that since this is in a resolved promise and we want to propagate the error, we will leave this as is...this can be tabled for a future discussion

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: since i didn't touch this function, i can change this in the next ticket regarding fixing the tests in this file

TrevorJoelHarris (approved):

packages/teams-js/test/public/media.spec.ts
erinha: Can we change this to `expect(mediaError).toEqual({ ErrorCode.SIZE_EXCEEDED });` ?

PR #899, opened 2022-02-03
Added missing callback for registerFocusEnterHandler in test app

Ella-ly:


apps/teams-test-app/src/components/PagesAPIs.tsx
```diff
@@ -126,12 +127,21 @@ const RegisterFocusEnterHandler = (): React.ReactElement =>
...
+          return true;
+        });
+        return 'registered';
+      },
+      withCallback: setResult => {
+        registerFocusEnterHandler(navigateForward => {
+          setResult('successfully called with navigateForward:' + navigateForward);
+          return true;
+        });
+        return 'registered';
```
jadahiya-MSFT, line 143: Will this output correctly or does this need to be setResult('registered')?

Ella-ly: I guess I need to clean up my mind, thanks!

MarcinJuraszek (approved):

PR #903, opened 2022-02-04
Added ipadOS to HostClientType

Vinodh-G:
Feature: https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/1774162
iPad Experience 
As a part of above feature we are sending the hostClientType via getContext as "ipados" if the client device is iPad.
Hence adding the same constant on the enum HostClientType.

### DIFF
--- a/src/public/constants.ts
+++ b/src/public/constants.ts
@@ -3,6 +3,7 @@ export enum HostClientType {
   web = 'web',
   android = 'android',
   ios = 'ios',
+  ipados = 'ipados',
   /**
    * @deprecated Use teamsRoomsWindows instead.
    */


src/public/constants.ts
```diff
@@ -3,6 +3,7 @@ export enum HostClientType {
   web = 'web',
   android = 'android',
   ios = 'ios',
+  ipados = 'ipados',
```
aeunms, line 6: As per [camelCase guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) and for readability, I would recommend changing the naming from 'ipados' to 'iPadOS'

aeunms (changes requested):

src/public/constants.ts
Vinodh-G: @aeunms the client in ios is already sending the getcontext info for hostClientType as ipados for client type iPad, and it's been used by a lot of platform developers since this is in ring 4, and also the naming convention for ios should be iOS instead of ios.
Can we go ahead now?

I will work on follow up PR, changing both getContext hostclientType to ios -> iOS and ipados -> iPadOS on the client code and parallelly raise PR here as well, and this requires documentation update communications to all the developers.

ydogandjiev: While I do agree with @aeunms's feedback above, I think it's too late to rename this now since it would be a significant breaking change for developers. We could choose to rename the enum items themselves but leave the values with v2 of the SDK since that's our only upcoming chance to make a breaking change from a developer perspective. But the underlying values being sent from the Teams Clients shouldn't change since that would break any apps that have taken a dependence on them.

@Vinodh-G, in the future please make sure to make the App/Client SDK changes first before implementing them in the Mobile Clients. That would allow feedback like this to be provided before the code gets written.

ydogandjiev (approved):

src/public/constants.ts
TrevorJoelHarris: Agree 100% with Yuri. Will approve this change because it's just changing a type that already exists and is in use. In the future, please make these types of changes in the App SDK first to avoid running into this kind of hold up.

TrevorJoelHarris (approved):

PR #907, opened 2022-02-04
cleaning up old release files

ramourt:
These are some old files used for releasing that we don't need anymore.


aeunms (approved):
LGTM!

erinha (approved):
Yay! I love when files are deleted. 😆

PR #910, opened 2022-02-05
public/sharing.spec.ts Refactoring initializeContext 

shrshindeMSFT:
Updated `initializeContext('content')` to `initializeContext(FramContexts.content)` to restrict the code behavior to **FrameContexts.content**.


packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -1,10 +1,12 @@
 import { app } from '../../src/public/app';
 import { ErrorCode } from '../../src/public/interfaces';
+import { FrameContexts } from '../../src/public/constants';
 import { sharing } from '../../src/public/sharing';
 import { Utils } from '../utils';
 
 describe('sharing_v1', () => {
   const utils = new Utils();
+  const allowedContext = ['content', 'sidePanel', 'task', 'stage', 'meetingStage'];
```
aeunms, line 9: Why is it that only these five out of the eight FrameContexts are allowed in this test?

shrshindeMSFT: In sharing uses the above context and rest 3 are not allowed.
_The below code is from the sharing.ts_
 ` ensureInitialized(
      FrameContexts.content,
      FrameContexts.sidePanel,
      FrameContexts.task,
      FrameContexts.stage,
      FrameContexts.meetingStage,
    );`
    
   Also, unit test cases messages are misleading so to avoid the confusion I have limited the test cases for the allowed context.

packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -23,8 +25,31 @@ describe('sharing_v1', () => {
...
+            content: [
+              {
+                type: 'URL',
+                url: 'https://www.microsoft.com',
+                preview: true,
+                message: 'Test',
+              },
+            ],
+          };
+          expect(() => sharing.shareWebContent(shareRequest)).toThrowError(
```
aeunms, line 44: Maybe we can use the constant you made in the message? Using a template from Lakhveer's PR: 

```
 `This call is only allowed in following contexts: ${JSON.stringify(
            allowedContext,
          )}. Current context: "${context}".`,
```

shrshindeMSFT: Done

shrshindeMSFT: Done

packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -23,8 +25,33 @@ describe('sharing_v1', () => {
     }
   });
 
+  Object.keys(FrameContexts)
+    .map(key => FrameContexts[key])
```
aeunms, line 29: Nit: I think this is fine as is, but for future tests you could use Object.values(FrameContexts) without doing the extra mapping.

aeunms (dismissed):
LGTM, thanks :)

packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -358,12 +409,12 @@ describe('sharing_v2', () => {
 
     const promise = sharing.shareWebContent(shareRequest as any);
     const shareMessage = utils.findMessageByFunc(sharing.SharingAPIMessages.shareWebContent);
-    expect(shareMessage).toBeNull();
+    // expect(shareMessage).toBeNull();
```
erinha, line 412: Is this change intentional?

aeunms (dismissed):

packages/teams-js/test/public/sharing.spec.ts
erinha, line 9: Can we name this variable `allowedContexts` since it's an array of contexts?

packages/teams-js/test/public/sharing.spec.ts
erinha: Can you please update the message to be "This call is only allowed in the following following contexts:" ?

packages/teams-js/test/public/sharing.spec.ts
shrshindeMSFT: Sure.

packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -23,8 +25,33 @@ describe('sharing_v1', () => {
...
+              {
+                type: 'URL',
+                url: 'https://www.microsoft.com',
+                preview: true,
+                message: 'Test',
+              },
+            ],
+          };
+          expect(() => sharing.shareWebContent(shareRequest)).toThrowError(
+            `This call is only allowed in following contexts: ${JSON.stringify(
```
erinha, line 45: Do we have tests that validate that the `sharing.shareWebContent` function cannot be called from disallowed contexts? (The reverse of these ones)

packages/teams-js/test/public/sharing.spec.ts
erinha: Thanks! Please also update the same variable in the v2 tests.

packages/teams-js/test/public/sharing.spec.ts
shrshindeMSFT: Yes, we do already have them in place.

packages/teams-js/test/public/sharing.spec.ts
erinha: Actually, one further update. It would be useful to see in the test failure message which function is failing the test, e.g. "sharing.shareWebContent is only allowed in the following contexts:"

shrshindeMSFT: I was thinking to categorize all the test functions in describe block so that we would have some nice clarity.
For e.g.,
` describe('should test settings.getSettings function', () => {
    const allowedContexts = [
      FrameContexts.content,
      FrameContexts.settings,
      FrameContexts.remove,
      FrameContexts.sidePanel,
    ];
    const expectedSettings: settings.Settings = {
      suggestedDisplayName: 'someSuggestedDisplayName',
      contentUrl: 'someContentUrl',
      websiteUrl: 'someWebsiteUrl',
      entityId: 'someEntityId',
    };

      if (!allowedContexts.some(allowedContexts => allowedContexts == context)) {
        it(`should not allow calls from ${context} context`, async () => {
          expect(() => {
            settings.getSettings(settings => {
              expect(settings).toBe(expectedSettings);
            });
          }).toThrowError(
            `This call is only allowed in following contexts: ${JSON.stringify(
              allowedContexts,
            )}. Current context: "${context}".`,
          );
        });
      }
    });

    it('should successfully get settings', async () => {
      await utils.initializeWithContext('settings');
      settings.getSettings(settings => {
        expect(settings).toBe(expectedSettings);
      });

      const message = utils.findMessageByFunc('settings.getSettings');
      expect(message).not.toBeNull();

      utils.respondToMessage(message, expectedSettings);
    });
  });` 

Output: -
                                      should test settings.setValidityState function
@microsoft/teams-js:       √ should not allow calls from content context (58 ms)
@microsoft/teams-js:       √ should not allow calls from authentication context (4 ms)
@microsoft/teams-js:       √ should not allow calls from task context (3 ms)
@microsoft/teams-js:       √ should not allow calls from sidePanel context (2 ms)
@microsoft/teams-js:       √ should not allow calls from stage context (2 ms)
@microsoft/teams-js:       √ should not allow calls from meetingStage context (3 ms)
@microsoft/teams-js:       √ should successfully set validity state to true (3 ms)
@microsoft/teams-js:       √ should successfully set validity state to false (2 ms)

something on the above lines.

erinha: I love this idea!

erinha (changes requested):
I'd love to see the change you proposed by organizing the tests in describe groups.

packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -23,12 +31,335 @@ describe('sharing_v1', () => {
...
+            expect(() => sharing.shareWebContent(shareRequest)).toThrowError(
+              `This call is only allowed in following contexts: ${JSON.stringify(
+                allowedContexts,
+              )}. Current context: "${frameContext}".`,
+            );
+          });
+        }
+      });
+
+    it('sharing.shareWebContent should successfully call the callback function when given the share web content in correct format - success scenario', done => {
```
erinha, line 60: It looks like we only test the `sharing.shareWebContent` function success scenario with FrameContexts.content. Since we allow content, sidePanel, task, stage, and meetingStage, can you add the success scenario with those contexts as well please?

erinha (changes requested):
I love the improvements you've made here. One last request for testing success cases with other FrameContexts.

packages/teams-js/test/public/sharing.spec.ts
shrshindeMSFT: I have added it for both v1 and v2 please check the latest commit.

ramourt (approved):
lgtm

erinha (approved):
Looks great, thanks!

PR #911, opened 2022-02-05
meeting.spec.ts - Fixing asnyc tests and adding new unit tests for all context types (e.g. setting, side panel etc)

lakhveerk:


packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -23,142 +23,228 @@ describe('meeting', () => {
...
     });
 
-    it('should return error code 500', async () => {
-      await desktopPlatformMock.initializeWithContext('meetingStage');
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
+      }
+      it(`should successfully send the toggleIncomingClientAudio message. context: ${context}`, async () => {
+        await desktopPlatformMock.initializeWithContext(context);
```
shrshindeMSFT, line 52: In here are we using FramelessMock for desktop platform? Also, to be consistent we need to check if frameless is dedicated to mobile or desktop client.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -23,142 +23,228 @@ describe('meeting', () => {
...
-          id: callbackId,
-          args: [{ errorCode: ErrorCode.INTERNAL_ERROR }, null],
-        },
-      } as DOMMessageEvent);
-      await expect(promise).rejects.toEqual({ errorCode: ErrorCode.INTERNAL_ERROR });
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
+      }
+      it(`should successfully send the getIncomingClientAudio message. context: ${context}`, async () => {
```
shrshindeMSFT, line 118: Add specific platform name in the test case as it'll be helpful while debugging.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -23,142 +23,228 @@ describe('meeting', () => {
...
+        );
+      });
     });
 
-    it('should return error code 500', async () => {
-      await desktopPlatformMock.initializeWithContext('meetingStage');
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
+      }
```
erinha, line 50: See my comment above about changing the if statement and not using `return`

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -23,142 +23,228 @@ describe('meeting', () => {
...
     it('should not allow calls before initialization', () => {
       expect(() => meeting.toggleIncomingClientAudio()).toThrowError('The library has not yet been initialized');
     });
 
-    it('should successfully toggle the incoming client audio', async () => {
-      await desktopPlatformMock.initializeWithContext('sidePanel');
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
+      }
```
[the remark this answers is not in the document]
Jessica-Koch: possibly stupid question, why do we need an if statement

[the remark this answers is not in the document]
lakhveerk: if I do the return statement, I see options to debug/Run for the test, but if I nest the test inside for loop, I don't see the debug/run option. Let me search if there is any other better option online
image
image

packages/teams-js/test/public/meeting.spec.ts
lakhveerk: Should we follow the same pattern in other files like mail.spec or should we leave them as they are already done?

packages/teams-js/test/public/meeting.spec.ts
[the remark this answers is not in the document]
lakhveerk: @Jessica-Koch because we want an API to throw error if we initialize with a context in which capability is not supported in (and works fine in other cases).

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -355,201 +511,332 @@ describe('meeting', () => {
...
-      utils.sendMessage('meeting.liveStreamChanged', { isStreaming: true });
-
-      expect(handlerCalled).toBe(true);
-      expect(returnedLiveStreamState).not.toBeNull();
-      expect(returnedLiveStreamState).toEqual({ isStreaming: true });
+    const allowedContexts = [FrameContexts.sidePanel];
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        it('should successfully register a handler for when live stream is changed', async () => {
+          await framedPlatformMock.initializeWithContext(FrameContexts.sidePanel);
```
shrshindeMSFT, line 518: We need the code to pass the context change `FrameContexts.sidePanel` => `context`

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -303,42 +430,71 @@ describe('meeting', () => {
...
-          args: [null, { isStreaming: false }],
-        },
-      } as DOMMessageEvent);
-
-      await expect(promise).resolves;
+    const allowedContexts = [FrameContexts.sidePanel];
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        it('should successfully send the requestStartLiveStreaming message.', async () => {
+          await framelessPlatformMock.initializeWithContext(FrameContexts.sidePanel);
```
shrshindeMSFT, line 437: We want code to pass the context here change `FrameContexts.sidePanel` => `context`

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -303,42 +430,71 @@ describe('meeting', () => {
...
+
+          const requestStopLiveStreamingMessage = framelessPlatformMock.findMessageByFunc(
+            'meeting.requestStopLiveStreaming',
+          );
+          expect(requestStopLiveStreamingMessage).not.toBeNull();
+          expect(requestStopLiveStreamingMessage.args.length).toEqual(0);
+        });
+
+        it('should throw if the requestStopLiveStreaming message sends and fails', async () => {
+          await framelessPlatformMock.initializeWithContext(FrameContexts.sidePanel);
```
shrshindeMSFT, line 449: Same as above.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -303,42 +430,71 @@ describe('meeting', () => {
...
+              id: callbackId,
+              args: [{ errorCode: ErrorCode.INTERNAL_ERROR }, null],
+            },
+          } as DOMMessageEvent);
+
+          await expect(promise).rejects.toEqual({ errorCode: ErrorCode.INTERNAL_ERROR });
+        });
+
+        it('should resolve the promise after succesfully sending the meeting.requestStopLiveStreaming call', async () => {
+          await framelessPlatformMock.initializeWithContext(FrameContexts.sidePanel);
```
shrshindeMSFT, line 469: same as above

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -355,201 +511,332 @@ describe('meeting', () => {
...
-      expect(shareAppContentToStageMessage.args).toContain(requestUrl);
-      expect;
+      expect(() => meeting.shareAppContentToStage('')).toThrowError('The library has not yet been initialized');
+    });
+
+    const allowedContexts = [FrameContexts.sidePanel];
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        it('should successfully send the shareAppContentToStage message.', async () => {
+          framelessPlatformMock.initializeWithContext('sidePanel');
```
shrshindeMSFT, line 558: Can you please change the static context string to context variable in required locations.

shrshindeMSFT (changes requested):
Please change the static initialization context with respective variable in the code.

packages/teams-js/test/public/meeting.spec.ts
lakhveerk: It was how they do it before. I changed the namings now!

packages/teams-js/test/public/meeting.spec.ts
lakhveerk: Thank you so much for catching this. It was on my notes to change, but missed somehow. Will do it

packages/teams-js/test/public/meeting.spec.ts
lakhveerk: I have not added both platforms testing as part of this PR. This was only for fixing the async/await and to add left out unit tests (only for the platform that we were testing here.).

shrshindeMSFT (approved):

PR #913, opened 2022-02-08
public/tasks.spec.ts Error Validation Tests 

shrshindeMSFT:
Adding more unit test cases for **_FrameContext_** and cleaning up the code.


packages/teams-js/test/public/tasks.spec.ts
```diff
@@ -24,49 +25,29 @@ describe('tasks', () => {
...
-
-    it('should not allow calls from task context', async () => {
-      await utils.initializeWithContext('task');
-
-      const taskInfo: TaskInfo = {};
-      expect(() => tasks.startTask(taskInfo)).toThrowError(
-        'This call is only allowed in following contexts: ["content","sidePanel","meetingStage"]. Current context: "task".',
-      );
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContexts => allowedContexts === context)) {
```
Jessica-Koch, line 36: sorry if this is obvious, having this if here confuses me, I feel like in a test we should only be testing cases in which the if condition is true, and since we have control of the context that we are executing a test, we can ensure this is the case.  I'm not sure if this makes sense, I just don't know why there is an `if` there

shrshindeMSFT: Here we are checking for all the context which are not allowed by `tasks.startTask()` function. Also, in future use case if we have some new context, we would like to include that.

TrevorJoelHarris: I think this code is okay. The test itself is the the code inside the it(...) block, and each one is testing a single thing. We're just using some magic to generate a certain number of tests based on which contexts are allowed. It ends up (we hope) saving us a lot of typing and updating tests in the event that the set of valid frame contexts change.

TrevorJoelHarris: Jess' point about it being slightly confusing to read is something we should definitely think about and always keep in mind when we're writing unit tests though: how can we make these easy to understand and maintain. I think making an improvement in clarity is outside the scope of shreyas' PR since this is an established pattern we're using in other unit test files (but don't let that stop you if you have a good idea of how to make it easier to read)

TrevorJoelHarris (approved):

PR #915, opened 2022-02-08
Jekoch/throwing errors v2

Jessica-Koch:
* removed duplicate test blocks that I introduced in https://github.com/OfficeDev/microsoft-teams-library-js/pull/898/files#
* removed unnecessary expect.assertions() as they are only useful when testing code within a catch block
* refactored notifyEventToHost to follow a pattern illustrated elsewhere in the code


Jessica-Koch:
Learnings: While jest will not allow two tests within a describe block to be named the same, titles of describe blocks are not recognized as being the same, this means that a test like the following will not throw an error
```
describe('myModules', () => {

    describe('module a', () => {
        ...some tests
    })

    describe('module a', () => {
        ...some tests
    })
})
```
What is stranger is that in the test output, all the tests will be collapsed under a single 'module a' header.  This made the duplicated tests go undetected

packages/teams-js/test/public/media.spec.ts
```diff
@@ -755,9 +750,11 @@ describe('media', () => {
           errorCode: ErrorCode.INTERNAL_ERROR,
         };
 
-        await new media.VideoController().stop((e: SdkError) => {
-          expect(e).toEqual(err);
-        });
+        try {
+          new media.VideoController().stop(emptyCallback);
+        } catch (error) {
+          expect(error).toEqual(err);
```
erinha, line 756: I think this test should have `expect.assertions(4)` at the top (if my count is right), since we are adding an `expect` in a `catch` block

erinha: Or more because of the `mobilePlatformMock.initializeWithContext` (which is whole other story, don't get me started...)

packages/teams-js/src/public/media.ts
```diff
@@ -434,27 +434,31 @@ export namespace media {
...
-          throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
-        })
-        .then(() => {
-          const params: MediaControllerParam = { mediaType: this.getMediaType(), mediaControllerEvent: mediaEvent };
+      try {
+        throwExceptionIfMobileApiIsNotSupported(nonFullScreenVideoModeAPISupportVersion);
+      } catch (err) {
+        const wrappedRejectedErrorFn: InputFunction<void> = () => Promise.reject(err);
 
+        return callCallbackWithSdkErrorFromPromiseAndReturnPromise(wrappedRejectedErrorFn, callback);
```
TrevorJoelHarris, line 442: I think if we hit this line, then callback won't be passed the SdkError (looking at line 233 of utils.callCallbackWithSdkErrorFromPromiseAndReturnPromise). If I'm right, we could do something like wrap callback and pass it in like so:

const wrappedCallback = () => callback(err);
return callCallbackWithSdkErrorFromPromiseAndReturnPromise(wrappedRejectedErrorFn, wrappedCallback);

packages/teams-js/test/public/media.spec.ts
```diff
@@ -745,8 +742,6 @@ describe('media', () => {
       });
 
       it('videoController stop function returns SdkError to callback when parent rejects message"', async () => {
```
TrevorJoelHarris, line 744: I don't think this test is still testing what the title says it is. Title says we're checking that SdkError gets passed to callback, but test is testing whether we catch the error when calling.

Are there tests testing the callback still? On a quick scan I didn't see any.

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: this is the reason that the expect.assertions() was created, when I added it and the assertions should be 7 because there are 3 outside of this file, the catch was not getting hit

erinha: Are you saying that when you add `expect.assertions(7)` to the top of this test, the catch does not get hit?

packages/teams-js/src/public/media.ts
Jessica-Koch: We spoke about this offline, in the utility function, the error that is passed into the callback is the one that you receive from the rejected promise, if I write it the way you suggested, it will, still work, but i'll have to add an additional check to make sure that callback is defined in the cases where the callback is not provided.  I think this is extra code that isn't necessary

packages/teams-js/src/public/media.ts
```diff
@@ -434,27 +434,31 @@ export namespace media {
...
+      }
+
+      const params: MediaControllerParam = {
+        mediaType: this.getMediaType(),
+        mediaControllerEvent: mediaEvent,
+      };
+
+      const wrappedFunction = (): Promise<void> =>
+        Promise.resolve(
```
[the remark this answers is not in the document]
Jessica-Koch: you're right, you can still chain off of the Promise.resolve, but it resolves immediately, i'll change it

packages/teams-js/test/public/media.spec.ts
```diff
@@ -745,19 +742,16 @@ describe('media', () => {
...
```
erinha, line 744: Can you remove the `"` after "message" in the test name? (I'm a bit surprised that doesn't cause an error actually)

packages/teams-js/test/public/media.spec.ts
```diff
@@ -745,19 +742,16 @@ describe('media', () => {
...
-        expect.assertions(7);
-
+        expect.assertions(7); // 3 assertions in initializeWithContext + 4 local
         await mobilePlatformMock.initializeWithContext(FrameContexts.content, HostClientType.android);
         mobilePlatformMock.setClientSupportedSDKVersion(nonFullScreenVideoModeAPISupportVersion);
-
+
```
erinha: nit: extra whitespace

packages/teams-js/test/public/media.spec.ts
Jessica-Koch: it doesn't cause an error because there is no other '"' in the string, but I'm not sure why it is there in the first place

packages/teams-js/src/public/media.ts
```diff
@@ -434,27 +434,33 @@ export namespace media {
...
+
+      const params: MediaControllerParam = {
+        mediaType: this.getMediaType(),
+        mediaControllerEvent: mediaEvent,
+      };
+
+      const wrappedFunction = (): Promise<void> =>
+        new Promise(resolve =>
+          resolve(
+            sendMessageToParent('media.controller', [params], (err?: SdkError) => {
```
erinha, line 453: I think we want to use `sendAndHandleSdkError` for this code which calls `sendMessageToParentAsync`, rather than `sendMessageToParent`

packages/teams-js/src/public/media.ts
TrevorJoelHarris: Yes, agree, we spoke and that makes sense

TrevorJoelHarris (approved):
Looks good, thanks for making those changes

PR #921, opened 2022-02-11
Jekoch/media tests

Jessica-Koch:
In this PR: 

* convert tests to use async await 
* add missing `isSupported` test 
* add missing tests for  `notifyEventToApp`
* add missing line coverage
* ensure tests are not issuing false positives

**Due to the amount of whitespace changes when moving from using the `done` callback to async/await, I left a comment where I wrote an entirely new test or made a change that needs attention.  If it is not marked, it means it's moving to async await or changing test to follow guidelines linked above**


packages/teams-js/test/public/media.spec.ts
```diff
@@ -46,20 +47,30 @@ describe('media', () => {
     jest.clearAllMocks();
   });
 
+  describe('isSupported', () => {
+    it('returns true if media is supported', () => {
+      Object.defineProperty(runtime.supports, 'media', { value: true });
```
shrshindeMSFT, line 52: When run in VS Code the error is this => TypeError: Cannot redefine property: media.
In Jest CLI the test are passing and we have a conflict. Is it acceptable?

Jessica-Koch: this may be a node version, I ran this in VSCode to test and did not receive an error.  This was the only way I found in the docs to edit a readonly property 

Maybe we can discuss this offline to make sure I understand

packages/teams-js/test/public/media.spec.ts
```diff
@@ -953,153 +963,253 @@ describe('media', () => {
...
-          const mediaOutput: media.Media = new media.Media();
-          mediaOutput.content = '1234567';
-          mediaOutput.mimeType = 'image/jpeg';
-          mediaOutput.format = media.FileFormat.ID;
-          mediaOutput.getMedia((error: SdkError, blob: Blob) => {
-            getStringContainedInBlob(blob).then(res => {
-              expect(res).toEqual(stringMediaData);
-              done();
-            });
+      it('getMedia via handler call returns error when mediaResult contains SDK error', async () => {
```
Jessica-Koch, line 1028: This is a new test to add coverage to the error handling

packages/teams-js/test/public/media.spec.ts
```diff
@@ -953,153 +963,253 @@ describe('media', () => {
...
-          const mediaResults = Array.of(firstMediaResult, secondMediaResult);
+        for (let i = 0; i < mediaResults.length; ++i) {
+          callHandler(getMediaHandlerName, [JSON.stringify(mediaResults[i])]);
+        }
+      });
 
-          for (let i = 0; i < mediaResults.length; ++i) {
-            callHandler(getMediaHandlerName, [JSON.stringify(mediaResults[i])]);
-          }
+      it('getMedia calls with error when data is null via the handler', async () => {
```
Jessica-Koch, line 1010: This is a new test to add coverage to the error handling

packages/teams-js/test/public/media.spec.ts
```diff
@@ -953,153 +963,253 @@ describe('media', () => {
...
-              args: [undefined, undefined],
+              id: message.id,
+              args: [mediaResults[i]],
+              isPartialResponse: i < mediaResults.length - 1,
             },
           } as DOMMessageEvent);
+        }
+      });
+
+      it('getMedia calls with error with MediaCallback when data is null', async () => {
```
Jessica-Koch, line 1162: name of this test was changed to specify this test when data is null

packages/teams-js/test/public/media.spec.ts
```diff
@@ -953,153 +963,253 @@ describe('media', () => {
...
-          mediaOutput.mimeType = 'image/jpeg';
-          mediaOutput.format = media.FileFormat.ID;
-          mediaOutput.getMedia((error: SdkError, blob: Blob) => {
-            expect(error.errorCode).toBe(500);
-            expect(error.message).toEqual('data received is null');
-            expect(blob).toBeFalsy();
-            done();
-          });
+      it('getMedia returns error when called via the callback', async () => {
+        await mobilePlatformMock.initializeWithContext(FrameContexts.content);
```
Jessica-Koch, line 1118: new test, used the USER_ABORT as just a random error code to add to handle the case where there IS a media result, but there is also an error in that object

packages/teams-js/test/public/media.spec.ts
```diff
@@ -1215,6 +1333,47 @@ describe('media', () => {
         expect(blobContents).toEqual(stringMediaData);
       }
 
+      async function inValidateGetMediaMessageAndResults(
```
Jessica-Koch, line 1336: New function to handle the error cases for the v2 tests.

I could have made this into one function, but I couldn't find a clean way to pass whether or not the tests were supposed to test for a success case or a rejected case, I would also have had to pass in the mediaResult array but this is straightforward.

If anyone has a suggestion, I would love to hear it

packages/teams-js/test/public/media.spec.ts
```diff
@@ -1235,6 +1394,26 @@ describe('media', () => {
         );
       });
 
+      it('getMedia calls with error with MediaCallback', async () => {
```
Jessica-Koch, line 1397: new test using new function above

packages/teams-js/test/public/media.spec.ts
```diff
@@ -1249,48 +1428,82 @@ describe('media', () => {
           },
         );
       });
+
+      it('getMedia via the handler rejects if there is no data', async () => {
```
Jessica-Koch, line 1432: This is a new test to add coverage to the error handling

packages/teams-js/test/public/media.spec.ts
```diff
@@ -1249,48 +1428,82 @@ describe('media', () => {
...
+
+        callHandler(getMediaHandlerName, [JSON.stringify({})]);
+
+        return expect(getMediaPromise).rejects.toEqual({
+          errorCode: ErrorCode.INTERNAL_ERROR,
+          message: 'data received is null',
+        });
+      });
+
+      it('getMedia calls with error via Handler', async () => {
```
Jessica-Koch, line 1457: New test using the helper function above

packages/teams-js/test/public/media.spec.ts
```diff
@@ -543,7 +561,35 @@ describe('media', () => {
         expect(mediaError).toBeFalsy();
         expect(callbackSpy).not.toHaveBeenCalled();
       });
+      it('should invoke video callback for MediaControllerEvent when registered', async () => {
```
Jessica-Koch, line 564: new test to test behavior of notifyEventToApp

packages/teams-js/test/public/media.spec.ts
```diff
@@ -790,6 +836,38 @@ describe('media', () => {
 
         expect(sendMessageToParentSpy).not.toHaveBeenCalled();
       });
+
+      it('videoController notifyEventToApp should return if no callback is provided', async () => {
+        await mobilePlatformMock.initializeWithContext(FrameContexts.content, HostClientType.android);
+        mobilePlatformMock.setClientSupportedSDKVersion(originalDefaultPlatformVersion);
+
```
Jessica-Koch, line 843: new tests

packages/teams-js/test/public/media.spec.ts
shrshindeMSFT: Sure I'd like to know more about it.

Jessica-Koch: so after speaking briefly, I realized that @shrshindeMSFT was saying that this creates the typeerror but the tests pass.  I was unable to create a spy or mock implementation to get around this issue, not saying that it is not possible, I just couldn't figure out how to do it.  The solution I found came from https://github.com/facebook/jest/issues/2227

erinha, line 52: I think this should be setting the value to '{}', e.g.,:
`Object.defineProperty(runtime.supports, 'media', { value: {} });`
because that is closer to how the actual code sets the value when a capability is supported (e.g., it just checks that it's not undefined).

packages/teams-js/test/public/media.spec.ts
```diff
@@ -46,20 +47,30 @@ describe('media', () => {
     jest.clearAllMocks();
   });
 
+  describe('isSupported', () => {
+    it('returns true if media is supported', () => {
+      Object.defineProperty(runtime.supports, 'media', { value: true });
+      expect(media.isSupported()).toBe(true);
+    });
```
erinha, line 54: nit: please add a newline between the test cases

packages/teams-js/test/public/media.spec.ts
erinha: I discovered that there is a utils method designed for this purpose. Can you please not use `Object.defineProperty` and instead use `utils.setRuntimeConfig({ apiVersion: 1, supports: { media: {} } });`?

Check out calendar.spec.ts as a reference (the `isSupported` tests live at the bottom of that test file).

Jessica-Koch: sure, thanks,and that is a REALLY weird case that it only sometimes throws....impressive that you figured that out

Jessica-Koch: I'm resolving this since we aren't using defineProperty anymore

TrevorJoelHarris: I checked and it looks like these changes have now been made

TrevorJoelHarris (approved):
Since Erin is out today, went through and verified that all the changes she asked for have been done. Looks good!

PR #930, opened 2022-02-15
Adding unit tests to video.spec

lakhveerk:


packages/teams-js/test/public/video.spec.ts
```diff
@@ -23,38 +23,413 @@ describe('video', () => {
...
+    const videoFrameConfig: video.VideoFrameConfig = {
+      format: video.VideoFrameFormat.NV12,
+    };
+
+    const allowedContexts = [FrameContexts.sidePanel];
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
+      }
+      it('DESKTOP - should not allow registerForVideoFrame calls from the wrong context', async () => {
```
Jessica-Koch, line 41: would it be unreasonable to have different describe blocks for mobile and desktop within the registerforVideoFrame describe block? Just asking opinions

TrevorJoelHarris: This seems like a good idea to me

lakhveerk: Hey @Jessica-Koch, Is it okay to close the comment? We can talk further if you still have questions

packages/teams-js/test/public/video.spec.ts
```diff
@@ -3,18 +3,18 @@ import { app } from '../../src/public/app';
...
-  const mobilePlatformMock = new FramelessPostMocks();
-  const utils = new Utils();
+  const framelessPlatformMock = new FramelessPostMocks();
+  const framedPlatformMock = new Utils();
 
   beforeEach(() => {
-    mobilePlatformMock.messages = [];
-
-    // Set a mock window for testing
-    app._initialize(mobilePlatformMock.mockWindow);
```
aeunms, line 17: It seems we still `app._uninitialize` in afterEach, did we want to keep the `app._initialize`? Either way, maybe we should have both of them or neither of them.

packages/teams-js/test/public/video.spec.ts
```diff
@@ -23,38 +23,426 @@ describe('video', () => {
...
+    it('FRAMELESS - should successfully send registerForVideoFrame message', async () => {
+      await framelessPlatformMock.initializeWithContext(FrameContexts.sidePanel);
+      video.registerForVideoFrame(emptyVideoEffectCallback, videoFrameConfig);
+      const message = framelessPlatformMock.findMessageByFunc('video.registerForVideoFrame');
+      expect(message).not.toBeNull();
+      expect(message.args.length).toBe(1);
+      expect(message.args).toEqual([videoFrameConfig]);
+    });
+
+    it('FRAMED - should successful register video frame handler', async () => {
```
aeunms, line 81: Nit: typo 'successful'. Saw this a few times in other parts of the file too.

packages/teams-js/test/public/video.spec.ts
```diff
@@ -23,38 +23,426 @@ describe('video', () => {
...
+          args: [videoFrameMock],
+        },
+      } as DOMMessageEvent);
+      expect(handlerInvoked).toBeTruthy();
+      expect(returnedVideoFrame).toEqual(videoFrameMock);
+    });
+
+    it('FRAMED - should invoke video frame event handler and successfully send videoFrameProcessed', async () => {
+      await framedPlatformMock.initializeWithContext(FrameContexts.sidePanel);
+      const errorMessage = 'error occurs';
```
aeunms, line 159: I think the error message could be more descriptive. (Same goes for the other 'error occurs' error messages in the file.) Maybe replace it with something like 'VideoFrame was unexpectedly null'. Different messages for different tests if necessary.

packages/teams-js/test/public/video.spec.ts
lakhveerk: We call `initializeWithContext` in every API, and in `initializeWithContext`  function, first thing we do is call app._initialize. In my opinion, this is how it should be becuase `app._initialize` call depends upon the client that we are mocking. For all the iframed clients, we want to use utils.mockWindow and for frameless, we want to use framelessPosttMock.mockWindow. So, I deleted it from top. 

we still need to uninitialize  to clear states between tests.

packages/teams-js/test/public/video.spec.ts
lakhveerk: Thanks for catching. Will fix it.

packages/teams-js/test/public/video.spec.ts
lakhveerk: This error is coming back from host side. The purpose here is to show to display the same error that we get. That is the very reason, it is named, 'error occurs' because it could be any error and should be displayed correctly. We can talk about this and the other comment on a call.

packages/teams-js/test/public/video.spec.ts
aeunms: Gotcha. That sounds good. :)

packages/teams-js/test/public/video.spec.ts
lakhveerk: I changed the error name like we talked

aeunms: Discussed offline. error message is now specified.

aeunms (approved):
:shipit:

PR #932, opened 2022-02-15
public/settings.spec.ts

shrshindeMSFT:
- Cleaning up the code 
- Adding new test cases for settings.spec.ts 
- Refactoring unit test cases


packages/teams-js/test/public/settings.spec.ts
```diff
@@ -71,46 +57,131 @@ describe('settings', () => {
...
+    ];
+    const expectedSettings: settings.Settings = {
+      suggestedDisplayName: 'someSuggestedDisplayName',
+      contentUrl: 'someContentUrl',
+      websiteUrl: 'someWebsiteUrl',
+      entityId: 'someEntityId',
+    };
+
+    it('settings.getSettings should not allow calls before initialization', () => {
       const expectedSettings: settings.Settings = {
```
erinha, line 75: This is also defined on line 67-72. Should it be removed from within the test case?

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -71,46 +57,131 @@ describe('settings', () => {
...
+    });
 
-      utils.respondToMessage(message, expectedSettings);
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContexts => allowedContexts == context)) {
+        it(`settings.getSettings does not allow calls from ${context} context`, async () => {
+          await utils.initializeWithContext(context);
+          expect(() => {
+            settings.getSettings(settings => {
+              expect(settings).toBe(expectedSettings);
```
erinha, line 94: I don't think this is correct. We don't want to validate that the settings are correct in the tests where we expect it to throw an error. I'm concerned that the tests are passing with both "expect success" and "expect error" in the same block of code. If we get to line 94, the test should fail.

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -71,46 +57,131 @@ describe('settings', () => {
...
+            `This call is only allowed in following contexts: ${JSON.stringify(
+              allowedContexts,
+            )}. Current context: "${context}".`,
+          );
+        });
+      } else {
+        it(`settings.getSettings should successfully get settings from ${context} context`, async () => {
+          await utils.initializeWithContext(context);
+          settings.getSettings(settings => {
+            expect(settings).toBe(expectedSettings);
```
erinha, line 106: This check makes sense here in the success case.

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -71,46 +57,131 @@ describe('settings', () => {
...
+        });
+      }
     });
   });
 
-  it('should successfully register a save handler', () => {
-    utils.initializeWithContext('settings').then(() => {
+  describe('Testing settings.registerOnSaveHandler function', () => {
+    const allowedContexts = [FrameContexts.settings];
+    it('settings.registerOnSaveHandler should not allow calls before initialization', () => {
```
erinha, line 160: nit: newline before the start of the test case

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -71,46 +57,131 @@ describe('settings', () => {
...
+      if (!allowedContexts.some(allowedContexts => allowedContexts == context)) {
+        it(`settings.registerOnSaveHandler does not allow calls from ${context} context`, async () => {
+          await utils.initializeWithContext(context);
+          let handlerCalled = false;
+          expect(() => settings.registerOnSaveHandler(() => (handlerCalled = true))).toThrowError(
+            `This call is only allowed in following contexts: ${JSON.stringify(
+              allowedContexts,
+            )}. Current context: "${context}".`,
+          );
+        });
```
erinha, line 179: You could add a check that handlerCalled is still false.

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -120,10 +191,10 @@ describe('settings', () => {
 
       expect(handlerCalled).toBe(true);
     });
-  });
 
-  it('should successfully register a remove handler', () => {
-    return utils.initializeWithContext('settings').then(() => {
+    it('settings.registerOnSaveHandler should successfully register a remove handler', async () => {
```
erinha, line 195: I know you didn't originally write this test, but what exactly is a "remove handler"?

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -138,120 +209,154 @@ describe('settings', () => {
...
-    expect(handlerCalled).toBe(true);
-    const message = utils.findMessageByFunc('settings.save.failure');
-    expect(message).not.toBeNull();
-    expect(message.args.length).toBe(1);
-    expect(message.args[0]).toBe('someReason');
+      let handlerCalled = false;
+      settings.registerOnSaveHandler(saveEvent => {
+        saveEvent.notifySuccess();
+        expect(() => saveEvent.notifySuccess()).toThrowError('The SaveEvent may only notify success or failure once.');
+        expect(() => saveEvent.notifyFailure()).toThrowError('The SaveEvent may only notify success or failure once.');
```
erinha, line 273: I think another valid test case is doing the same sequence of actions except the first thing that is called is saveEvent.notifyFailure (like on line 271).

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -138,120 +209,154 @@ describe('settings', () => {
...
+          expect(() => settings.registerOnRemoveHandler(() => true)).toThrowError(
+            `This call is only allowed in following contexts: ${JSON.stringify(
+              allowedContexts,
+            )}. Current context: "${context}".`,
+          );
+        });
+      }
+    });
+
+    it('should successfully register a remove handler', async () => {
```
erinha, line 301: I don't think this test ('should successfully register a remove handler') is actually verifying anything?

packages/teams-js/test/public/settings.spec.ts
erinha: Can you add such a test case please?

packages/teams-js/test/public/settings.spec.ts
erinha: I think adding that check is worthwhile

packages/teams-js/test/public/settings.spec.ts
erinha: In fact, if you could add an expect that fails if this line of code is reached, that would be preferable.

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -138,120 +209,154 @@ describe('settings', () => {
...
-      handlerCalled = true;
+      expect(handlerCalled).toBe(true);
+      const message = utils.findMessageByFunc('settings.remove.failure');
+      expect(message).not.toBeNull();
+      expect(message.args.length).toBe(1);
+      expect(message.args[0]).toBe('someReason');
     });
+  });
 
+  it('should successfully notify success on save when there is no registered handler', async () => {
```
erinha, line 356: I don't think this test case belongs with the removeHandler test cases. I think it's better grouped at the very beginning in the describe("settings") test group.

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -20,48 +21,33 @@ describe('settings', () => {
...
-
-  it('should not allow calls from the wrong context', () => {
-    utils.initializeWithContext('content').then(() => {
-      expect(() => settings.setValidityState(true)).toThrowError(
-        'This call is only allowed in following contexts: ["settings","remove"]. Current context: "content".',
-      );
+  describe('Testing settings.setValidityState function', () => {
+    const allowedContexts = [FrameContexts.settings, FrameContexts.remove];
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContexts => allowedContexts == context)) {
```
erinha, line 27: Please use "==="

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: Removed duplicates.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: Added in the latest commit.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: Grouped to the top for the settings group.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: Modified in the latest commit.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: Added in the latest commit.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: I fix it by sending empty callback function can you please check this in the latest commit.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: It was issue with naming convention I have updated it.

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -6,6 +7,9 @@ describe('settings', () => {
   // Use to send a mock message from the app.
 
   const utils = new Utils();
+  const emptyCallback = (): void => {
```
jadahiya-MSFT, line 10: Why is this needed?

packages/teams-js/test/public/settings.spec.ts
```diff
@@ -71,46 +70,136 @@ describe('settings', () => {
...
+      }).toThrowError('The library has not yet been initialized');
+    });
 
-      const message = utils.findMessageByFunc('settings.getSettings');
-      expect(message).not.toBeNull();
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContexts => allowedContexts === context)) {
+        it(`settings.getSettings does not allow calls from ${context} context`, async () => {
+          await utils.initializeWithContext(context);
+          expect(() => settings.getSettings(emptyCallback)).toThrowError(
```
jadahiya-MSFT, line 99: Can you verify if this works as intended? Sometimes `toThrowError` provides a positive test pass when the output isn't correct.

shrshindeMSFT: Yes, I have verified the test case in both positive test pass and fail.

packages/teams-js/test/public/settings.spec.ts
shrshindeMSFT: We need this to send it as a parameter to `getSettings` function.

packages/teams-js/test/public/settings.spec.ts
jadahiya-MSFT: Awesome

packages/teams-js/test/public/settings.spec.ts
jadahiya-MSFT: does this not work?
```
getSettings( () => {}).toThrowError(...
```

shrshindeMSFT: TypeScript doesn't allow and throws error.

jadahiya-MSFT (approved):

erinha (approved):
:shipit:

PR #934, opened 2022-02-15
Removing manual conditions

ramourt:
Removing manual conditionals on the CI builds. 
If CI builds are triggered manually, they would eventually create releases, so these conditionals should also be enabled for manual builds on 2.0-preview or release/* branches.


tools/yaml-templates/security.yml
```diff
@@ -35,9 +35,8 @@ steps:
       IsShipped: false
       DropsToRetain: 'CodeAnalysisLogs'
     condition: and(
-      in(variables['Agent.JobStatus'], 'Succeeded', 'SucceededWithIssues'),
```
aeunms, line 38: (Asking because I genuinely don't know) why do we want to remove the succeeded condition?

ramourt: It's a good question.
This condition doesn't really do anything. A condition of succeeded or succeededWithIssues is the default behavior. We should either set this to succeeded only, or to nothing.

From here: [Conditions](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml)
> By default, a step runs if nothing in its job has failed yet and the step immediately preceding it has finished

aeunms: We have a bunch of these across the yml files- what do you think about me creating a ticket under engineering improvement to take care of these in another PR?

erinha: @aeunms That would be great, thank you!

tools/yaml-templates/build-test-publish.yml
```diff
@@ -23,8 +23,7 @@ steps:
     condition: and(
       eq(variables['Build.SourceBranch'], 'refs/heads/2.0-preview'),
       in(variables['Agent.JobStatus'], 'Succeeded', 'SucceededWithIssues'),
```
erinha, line 25: Should this also be removed since you're touching this file?

erinha:
So I understand, does this mean that you cannot trigger a CI pipeline manually? I don't think so, I think it means that the steps you are modifying will not execute if the pipeline is triggered manually. Or am I totally confused?

ramourt:
> So I understand, does this mean that you cannot trigger a CI pipeline manually? I don't think so, I think it means that the steps you are modifying will not execute if the pipeline is triggered manually. Or am I totally confused?

@erinha I removed the condition on these that were set to not run on manual runs. ie: if we did manual runs for the 2.0-preview CI, the version wouldn't be set to next-dev. We need this to fix the existing issue we talked about regarding to release pipelines running in parallel and attempting to set the same version. That's why I had to remove the manual condition skip for some of these tasks.

tools/yaml-templates/security.yml
ramourt: Thanks @aeunms !

erinha:
> > So I understand, does this mean that you cannot trigger a CI pipeline manually? I don't think so, I think it means that the steps you are modifying will not execute if the pipeline is triggered manually. Or am I totally confused?
> 
> @erinha I removed the condition on these that were set to not run on manual runs. ie: if we did manual runs for the 2.0-preview CI, the version wouldn't be set to next-dev. We need this to fix the existing issue we talked about regarding to release pipelines running in parallel and attempting to set the same version. That's why I had to remove the manual condition skip for some of these tasks.

Ah I see. Thank you for clarifying.

erinha (approved):
:shipit:

PR #935, opened 2022-02-15
Adding e2e test after release

ramourt:
This test will only run in the final stage of release. 

Once the release branch has been created and updates versions/integrity hashes, we will create a cdn/npm deployment from it. 

A PR will also be made to merge that branch back into 2.0-preview. This PR will run an extra check to run the test app (pulling the cdn bundle) as an additional e2e test. The PR is expected to fail initially (until the cdn deployment is complete). A re-run after deployment should succeed however.

We only want to run this test once in this stage. Future PRs should not run this test as the code becomes out of data with current code versions.


aeunms (approved):
LGTM

PR #936, opened 2022-02-16
lakhveer/people validating unit tests 

lakhveerk:


packages/teams-js/test/public/people.spec.ts
```diff
@@ -26,270 +26,226 @@ describe('peoplePicker', () => {
...
+  const allowedContexts = [FrameContexts.content, FrameContexts.task, FrameContexts.settings];
+  describe('peoplePicker', () => {
+    /**
+     * People Picker tests
+     */
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
+      }
+      it('should not allow selectPeople calls from the wrong context', async () => {
```
TrevorJoelHarris, line 38: Let's add the name of the context to the test the same way you did in the test after this. It's minor, but it can be a real help if a test fails to be able to tell exactly what the problem is just from reading the test name.

TrevorJoelHarris (approved):
Looks good to me!

Jessica-Koch:
I just wanted to ask, are we moving all tests to async/await, I felt like we were, so we were not going to be using the `done` callback.  If that's the case, I would update the tests that use `done` to use async await

lakhveerk:
> 'should not allow selectPeople calls from the wrong contex

We are using `done` only in V1 test cases where we had callbacks (not promises). All the new test cases (with promise) are tested using async/await

lakhveerk:
> We are using `done` only in V1 test cases where we had callbacks (not promises). All the new test cases (with promise) are tested using async/await

We are using done only in V1 test cases where we had callbacks (not promises). All the new test cases (with promise) are tested using async/await

packages/teams-js/test/public/people.spec.ts
```diff
@@ -26,270 +26,233 @@ describe('peoplePicker', () => {
...
+  describe('peoplePicker', () => {
+    /**
+     * People Picker tests
+     */
+    it('should not allow selectPeople calls before initialization', async () => {
+      expect(() => people.selectPeople()).toThrowError('The library has not yet been initialized');
+    });
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
```
shrshindeMSFT, line 50: Instead of using return can you please put the test cases where we initialize it using the allowed contexts.

packages/teams-js/test/public/people.spec.ts
```diff
@@ -26,270 +26,233 @@ describe('peoplePicker', () => {
...
-  });
-
-  it('should allow selectPeople calls with undefined peoplePickerInputs', async () => {
-    await mobilePlatformMock.initializeWithContext(FrameContexts.task);
-    mobilePlatformMock.setClientSupportedSDKVersion(minVersionForSelectPeople);
-    return expect(people.selectPeople(undefined)).resolves;
-  });
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
```
shrshindeMSFT, line 66: Similarly, avoid using return and put test cases in the required if and else block.

packages/teams-js/test/public/people.spec.ts
```diff
@@ -26,270 +26,233 @@ describe('peoplePicker', () => {
...
-  it('selectPeople call in default version of platform support fails', done => {
-    mobilePlatformMock.initializeWithContext(FrameContexts.task).then(() => {
-      mobilePlatformMock.setClientSupportedSDKVersion(originalDefaultPlatformVersion);
-      people.selectPeople((error: SdkError, people: people.PeoplePickerResult[]) => {
-        expect(error).not.toBeNull();
-        expect(error.errorCode).toBe(ErrorCode.OLD_PLATFORM);
-        done();
+    Object.values(FrameContexts).forEach(context => {
+      if (!allowedContexts.some(allowedContext => allowedContext === context)) {
+        return;
```
shrshindeMSFT, line 163: Same as above place the code in respective if and else block.

packages/teams-js/test/public/people.spec.ts
```diff
@@ -26,270 +26,233 @@ describe('peoplePicker', () => {
...
-      ];
-      mobilePlatformMock.respondToMessage({
-        data: {
-          id: callbackId,
-          args: [undefined, result],
-        },
-      } as DOMMessageEvent);
-    });
-  });
+      it(`selectPeople calls with valid peoplePickerInput. context: ${context}`, done => {
```
shrshindeMSFT, line 207: Can we use async and await instead of callbacks?

packages/teams-js/test/public/people.spec.ts
```diff
@@ -26,270 +26,233 @@ describe('peoplePicker', () => {
...
-      expect(message.args.length).toBe(1);
-
-      const callbackId = message.id;
-      mobilePlatformMock.respondToMessage({
-        data: {
-          id: callbackId,
-          args: [{ errorCode: ErrorCode.INTERNAL_ERROR }],
-        },
-      } as DOMMessageEvent);
+      it(`selectPeople calls with error. context: ${context}`, done => {
```
shrshindeMSFT, line 229: Can we use async and await here instead of callbacks?

shrshindeMSFT (changes requested):

packages/teams-js/test/public/people.spec.ts
lakhveerk: In v1, we have callbacks. I am having two groups of testing. With promises for v2, and with callbacks for v1

shrshindeMSFT: Okay cool.

lakhveerk: IS it okay to close these comments?

shrshindeMSFT (approved):

packages/teams-js/test/public/people.spec.ts
lakhveerk: Closing this, since you have approved the PR :)

lakhveerk:
> I just wanted to ask, are we moving all tests to async/await, I felt like we were, so we were not going to be using the `done` callback. If that's the case, I would update the tests that use `done` to use async await

We will talk more about this in our afternoon meeting, and if we decide, we do need to change something, I will open a new PR.

PR #945, opened 2022-02-22
Validating unit test cases for private/bot.spec.ts

shrshindeMSFT:
- Added missing test cases for `isSupported` function and different contexts.
- Adding more unit test cases for better code coverage and checking for failing unit tests.


packages/teams-js/test/private/bot.spec.ts
```diff
@@ -178,74 +186,88 @@ describe('bot', () => {
...
-    expect(message.args).toContain(request);
-
-    // simulate response
-    const data = {
-      success: false,
-      response: 'Bot authorization was unsuccessful',
-    };
-
-    utils.respondToMessage(message, data.success, data.response);
+  describe('Testing bot.isSupported function', () => {
```
jadahiya-MSFT, line 262: nit: Don't think we need function in here. Testing `bot.isSupported` should be enough

shrshindeMSFT: We are trying to change the standard format so that we can group them together and be informative in JEST CLI

jadahiya-MSFT (approved):

PR #949, opened 2022-02-23
Add Menus capability to test app for E2E testing

jekloudaMSFT:


apps/teams-test-app/src/components/MenusAPIs.tsx
```diff
@@ -0,0 +1,84 @@
...
+            throw new Error('ID and Title are required for each viewConfiguration');
+          }
+        });
+      },
+      submit: async (input, setResult) => {
+        menus.setUpViews(input, id => {
+          setResult('handler called with id: ' + id);
+          return true;
+        });
+        return 'Completed';
```
jadahiya-MSFT, line 23: Is Menus a new API for v2 or does it also exist in v1? If it also exists in v1 we will need to add `withPromise` and `withCallback` calls to the `submit`.

packages/teams-js/src/private/menus.ts
```diff
@@ -89,7 +89,7 @@ export namespace menus {
      * @hidden
      * Interface to show list of items on selection of menu item.
      */
-    public viewData: ViewData;
+    public viewData?: ViewData;
```
jadahiya-MSFT, line 92: Is this an intentional change?

jekloudaMSFT: Yes, I may be misunderstanding this API, but it looks like this should be an optional property. If `viewData` is a required property of `menuItem` and `menuItems` is a required property of `viewData` then it seems like it would be impossible to have some terminal menu item that doesn't open another menu.

jadahiya-MSFT: Got it thanks

apps/teams-test-app/src/components/MenusAPIs.tsx
jadahiya-MSFT: Discussed offline, this shouldn't be an issue. Resolving conversation

jadahiya-MSFT (approved):

PR #955, opened 2022-02-25
Replace instances of throwing strings to throw errors instead

aeunms:
Solves bug number #5720429 in ADO.

While Javascript allows throwing objects that aren't Error objects, it is still good style to throw Errors to capture all of the properties that can typically be expected from what is thrown and also to have a more descriptive stack trace.


lakhveerk (approved):
LGTM!

PR #957, opened 2022-02-28
Jadahiya/git magic feb 28 2022

jadahiya-MSFT:
Here is the [collection of changes being brought in from v1](https://github.com/OfficeDev/microsoft-teams-library-js/compare/d9a8669171184e1c2266d98c4f279e86859143e1...127586af6ddcb159a957d4c08100e74c46c6ee4c#)

### DIFF
--- a/packages/teams-js/test/public/calendar.spec.ts
+++ b/packages/teams-js/test/public/calendar.spec.ts
@@ -158,7 +158,7 @@ describe('calendar', () => {
       };
 
       utils.respondToMessage(openCalendarItemMessage, data.success);
-      expect(openCalendarItemPromise).resolves;
+      await expect(openCalendarItemPromise).resolves.not.toThrow();
     });
   });
 
@@ -257,7 +257,7 @@ describe('calendar', () => {
       };
 
       utils.respondToMessage(composeMeetingMessage, data.success);
-      expect(composeMeetingPromise).resolves;
+      await expect(composeMeetingPromise).resolves.not.toThrow();
     });
   });
 


packages/teams-js/src/public/stageView.ts
```diff
@@ -0,0 +1,64 @@
...
+    websiteUrl?: string;
+
+    /**
+     * The entity ID of the Teams app content being opened.
+     */
+    entityId?: string;
+  }
+
+  /**
+   * @private
```
aeunms, line 46: I know this is still a draft so no rush, but I thought it would be easier to let you know sooner than later: we have to change the `@private` tags from v1 to `@hidden` in v2 because we use a different docs generator and that's the tag that will correctly hide the docs as intended. (This wasn't anywhere in the git magician guide yet I think, so totally not something you could have known. I'm adding to a list of things to add to the wiki guide and these will be good additions!)

change/@microsoft-teams-js-e23ac78a-fff1-4ea1-919e-9b9e06ad3af7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "gitmagic additions to 2.0-preview for February 28, 2022",
```
aeunms, line 3: For all git magic PR changefiles, let's describe the changes in the source code that's being brought in on top of explaining that these are changes being brought in from master. (Another thing I'll be adding to our wiki- I think it would be good to start each git magic changefile comment with the same git magic 'title'. today we can start by just appending `(Changes merged in from v1)` to the beginning of the changefile comment. We can discuss what this should be in the retro too.)

packages/teams-js/test/public/tasks.spec.ts
```diff
@@ -126,7 +126,12 @@ describe('tasks', () => {
   });
 
   describe('updateTask', () => {
-    const allowedContexts = [FrameContexts.task];
+    const allowedContexts = [
```
aeunms, line 129: Personally I'm still unsure whether these changes should go in dialog vs. tasks vs. both. I'm leaning towards just dialog since tasks.updateTask is deprecated and what's really changing underneath the hood is in dialog, and having it in both seems duplicated. If you have a strong-ish opinion for any of these options, let me know! If not, I think we could move it to dialog and make a ticket to have someone else double-check what the situation here and other similar circumstances should generally point to. (i.e. changes in deprecated functions from v1 should be tested in the new function's test file vs. the deprecated function's test file)

aeunms (changes requested):
Thanks for doing awesome, awesome work here!! This period was full of heavy, complicated changes and you were so thorough. 

I left some comments that might warrant a new work item or two. Let's start a thread in our Teams channel and link any newly created work items there (not here though, as per our internal guidelines). If there turns out to be no new work items, please mention that there too.

packages/teams-js/test/public/tasks.spec.ts
jadahiya-MSFT: If we don't propagate this here, the tests will fail due to tasks calling `dialog.open` under the hood. Sadly I don't think we can remove it just yet, but we can definitely make a ticket to remove it in the future!

change/@microsoft-teams-js-e23ac78a-fff1-4ea1-919e-9b9e06ad3af7.json
jadahiya-MSFT: That's a great suggestion! I'll append that to the comment :)

change/@microsoft-teams-js-e23ac78a-fff1-4ea1-919e-9b9e06ad3af7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "gitmagic additions to 2.0-preview for February 28, 2022. Changes merged in from V1 include stageView implementation and modifications to tasks",
```
aeunms, line 3: Since this is the only thing that people will see in the changelog about this change, I was hoping we would be super specific. Something like: (sorry, I should have included an example the first time around):

> ...Changes merged in from V1 include stageView implementation and modifying `dialog.resize()` and `dialog.submit()` to also work in FrameContexts `sidePanel`, `content`, `meetingStage`, and `task` rather than just `task`.

jadahiya-MSFT: Got it, going forward we want to be super specific. I didn't know the level of specificity we were aiming for, I'll make that change now

aeunms: Yeah IMO it's a bit counterintuitive that we'd want it to be so specific considering it's only done on one line and you can't really use bullet points. Hopefully we can come up with something better for documenting on the changelog if this turns out to be a big issue.

aeunms (dismissed):
Thank you! :shipit:

PR #967, opened 2022-03-01
public/navigation.spec.ts

shrshindeMSFT:
- Adding test cases for FrameContexts 
- Fixing unit testcases with Async and await

### DIFF
--- a/packages/teams-js/test/public/navigation.spec.ts
+++ b/packages/teams-js/test/public/navigation.spec.ts
@@ -1,5 +1,6 @@
+import exp from 'constants';
 import * as utilFunc from '../../src/internal/utils';
-import { FrameContexts } from '../../src/public';
+import { FrameContexts, pages } from '../../src/public';
 import { navigateBack, navigateCrossDomain, navigateToTab, returnFocus } from '../../src/public/navigation';
 import { _uninitialize } from '../../src/public/publicAPIs';
 import { Utils } from '../utils';
@@ -31,6 +32,13 @@ describe('MicrosoftTeams-Navigation', () => {
 
     Object.values(FrameContexts).forEach(context => {
       if (allowedContexts.some(allowedContexts => allowedContexts === context)) {
+        it(`navigation.returnFocus should successfully call pages.returnFocus when initialized with ${context} context`, async () => {
+          await utils.initializeWithContext(context);
+          const pagesReturnFocus = jest.spyOn(pages, 'returnFocus');
+          returnFocus(true);
+          expect(pagesReturnFocus).toHaveBeenCalled();
+        });
+
         it(`navigation.returnFocus should successfully returnFocus when set to true and initialized with ${context} context`, async () => {
           await utils.initializeWithContext(context);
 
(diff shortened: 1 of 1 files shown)


packages/teams-js/test/public/navigation.spec.ts
```diff
@@ -21,134 +22,173 @@ describe('MicrosoftTeams-Navigation', () => {
...
-      expect(returnFocusMessage.args[0]).toBe(true);
-    });
-  });
-  describe('navigateCrossDomain', () => {
-    it('should not allow calls before initialization', () => {
-      expect(() => navigateCrossDomain('https://valid.origin.com')).toThrowError(
-        'The library has not yet been initialized',
-      );
+    it('navigation.returnFocus should not allow calls before initialization', () => {
+      expect(() => returnFocus(true)).toThrowError('The library has not yet been initialized');
```
lakhveerk, line 29: returnFocus just calls the pages.returnFocus function.  
```
export function returnFocus(navigateForward?: boolean): void {
  pages.returnFocus(navigateForward);
}
```
We should test pages.returnFocus in pages.spec file. I don't think we need to duplicate the tests here

packages/teams-js/test/public/navigation.spec.ts
```diff
@@ -21,134 +22,173 @@ describe('MicrosoftTeams-Navigation', () => {
...
+    it('navigation.navigateToTab should not allow calls before initialization', () => {
+      expect(() => navigateToTab(null)).toThrowError('The library has not yet been initialized');
     });
 
-    it('should allow calls from task context', async () => {
-      await utils.initializeWithContext('task');
-
-      navigateCrossDomain('https://valid.origin.com');
+    Object.values(FrameContexts).forEach(context => {
+      it(`navigation.navigateToTab should register the navigateToTab action when initialized with ${context} context`, () => {
```
lakhveerk, line 83: Same here. Actual API in pages file can check for the registration. Here we can mock a return value from pages.tabs.navifateToTab and make sure that we get true when it is completed successfully and an error message when there is an error.

packages/teams-js/test/public/navigation.spec.ts
```diff
@@ -21,134 +22,173 @@ describe('MicrosoftTeams-Navigation', () => {
...
-      utils.initializeWithContext('settings').then(() => {
-        navigateCrossDomain('https://invalid.origin.com', (success, reason) => {
-          expect(success).toBeFalsy();
-          expect(reason).toBe(
-            'Cross-origin navigation is only supported for URLs matching the pattern registered in the manifest.',
-          );
-          done();
+          navigateCrossDomain('https://valid.origin.com');
+
+          const navigateCrossDomainMessage = utils.findMessageByFunc('navigateCrossDomain');
```
lakhveerk, line 122: same here and other APIs in this file. I am sorry if I am getting it totally wrong. I think other people's review might help!

packages/teams-js/test/public/navigation.spec.ts
shrshindeMSFT: These are deprecated function and are internally replaced by `pages.returnFocus`. Here in this unit test case, we are checking even if we call deprecated function it should work.

lakhveerk: Why just checking that `pages.returnFocus` has been called (or not) does not work? I am just trying to understand.

shrshindeMSFT: It will work but the scope is to see the expected output when depreciated function is used and is working accordingly.

shrshindeMSFT: Check the latest commit I have added new test cases as you mentioned above.

packages/teams-js/test/public/navigation.spec.ts
lakhveerk: I see that you added code to check navigation.navigateToTab should successfully call pages.tabs.nagivateToTab. Don't we need to test .then which returns true and and .catch part for error?

lakhveerk: What I meant here was that we need to pass onComplete with error or with success case

jadahiya-MSFT: I agree that we might need a test that checks when we expect this sequence to fail/throw an error.

jadahiya-MSFT (approved):

PR #968, opened 2022-03-02
Add registerDetectSpeakingStateChangedHandler API

arocc:
Add client SDK changes to support registerDetectSpeakingStateChangedHandler API.

### DIFF
--- a/src/public/meeting.ts
+++ b/src/public/meeting.ts
@@ -274,7 +274,7 @@ export namespace meeting {
     if (!callback) {
       throw new Error('[share app content to stage] Callback cannot be null');
     }
-    ensureInitialized(FrameContexts.sidePanel);
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
     sendMessageToParent('meeting.shareAppContentToStage', [appContentUrl], callback);
   }
 
@@ -294,7 +294,7 @@ export namespace meeting {
     if (!callback) {
       throw new Error('[get app content stage sharing capabilities] Callback cannot be null');
     }
-    ensureInitialized(FrameContexts.sidePanel);
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
     sendMessageToParent('meeting.getAppContentStageSharingCapabilities', callback);
   }
 
@@ -312,7 +312,7 @@ export namespace meeting {
     if (!callback) {
       throw new Error('[stop sharing app content to stage] Callback cannot be null');
     }
-    ensureInitialized(FrameContexts.sidePanel);
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
     sendMessageToParent('meeting.stopSharingAppContentToStage', callback);
   }
 
@@ -329,7 +329,7 @@ export namespace meeting {
     if (!callback) {
       throw new Error('[get app content stage sharing state] Callback cannot be null');
     }
-    ensureInitialized(FrameContexts.sidePanel);
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
     sendMessageToParent('meeting.getAppContentStageSharingState', callback);
   }
 
@@ -344,8 +344,6 @@ export namespace meeting {
       throw new Error('[registerSpeakingStateChangeHandler] Handler cannot be null');
     }
     ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
-    registerHandler('meeting.speakingStateChanged', (speakingState: ISpeakingState) => {
-      handler(speakingState);
-    });
+    registerHandler('meeting.speakingStateChanged', handler);
   }
 }
--- a/src/public/meeting.ts
+++ b/src/public/meeting.ts
@@ -132,7 +132,7 @@ export namespace meeting {
     /**
      * returns speaking state of participants
      */
-    participantSpeakingStates: [IParticipantSpeakingState] | undefined;
+    participantSpeakingStates?: IParticipantSpeakingState[];
   }
 
   export enum MeetingType {
--- a/src/public/meeting.ts
+++ b/src/public/meeting.ts
@@ -110,29 +110,12 @@ export namespace meeting {
      */
     isAppSharing: boolean;
   }
-
-  export interface IParticipantSpeakingState {
-    /**
-     * id of participant
-     */
-    id: string;
-
-    /**
-     * indicates whether a participant is speaking
-     */
-    isSpeaking: boolean;
-  }
-
   export interface ISpeakingState {
     /**
-     * indicates whether one or more participants in a meeting are speaking
+     * indicates whether one or more participants in a meeting are speaking, or
+     * if no participants are speaking
      */
     isSpeakingDetected: boolean;
-
-    /**
-     * returns speaking state of participants
-     */
-    participantSpeakingStates?: IParticipantSpeakingState[];
   }
 
   export enum MeetingType {
@@ -351,20 +334,17 @@ export namespace meeting {
   }
 
   /**
-   * Registers a handler for changes to speaking array.
+   * Registers a handler for changes to paticipant speaking states. If any participant is speaking, isSpeakingDetected
+   * will be true. If no participants are speaking, isSpeakingDetected will be false.
    * @param handler The handler to invoke when the speaking state of any participant changes (start/stop speaking).
-   * @param includeParticipants is a flag to opt-in to receive participant ids of who is speaking.
-   * The default value is false. If the participant ids are needed, register for the permission:
-   * OnlineMeetingParticipant.Read.Chat in the app's manifest.
    */
   export function registerDetectSpeakingStateChangedHandler(
-    handler: (speakingState: ISpeakingState) => void,
-    includeParticipants?: boolean,
+    handler: (error: SdkError | null, speakingState: ISpeakingState) => void,
   ): void {
     if (!handler) {
       throw new Error('[registerDetectSpeakingStateChangedHandler] Handler cannot be null');
     }
     ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
-    registerHandler('meeting.speakingStateChanged', handler, true, [includeParticipants]);
+    registerHandler('meeting.speakingStateChanged', handler, true);
   }
 }
--- a/src/public/meeting.ts
+++ b/src/public/meeting.ts
@@ -339,13 +339,12 @@ export namespace meeting {
    * at a time. A subsequent registration replaces an existing registration.
    * @param handler The handler to invoke when the speaking state of any participant changes (start/stop speaking).
    */
-  export function registerDetectSpeakingStateChangedHandler(handler: (speakingState: ISpeakingState) => void): void {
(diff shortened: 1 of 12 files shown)


src/public/meeting.ts
```diff
@@ -111,6 +111,30 @@ export namespace meeting {
...
+  export interface ISpeakingState {
+    /**
+     * indicates whether one or more participants in a meeting are speaking
+     */
+    isSpeakingDetected: boolean;
+
+    /**
+     * returns speaking state of participants
+     */
+    participantSpeakingStates: [IParticipantSpeakingState] | undefined;
```
MarcinJuraszek, line 135: `[IParticipantSpeakingState]` means an array with a single element of type `IParticipantSpeakingState`. What I think you wanted to do is IParticipantSpeakingState[]`.

Also, do we need undefined? Why not just empty array for cases where there are no states?

arocc: Oh yes, thank you for catching that! And empty array works well too -- will update it.

arocc: Updated and also made this optional, as we would only be returning it if the developer specified to include participants and they have the necessary permission.

src/public/meeting.ts
```diff
@@ -325,4 +349,22 @@ export namespace meeting {
...
+  /**
+   * Registers a handler for changes to speaking array.
+   * @param handler The handler to invoke when the array of participants speaking changes. A value
+   * of 0 (no participants speaking) or 1 (one or more participants are speaking) is provided.
+   * @param includeParticipants is a flag to opt-in to receive participant ids of who is speaking.
+   * Default value is false.
+   */
+  export function registerDetectSpeakingStateChangedHandler(
+    handler: (isSpeakingDetected: ISpeakingState) => void,
+    includeParticipants?: boolean,
```
ryanbliss, line 362: Not sure if matters, but is it more common to just do includeParticipants = false to make it more implicit what the default value is? I'm not sure I've ever seen an optional Boolean before, but both ways should work.

src/public/meeting.ts
ryanbliss: Heh, looks like my syntax goof carried over. My first serious programming language was Swift, where the `[T]` is a perfectly valid shorthand declaration 😂

src/public/meeting.ts
MarcinJuraszek, line 362: Why would the caller specify `false` here instead of just ignoring the information in the callback they provide? Does it make it cheaper on the Teams' side not to send participants info?

src/public/meeting.ts
```diff
@@ -325,4 +349,22 @@ export namespace meeting {
...
+
+  /**
+   * Registers a handler for changes to speaking array.
+   * @param handler The handler to invoke when the array of participants speaking changes. A value
+   * of 0 (no participants speaking) or 1 (one or more participants are speaking) is provided.
+   * @param includeParticipants is a flag to opt-in to receive participant ids of who is speaking.
+   * Default value is false.
+   */
+  export function registerDetectSpeakingStateChangedHandler(
+    handler: (isSpeakingDetected: ISpeakingState) => void,
```
MarcinJuraszek, line 361: The parameter name being `isSpeakingDetected` is confusing, because it's actually `ISpeakingState`, so you'd have to do `isSpeakingDetected.isSpeakingDetected` to actually get a `true`/`false` value. That doesn't seem right.

src/public/meeting.ts
```diff
@@ -325,4 +349,22 @@ export namespace meeting {
     ensureInitialized(FrameContexts.sidePanel);
     sendMessageToParent('meeting.getAppContentStageSharingState', callback);
   }
+
+  /**
+   * Registers a handler for changes to speaking array.
+   * @param handler The handler to invoke when the array of participants speaking changes. A value
+   * of 0 (no participants speaking) or 1 (one or more participants are speaking) is provided.
+   * @param includeParticipants is a flag to opt-in to receive participant ids of who is speaking.
+   * Default value is false.
```
MarcinJuraszek, line 358: This comment doesn't make much sense either. It says you get `0` or `1`, which isn't what the types below suggest. And the second comment says `to receive participant ids` - receive where, if the callback gets a `0` or `1`?

src/public/meeting.ts
arocc: @ryanbliss all good! I didn't notice until Marcin pointed it out, my bad too 😄

src/public/meeting.ts
arocc: No, this was more to do with permissions. Getting the participants info would require an additional permission in the app manifest.

src/public/meeting.ts
arocc: Thanks Marcin! The idea for ISpeakingState was to communicate if anyone is speaking, and also (if the developer asks and has the necessary permission) to send an array of all the participant ids and their `isSpeaking` (0 or 1). Maybe a better name for this parameter would be `speakingState`? And if a developer only wants to know if _anyone_ is speaking they can do `speakingState.isSpeakingDetected`.

In the case of the developer wanting participant info too, do you think it would make more sense to return an array of just the ids of participants that are currently speaking or an array of all participants and whether or not they are speaking? Would be happy to discuss on a call too if you'd like.

src/public/meeting.ts
arocc: That's true, this should instead align with the types below. The callback should get the type `ISpeakingState ` -- with a boolean for `isSpeakingDetected` and if requested by the developer (using the `includeParticipants` flag), an array of `IParticipantSpeakingState `.

src/public/meeting.ts
ryanbliss: +1 for speakingState instead of isSpeakingDetected for the response object.

src/public/meeting.ts
arocc: @ryanbliss to follow up on this part, I believe it's more common to handle the default false value on the teams side. For example, in location.ts, there is an optional boolean to show a map: https://github.com/OfficeDev/microsoft-teams-library-js/blob/a3ae40a9606803c68e280a022fe9849b66901e99/src/public/location.ts#L21 . Then on the teams side there is some logic in the hook file that handles this API for the default false value. From what I could find, I didn't see any example in the SDK where we set default values on this side.

src/public/meeting.ts
arocc: Hi @MarcinJuraszek I updated the comment, please feel free to let me know if you have any thoughts on the change.

src/public/meeting.ts
arocc: @MarcinJuraszek @ryanbliss -- I updated this parameter name to `speakingState` to hopefully be clearer.

src/public/meeting.ts
ryanbliss: @arocc good to know, thank you!

MarcinJuraszek: hmm, somehow that doesn't sit well with me. Should there be two separate APIs instead?

arocc: Hi @MarcinJuraszek thank you for the feedback! I updated it to remove that part of including participants. If needed in the future, it can be two separate APIs. For now, we will only expose if anyone is speaking, or no one is speaking.

src/public/meeting.ts
```diff
@@ -325,4 +332,21 @@ export namespace meeting {
...
+   * at a time. A subsequent registration replaces an existing registration.
+   * @param handler The handler to invoke when the speaking state of any participant changes (start/stop speaking).
+   */
+  export function registerDetectSpeakingStateChangedHandler(handler: (speakingState: ISpeakingState) => void): void {
+    if (!handler) {
+      throw new Error('[registerDetectSpeakingStateChangedHandler] Handler cannot be null');
+    }
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
+    registerHandler('meeting.speakingStateChanged', (speakingState: ISpeakingState) => {
+      ensureInitialized();
```
ydogandjiev, line 348: I don't think this second ensureInitialized call is necessary. Initialization of the SDK is only ever done one time and for this handler to be invoked, the SDK must already have been initialized since you checked that on line 346.

src/public/meeting.ts
```diff
@@ -110,6 +110,13 @@ export namespace meeting {
      */
     isAppSharing: boolean;
   }
+  export interface ISpeakingState {
+    /**
+     * indicates whether one or more participants in a meeting are speaking, or
```
ydogandjiev, line 115: NIT: Capitalize "indicates" --> "Indicates"; these comments automatically end up in our public docs so it's good to make them consistent

ydogandjiev (changes requested):

src/public/meeting.ts
arocc: Hi @ydogandjiev thank you for the feedback! I noticed other interfaces in this file more often start with lowercase and was following that convention -- should the others be capitalized too?

arocc, line 115: Hi @ydogandjiev, I included the comment for ISpeakingState here on the following lines.

src/public/meeting.ts
```diff
@@ -110,6 +110,13 @@ export namespace meeting {
      */
     isAppSharing: boolean;
   }
+  export interface ISpeakingState {
```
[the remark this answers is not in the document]
arocc: Added new comment on that line where the description is!

src/public/meeting.ts
```diff
@@ -325,4 +332,21 @@ export namespace meeting {
...
     sendMessageToParent('meeting.getAppContentStageSharingState', callback);
   }
+
+  /**
+   * Registers a handler for changes to paticipant speaking states. If any participant is speaking, isSpeakingDetected
+   * will be true. If no participants are speaking, isSpeakingDetected will be false. Only one handler can be registered
+   * at a time. A subsequent registration replaces an existing registration.
+   * @param handler The handler to invoke when the speaking state of any participant changes (start/stop speaking).
+   */
+  export function registerDetectSpeakingStateChangedHandler(handler: (speakingState: ISpeakingState) => void): void {
```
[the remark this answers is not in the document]
arocc: Agree -- I think that name is better, thank you! Updated now.

TrevorJoelHarris (approved):

ydogandjiev (approved):

src/public/meeting.ts
```diff
@@ -325,4 +332,20 @@ export namespace meeting {
...
+   * will be true. If no participants are speaking, isSpeakingDetected will be false. Only one handler can be registered
+   * at a time. A subsequent registration replaces an existing registration.
+   * @param handler The handler to invoke when the speaking state of any participant changes (start/stop speaking).
+   */
+  export function registerSpeakingStateChangeHandler(handler: (speakingState: ISpeakingState) => void): void {
+    if (!handler) {
+      throw new Error('[registerSpeakingStateChangeHandler] Handler cannot be null');
+    }
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
+    registerHandler('meeting.speakingStateChanged', (speakingState: ISpeakingState) => {
```
ydogandjiev, line 347: NIT: I think this can be simplified to:
```
registerHandler('meeting.speakingStateChanged', handler);
```

ydogandjiev (approved):

PR #973, opened 2022-03-03
Changes for automated versioning

ramourt:


erinha:
This PR is necessary before we release v2.0.0. Once this is merged to main, I can create the release branch for 2.0.0.

AE-MS:
>     throw new Error('MicrosoftTeams.min.js integrity hash value was not parsed');

Some part of me says that this filename should be a defined constant that we refer to everywhere but I'm guessing this isn't the place to make that change.

---
Refers to: tools/cli/preRelease.js:38 in 48199c4. [](commit_id = 48199c4bbec822f4164f173e006a29221ab7980d, deletion_comment = False)

AE-MS (dismissed):
:shipit:

aeunms (changes requested):
We will have some weird dev packages released if we don't modify the prepNextDevRelease script or disable it temporarily from the build-test-publish.yml file.

tools/yaml-templates/build-test-publish.yml
```diff
@@ -21,8 +21,8 @@ steps:
     displayName: 'Check that changefile was created if needed'
     condition: and(
       eq(variables['Build.Reason'], 'PullRequest'),
-      not(startsWith(variables['System.PullRequest.SourceBranch'], 'release/2.0.0-beta')),
-      not(startsWith(variables['System.PullRequest.TargetBranch'], 'release/2.0.0-beta'))
+      not(startsWith(variables['System.PullRequest.SourceBranch'], 'release/')),
+      not(startsWith(variables['System.PullRequest.TargetBranch'], 'release/'))
       )
     inputs:
       Arguments: 'beachball check'
```
aeunms, line 28: My recommendation would be to simply comment out lines 30-37 here to avoid dev releases with inaccurate 2.0.0-beta.# naming.

erinha: Thanks for pointing this out! I was reasoning over this and I was thinking that until we release v2.0.0, we still want the beta next-dev releases from main. They will continue to query @next-dev and compare with what is in package.json, which will be 2.0.0-beta.7 until we release 2.0. 

Now that I'm typing this out, I'm wondering if that is indeed what we want because we may release @next-dev code changes that aren't present in 2.0.0, but will be present in 2.0.1 (or 2.1, whichever we determine via beachball).

aeunms: Right, I thought that it would be consistent to publish the dev packages as 2.0.1-*.

erinha: I think it's best to wait until after 2.0.0 to continuing publishing @next-dev releases from `main`. I've disabled the release pipeline and updated this file to skip this step.

aeunms (approved):
🚢

erinha:
> > ```
> > throw new Error('MicrosoftTeams.min.js integrity hash value was not parsed');
> > ```
> 
> Some part of me says that this filename should be a defined constant that we refer to everywhere but I'm guessing this isn't the place to make that change.
> 

I agree. I opened a work item to track fixing this.

AE-MS (approved):
:shipit:

PR #977, opened 2022-03-07
Create new subcategory for conversations and add in openChat functionality

jadahiya-MSFT:
Implemented changes to the chat functionality based off of offline discussions


packages/teams-js/src/private/chat.ts
```diff
@@ -18,64 +18,47 @@ import { ChatMembersInformation } from './interfaces';
  */
 export namespace chat {
   /**
-   * @hidden
-   * Hide from docs
-   * --------------
-   * Allows the user to start or continue a conversation with each subentity inside the tab
    *
+   * Allows the user to start a direct chat with a user
+   * @members = length 1
```
aeunms, line 23: what is the @member tag here? I can't find it on the official Typedoc docs.

packages/teams-js/src/private/chat.ts
```diff
@@ -18,64 +18,47 @@ import { ChatMembersInformation } from './interfaces';
...
    * @returns Promise resolved upon completion
    */
-  export function openConversation(openConversationRequest: OpenConversationRequest): Promise<void> {
+
+  export function openChat(openChatRequest: OpenChatRequest): Promise<void> {
     return new Promise<void>(resolve => {
+      if (openChatRequest.members.length !== 1) {
```
TrevorJoelHarris, line 29: It would be nice if we could make this "impossible" for app developers to call incorrectly. Maybe if we introduce a second interface that inherits from OpenChatRequest, something like "OpenSingleChatRequest". Then have the new one take only a single member while OpenChatGroupRequest takes an array. If we did that, we probably wouldn't export OpenChatRequest since no app developers should be passing it in (they should be using one of the two interfaces that extend it)

packages/teams-js/src/private/chat.ts
```diff
@@ -18,64 +18,47 @@ import { ChatMembersInformation } from './interfaces';
  */
 export namespace chat {
   /**
-   * @hidden
-   * Hide from docs
-   * --------------
-   * Allows the user to start or continue a conversation with each subentity inside the tab
    *
+   * Allows the user to start a direct chat with a user
```
TrevorJoelHarris, line 22: We should follow up with @erinha about what the right style and level of detail we are looking for in comments on exported functions. (It's also okay with me if you want to handle that when you go through and look at all of the comments and tests on the whole suite of deep linking issues)

packages/teams-js/src/public/runtime.ts
```diff
@@ -10,7 +10,9 @@ export interface IRuntime {
     readonly bot?: {};
     readonly calendar?: {};
     readonly call?: {};
-    readonly chat?: {};
+    readonly chat?: {
```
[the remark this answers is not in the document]
TrevorJoelHarris: (If there is, that would happen in a separate checkin since this it's in a different repo, just wanted to check)

packages/teams-js/src/private/chat.ts
aeunms: I don't think it generates anything as is because it's @hidden. When I check out the branch and remove the @hidden tag to run `yarn docs` I get this:
image

So it won't break, but I feel like this isn't the way we want to specify required length of members. (Seeing now that members is probably just openChatRequest's members and not a Typedoc tag like @remarks, @hidden, etc.) I would expect to see something like 

```
@param openChatRequest The request to open a chat for. Length must be equal to 1.
```

And maybe including more details pending additional implementation on what happens with Trevor's other comment below on making it impossible for app devs to call this API incorrectly.

TrevorJoelHarris (changes requested):
Let me know when ready for review again

packages/teams-js/test/private/chat.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('chat', () => {
         title: 'someTitle',
```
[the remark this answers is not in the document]
jadahiya-MSFT: I will add a new work item that tracks this, but we will need to wait for the service side code to go in before we can fully test the results of the Promises. I've added in tests that we can currently perform though. I could additionally mock up these tests and just add a `it.skip` to them for now.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -577,6 +577,53 @@ export interface DialogInfo {
  */
 export type TaskInfo = DialogInfo;
 
+/**
+ *
+ * @internal
```
TrevorJoelHarris, line 582: We had a discussion with Lakhveer about where interfaces should go in standup on wednesday. In general, there's no strict guidance on where exported interfaces should go, what we landed on was sort of "if you think the interface is probably going to be limited to just one capability, consider putting it in that capability, while mostly interfaces.ts should be for "shared capabilities". We haven't been super careful about this in the past, so I don't think you need to move these, but it might make slightly more sense to move these into chat since they feel like they are probably only ever going to be for that capability.

change/@microsoft-teams-js-cf7d9559-bf41-4407-bbc2-f1d5cc910a36.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Moving chat.openConversation and chat.closeConversation into a subcategory chat.conversation. Added in new APIs chat.openChat and chat.openGroupChat to leverage deeplinking to open Teams chats with one or more user",
```
TrevorJoelHarris, line 3: I might replace the word "leverage" with "replace". These logs are for developers looking to update their apps to use the new functionality. While under the covers we might still be support some deep linking scenarios, I think it makes more sense for app devs to think of this change as a replacement for the teams-specific deep linking model

packages/teams-js/src/private/chat.ts
```diff
@@ -21,61 +21,53 @@ export namespace chat {
    * @hidden
    * Hide from docs
    * --------------
-   * Allows the user to start or continue a conversation with each subentity inside the tab
+   * Allows the user to open a chat with a single user using a deep link and
```
TrevorJoelHarris, line 24: I think it would make sense to remove the deep link mentions from the comments. I think from the app dev perspective this isn't deep link related anymore: they don't pass in URLs and no longer need to know about them.

Otherwise these comments look *fantastic* to me. Love the detail!

change/@microsoft-teams-js-cf7d9559-bf41-4407-bbc2-f1d5cc910a36.json
jadahiya-MSFT: Good idea

packages/teams-js/src/public/interfaces.ts
jadahiya-MSFT: Sounds good, I'll move that now

packages/teams-js/test/private/chat.spec.ts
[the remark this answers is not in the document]
TrevorJoelHarris: I would assume that these are unit tests, and so the service side doesn't have to be implemented yet because you're just mocking it out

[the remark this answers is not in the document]
jadahiya-MSFT: Whoops, yes this is correct. I've added in the necessary tests

packages/teams-js/src/private/chat.ts
```diff
@@ -21,61 +21,53 @@ export namespace chat {
    * @hidden
    * Hide from docs
    * --------------
-   * Allows the user to start or continue a conversation with each subentity inside the tab
+   * Allows the user to open a chat with a single user and allows
+   * for the user to specify the message they wish to send.
    *
```
TrevorJoelHarris, line 26: We should add openChatRequest to the comments using @param

packages/teams-js/src/private/chat.ts
```diff
@@ -21,61 +21,53 @@ export namespace chat {
...
-
   /**
    * @hidden
    * Hide from docs
    * --------------
-   * Allows the user to close the conversation in the right pane
+   * Allows the user to create a chat with multiple users (2+) and allows
+   * for the user to specify a message and name the topic of the conversation. If
+   * only 1 user is provided into users array default back to origin openChat.
+   *
```
TrevorJoelHarris, line 46: We should add openChatRequest to the comments using @param

packages/teams-js/src/private/chat.ts
```diff
@@ -21,61 +21,53 @@ export namespace chat {
...
+          topic: openChatRequest.topic,
+        });
+        resolve(sendPromise);
+      }
+    });
   }
-
   /**
    * @hidden
    * Hide from docs
```
TrevorJoelHarris, line 73: Looks like you didn't change this, but the callback param for getChatMembers doesn't exist anymore we should probably delete this part of the comment

TrevorJoelHarris (approved):

packages/teams-js/src/private/chat.ts
TrevorJoelHarris: Sorry I promise this is the last one. If we make this:

@param openChatRequest {@link OpenSingleChatRequest} - a request object that contains a user's email as well as an optional message parameter.

Then i believe the documentation will include a link to the definition of OpenSingleChatRequest

packages/teams-js/src/private/chat.ts
TrevorJoelHarris: If we make this:

@param openChatRequest {@link OpenGroupChatRequest} - a request object that contains a list of user emails as well as optional parameters for message and topic (display name for the group chat).

Then i believe the documentation will include a link to the definition of OpenGroupChatRequest

TrevorJoelHarris (dismissed):

packages/teams-js/src/private/chat.ts
jadahiya-MSFT: Good idea

TrevorJoelHarris (approved):

PR #982, opened 2022-03-08
public/pages.spec.ts

shrshindeMSFT:
- Adding missing unit test cases for pages module 
- Fixing async/await unit test cases


packages/teams-js/test/public/pages.spec.ts
```diff
@@ -1,9 +1,16 @@
+import { OBJECTMEMBER_TYPES } from '@babel/types';
```
jadahiya-MSFT, line 1: is this import used?

shrshindeMSFT: Fixed it in latest commit

jadahiya-MSFT (approved):

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+        utils.setRuntimeConfig({
+          apiVersion: 1,
+          supports: { pages: { tabs: {} } },
+        });
+        expect(pages.tabs.isSupported()).toBeTruthy();
+      });
+    });
+  });
+
+  describe('Testing pages.config namespace', () => {
```
erinha, line 668: I just noticed that we have a test file called config.spec.ts, which apperas to have pages.config/settings capability tests in it. Could you please take a look at that file and reconcile it with these changes?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+        utils.setRuntimeConfig({
+          apiVersion: 1,
+          supports: { pages: { backStack: {} } },
+        });
+        expect(pages.backStack.isSupported()).toBeTruthy();
+      });
+    });
+  });
+
+  describe('Testing pages.fullTrust namespace', () => {
```
erinha, line 1073: I also just noticed that we have test/private/pages.spec.ts where the pages.fullTrust capability appears to have tests. Can you also please reconcile these tests with that file?

shrshindeMSFT: Do I need to remove those test cases in `test/private/pages.spec.ts` file?

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: I noticed that both are almost the same and we can keep both of them and it's not going to be a problem. But to avoid redundancy I can remove the `config.spec.ts`.

shrshindeMSFT: I've removed the duplicate files in the code base and updated in the latest commit

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -24,98 +29,216 @@ describe('AppSDK-TeamsAPIs', () => {
...
-      const navigateToAppMessage = utils.findMessageByFunc('pages.navigateToApp');
-      utils.respondToMessage(navigateToAppMessage, true);
-      await promise;
+      it(`pages.registerFocusEnterHandler should not invoke focus enter handler when set to false initialized with ${context} context`, async () => {
+        await utils.initializeWithContext(context);
 
-      expect(navigateToAppMessage).not.toBeNull();
-      expect(navigateToAppMessage.args[0]).toStrictEqual(navigateToAppParams);
+        let handlerInvoked = true;
+        pages.registerFocusEnterHandler((x: boolean) => {
```
Jessica-Koch, line 109: I assume that these functions are supposed to return a boolean, because we aren't doing anything with it

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -24,98 +29,216 @@ describe('AppSDK-TeamsAPIs', () => {
...
+          expect(utils.processMessage).toBeDefined();
+          expect(utils.messages.length).toBe(2);
+
+          let initMessage = utils.findMessageByFunc('initialize');
+          expect(initMessage).not.toBeNull();
+          expect(initMessage.id).toBe(0);
+          expect(initMessage.func).toBe('initialize');
+          expect(initMessage.args.length).toEqual(1);
+          expect(initMessage.args[0]).toEqual(version);
+          let message = utils.findMessageByFunc('setFrameContext');
```
Jessica-Koch, line 175: are initMessage or message getting reassigned and i'm missing it?

Jessica-Koch (approved):

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: We are just checking that the messages are initialized correctly and have expected arguments.

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Yes, are just verifying that the handler is invoked or not by flowing through the code.

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+          await promise;
+
+          expect(navigateToAppMessage).not.toBeNull();
+          expect(navigateToAppMessage.args[0]).toStrictEqual(navigateToAppParams);
+        });
+
+        it('pages.navigateToApp should successfully send an executeDeepLink message for legacy teams clients', async () => {
+          await utils.initializeWithContext(context);
+
+          const promise = pages.navigateToApp(navigateToAppParams);
```
erinha, line 394: Suggest setting the runtime.isLegacyTeams to true so the test is more explicit and not relying on default behavior. Although actually it might be good to add a test that validates the default behavior as well, in case it ever changes.

erinha:
>     // Commenting out these tests as url validation is not implemented

Do you know if these should still be commented out or if they can be re-enabled now?

---
Refers to: packages/teams-js/test/public/pages.spec.ts:247 in 495cdb6. [](commit_id = 495cdb671ae25be509045873beb510bf81371521, deletion_comment = False)

shrshindeMSFT:
Yes it's still need to be commented

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+
+      Object.values(FrameContexts).forEach(context => {
+        it(`pages.tabs.navigateToTab should register the navigateToTab action when initialized with ${context} context`, async () => {
+          await utils.initializeWithContext(context);
+          pages.tabs.navigateToTab(tabInstance);
+          const navigateToTabMsg = utils.findMessageByFunc('navigateToTab');
+          expect(navigateToTabMsg).not.toBeNull();
+          expect(navigateToTabMsg.args[0]).toBe(tabInstance);
+        });
+        it(`pages.tabs.navigateToTab should register the navigateToTab action when initialized with ${context} context - success case`, async () => {
```
erinha, line 551: I don't see a test case that validates that an error is thrown with message: "Invalid internalTabInstanceId and/or channelId were/was provided"

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+          await utils.initializeWithContext(context);
+          const promise = pages.tabs.getTabInstances();
+          let message = utils.findMessageByFunc('getTabInstances');
+
+          utils.respondToMessage(message, expectedTabInstanceParameters);
+          expect(message).not.toBeNull();
+          expect(promise).resolves.toBe(expectedTabInstanceParameters);
+        });
+
+        it(`pages.tabs.getTabInstances should be undefined getTabInstance when parameters are passed and  initialized with ${context} context`, async () => {
```
erinha, line 585: The test says it should be "undefined getTabInstance", but I don't see that as the expected value? (Unless I'm missing something.) Also there is an extra space between "and" and "initialized"

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+          await utils.initializeWithContext(context);
+          const promise = pages.tabs.getMruTabInstances();
+          let message = utils.findMessageByFunc('getMruTabInstances');
+
+          utils.respondToMessage(message, expectedTabInstanceParameters);
+          expect(message).not.toBeNull();
+          expect(promise).resolves.toBe(expectedTabInstanceParameters);
+        });
+
+        it(`pages.tabs.getMruTabInstances should be undefined getTabInstance when parameters are passed and  initialized with ${context} context`, async () => {
```
erinha, line 630: Same comment as in getTabInstances:
The test says it should be "undefined getTabInstance", but I don't see that as the expected value? (Unless I'm missing something.) Also there is an extra space between "and" and "initialized"

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+
+          utils.respondToMessage(message);
+          expect(message).not.toBeNull();
+          expect(promise).resolves.toBeUndefined();
+        });
+      });
+    });
+
+    describe('Testing pages.tabs.isSupported function', () => {
+      it('pages.tabs.isSupported should return false if the runtime says pages.tabs is not supported', () => {
```
erinha, line 653: Can you add an additional test that checks that `pages.tabs.isSupported` returns false when `supports: {}` please?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+    });
+  });
+
+  describe('Testing pages.config namespace', () => {
+    describe('Testing pages.config.initialize function', () => {
+      it('pages.config.initialize should successfully register settings.save and settings.remove handler', async () => {
+        pages.config.initialize();
+        const messageForSettingeSaveHandler = utils.findMessageByFunc('settings.save');
+        const messageForSettingeRemoveHandler = utils.findMessageByFunc('settings.remove');
+        expect(messageForSettingeSaveHandler).toBeNull();
```
erinha, line 674: I'm a bit confused about how this validates that pages.config.initialize registered the settings.save and settings.remove handlers.

erinha: For example, if I comment out `pages.config.initialize` and run this test, it still passes.

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+            );
+          });
+        }
+      });
+    });
+
+    describe('Testing pages.config.setConfig function', () => {
+      const allowedContexts = [FrameContexts.content, FrameContexts.settings, FrameContexts.sidePanel];
+
+      let settingsObj: pages.InstanceConfig = {
```
erinha, line 724: Please use `const` instead of `let`. Incidentally when I synced your branch, my VS Code complained about this via the Prettier extension. Do you not have that installed?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+          'The library has not yet been initialized',
+        );
+      });
+
+      Object.values(FrameContexts).forEach(context => {
+        if (allowedContexts.some(allowedContexts => allowedContexts === context)) {
+          it(`pages.config.setConfig should successfully set settings when initialized with ${context} context`, async () => {
+            await utils.initializeWithContext(context);
+            pages.config.setConfig(settingsObj);
+            let message = utils.findMessageByFunc('settings.setSettings');
```
erinha, line 742: `const` instead of `let` please

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+
+      let settingsObj: pages.InstanceConfig = {
+        suggestedDisplayName: 'someSuggestedDisplayName',
+        contentUrl: 'someContentUrl',
+        websiteUrl: 'someWebsiteUrl',
+        entityId: 'someEntityId',
+      };
+
+      it('pages.config.setConfig should not allow calls before initialization', () => {
+        expect(() => pages.config.setConfig({} as pages.InstanceConfig)).rejects.toThrowError(
```
erinha, line 732: It's interesting. I was curious if calling `pages.config.setConfig` with `{} as pages.InstanceConfig` was valid and it turns out it is (which makes me think we should have a test that takes an empty `pages.InstanceConfig`). It's a little surprising because `pages.InstanceConfig` has a required property of `contentUrl`, but you can pass an empty object without that required property. I wonder if that should somehow result in an error?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+              )}. Current context: "${context}".`,
+            );
+          });
+        }
+      });
+    });
+
+    describe('Testing pages.config.isSupported function', () => {
+      it('pages.config.isSupported should return false if the runtime says pages.config is not supported', () => {
+        utils.setRuntimeConfig({ apiVersion: 1, supports: { pages: {} } });
```
erinha, line 983: Can you please add the test case that verifies `pages.config.isSupported` returns false when `supports: {}`?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+          let navigateBackMessage = utils.findMessageByFunc('navigateBack');
+          expect(navigateBackMessage).not.toBeNull();
+          expect(handlerInvoked).toBe(true);
+        });
+      });
+    });
+
+    describe('Testing pages.backStack.isSupported function', () => {
+      it('pages.backStack.isSupported should return false if the runtime says pages.backStack is not supported', () => {
+        utils.setRuntimeConfig({ apiVersion: 1, supports: { pages: {} } });
```
erinha, line 1059: Can you please add a test case that verifies that `pages.backStack.isSupported` returns false when we set the runtime config to `supports: {}`?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+              )}. Current context: "${context}".`,
+            );
+          });
+        }
+      });
+    });
+
+    describe('Testing pages.fullTrust.isSupported function', () => {
+      it('pages.fullTrust.isSupported should return false if the runtime says pages.fullTrust is not supported', () => {
+        utils.setRuntimeConfig({ apiVersion: 1, supports: { pages: {} } });
```
erinha, line 1129: Can you please add a test that `pages.fullTrust.isSupported` returns false when we set the runtime config to `supports: {}`?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -210,42 +335,917 @@ describe('AppSDK-TeamsAPIs', () => {
...
+              )}. Current context: "${context}".`,
+            );
+          });
+        }
+      });
+    });
+
+    describe('Testing pages.appButton.isSupported function', () => {
+      it('pages.appButton.isSupported should return false if the runtime says pages.appButton is not supported', () => {
+        utils.setRuntimeConfig({ apiVersion: 1, supports: { pages: {} } });
```
erinha, line 1238: I think you know what my comment is here. :)

erinha (changes requested):
Thanks for adding all these tests and consolidating the 3 pages test files into one! Left several comments to  address.

erinha:
BTW,  I don't think I left comments for all the places that are using `let` but should be using `const`. It would be good to double-check with the Prettier extension installed.

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Added in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Added in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Added in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Added in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Apparently when passed in empty object we do get empty object in return adding that test case doesn't checks or verifies anything.

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Added in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Fixed in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Fixed in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: Added in the latest commit

packages/teams-js/test/public/pages.spec.ts
shrshindeMSFT: We already have the test for default behavior, added the `isLegacy` to true in latest commit.

erinha (dismissed):
Looks great, thanks for making the changes and adding these tests!

jadahiya-MSFT (approved):

erinha (approved):

PR #985, opened 2022-03-10
Lakhveer/callbacks

lakhveerk:
In appWindow, the postMessage functions use promise and wait for a response before triggering their callback, but these promises never get resolved. Reverting this back to callbacks like we had in v1.

### DIFF
--- a/change/@microsoft-teams-js-2e19994e-bf6c-4039-96ba-e01f9c9387e0.json
+++ b/change/@microsoft-teams-js-2e19994e-bf6c-4039-96ba-e01f9c9387e0.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "In `appWindow.ts` file, `ChildAppWindow` and `ParentAppWindow` are convereted back to synchronous calls because the promise was never being resolved.",
+  "comment": "In `appWindow.ts` file, `ChildAppWindow` and `ParentAppWindow` are converted back to synchronous calls because the promise was never being resolved.",
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-2e19994e-bf6c-4039-96ba-e01f9c9387e0.json
+++ b/change/@microsoft-teams-js-2e19994e-bf6c-4039-96ba-e01f9c9387e0.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "In `appWindow.ts` file, `ChildAppWindow` and `ParentAppWindow` are convereted back to synchronous calls because the promise was never being resolved.",
+  "packageName": "@microsoft/teams-js",
+  "email": "lakhveerkaur@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/public/appWindow.ts
+++ b/packages/teams-js/src/public/appWindow.ts
@@ -30,7 +30,7 @@ export interface IAppWindow {
 
 export class ChildAppWindow implements IAppWindow {
   /**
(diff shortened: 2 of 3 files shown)


change/@microsoft-teams-js-2e19994e-bf6c-4039-96ba-e01f9c9387e0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "In `appWindow.ts` file, `ChildAppWindow` and `ParentAppWindow` are convereted back to synchronous calls because the promise was never being resolved.",
```
TrevorJoelHarris, line 3: extra 'e' in converted

packages/teams-js/src/public/appWindow.ts
```diff
@@ -36,39 +31,20 @@ export interface IAppWindow {
...
-  }
-  public postMessageHelper(message: any): Promise<void> {
-    return new Promise<void>(resolve => {
-      resolve(send('messageForChild', message));
-    });
+    sendMessageToParent('messageForChild', [message], onComplete ? onComplete : getGenericOnCompleteHandler());
   }
-
+  /**
+   * Add a listener that will be called when an event is received from this AppWindow.
```
TrevorJoelHarris, line 43: Would this more correctly be "Add a listener that will be called when an event is received from the ChildAppWindow"?

packages/teams-js/src/public/appWindow.ts
```diff
@@ -85,39 +61,21 @@ export class ParentAppWindow implements IAppWindow {
...
-  }
-  public postMessageHelper(message: any): Promise<void> {
-    return new Promise<void>(resolve => {
-      resolve(send('messageForParent', message));
-    });
+    sendMessageToParent('messageForParent', [message], onComplete ? onComplete : getGenericOnCompleteHandler());
   }
 
+  /**
+   * Add a listener that will be called when an event is received from this AppWindow.
```
TrevorJoelHarris, line 74: Would it be more correct to say "Add a listener that will be called when an event is received from the ParentAppWindow"?

packages/teams-js/src/public/appWindow.ts
```diff
@@ -36,39 +31,20 @@ export interface IAppWindow {
...
-  /**
-   * @deprecated
-   * As of 2.0.0-beta.1, please use {@link ChildAppWindow.postMessage ChildAppWindow.postMessage(message: any): Promise\<void\>} instead.
-   * @param message - The message to send
-   * @param onComplete - The deprecated way of invoking a callback to know if the postMessage has been success/failed.
+   * @param onComplete - The callback to know if the postMessage has been success/failed.
    */
-  public postMessage(message: any, onComplete: (status: boolean, reason?: string) => void): void;
-  public postMessage(message: any, onComplete?: (status: boolean, reason?: string) => void): Promise<void> {
+  public postMessage(message: any, onComplete?: (status: boolean, reason?: string) => void): void {
```
TrevorJoelHarris, line 38: should onComplete be a required parameter? It looks like in the old version of this class it was required, so I was curious about whether what you are doing on line 40 (getGenericOnCompleteHandler) was actually changing how this worked?

lakhveerk: I cannot seem to find a release where it was required. Do you remember what release was it?

TrevorJoelHarris: I was going off of the old signature (line 49 before your change) where onComplete was a required parameter

lakhveerk: This was optional in v1, like we talked, we are going to keep it optional going forward.

TrevorJoelHarris (approved):

PR #986, opened 2022-03-11
public/authentication.spec.ts

shrshindeMSFT:
- Added missing unit test cases for various functions.
- Adding missing unit test cases for various `FrameContexts`
- Adding missing unit test cases for missing `hostClientType`


packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -31,41 +51,221 @@ describe('authentication', () => {
...
-      url: 'https://someurl/',
-      width: 100,
-      height: 200,
-    };
+  describe('Testing authentication.initialize function', () => {
+    it('authentication.initialize should successfully register authentication.authenticate.success/failure handler', async () => {
+      authentication.initialize();
+      const messageForAuthenticationSuccess = utils.findMessageByFunc('authentication.authenticate.success');
+      const messageForAuthenticationFailure = utils.findMessageByFunc('authentication.authenticate.failure');
+      expect(messageForAuthenticationSuccess).toBeNull();
```
aeunms, line 59: (I'm asking because I'm genuinely curious, I'm just trying to learn more about this area:) why would we expect the message to be null here? I looked through the files and it seems like we would want that message to have been sent and kept by utils.

shrshindeMSFT: When we initialize something we don't respond with any function we just tell to initialize it and not send any message.

Jessica-Koch (approved):

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -90,647 +412,480 @@ describe('authentication', () => {
...
-    expect(windowAssignSpyCalled).toBe(true);
-  });
+      it(`authentication.getUser should throw error in getting user profile with ${context} context`, async () => {
+        await utils.initializeWithContext(context);
+        const promise = authentication.getUser();
+        const message = utils.findMessageByFunc('authentication.getUser');
+        expect(message).not.toBeNull();
+        expect(message.id).toBe(1);
+        expect(message.args[0]).toBe(undefined);
+        utils.respondToMessage(message, false, mockResult);
```
aeunms, line 627: I'm a bit confused by these couples tests. The test name in 620 and 631 says that getUser should throw error and also successfully execute in the same context. The only difference I see is the difference in respondToMessage. Should these tests be named something like "getUser should ____ with ____ context when _____"? (Although I'm still not sure I understand what the when _____ is in either of these)

shrshindeMSFT: In respondToMessage we set it to false and promise throws the error. Where as in 631 test cases it resolves the error.

jadahiya-MSFT (approved):

PR #987, opened 2022-03-11
Add missing hostClientType value in comment

steffenloesch:
The enum in `constants.ts` has the `surfaceHub` value, but the comment in `interfaces.ts` does not. This change fixes the comment.


aeunms (approved):
Thank you for the catch! 👍🏼

lakhveerk (approved):

PR #989, opened 2022-03-11
Lakhveer/open

lakhveerk:
1. Dialog capability has been split into a main capability (dialog) for supporting HTML-based dialogs and a subcapability dialog.bot for bot based dialogs. For now, dialog capability does not support adaptive card based dialogs
2. Changed Dialog.open function is changed to
- take UrlDialogInfo instead of DialogInfo to enforce only HTML based dialogs
-  take messageFromChildHandler optional parameter,
- return equivalent of ChildAppWindow.postMessage (instead of a childAppWindow),
- submitHandler takes a single object containing both err and result
- dialog.bot.open has the same function signature except it takes BotUrlDialogInfo instead of UrlDialogInfo
4. tasks.startTask works like the same but does not call dialog.open. 
5. Updated open function in test-app according to the new parameter and return function. 
6. Updated unit tests for open function because of different submitHandler parameters.

### DIFF
--- a/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
+++ b/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Dialog.open function takes an extra optionl parameter named messageForChildHandler and retuns a handler back instead of a ChildAppWindow. ",
+  "packageName": "@microsoft/teams-js",
+  "email": "lakhveerkaur@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
+++ b/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
@@ -1,6 +1,6 @@
 {
   "type": "major",
-  "comment": "Dialog.open function takes an extra optionl parameter named messageForChildHandler and retuns a handler back instead of a ChildAppWindow. ",
+  "comment": "Dialog.open function takes an extra optionl parameter named messageFromChildHandler and returns a handler back instead of a ChildAppWindow.",
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
+++ b/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
@@ -1,6 +1,12 @@
 {
   "type": "major",
-  "comment": "Dialog.open function takes an extra optionl parameter named messageFromChildHandler and returns a handler back instead of a ChildAppWindow.",
+  "comment": [
+    "1. Dialog capability has been split into a main capability (dialog) for supporting HTML-based dialogs and a subcapability dialog.bot for bot based dialogs. For now, dialog capability does not support adaptive card based dialogs",
+    "\n  2. dialog.open has been changed to take UrlDialogInfo instead of DialogInfo to enforce only HTML based dialogs",
+    "\n  3. dialog.open takes an extra optional parameter named messageFromChildHandler which is triggered if dialog sends a message to the app",
+    "\n  4. dialog.open returns a function that can be used to send messages to the dialog instead of returning a ChildAppWindow",
+    "\n  5. dialog.bot.open has the same function signature except it takes BotUrlDialogInfo instead of UrlDialogInfo"
+  ],
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
+++ b/change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
@@ -2,10 +2,11 @@
   "type": "major",
   "comment": [
     "1. Dialog capability has been split into a main capability (dialog) for supporting HTML-based dialogs and a subcapability dialog.bot for bot based dialogs. For now, dialog capability does not support adaptive card based dialogs",
-    "\n  2. dialog.open has been changed to take UrlDialogInfo instead of DialogInfo to enforce only HTML based dialogs",
-    "\n  3. dialog.open takes an extra optional parameter named messageFromChildHandler which is triggered if dialog sends a message to the app",
-    "\n  4. dialog.open returns a function that can be used to send messages to the dialog instead of returning a ChildAppWindow",
-    "\n  5. dialog.bot.open has the same function signature except it takes BotUrlDialogInfo instead of UrlDialogInfo"
+    "\n  2. dialog.open takes a UrlDialogInfo instead of DialogInfo to enforce only HTML based dialogs",
+    "\n  3. callback submitHandler takes a single object parameter containing both error and result",
+    "\n  4. dialog.open takes one more optional parameter named messageFromChildHandler which is triggered if dialog sends a message to the app",
+    "\n  5. dialog.open returns a function that can be used to send messages to the dialog instead of returning a ChildAppWindow",
+    "\n  6. dialog.bot.open has the same function signature except it takes BotUrlDialogInfo instead of UrlDialogInfo"
   ],
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -18,33 +18,33 @@ import { runtime } from './runtime';
  *
  * @beta
  */
-export interface Result {
+export interface SdkResponse {
   err: string;
   result: string | object;
 }
 
 export namespace dialog {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  type PostMessageChannel = (message: any) => void;
-  type DialogSubmitHandler = (result: Result) => void;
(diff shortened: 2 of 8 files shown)


change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Dialog.open function takes an extra optionl parameter named messageForChildHandler and retuns a handler back instead of a ChildAppWindow. ",
```
lakhveerk, line 3: I will fix the `return` spelling. just noticed now

jadahiya-MSFT: You can do that inside of Github itself if you'd like

packages/teams-js/src/public/dialog.ts
```diff
@@ -16,21 +18,45 @@ import { runtime } from './runtime';
...
-  ): IAppWindow {
+    submitHandler?: DialogSubmitHandler,
+    messageForChildHandler?: PostMessageChannel,
+  ): PostMessageChannel {
     ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
 
-    sendMessageToParent('tasks.startTask', [dialogInfo], submitHandler);
-    return new ChildAppWindow();
+    if (messageForChildHandler) {
+      registerHandler('messageForParent', messageForChildHandler);
```
TrevorJoelHarris, line 47: I think there's a separate item tracking this, but just in case: we should make sure to also unregister the handler when dealing with the submit call

packages/teams-js/src/public/tasks.ts
```diff
@@ -27,7 +29,10 @@ export namespace tasks {
     taskInfo: TaskInfo,
     submitHandler?: (err: string, result: string | object) => void,
   ): IAppWindow {
-    return dialog.open(getDialogInfoFromTaskInfo(taskInfo), submitHandler);
+    ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
```
TrevorJoelHarris, line 32: It might be slightly less error prone to still call dialog.open here and then ignore its return value and just return ChildAppWindow out of this. It would keep the two flows (task.startTask and dialog.open) more in sync which might make it easier for people to convert in the future. Code I'm imagining here would be something like:

dialog.open(getDialogInfoFromTaskInfo(taskInfo), submitHandler ? (res: Result) => submitHandler(result.err, result.result) : undefined);
return new ChildAppWindow();

What do you think?

packages/teams-js/src/public/dialog.ts
```diff
@@ -16,21 +18,45 @@ import { runtime } from './runtime';
...
  * @beta
  */
+export interface Result {
+  err: string;
+  result: string | object;
+}
+
 export namespace dialog {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  type PostMessageChannel = (message: any) => void;
```
TrevorJoelHarris, line 28: Should these types also be exported? And if so, do we normally put them in the capability files, or do we have a place specifically for types? (Same question for Result above)

packages/teams-js/src/public/dialog.ts
```diff
@@ -16,21 +18,45 @@ import { runtime } from './runtime';
  *
  * @beta
  */
+export interface Result {
```
TrevorJoelHarris, line 21: I know you took this name (Result) from some of the docs we looked at last week, but do you think there's a better name we could use? Result.result feels a little odd. Maybe something like SdkResponse? ErrorOrResult?

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -114,9 +114,10 @@ describe('Dialog', () => {
 
       let callbackCalled = false;
       const dialogInfo: DialogInfo = {};
-      dialog.open(dialogInfo, (err, result) => {
-        expect(err).toBeNull();
-        expect(result).toBe('someResult');
+      dialog.open(dialogInfo, resultObj => {
```
TrevorJoelHarris, line 117: Do we have unit tests for any of the new open functionality as well? (Passing in messageForChildHandler, making sure the new return value does what we expect it to, etc.)

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -131,9 +132,9 @@ describe('Dialog', () => {
 
       let callbackCalled = false;
       const dialogInfo: DialogInfo = {};
-      dialog.open(dialogInfo, (err, result) => {
-        expect(err).toBe('someError');
-        expect(result).toBeUndefined();
+      dialog.open(dialogInfo, resultObj => {
```
TrevorJoelHarris, line 135: I don't know much about our end to end testing: do we also need to update those tests as part of this change? Can we do it in the same checkin as this?

TrevorJoelHarris (changes requested):
Left some comments for you, let me know when it's ready for re-review

packages/teams-js/src/public/dialog.ts
lakhveerk: Yeah, I will do it in a separate call :)

packages/teams-js/src/public/dialog.ts
lakhveerk: I have seen many exported types and interfaces inside capability files. To name a few:  bot, authentication, mail video. There are three interfaces' files for public, private and internal. Not sure what the rule is. 
IS Erin good person to ask about this?

lakhveerk: and yes, I think, we might need to use the types outside, and need export. I will look more into this

packages/teams-js/test/public/dialog.spec.ts
lakhveerk: Yes, and it is on my list. I did add the new open function to test-app. I will add the tests on the other end as well but before that, I need to rename existing tests (renaming will fail the pipelines, so I have to figure some way).

packages/teams-js/src/public/tasks.ts
lakhveerk: That makes sense. I will make the change.

packages/teams-js/test/public/dialog.spec.ts
lakhveerk: I added some more unit tests. Please check them out.

change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Dialog.open function takes an extra optionl parameter named messageFromChildHandler and returns a handler back instead of a ChildAppWindow.",
```
TrevorJoelHarris, line 3: "optional" is missing an a

packages/teams-js/src/public/dialog.ts
```diff
@@ -16,21 +17,44 @@ import { runtime } from './runtime';
...
 export namespace dialog {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  export type PostMessageChannel = (message: any, onComplete?: (status: boolean, reason?: string) => void) => void;
+  export type DialogSubmitHandler = (result: SdkResponse) => void;
   /**
    * Allows an app to open the dialog module.
    *
-   * @param dialogInfo - An object containing the parameters of the dialog module
+   * @param urlDialogInfo - An object containing the parameters of the dialog module
    * @param submitHandler - Handler to call when the task module is completed
```
TrevorJoelHarris, line 33: This comment should probably not say task module anymore. Maybe something like "This handler is called when the dialog has been submitted or closed"

packages/teams-js/src/public/dialog.ts
```diff
@@ -16,21 +17,44 @@ import { runtime } from './runtime';
...
+   * @param messageFromChildHandler - Handler that triggers if dialog tries to send a message to the app.
+   *
+   * @returns a Handler that is triggerd to send a message to the dialog.
```
TrevorJoelHarris, line 36: Maybe this comment should be "@returns a function that can be used to send messages to the dialog"

packages/teams-js/src/public/tasks.ts
```diff
@@ -55,34 +70,36 @@ export namespace tasks {
...
+      title: taskInfo.title,
+      fallbackUrl: taskInfo.fallbackUrl,
+      completionBotId: taskInfo.completionBotId,
+    };
+    return botUrldialogInfo;
   }
 }
+
+export function getDefaultSizeIfNotProvided(taskInfo: TaskInfo): TaskInfo {
+  taskInfo.height = taskInfo.height ? taskInfo.height : TaskModuleDimension.Medium;
```
TrevorJoelHarris, line 102: I think in the doc we shared with partners we mentioned that we would default to small. I don't think it's particularly important that we stick to that, but why the change?

TrevorJoelHarris (changes requested):
Let me know when ready for sign off

packages/teams-js/src/public/tasks.ts
lakhveerk: Oh, I am sorry, I missed that. I will change it to small

packages/teams-js/src/public/dialog.ts
lakhveerk: Yes, makes more sense!

packages/teams-js/src/public/dialog.ts
lakhveerk: Will change!

change/@microsoft-teams-js-9c1b77aa-a0e6-4099-9eae-9fff30c8f94e.json
lakhveerk: Thank you for catching

lakhveerk: My excuse is English is not my first language ;)

TrevorJoelHarris (approved):

PR #991, opened 2022-03-12
Lakhveer/move functions:  Copied ParentAppWindow functionality into dialog capability

lakhveerk:
In Dialogs, 
ParentAppWindow.postMessage becomes dialog.sendMessageToParent(message: any): void
ParentAppWindow.addEventListener becomes dialog.registerOnMessageFromParent

### DIFF
--- a/apps/teams-test-app/src/components/DialogAPIs.tsx
+++ b/apps/teams-test-app/src/components/DialogAPIs.tsx
@@ -190,7 +190,7 @@ const DialogAPIs = (): ReactElement => {
       name: 'registerForParentMessage_v2',
       title: 'registerForParentMessage_v2',
       onClick: async setResult => {
-        dialog.registerOnMessageFromParent('message', (message: string) => {
+        dialog.registerOnMessageFromParent((message: string) => {
           setResult(message);
         });
         return 'Completed';
--- a/apps/teams-test-app/src/components/DialogAPIs.tsx
+++ b/apps/teams-test-app/src/components/DialogAPIs.tsx
@@ -147,6 +147,31 @@ const DialogAPIs = (): ReactElement => {
       },
     });
 
+  const SendMessageToParent_v2 = (): ReactElement =>
+    ApiWithTextInput<string>({
+      name: 'sendMessageToParent_v2',
+      title: 'sendMessageToParent_v2',
+      onClick: {
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        validateInput: () => {},
+        submit: async (message, setResult) => {
+          const onComplete = (status: boolean, reason?: string): void => {
+            if (!status) {
+              if (reason) {
+                setResult(JSON.stringify(reason));
+              } else {
+                setResult("Status is false but there's no reason?! This shouldn't happen.");
+              }
+            } else {
+              setResult('Message sent to parent');
+            }
+          };
+          dialog.sendMessageToParentFromDialog(message, onComplete);
+          return '';
+        },
+      },
+    });
+
   const RegisterForParentMessage = (): ReactElement =>
     ApiWithoutInput({
       name: 'registerForParentMessage',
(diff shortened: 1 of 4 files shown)


apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -147,6 +147,31 @@ const DialogAPIs = (): ReactElement => {
       },
     });
 
+  const SendMessageToParent_v2 = (): ReactElement =>
```
TrevorJoelHarris, line 150: I don't know much about this file: Why make _v2 versions of the functions rather than convert the existing functions to use the new dialog.* functions you're adding?

TrevorJoelHarris: (I'm not saying what you've done is wrong, I'd just love to learn more about what this file is used for)

packages/teams-js/src/public/dialog.ts
```diff
@@ -62,6 +65,34 @@ export namespace dialog {
     sendMessageToParent('tasks.completeTask', [result, Array.isArray(appIds) ? appIds : [appIds]]);
   }
 
+  /**
```
TrevorJoelHarris, line 68: @erinha Are there documentation guidelines we can use to verify we've got documentation that meets our standards?

packages/teams-js/src/public/dialog.ts
```diff
@@ -62,6 +65,34 @@ export namespace dialog {
...
+
+  /**
+   * Fucntion to call when an event is received from the Parent
+   *
+   * @param type - The event to listen to. Currently the only supported type is 'message'.
+   * @param listener - listener - The listener that will be called.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  export function registerOnMessageFromParent(type: string, listener: (message: any) => void): void {
+    if (type === 'message') {
```
TrevorJoelHarris, line 91: Would it make sense to remove the "type" parameter? It doesn't seem to be doing very much for us

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -220,4 +221,45 @@ describe('Dialog', () => {
       expect(submitMessage.args).toEqual(['someResult', ['someAppId']]);
     });
   });
+  describe('sendMessageToParentFromDialog', () => {
```
TrevorJoelHarris, line 224: can we verify that sendMessageToParent was called correctly using unit tests?

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -220,4 +221,45 @@ describe('Dialog', () => {
...
+            expect(() => dialog.sendMessageToParentFromDialog('message', emptyCallback)).toThrowError(
+              `This call is only allowed in following contexts: ${JSON.stringify(
+                allowedContexts,
+              )}. Current context: "${frameContext}".`,
+            );
+          });
+        }
+      });
+  });
+  it('registerOnMessageFromParent', () => {
```
TrevorJoelHarris, line 258: this should probably be in its own describe block, and the test itself should have a more descriptive name like the others you wrote above.

TrevorJoelHarris (changes requested):
Let me know when ready for review again

packages/teams-js/src/public/dialog.ts
lakhveerk: We have used tsdoc style in past for document comments. https://tsdoc.org/
But it might have changed. I am open Erin's advice! :)

apps/teams-test-app/src/components/DialogAPIs.tsx
lakhveerk: I think it would make sense to have two separate files for tasks and dialogs for testing. I am doing v2, so the e2e doesn't break. 
Once I am done with core work of changing APIs, I will create a PR for e2e testing. The benefit is, I would not have to deal with breaking e2e tests in every PR. 
If I don't do it, I would have to open a PR on host side, every time

packages/teams-js/src/public/dialog.ts
lakhveerk: Yeah, I think we can delete it for this function.
I read it wrong, did all the research and then figured you are asking to change it in our new function, not in the appWindow file. 😂

lakhveerk: Does it make sense to ensureInitialized here

packages/teams-js/test/public/dialog.spec.ts
lakhveerk: It would be nice to talk about it. Are we worrying because these calls would be generated from a different instance of appSDK

TrevorJoelHarris: It looks like you have tests for this on 239, maybe I missed them last time

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: Please do follow up with Erin and ask if we need to do anything to get the comments looked at by our documentation people :)

TrevorJoelHarris (approved):

PR #994, opened 2022-03-14
Lakhveer/updateResize Adding a sub capability named update in dialogs.

lakhveerk:
1. Created a sub capability named update in dialogs. Moved resize functions in there and added isSupported function. 
2. Made changes in tasks.update, so it would keep working as it was before.
3. Fixed Unit testing and test-app file to support the sub capability.

### DIFF
--- a/apps/teams-test-app/src/components/DialogAPIs.tsx
+++ b/apps/teams-test-app/src/components/DialogAPIs.tsx
@@ -69,24 +69,6 @@ const DialogAPIs = (): ReactElement => {
       },
     });
 
-  const UpdateTaskModule = (): ReactElement =>
-    ApiWithTextInput<TaskInfo>({
-      name: 'updateTaskModule',
-      title: 'Update Task Module',
-      onClick: {
-        validateInput: input => {
-          if (input.height === undefined && input.width === undefined) {
-            throw new Error('Height and width undefined');
-          }
-        },
-        submit: async (taskInfo, setResult) => {
-          tasks.updateTask(taskInfo);
-          setResult('Teams client SDK call tasks.updateTask was called');
-          return '';
-        },
-      },
-    });
-
   const ResizeDialog = (): ReactElement =>
     ApiWithTextInput<DialogSize>({
       name: 'dialogResize',
@@ -309,7 +291,6 @@ const DialogAPIs = (): ReactElement => {
       <CheckDialogCapability />
       <StartTask />
       <OpenDialog />
-      <UpdateTaskModule />
       <ResizeDialog />
       <CheckDialogResizeCapability />
       <SubmitDialogWithInput />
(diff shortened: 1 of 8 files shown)


change/@microsoft-teams-js-f4f055db-c85c-4ea2-b470-278759a1aafb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "In `dialog.ts` moved `Resize` functionality in a subcapability named `update`.",
```
aeunms, line 3: I think it would be good to specify the capability since the client would be calling it from the capability rather than looking at the file name. So maybe something like:

"`dialog.resize()` function has been moved to a new dialog subcapability and is now `dialog.update.resize()`."

aeunms, line 3: I think it would be good to specify the capability since the client would be calling it from the capability rather than looking at the file name. So maybe something like:

"`dialog.resize()` function has been moved to a new dialog subcapability and is now `dialog.update.resize()`."

aeunms, line 3: Another change that looks like it should be included is that the input type is now different. The parameter has changed from type DialogInfo to DialogSize.

aeunms, line 3: Another change that should be included is that the input type is now different. The parameter has changed from type DialogInfo to DialogSize.

packages/teams-js/src/public/tasks.ts
```diff
@@ -85,4 +92,10 @@ export namespace tasks {
       return TaskModuleDimension.Small;
     }
   }
+
+  export function getDefaultSizeIfNotProvided(taskInfo: TaskInfo): TaskInfo {
```
TrevorJoelHarris, line 96: Same question I left on the other PR: we had mentioned in the doc we showed to partners that we would default to Small. Why the change?

TrevorJoelHarris (changes requested):
Just waiting on discussion about default dialog size, then I will sign off

packages/teams-js/src/public/tasks.ts
lakhveerk: It was not intentional. Sorry, I missed it from the doc. Will change to small

apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -68,21 +69,41 @@ const DialogAPIs = (): ReactElement => {
       },
     });
 
+  const UpdateTaskModule = (): ReactElement =>
+    ApiWithTextInput<TaskInfo>({
+      name: 'updateTaskModule',
+      title: 'Update Task Module',
+      onClick: {
```
jadahiya-MSFT, line 76: This seems unnecessary since you accomplish the same goal with the `withCallback` call

jadahiya-MSFT (changes requested):
Everything looks good, I just think we can remove `UpdateTaskModule` UI since it works the same as `ResizeDialog`

apps/teams-test-app/src/components/DialogAPIs.tsx
lakhveerk: Deleted. I think it is a little confusing that we are using 'withPromise' for v2 even though, we actually are using callbacks.

jadahiya-MSFT (approved):

lakhveerk:
> 

I have changed the size to small :) and it is already merged as part of open API

TrevorJoelHarris (approved):

PR #995, opened 2022-03-15
public/pages.spec.ts.frameless

shrshindeMSFT:
- Adding Frameless test suite case for all the pages functions.


aeunms:
Nit: branch naming usually follows the <alias>/<feature> convention. This doesn't matter for the actual code but just FYI

Jessica-Koch (commented):
Is all of this new or are there white space changes?  Would it be possible to point out the parts that are different, it's really hard to review something this big

shrshindeMSFT:
> Is all of this new or are there white space changes? Would it be possible to point out the parts that are different, it's really hard to review something this big

aeunms:
Seconded that I'm having a hard time reviewing due to the size 😅 I ended up utilizing my VS Code GitHub extension which does a better job showing the differences. @shrshindeMSFT Isn't the set of framed tests the same from before, just a tab added to each line?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -156,96 +163,1296 @@ describe('AppSDK-TeamsAPIs', () => {
...
+    afterEach(() => {
+      // Reset the object since it's a singleton
+      if (app._uninitialize) {
+        app._uninitialize();
       }
     });
 
-    it('should successfully navigate cross-origin', async () => {
-      await utils.initializeWithContext(FrameContexts.content);
+    describe('Testing pages.returnFocus function', () => {
```
Jessica-Koch, line 277: :nit - I'm not sure it's necessary to preface all the tests with 'testing' since this is a test file, testing is implied

aeunms: I don't think I've seen other PRs use this convention, might be a good thing to bring up on our next team sync to remind folks if this has been decided to be our standard

shrshindeMSFT: We have discussed this in PR 910 - https://github.com/OfficeDev/microsoft-teams-library-js/pull/910
Can you please have a look at the comments?

aeunms: I can't seem to find the thread on the "Testing *" conversation, would you be able to put up a screenshot of it since deeplinking doesn't work well in Github resolved comment threads? 

Also, I believe you, I think we just have to make sure everyone else follows the convention 😋

Jessica-Koch: I agree with @aeunms , I've updated many test files and this is the first that I've heard of this, and I have not followed this convention in any test files, so we should talk about it

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -156,96 +163,1296 @@ describe('AppSDK-TeamsAPIs', () => {
...
+            expect(navigateCrossDomainMessage).not.toBeNull();
+            framelessPostMocks.respondToMessage({
+              data: {
+                id: navigateCrossDomainMessage.id,
+                args: ['https://valid.origin.com'],
+              },
+            } as DOMMessageEvent);
+
+            await expect(promise).resolves.not.toThrow();
+            await expect(promise).resolves.toBeUndefined();
```
Jessica-Koch, line 523: stupid question, how does this work?  wouldn't the promise already be resolved at the second test, also is the second test that useful?

Jessica-Koch (approved):

jadahiya-MSFT (approved):

PR #999, opened 2022-03-16
Improvements to openDownloadFolder API

HarshJain27:
This PR contains improvements to openDownloadFolder API which is added in [this](https://github.com/OfficeDev/microsoft-teams-library-js/pull/858) PR. The improvements are :
1. Added file path open support 
2. Added callback to openDownloadFolder

### DIFF
--- a/src/private/files.ts
+++ b/src/private/files.ts
@@ -487,11 +487,17 @@ export namespace files {
    * @private
    * Hide from docs
    *
-   * Open download preference folder
+   * Open download preference folder or filePath if its value is non-null
+   * @param filePath: Path of the file which should be opened
+   * @param callback Callback that will be triggered post open download folder/path
    */
-  export function openDownloadFolder(): void {
+  export function openDownloadFolder(filePath: string, callback: (error?: SdkError) => void): void {
     ensureInitialized(FrameContexts.content);
 
-    sendMessageToParent('files.openDownloadFolder', []);
+    if (!callback) {
+      throw new Error('[files.openDownloadFolder] Callback cannot be null');
+    }
+
+    sendMessageToParent('files.openDownloadFolder', [filePath], callback);
   }
 }
--- a/src/private/files.ts
+++ b/src/private/files.ts
@@ -487,8 +487,8 @@ export namespace files {
    * @private
    * Hide from docs
    *
-   * Open download preference folder or filePath if its value is non-null
-   * @param filePath: Path of the file which should be opened
+   * Open download preference folder or folder containing filePath if its value is non-null
+   * @param filePath: Path of the file whose containing folder should be opened
    * @param callback Callback that will be triggered post open download folder/path
    */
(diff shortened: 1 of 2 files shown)


src/private/files.ts
```diff
@@ -487,11 +487,17 @@ export namespace files {
    * @private
    * Hide from docs
    *
-   * Open download preference folder
+   * Open download preference folder or filePath if its value is non-null
+   * @param filePath: Path of the file which should be opened
```
ydogandjiev, line 491: This is very confusing. Will this API be opening the file? If so, that's a huge security risk.

ydogandjiev (changes requested):

src/private/files.ts
HarshJain27: Depending on the **filePath** parameter value, if it is null, it will open the download preference folder configured in Teams client and if non null it will open the folder in which file is present which is provided as filePath param.

HarshJain27: The API will open the File Path in one of the scenario as mentioned in above comment, can you elaborate more on the security risk, if any. And how can i go about this feature if there is a security risk.

ydogandjiev (approved):

src/private/files.ts
```diff
@@ -487,11 +487,17 @@ export namespace files {
    * @private
    * Hide from docs
    *
-   * Open download preference folder
+   * Open download preference folder or folder containing filePath if its value is non-null
+   * @param filePath: Path of the file whose containing folder should be opened
```
TrevorJoelHarris, line 491: I'm not sure I understand why you are passing a path to a file into an "open a folder" function. Shouldn't you be passing in a folder path? The opening description is a little confusing:

"Open download preference folder or folder containing filePath if its value is non-null"

Does that mean that callers are intended to pass in a null filePath if they want to open the download preference folder? If that's the case, shouldn't filePath be an optional parameter?

test/private/files.spec.ts
```diff
@@ -467,27 +467,38 @@ describe('files', () => {
 
   describe('openDownloadFolder', () => {
     it('should not allow calls before initialization', () => {
-      expect(() => files.openDownloadFolder()).toThrowError(
```
TrevorJoelHarris, line 470: should have unit tests that pass in a value for filepath, all of these tests seem to pass in null for filepath

TrevorJoelHarris (changes requested):
Would like more info about the filePath parameter and why it's a required parameter that is always set to null

src/private/files.ts
HarshJain27: I agree with the fact that the callers should be given flexibility to not pass null for opening download preference folder. For now i have assigned default null value to filePath variable and in v2.0 will make sure to make it optional.

test/private/files.spec.ts
HarshJain27: Added the unit test as suggested.

TrevorJoelHarris (approved):

PR #1000, opened 2022-03-16
promisify getFileDownloads

Jessica-Koch:
promisify getFileDownloads and add new tests/fix existing ones

### DIFF
--- a/change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
+++ b/change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "promisify files.getFileDownloads",
+  "comment": "added promisified files.getFileDownloads",
   "packageName": "@microsoft/teams-js",
   "email": "jekoch@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
+++ b/change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "promisify getFileDownloads",
+  "packageName": "@microsoft/teams-js",
+  "email": "jekoch@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
+++ b/change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "promisify getFileDownloads",
+  "comment": "promisify files.getFileDownloads",
   "packageName": "@microsoft/teams-js",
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/private/files.ts
```diff
@@ -524,14 +525,25 @@ export namespace files {
...
    */
-  export function getFileDownloads(callback: (error?: SdkError, files?: IFileItem[]) => void): void {
+  export function getFileDownloads(): Promise<IFileItem[]>;
+  /**
+   * @hidden
+   * Hide from docs
+   * ------
+   *
+   * @deprecated
+   * As of 2.0.0-beta.3, please use {@link file.getFileDownloads file.getFileDownloads(): Promise\<IFileItem[]\>} instead.
```
aeunms, line 535: I think this should be beta.4 since the current version is beta.3.

change/@microsoft-teams-js-b80ee3ad-b996-4147-b9a9-34e6609afbb1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "promisify getFileDownloads",
```
aeunms, line 3: I think it would be helpful to include the capability, so something like "promisify files.getFileDownloads()"

aeunms: Or actually, since we offer the deprecated function for back compat, maybe it would be clearer to say "added a promisified files.getFileDownloads" or something along those lines.

Jessica-Koch: done

packages/teams-js/test/private/files.spec.ts
```diff
@@ -483,40 +483,67 @@ describe('files', () => {
...
+          },
+        ];
+        let called = false;
+
+        const callback = jest.fn((err, fileList) => {
+          called = true;
+
+          expect(err).toBeFalsy();
+          expect(fileList).toEqual(mockFileDownloads);
+          expect(called).toBe(true);
```
aeunms, line 511: Should this line be outside the callback since `expect(called).toBe(true)` will always be true in the callback?

aeunms (commented):
Done with first round of comments!

packages/teams-js/test/private/files.spec.ts
aeunms: Boof github still has a duplicate comment bug

aeunms: Boof github still has a duplicate comment bug

Jessica-Koch: technically, I don't even need to test if it's called because i'm already doing that with the expect.assertions(which I know if technically for catch blocks).  If I remove the call to the callback, the test will fail saying there is only one assertion

aeunms: Hmm maybe we don't even need called then? And have just three assertions

Jessica-Koch: so I understand that it looks redundant, but it actually isn't because the assertions within the callback won't ever run if the callback isn't called.  If we set a variable and changed it outside of any function, this would be useless, but because the value never gets 

We can't check if called is true outside because the callback function is actually called AFTER the test is run...just to show this, I added a bunch of console statements to illustrate this

image

aeunms (approved):

PR #1002, opened 2022-03-17
add missing functions to the teams test app

Jessica-Koch:
Add following to the teams test app:
* enum FileDownloadStatus
* interface IFileItem
* getFileDownloads()
* openDownloadFolder()

### DIFF
--- a/apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
@@ -1,4 +1,4 @@
-import { FileOpenPreference, FilePreviewParameters, files } from '@microsoft/teams-js';
+import { FileOpenPreference, FilePreviewParameters, files, SdkError } from '@microsoft/teams-js';
 import React, { ReactElement } from 'react';
 
 import { ApiWithCheckboxInput, ApiWithoutInput, ApiWithTextInput } from '../utils';
@@ -169,6 +169,7 @@ enum FileDownloadStatus {
   Downloading = 'Downloading',
   Failed = 'Failed',
 }
+
 interface IFileItem {
   objectId?: string;
   path?: string;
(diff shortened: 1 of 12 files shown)


apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
```diff
@@ -164,6 +164,59 @@ const CopyMoveFiles = (): ReactElement =>
...
+  ApiWithTextInput<IFileItem[]>({
+    name: 'getFileDownloads',
+    title: 'Get File Downloads',
+    onClick: {
+      validateInput: () => {
+        return;
+      },
+      submit: {
+        withCallback: (input, setResult) => {
+          const callback = (error, files): void => {
```
aeunms, line 192: Right now our linting is kinda wacky and won't tell you to do this, but I Think we'd want the parameter types to be specified.

apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
```diff
@@ -164,6 +164,59 @@ const CopyMoveFiles = (): ReactElement =>
...
+  path?: string;
+  sizeInBytes?: number;
+  status?: FileDownloadStatus;
+  timestamp: Date;
+  title: string;
+  extension: string;
+}
+
+const GetFileDownloads = (): ReactElement =>
+  ApiWithTextInput<IFileItem[]>({
```
aeunms, line 183: It looks like we won't take input from the user in this case, maybe we should use ApiWithoutInput instead?

apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
```diff
@@ -164,6 +164,39 @@ const CopyMoveFiles = (): ReactElement =>
...
+          } else {
+            setResult(JSON.stringify(files));
+          }
+        };
+
+        files.getFileDownloads(callback);
+      },
+      withPromise: async () => {
+        await files.getFileDownloads();
+        return 'files.getFileDownloads complete';
```
aeunms, line 185: Sorry I didn't catch this before, but since the resolved promise provides the files.IFileItem[] I wonder if we could display that for the withPromise onClick too?

aeunms (approved):
LGTM!

PR #1003, opened 2022-03-17
add npmrc and enforce node and npm versions

Jessica-Koch:
Enforce a node version above 16, lts is 16.14.  When a user goes to run a yarn install in any of the directories, an error will be returned: 
image

I would like input on whether I should add an upper constraint of 17.0, v16 goes into maintenance in october: 
image

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,3 @@
+# will force display errors if there is a version mismatch
+#  with node or npm defined in the package.json
+engine-strict=true
--- a/package.json
+++ b/package.json
@@ -7,6 +7,10 @@
     "packages/*",
     "tools/**/*"
   ],
+  "engines": {
+    "node": ">=16.0.0",
+    "npm": ">=8.0.0"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/OfficeDev/microsoft-teams-library-js/tree/2.0-preview"


aeunms:
When you say upper constraint of 17, do you mean versions less than 17? Since odd versions are not stable

TrevorJoelHarris:
Supporting 16 for now makes sense since it's the active one. As Angela pointed out, since 17 isn't stable I don't think we want to restrict to just 17.

Personally, I wouldn't block people using 17 if they want to unless we know it will cause a problem.

TrevorJoelHarris (approved):

erinha:
I believe we have github actions that verify that our build works on node versions 12 and 14, which are now failing. Additionally, our pipeline runs on node 14, so that appears to be failing as well. Those will have to be updated before this change is merged. 

Note that the github actions that verify the build works on node 12 and 14 are the same as those running in v1, so we should double-check that there isn't a reason we should continue supporting 12 and 14.

.npmrc
```diff
@@ -0,0 +1,3 @@
+# will force display errors if there is a version mismatch
+#  with node or npm defined in the package.json
+engine-strict=true
```
erinha, line 3: What is the effect if we don't have `engine-strict=true`? Can we display warnings without failing the build?

erinha (changes requested):
Appreciate the intent here to help us all stay consistent/avoid problems due to node versions! Question about `engine-strict=true` and comment about failing pipelines/checks.

package.json
```diff
@@ -7,6 +7,10 @@
     "packages/*",
     "tools/**/*"
   ],
+  "engines": {
+    "node": ">=16.0.0",
+    "npm": ">=8.0.0"
```
erinha, line 12: This is probably a stupid question, but since we use yarn instead of npm, what is the reason to include this here?

erinha:
> Supporting 16 for now makes sense since it's the active one. As Angela pointed out, since 17 isn't stable I don't think we want to restrict to just 17.
> 
> Personally, I wouldn't block people using 17 if they want to unless we know it will cause a problem.

Jess, were you asking if we should block people from using versions greater than 17. So that for example, when version 18 is released, if anyone upgraded too soon, it would cause failures and notify us that we need to update it? Or am I misunderstanding?

.npmrc
Jessica-Koch: It force displays the errors in the console, we need this in order to see them, the .npmrc file defines the settings on how NPM should behave when running commands.

package.json
Jessica-Koch: Not a stupid question, yarn is installed via npm, so this impacts yarn (I tested it because I was confused at first), the screenshot above is running a yarn command in the console.

Jessica-Koch:
I can't respond to comments in this UI so I'll respond to them here

@aeunms - I didn't realize that the odd versions weren't stable so that is interesting!  I meant making it so a dev can only use a node version that is greater than 15 and less than 17.  Right now, I have it configured to only throw an error when the node version is less than 16. 

@TrevorJoelHarris - you answered my question, that's what I was asking, should we stop people from using 17, I know personally I was originally using 17 and it did throw errors, but maybe because it isn't stable, that should be expected.  I like the idea of allowing people to use the newer version, it makes updating later easier.  

@erinha - Yes I did notice that these pipelines are failing, you're totally right that these would have to be updated.  This brought up a question that I have which is, if we are testing the code against older versions of node, then why when a dev uses v14, for instance, do they get issues if the code will pass checks in our pipeline?  This may be obvious, I just don't understand.  I would love to have a conversation about the best way forward in terms of updating these pipelines and whatnot, who else I would need to talk to etc.

Jessica-Koch:
> > Supporting 16 for now makes sense since it's the active one. As Angela pointed out, since 17 isn't stable I don't think we want to restrict to just 17.
> > Personally, I wouldn't block people using 17 if they want to unless we know it will cause a problem.
> 
> @Jessica-Koch, were you asking if we should block people from using versions greater than 17. So that for example, when version 18 is released, if anyone upgraded too soon, it would cause failures and notify us that we need to update it? Or am I misunderstanding?

I was asking if we should stop people from using versions 17 and above, but honestly, because those aren't stable maybe it should be assumed that it may break the build.  I'm hesitant to restrict newer versions because I know from experience, it's easy to just never update the version of node and then you run into a problem when you notice and your version is 4 versions out of date.  If that makes any sense at all.  If we keep the upper bound open, then we can more easily test with newer node versions when 17 becomes stable (or I guess becomes 18?  referring to what @aeunms  said above)

erinha:
> why when a dev uses v14, for instance, do they get issues if the code will pass checks in our pipeline? 

@Jessica-Koch this is a great question. If they pass the pipeline I would expect them to pass in our code as well. Were people seeing local issues with node versions 12 and 14? I thought it was just unsupported versions.

erinha:
> I would love to have a conversation about the best way forward in terms of updating these pipelines and whatnot, who else I would need to talk to etc.

To update the ADO pipelines, you just need to edit the yaml files that reference the node version. For removing the github action checks that validate the build on node versions 12 and 14, I would recommend getting approval/input from @ydogandjiev.

.npmrc
erinha: That's helpful, thanks. Do you know if it's possible to make it display warnings but not errors?

Jessica-Koch: I think I answered a different question than the one you asked, i'll look into this because i'm not sure

Jessica-Koch: My thought though is that adding the strict true is better because it's alerting a problem before it starts, if we make it a warning, it may end up breaking the build later and just kicks the problem down the road.  If this is a warning, i'm not sure I see the point of changing this at all...warnings get ignored (like the ones on the main page of this repo warning us about vulnerabilities that still have gone unchanged)

Jessica-Koch: I may not be looking at the benefits of having a warning correctly, but the way I'm seeing it, just making it a warning so it passes the build isn't the best justification, but I am totally open to being wrong about this

Jessica-Koch:
I am closing this as I don't think it is useful and forgot to close this with the ticket

PR #1015, opened 2022-03-23
Lakhveer/dialogs Add framed and frameless to unit tests

lakhveerk:
1. Previously, we were using Utils file to mock desktopPlatform. However, it can be any framed platform (e.g. web). 
Replaced keyword desktopPlatformMock in these instances to framedMock. 

2. Added unit tests for framlessMock scnerios. Communication goes through a slightly diff route if we use utils vs frameless to initialize the app. We decided that it is a good idea to test everything using both ways. 

3. Fixed some other tests


shrshindeMSFT (approved):

TrevorJoelHarris:
It might be good to leave a comment in this PR outlining *why* you changed from using the utils file to using framedmock. It could be helpful to people looking to make unit test changes in the future

TrevorJoelHarris (approved):
Please add a comment like I asked for above, otherwise looks good to me

PR #1021, opened 2022-03-24
deleting callbacks in dialogs from type PostMessageChannel and sendMessageToParentFromDialog

lakhveerk:
These callbacks never get called and should not be there.


packages/teams-js/src/public/dialog.ts
```diff
@@ -18,14 +17,15 @@ import { runtime } from './runtime';
  *
  * @beta
  */
+
 export interface SdkResponse {
-  err: string;
-  result: string | object;
+  err?: string;
+  result?: string | object;
```
lakhveerk, line 23: Is this okay, or should I enforce that one of them has to be true?

MarcinJuraszek (approved):

jadahiya-MSFT (approved):

PR #1023, opened 2022-03-25
Add .npmrc file for publishing

erinha:
This change adds a .npmrc file for publishing and ensures that .npmrc is included in the published build artifacts so the release pipeline can use it.


AE-MS (dismissed):
:shipit:

shrshindeMSFT (approved):

PR #1026, opened 2022-03-28
Disable the post bundle analysis task

erinha:
Disable the command that posts the bundle analysis report because it's failing due to an external issue. Prefer that we stop seeing a failure to post, which we might get used to. 

When that issue is resolved, I will re-enable the task.


aeunms (approved):

AE-MS (approved):
:shipit:

PR #1030, opened 2022-03-29
Enabling E2E tests to run against updated Dialog APIs

jadahiya-MSFT:
- Removed V2 Test functions
- Added in new `sendToDialog` API for test purposes
- Ensuring backwards compatibility


packages/teams-js/src/public/dialog.ts
```diff
@@ -98,6 +98,19 @@ export namespace dialog {
     sendMessageToParent('messageForParent', [message]);
   }
 
+  /**
```
TrevorJoelHarris, line 101: Is this being added just for testability? If so, it might be worth talking about. In general I don't like adding exported public functions that aren't actually needed for app developers because it makes the sdk surface harder to understand (this is the same as the PostMessageChannel that open returns if I'm not mistaken)

jadahiya-MSFT: This used to be for testability only, but I think we want to return a `Promise<void>` again for `dialog.open` and so we would need to extract this capability out.

apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -115,47 +100,30 @@ const DialogAPIs = (): ReactElement => {
...
-            childWindow.postMessage(message, onComplete);
+              };
+              childWindow.postMessage(message, onComplete);
+            } else {
+              setResult("childWindow doesn't exist");
+            }
+            return '';
           } else {
-            setResult("childWindow doesn't exist");
+            setResult('A post call to child is initiated');
```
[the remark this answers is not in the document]
jadahiya-MSFT: You're right this is a good catch, I'll make that change. Thanks :)

packages/teams-js/src/public/dialog.ts
lakhveerk: Are you planning to make those changes in a separate PR? (Turning the return to Promise<void> and resolving the promise when dialog registers to hear from app)?

jadahiya-MSFT: Promisifying will be in a separate PR, this PR enables us to test any changes and ensure we aren't breaking anything in the process.

packages/teams-js/src/public/dialog.ts
```diff
@@ -161,23 +167,29 @@ export namespace dialog {
...
         removeHandler('messageForParent');
       });
+    }
 
+    /**
+     *  Send message to the dialog from the parent
+     *
+     * @param message - The message to send
+     */
+    export function sendMessageToDialog(
```
lakhveerk, line 187: I think, we don't need a separate a function for `bot` since this not different than the other outside function. What do you think?

jadahiya-MSFT: I didn't know if this was needed or not, I will remove it. Good catch :)

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -95,17 +95,17 @@ describe('Dialog', () => {
...
-          const sendMessageToDialogHandler = dialog.open(dialogInfo, emptyCallback, emptyCallback);
-          sendMessageToDialogHandler('exampleMessage');
+          dialog.open(dialogInfo, emptyCallback, emptyCallback);
+          dialog.sendMessageToDialog('exampleMessage');
           const message = framedMock.findMessageByFunc('messageForChild');
           expect(message).not.toBeUndefined();
           expect(message.args).toStrictEqual(['exampleMessage']);
         });
 
         it(`FRAMELESS: should initiate the post message to dialog using returned function. context: ${context}`, async () => {
```
lakhveerk, line 105: We can change the test description according to new functionality :)

lakhveerk: And, I think, for unit testing, we don't need to open a dialog to test `sendMessageToDialog`

lakhveerk (dismissed):

apps/teams-test-app/src/components/DialogAPIs.tsx
```diff
@@ -78,17 +58,14 @@ const DialogAPIs = (): ReactElement => {
...
-            setResult('Teams client SDK call dailog.update.resize was called');
-            return '';
-          },
-          //For V1 back compatibility
-          withCallback: (taskInfo, setResult) => {
-            tasks.updateTask(taskInfo);
-            setResult('Teams client SDK call tasks.updateTask was called');
-          },
+          }
+          setResult('Teams client SDK call dailog.update.resize was called');
```
TrevorJoelHarris, line 67: I'm going to assume fixing this typo is hard in some way because some test somewhere is looking for the string "dailog". But if it's fixable that would be neat!

TrevorJoelHarris (dismissed):

apps/teams-test-app/src/components/DialogAPIs.tsx
jadahiya-MSFT: Good catch, I think that was just an oversite on my part. I will make the necessary adjustments

TrevorJoelHarris (approved):

lakhveerk (approved):

PR #1036, opened 2022-03-30
Updating comments in dialogs.ts file

lakhveerk:


packages/teams-js/src/public/dialog.ts
```diff
@@ -86,14 +96,14 @@ export namespace dialog {
...
-   * @param type - The event to listen to. Currently the only supported type is 'message'.
-   * @param listener - listener - The listener that will be called.
+   * @remarks
+   * This function is only called from inside of a dialog
+   *
+   * @param listener - The listener that will be triggered.
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  export function registerOnMessageFromParent(listener: (message: any) => void): void {
-    ensureInitialized();
```
lakhveerk, line 96: Inside appWindow, we don't ensure initialization in a similar (or just basically the same function), but I cannot see why. Any ideas? Now I am planning to add it back. Open for suggestions :) 

```
  public addEventListener(type: string, listener: (message: any) => void): void {
    if (type === 'message') {
      registerHandler('messageForChild', listener);
    }
  }
```

lakhveerk: In a team sync, we decided we should add the ensureInitialized here and in appWindow APIs. I have pushed a change and closing this comment here :)

jadahiya-MSFT (dismissed):

shrshindeMSFT (dismissed):

packages/teams-js/src/public/dialog.ts
```diff
@@ -17,21 +17,30 @@ import { runtime } from './runtime';
...
+   */
+  export interface ISdkResponse {
+    /**
+     * Error in case there is a failure before dialog submission
+     */
+    err?: string;
+
+    /**
+     * Result value that the dialog is submitted with
+     */
```
TrevorJoelHarris, line 32: I believe this comes from dialog.submit function right? If so, let's call attention to that.

packages/teams-js/src/public/dialog.ts
```diff
@@ -17,21 +17,30 @@ import { runtime } from './runtime';
  *
```
TrevorJoelHarris, line 17: Is the comment on line 16 true? I think some of these functions work in other FrameContexts (like task for example)

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: (There's a way to make a link to another function in the documentation, @link maybe?)

packages/teams-js/src/public/dialog.ts
```diff
@@ -17,21 +17,30 @@ import { runtime } from './runtime';
...
+     */
+    result?: string | object;
+  }
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   export type PostMessageChannel = (message: any) => void;
-  export type DialogSubmitHandler = (result: SdkResponse) => void;
+  export type DialogSubmitHandler = (result: ISdkResponse) => void;
+
   /**
    * Allows an app to open the dialog module.
```
TrevorJoelHarris, line 40: Might change this to something like "Allows app to open a dialog".

packages/teams-js/src/public/dialog.ts
```diff
@@ -17,21 +17,30 @@ import { runtime } from './runtime';
...
    *
    * @param urlDialogInfo - An object containing the parameters of the dialog module.
```
TrevorJoelHarris, line 42: Does this generate a link to the urlDialogInfo type, and does that type have sufficient documentation?

packages/teams-js/src/public/dialog.ts
```diff
@@ -17,21 +17,30 @@ import { runtime } from './runtime';
...
-   * @param submitHandler - This Handler is called when the dialog has been submitted or closed.
+   * @param submitHandler - Handler that triggers when the dialog has been submitted or closed.
```
TrevorJoelHarris, line 43: Probably worth again linking to the submit function and saying that this function will be called when a dialog calls the submit function OR when the user closes the dialog.

packages/teams-js/src/public/dialog.ts
```diff
@@ -75,7 +84,10 @@ export namespace dialog {
   /**
```
TrevorJoelHarris, line 84: On line 74, remove the word "contains" at the beginning

packages/teams-js/src/public/dialog.ts
```diff
@@ -86,21 +98,30 @@ export namespace dialog {
   }
 
   /**
-   * Fucntion to call when an event is received from the Parent
+   * Register a listener that wil be triggerd when an event is received from the parent
```
TrevorJoelHarris, line 101: "triggered". Maybe change the comment to "Register a listener that will be triggered when a message is received from the app that opened the dialog".

I assume this should also only be called from the dialog. Should we make sure that this is also ensureInitialized(FrameContexts.task), or does it have to be this way for backwards compatibility?

packages/teams-js/src/public/dialog.ts
```diff
@@ -86,21 +98,30 @@ export namespace dialog {
...
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  export function registerOnMessageFromParent(listener: (message: any) => void): void {
+  export function registerOnMessageFromParent(listener: PostMessageChannel): void {
     ensureInitialized();
     registerHandler('messageForChild', listener);
   }
 
+  /**
+   * Checks if dialog module is supported currently
```
TrevorJoelHarris, line 114: "is supported by the host"

packages/teams-js/src/public/dialog.ts
```diff
@@ -111,12 +132,30 @@ export namespace dialog {
       ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
       sendMessageToParent('tasks.updateTask', [dimensions]);
     }
+
+    /**
+     * Checks if dialog.update capability is supported currently
```
TrevorJoelHarris, line 137: "supported by the host"

packages/teams-js/src/public/dialog.ts
```diff
@@ -111,12 +132,30 @@ export namespace dialog {
...
+     *
+     * @returns boolean to represent whether dialog.update is supported
+     */
     export function isSupported(): boolean {
       return runtime.supports.dialog ? (runtime.supports.dialog.update ? true : false) : false;
     }
   }
 
+  /**
+   * Namespace to open the dialog using bot
```
TrevorJoelHarris, line 147: "open a dialog that sends results to the bot framework"

packages/teams-js/src/public/dialog.ts
```diff
@@ -111,12 +132,30 @@ export namespace dialog {
...
+   */
   export namespace bot {
+    /**
+     * Allows an app to open the dialog module using bot.
+     *
+     * @param botUrlDialogInfo - An object containing the parameters of the dialog module including completionBotId.
```
TrevorJoelHarris, line 153: Does this link to the BotUrlDialogInfo docs and are those docs finished/sufficient?

TrevorJoelHarris: (In general, same feedback as for dialog.open above)

packages/teams-js/src/public/dialog.ts
lakhveerk: does not seem true, thank you for catching. 

dialog.open works in 3 frame contexts,  but submit and resize work from 4 fram contexts. `sendMessageToParentFromDialog` works only on task. 

What should we update the remark to?

lakhveerk: does not seem true, thank you for catching. 

dialog.open works in 3 frame contexts,  but submit and resize work from 4 fram contexts. `sendMessageToParentFromDialog` works only on task. 

What should we update the remark to?

packages/teams-js/src/public/dialog.ts
```diff
@@ -139,6 +178,12 @@ export namespace dialog {
       };
       return sendMessageToDialog;
     }
+
+    /**
+     * Checks if dialog.bot capability is supported currently
```
TrevorJoelHarris, line 183: "supported by the host"

TrevorJoelHarris (changes requested):
Let me know when it's ready for re-review or if you have any questions

packages/teams-js/src/public/dialog.ts
lakhveerk: I am not very familiar with the frame contexts, but when we talked in the app SDK sync, Erin said it is fine to call ensureInitialized with no parameters (I hope I heard her right). 

Backward compatibility is also a good point, but we are changing it in the tasks file as well.
I have a PR up to add ensureInitialized in tasks file. If we know this should only work for FrameContexts.task, I can fix it in my PR as well.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -544,14 +547,17 @@ export interface DialogSize {
   width: DialogDimension | number;
 }
 
+/**
+ * Data structure to describe dialog information needed to open open a url based dialog.
```
lakhveerk, line 551: I added a summary line, but not sure how to create a link to the actual function without importing the function to this file. If we think, linking will help users, I will figure a way.

lakhveerk: Also was not sure how to write a summary for dialogInfo interface below

packages/teams-js/src/public/tasks.ts
```diff
@@ -18,7 +18,8 @@ import { BotUrlDialogInfo, DialogInfo, DialogSize, TaskInfo, UrlDialogInfo } fro
 export namespace tasks {
   /**
    * @deprecated
-   * As of 2.0.0-beta.4, please use {@link dialog.open(dialogInfo: DialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} instead.
+   * As of 2.0.0-beta.4, please use {@link dialog.open(urlDialogInfo: UrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} for url based dialogs
```
lakhveerk, line 21: developers still have to use tasks for card-based dialogs.  How do I say that this function is deprecated but still use it for card based?

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: I would probably remove this remark and if necessary add remarks to individual functions. For example, it might be worth calling out that you can't call dialog.open from a dialog

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: As far as I remember, this function is the "new" version of ParentAppWindow.addEventListener. I can't think of a reason that we would want anyone *except* a task/dialog context to register for messages from the parent app that created it. Could look at uses of ParentAppWindow or ask Yuri in office hours about it maybe?

Also okay to open a bug to investigate this if you don't want to hold up this PR.

packages/teams-js/src/public/interfaces.ts
TrevorJoelHarris: I do think you have to import the function in order to link to it. If dialog is already linking to this then maybe it's okay to not link in both directions (especially if this structure might be used in more than one capability or function).

packages/teams-js/src/public/dialog.ts
lakhveerk: opened a new work item for this :)

packages/teams-js/src/public/interfaces.ts
lakhveerk: I think, in dioalog.ts file, it is straight forward to find the dialogDimension interfaces. So, I did not link them. It makes my parameter summary a little smooth. 

In dialog.ts, do we prefer below, or do we want to leave out the `of type {@linkcode UrlDialogInfo}` part out? 
   * @param urlDialogInfo - An object of type {@linkcode UrlDialogInfo} containing the parameters of the dialog module.
   * @param submitHandler - Handler of type {@linkcode DialogSubmitHandler} that triggers when a dialog calls the {@linkcode submit} function or when the user closes the dialog.

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

packages/teams-js/src/public/tasks.ts
itstaby: After updating to the latest SDK, we are unable to use dialogs for cards. Is the recommendation here that we revert to tasks for those? And is this temporary?
This will require reverting many of the changes made for MetaOS support, I would prefer not doing that if you'll eventually add card support back to dialogs.

jadahiya-MSFT: You should talk about this on teams

PR #1044, opened 2022-04-01
Design back compat runtime config generator function

aeunms:
This PR is part of the work in revamping our isSupported() functionality. 

We have added a new function `generateBackCompatRuntimeConfig` to dynamically generate a runtime config for legacy host clients since they do not provide us with one. We previously had simply statically assigned a pre-formed runtime config for them as a workaround, but this runtime config should actually differ based on the host client type as well as their highest supported client sdk version.

The documentation on how to add any future APIs that utilize a version check has been added to CONTRIBUTING.md, but there will be further internal work done to make sure relevant contributors are aware of the new guidance here.


packages/teams-js/src/public/runtime.ts
```diff
@@ -121,19 +122,84 @@ export const teamsRuntimeConfig: IRuntime = {
...
+      HostClientType.android,
+      HostClientType.teamsRoomsAndroid,
+      HostClientType.teamsPhones,
+      HostClientType.teamsDisplays,
+    ],
+  },
+};
+
+export function generateBackCompatRuntimeConfig(highestSupportedVersion: string): IRuntime {
+  let newSupports = { ...teamsRuntimeConfig.supports, ...{} };
```
MarcinJuraszek, line 179: what does the `...{}` part do?

packages/teams-js/src/public/runtime.ts
```diff
@@ -121,19 +122,84 @@ export const teamsRuntimeConfig: IRuntime = {
...
     teamsCore: {},
     video: {},
   },
 };
 
+// object of version constants
+export const versionConstants = {
+  '1.9.0': {
+    capabilities: { location: {} },
+    hostClientTypes: [
```
MarcinJuraszek, line 139: this seems exactly the same as the one below. Can we extra it to only have the list once? Would that affect the bundle size?

packages/teams-js/src/public/runtime.ts
aeunms: This is the [Typescript spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and allows us to "spread out" the `supports` content and create a new object free of modification restrictions. Without the spreading, there's a lot of Typescript unhappiness with adding onto the supports field.

aeunms: The `...{}` is an empty object spread out which just allows us to use the new object creator.

packages/teams-js/src/public/runtime.ts
aeunms: What does it mean to "extra it"?

packages/teams-js/src/public/runtime.ts
aeunms: I checked on this to see if we could get rid of it (It's what I saw all over stack overflow so I hadn't questioned it haha) and I think we can get rid of the ...{}. Thanks for pointing it out!

packages/teams-js/src/public/runtime.ts
MarcinJuraszek: typo. Should have been 'extract'. As in have either a constant or a variable with the value and use it in two places.

packages/teams-js/src/public/runtime.ts
MarcinJuraszek: It's the spreading of an empty object part that I was a bit curious about. If it's not needed let's remove it.

aeunms: Agreed, done!

packages/teams-js/src/public/runtime.ts
aeunms: I've been tinkering with a couple ways to do this- I thought first that it would be awesome just to do `hostClientTypes: Object.values(HostClientType),` because I thought it was a clean and clear way to say "this works in all the host client types." And then I started to wonder if we would add any host client types in the future. The version constants would then automatically say they work in that host client type even if they aren't working there yet. Another option then is to declare a separate constant like you suggested, but I wasn't sure what to name it since the only thing that would separate it from any other host client types list would be the date/current version that I'm making it on, and that seems like an unfortunate way to name it. 

In the end though, this is suppose to serve an edge case for the few use cases for legacy hosts and by the time we might be adding new host client types it should hopefully be long in the past- so maybe it's a better choice to make it clean and clear to read. Would love to hear your thoughts

MarcinJuraszek: Good catch on the `Object.values(HostClientType)` issue. I think I would define a constant and name it something like `v1HostClientTypes` to make it clear 1) why it's defined 2) what it represents.

aeunms: Done! Thanks :)

shrshindeMSFT (approved):

CONTRIBUTING.md
```diff
@@ -38,6 +38,55 @@ times in N different commits. If there was some accidental reformatting or white
...
 the PR.
 
+### Adding an API that utilizes version checks for compatibility
+
+Here are the steps for adding an API that utilizes version checks (e.g. `if (!isCurrentSDKVersionAtLeast(captureImageMobileSupportVersion)...`):
+
+1. Add the API as a new capability or subcapability rather than adding to an existing capability.
+2. In [runtime.ts](packages/teams/src/public/runtime.ts), add an object describing the new capability and its compatibility requirements to `versionConstants`.
+
+Exa)
```
jadahiya-MSFT, line 48: Should this say example?

jadahiya-MSFT (approved):

PR #1046, opened 2022-04-04
Adding call types to meeting details

melu613:
This PR adds an additional type of MeetingDetailsCallType to support calls in meetings. This keeps the original MeetingType of getMeetingDetails, while adding the types of oneOnOneCall and groupCall to possible types returned by the getMeetingDetails API.

### DIFF
--- a/src/public/meeting.ts
+++ b/src/public/meeting.ts
@@ -267,7 +267,7 @@ export namespace meeting {
     if (!callback) {
       throw new Error('[share app content to stage] Callback cannot be null');
     }
-    ensureInitialized(FrameContexts.sidePanel);
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
     sendMessageToParent('meeting.shareAppContentToStage', [appContentUrl], callback);
   }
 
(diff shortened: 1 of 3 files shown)


MarcinJuraszek:
This is a breaking change. Apps which already use this API will have no idea what these new values mean once Teams starts returning them.

melu613:
> This is a breaking change. Apps which already use this API will have no idea what these new values mean once Teams starts returning them.

@MarcinJuraszek It is less of a breaking change than this would seem. The new values would only be returned in a 1:1 and group call app scenario. Apps in calls is currently only enabled till the internal Teams Ring 1, so there is no risk of public app users seeing these new types. 

That said, if you still think this is a breaking change, is there a way that we can stow this change before apps in calls reaches higher rings, using something like a versioning system?

src/public/meeting.ts
```diff
@@ -127,6 +127,11 @@ export namespace meeting {
     MeetNow = 'MeetNow',
   }
 
+  export enum MeetingDetailsCallType {
```
ydogandjiev, line 130: Why not just call this "CallType" to match the "MeetingType" syntax?

ydogandjiev (changes requested):

src/public/meeting.ts
melu613: CallType in the TMP codebase as used by the CMD team have different enums; they only differentiate between a one on one and conference call. A conference call in this case would include both the group call and a meeting. I wanted to avoid the potential confusion in the codebase, and named this enum something different. @ydogandjiev

MarcinJuraszek: We should not design the developer facing external API surface based on implementation details of how Teams works internally. We should design it to be the best it can be for the app developer, and deal with things internally as needed.

MarcinJuraszek (approved):

PR #1048, opened 2022-04-06
Mute the audio when you are in meetings.

vamsiv96:
* Allows an app to mute the incoming audio speaker in the meeting. Unmute is not allowed
   * @param callback Callback contains 2 parameters, error and result.
   * error can either contain an error of type SdkError, incase of an error, or null when toggle is successful
   * result can either contain the true/false value, incase of a successful toggle or null when the toggling fails
   * result: True means incoming audio is muted

### DIFF
--- a/src/public/meeting.ts
+++ b/src/public/meeting.ts
@@ -153,6 +153,21 @@ export namespace meeting {
   }
 
   /**
+   * Allows an app to mute the incoming audio speaker in the meeting. Unmute is not allowed
+   * @param callback Callback contains 2 parameters, error and result.
+   * error can either contain an error of type SdkError, incase of an error, or null when toggle is successful
+   * result can either contain the true/false value, incase of a successful toggle or null when the toggling fails
+   * result: True means incoming audio is muted
+   */
+  export function muteViaMeshApp(callback: (error: SdkError | null, result: boolean | null) => void): void {
+    if (!callback) {
+      throw new Error('[mute via mesh app] Callback cannot be null');
+    }
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
+    sendMessageToParent('muteViaMeshApp', callback);
+  }
+
+  /**
    * @private
    * Hide from docs
    * Allows an app to get the meeting details for the meeting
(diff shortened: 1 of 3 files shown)


src/public/meeting.ts
```diff
@@ -159,6 +159,21 @@ export namespace meeting {
...
   }
 
+  /**
+   * Allows an app to mute the incoming audio speaker in the meeting. Unmute is not allowed
+   * @param callback Callback contains 2 parameters, error and result.
+   * error can either contain an error of type SdkError, incase of an error, or null when toggle is successful
+   * result can either contain the true/false value, incase of a successful toggle or null when the toggling fails
+   * result: True means incoming audio is muted
+   */
+  export function muteViaMeshApp(callback: (error: SdkError | null, result: boolean | null) => void): void {
```
ronakbakshi, line 169: Update the name of the API. This API name is limiting to Mesh app only, but our API should be open to all apps. I suggest something like `muteMicrophone`

src/public/meeting.ts
```diff
@@ -159,6 +159,21 @@ export namespace meeting {
...
+    if (!callback) {
+      throw new Error('[mute via mesh app] Callback cannot be null');
+    }
+    ensureInitialized(FrameContexts.sidePanel, FrameContexts.meetingStage);
```
ronakbakshi, line 173: start a discussion with Me, Kumud, Lisa, Ajay and check for which contexts we should support this API?

MarcinJuraszek:
Did this go through SDK review process already?

src/public/meeting.ts
```diff
@@ -159,6 +159,20 @@ export namespace meeting {
     sendMessageToParent('toggleIncomingClientAudio', callback);
   }
 
+  /**
+   * Allows an app to mute the microphone in the meeting.
+   * @param callback Callback contains 2 parameters, error and result.
+   * error can either contain an error of type SdkError, incase of an error, or true when mute is successful
```
ronakbakshi: * error can either contain an error of type SdkError, incase of an error, or null when mute is successful
   * result can either contain the true value, incase of a successful mute or null when the mute fails
   * result: True means microphone is muted

TrevorJoelHarris:
The author didn't answer @MarcinJuraszek 's question about design review and hasn't updated this PR in almost 6 months. I am going to close it, please feel free to re-open the PR once you've gone through a design review for this change. Let me know if you need help going through that process.

PR #1049, opened 2022-04-06
Changes to the location capability in prep for device permissions work

TrevorJoelHarris:
- The location capability now has hasPermission() and requestPermission() functions. The app developer doesn't have to specify *which* permissions they are requesting, the host grants the correct permissions for all functions in this capability
-location.getLocation and location.showLocation are now strictly deprecated back-compat functions
location.getCurrentLocation gets the user's current location *without* showing any UI such as maps. Hosts can choose to implement this without having any map based UI
- location.map now contains all functions that show UI to the user
- location.map.chooseLocation shows the user's current location on the map and allows the user to manipulate the map in order to change that location. When the promise resolves the map has been closed and the function returns the location the user chose
- location.map.showLocation shows the user's current location on the map but does not allow the user to change the location

### DIFF
--- a/packages/teams-js/src/public/barcode.ts
+++ b/packages/teams-js/src/public/barcode.ts
@@ -24,7 +24,7 @@ export namespace barcode {
       }
 
       // Decided not to use the old message format here and totally break from media.ts
-      resolve(sendAndHandleSdkError('barcode.scanBarCode', barcodeConfig));
+      resolve(sendAndHandleSdkError('barcode.scan', barcodeConfig));
     });
   }
 
--- a/packages/teams-js/src/public/barcode.ts
+++ b/packages/teams-js/src/public/barcode.ts
@@ -2,7 +2,7 @@ import { sendAndHandleSdkError } from '../internal/communication';
 import { ensureInitialized } from '../internal/internalAPIs';
 import { validateScanBarCodeInput } from '../internal/mediaUtil';
 import { FrameContexts } from './constants';
-import { ErrorCode } from './interfaces';
+import { DevicePermission, DevicePermissionType, ErrorCode } from './interfaces';
 import { runtime } from './runtime';
 
 export namespace barcode {
@@ -23,15 +23,16 @@ export namespace barcode {
         throw { errorCode: ErrorCode.INVALID_ARGUMENTS };
       }
 
-      // could also not use the old message format here and totally break from media.ts, open to opinions
-      // actually thinking it might be good to totally break from media. Leaving in now to provoke discussion
-      resolve(sendAndHandleSdkError('media.scanBarCode', barcodeConfig));
+      // Decided not to use the old message format here and totally break from media.ts
+      resolve(sendAndHandleSdkError('barcode.scanBarCode', barcodeConfig));
     });
   }
 
   export function hasPermission(): Promise<boolean> {
+    const permissions: DevicePermission[] = [{ type: DevicePermissionType.Media }];
+
     return new Promise<boolean>(resolve => {
-      resolve(sendAndHandleSdkError('barcode.hasPermission'));
+      resolve(sendAndHandleSdkError('permission.has', permissions));
     });
   }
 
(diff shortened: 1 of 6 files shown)


packages/teams-js/src/public/location.ts
```diff
@@ -48,25 +48,71 @@ export namespace location {
...
+    });
+  }
+
+  // This should not trigger the "refresh the app scenario" because this is for setting things up
+  // for use through teamsjs-sdk 2.0. If the user DOES refresh the app after calling this the iframe
+  // would have the new allow parameters, but only the AppPermissions dialog should trigger the
+  // "ask the user to refresh" flow
+  export function requestPermission(): Promise<boolean> {
+    return new Promise<boolean>(resolve => {
+      resolve(sendAndHandleError('location.requestPermission'));
```
MarcinJuraszek, line 71: would it be more scalable to use `requestPermission` as message type and send the kind of permissions being request as parameters? I'm trying to think about it form the perspective of host and host sdk. Because the flow when permissions is requested shouldn't differ (show UI, allow use to toggle a button) I would expect the host having a singular implementation for these. Which means having them all come across the isolation boundary as separate messages, with separate message handlers, is a bit unexpected.

MarcinJuraszek: Same for `location.hasPermission` above.

packages/teams-js/src/public/location.ts
```diff
@@ -48,25 +48,71 @@ export namespace location {
...
+      throw { errorCode: ErrorCode.OLD_PLATFORM };
+    }
+    if (!props) {
+      throw { errorCode: ErrorCode.INVALID_ARGUMENTS };
+    }
+
+    callCallbackWithErrorOrResultFromPromiseAndReturnPromise<Location>(getLocationHelper, callback, props);
   }
 
   function getLocationHelper(props: LocationProps): Promise<Location> {
```
MarcinJuraszek, line 118: this is only used once now, which means it might be better to remove it and have the non-promise based `getLocation` use `sendMessageToParent(..., callback)` directly. Same below.

MarcinJuraszek: Right now we go from callback, to a promise, and back to a callback. That middle step seems unnecessary if all we need is a callback version.

TrevorJoelHarris: Agreed, I will leave this as an exercise for the reader / implementer to restore this to the original un-promisified back compat function correctly. Functionality here is basically correct but as marcin points out no reason to go back and forth between callback and promise.

packages/teams-js/src/public/barcode.ts
```diff
@@ -0,0 +1,51 @@
...
+import { validateScanBarCodeInput } from '../internal/mediaUtil';
+import { FrameContexts } from './constants';
+import { ErrorCode } from './interfaces';
+import { runtime } from './runtime';
+
+export namespace barcode {
+  export interface BarCodeConfig {
+    /**
+     * Optional; Lets the developer specify the scan timeout interval in seconds
+     * Default value is 30 seconds and max allowed value is 60 seconds
```
TrevorJoelHarris, line 12: Generally speaking, not crazy about defaults enforced in comments only. Will think about best way to address this

packages/teams-js/src/public/location.ts
TrevorJoelHarris: That makes sense to me, will tweak

packages/teams-js/src/public/location.ts
```diff
@@ -48,25 +48,76 @@ export namespace location {
...
-    return callCallbackWithErrorOrResultFromPromiseAndReturnPromise<Location>(getLocationHelper, callback, props);
+
+    if (!isCurrentSDKVersionAtLeast(locationAPIsRequiredVersion)) {
+      throw { errorCode: ErrorCode.OLD_PLATFORM };
+    }
+    if (!props) {
+      throw { errorCode: ErrorCode.INVALID_ARGUMENTS };
+    }
+
+    callCallbackWithErrorOrResultFromPromiseAndReturnPromise<Location>(getLocationHelper, callback, props);
```
Jessica-Koch, line 120: I think you want callCallbackWithSDKErrorOrResultFromPromiseAndReturnPromise since you're returning an sdk error but it seems both of those functions take sdkErrors so I could be wrong

PR #1052, opened 2022-04-08
appEntity capability support updates

aeunms:
Updates APIs in appEntity capability to throw not supported on platform error. There wasn't already a test for appEntity, so I created a new file. I have only added the test case regarding the new error being thrown as this work is top priority currently. The rest of the tests should be added through a separate ticket which I will create.


packages/teams-js/src/private/appEntity.ts
```diff
@@ -72,6 +73,10 @@ export namespace appEntity {
   ): void {
     ensureInitialized(FrameContexts.content);
 
+    if (!isSupported()) {
+      throw new Error(errorNotSupportedOnPlatform);
```
erinha, line 77: Since the isSupported work started, we decided to throw SdkError instead of Error to be consistent.

change/@microsoft-teams-js-59df1594-8bfb-454c-8223-13c6e8c1e73d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "appEntity APIs will now throw a NOT_SUPPORTED_ON_PLATFORM error if called while capability is not supported.",
```
erinha, line 3: This should be consistent with the other isSupported change files: "Adding isSupported check to appEntity capability"

erinha, line 3: This should be consistent with the other isSupported change files: "Adding isSupported check to appEntity capability"

erinha (changes requested):
Can you please update after pulling the latest from 2.0-preview? Thanks!

change/@microsoft-teams-js-59df1594-8bfb-454c-8223-13c6e8c1e73d.json
shrshindeMSFT: Fixed it in latest commit.

erinha (approved):
Looks good, thanks!

shrshindeMSFT (approved):

PR #1054, opened 2022-04-08
Jekoch/open download folder

Jessica-Koch:
* promisify openDownloadFolder
* update tests and add new ones


packages/teams-js/test/private/files.spec.ts
```diff
@@ -548,54 +550,123 @@ describe('files', () => {
...
+      it('should not allow calls before initialization', async () => {
+        expect(() => files.openDownloadFolder(null)).toThrowError('The library has not yet been initialized');
+      });
+
+      Object.keys(FrameContexts)
+        .map(k => FrameContexts[k])
+        .forEach(frameContext => {
+          it(`should not allow calls from ${frameContext} context`, async () => {
+            if (frameContext === FrameContexts.content) {
+              return;
```
shrshindeMSFT, line 629: Unit Test case guideline states we avoid the use of `return`, reference code : `Object.keys(FrameContexts).forEach(k => {
      const context = FrameContexts[k];
      if (allowedContexts.some(allowedContext => allowedContext === context)) {`
      add test cases in if block.

packages/teams-js/test/private/files.spec.ts
```diff
@@ -548,54 +550,123 @@ describe('files', () => {
...
+        expect(() => files.openDownloadFolder(null, emptyCallback)).toThrowError(
+          'The library has not yet been initialized',
+        );
+      });
+      Object.keys(FrameContexts)
+        .map(k => FrameContexts[k])
+        .forEach(frameContext => {
+          it(`should not allow calls from ${frameContext} context`, async () => {
+            if (frameContext === FrameContexts.content) {
+              return;
```
shrshindeMSFT, line 564: Unit Test case guideline states we avoid the use of return, reference code : Object.keys(FrameContexts).forEach(k => { const context = FrameContexts[k]; if (allowedContexts.some(allowedContext => allowedContext === context)) {
add test cases in if block.

shrshindeMSFT (changes requested):

packages/teams-js/test/private/files.spec.ts
Jessica-Koch: are you sure, i've seen that return case in other modules, calendar and mail being examples. I don't see this in the unit test case guidelines anywhere but I may be missing it.  I'm fine changing it, but I'd like to know where the guideline is so that I can avoid these in the future, i've been reading off of the guide @erinha  posted

shrshindeMSFT: Check the note where the preference is using `async/await` 
https://office.visualstudio.com/MetaOS/_wiki/wikis/MetaOS.wiki/43604/Unit-Test-Guidelines

shrshindeMSFT (approved):

change/@microsoft-teams-js-fd11873b-f192-4fe9-b95c-fa2cad38dbc6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "promisify openDownloadFolder and add appropriate tests",
```
erinha, line 3: Can you please capitalize the start of the comment?

erinha, line 3: Can you please capitalize the start of the comment?

erinha:
Thank you for doing this work! Since we are changing files back to what it is in v1, this PR is no longer needed.

PR #1056, opened 2022-04-11
Adding userSessionId as part of Context in getContextAPI

meethjain:
- Adding a userSessionID as a part of the Context Object via getContext API
- This id would be unique per user & per platform application
- It would be scoped to the login/logout session of the teams native app user & would be cleared after the user logs out of the teams native app.  

- This is part of a requirement from one of the partner apps (Assignments)

### DIFF
--- a/src/public/interfaces.ts
+++ b/src/public/interfaces.ts
@@ -294,7 +294,7 @@ export interface Context {
   /**
    * Unique ID which is tied to the login/logout session of the user for the Host Teams Native App.
    */
-   userSessionId?: string;
+  userSessionId?: string;
 
   /**
    * The user's role in the team.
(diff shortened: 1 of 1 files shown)


TrevorJoelHarris:
Hi @meethjain, apologies that no one responded to this earlier. A change like this would need to go through an internal design review. Also, the context structure you're adding this property to is no longer the most current one, that one is in app.ts so you would need to add it there instead.

In the design review, I'd be interested to hear why current context values can't be combined to get the effect you need.

I'm going to close this review because it's been open for a long time and you haven't reached out and tried to get any further reviews on it in the last 6 months, but if you are still planning on checking it in please let me know and I can help you set up a design review for your change.

PR #1058, opened 2022-04-11
Split private chat capability into a public chat capability and a private conversation capability

AE-MS:
There once was a private `conversation` capability. There was also a [deeplink-based way](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/build-and-test/deep-links#deep-linking-to-a-chat) of opening chats in Teams. As part of the effort to transform deeplink-based functionality into strongly-typed capabilities, new methods were added to TeamsJS to support creating chats (mimicking what was available via deeplinks). Since there already was a `conversation` capability at the time, these new chat-creation methods were grouped with that existing `conversation` capability and it was renamed `chat`.

This is a problem because the previously-only-available-via-deeplink chat functions should be public, since they were previously available publicly (via the deeplink syntax) and we want to encourage developers to switch to using strongly typed APIs instead of deeplinks for this purpose. However, we can't offer that encouragement if there is no public way to do it.

This change splits the private `chat` capability into two: a public `chat` capability (containing all the functions which do what the deeplink syntax used to do) and a private `conversation` capability that did what it always did before the chat+conversation merge.

This particular change is only about moving code/renaming files/tests. There is no actual code change in here -- just movement /renaming of functions. A subsequent iteration will update inline documentation and other necessary things like that (I wanted to keep that separate so it would be easier to review what changed).


packages/teams-js/src/public/chat.ts
```diff
@@ -0,0 +1,119 @@
...
+      ensureInitialized(FrameContexts.content);
+      const sendPromise = sendAndHandleError('chat.openChat', {
+        members: openChatRequest.user,
+        message: openChatRequest.message,
+      });
+      resolve(sendPromise);
+    });
+  }
+  /**
+   * @hidden
```
erinha, line 82: If this is public, do we still want to hide documentation from the public docs?

packages/teams-js/src/public/chat.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * @hidden
+ * Namespace to interact with the conversational subEntities inside the tab
+ *
+ * @alpha
+ */
+export namespace chat {
+  /**
+   * @hidden
```
erinha, line 61: If this is public, do we still want to hide documentation from the official docs?

erinha (changes requested):
Question about API reference documentation annotations.

packages/teams-js/src/public/chat.ts
erinha: Note: this applies to all the @hidden annotations

erinha: Ahhh. I see in the PR description you plan to tackle these later. Carry on. 🚀

packages/teams-js/src/public/chat.ts
erinha: To be fixed later.

erinha (approved):
Documentation to be fixed later (as per PR description). Approved!

PR #1061, opened 2022-04-11
Update chat documentation

AE-MS:
Update `chat` documentation to reflect public nature of `chat` capability, now that it has been refactored with the prior change.


erinha (approved):
This looks great! Only question is whether we should add reference documentation for `isSupported()`. @erikadoyle do you have an opinion?

TrevorJoelHarris (dismissed):

erinha (approved):

PR #1063, opened 2022-04-12
Create utility function for constructing a chat deep link URL

AE-MS:
Create utility function for constructing a chat deep link URL

This utility function will be used to create a deep link URL that will
be sent to the host if it is determined that the host is "legacy teams"
and would not understand the chat capability messages.
This change does not actually execute the deep link (that will be in
the next change), it only constructs the deep link. It also includes
unit tests to validate the construction.

### DIFF
--- a/packages/teams-js/src/internal/chatConstants.ts
+++ b/packages/teams-js/src/internal/chatConstants.ts
@@ -1,0 +1,4 @@
+export const deepLinkUrlPathForChat = '/l/chat/0/0';
+export const usersUrlParameterName = 'users';
+export const topicUrlParameterName = 'topic';
+export const messageUrlParameterName = 'message';
--- a/packages/teams-js/src/internal/chatConstants.ts
+++ b/packages/teams-js/src/internal/chatConstants.ts
@@ -1,4 +1,4 @@
-export const deepLinkUrlPathForChat = '/l/chat/0/0';
-export const usersUrlParameterName = 'users';
-export const topicUrlParameterName = 'topic';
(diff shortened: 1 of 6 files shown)


erinha (approved):

packages/teams-js/src/internal/chatConstants.ts
```diff
@@ -0,0 +1,4 @@
+export const deepLinkUrlPathForChat = '/l/chat/0/0';
```
TrevorJoelHarris, line 1: are these teams URL specific? If so it might be useful to name them or comment to that effect

packages/teams-js/src/internal/chatUtilities.ts
```diff
@@ -0,0 +1,20 @@
+import {
```
TrevorJoelHarris, line 1: Looking forward to the work Jay will have to do for the non-chat capabilities that are like this, would it make more sense to name this file something like deepLinkUtilities? That way similar functionality could be put in here

AE-MS: My take on things like this is that it's good to think about future extensibility but given that the main thing we would do to potentially support that in the future is just renaming, I'd prefer to wait until we have the second consumer of this and then rename it, rather than pre-emptively rename it and then somehow discover later that Jay's work has to be done in a slightly different way. Said differently: I hope this is helpful to Jay, and it should be very easy for him to rename it when he's ready.

TrevorJoelHarris (approved):

PR #1072, opened 2022-04-13
Change DeepLinkParameters not to use subEntity* anymore

MarcinJuraszek:


jadahiya-MSFT (approved):

jadahiya-MSFT:
Android E2E tests seem to be a bit flaky and might need to be rerun, I've talked with @jekloudaMSFT who is investigating why this is happening.

MarcinJuraszek:
I think in this case it's an actual test failure. I pushed a new commit to try to mitigate that.

jadahiya-MSFT (dismissed):

MarcinJuraszek:
OK. Looks like it wasn't my fault after all :)

jadahiya-MSFT (dismissed):

apps/teams-test-app/src/components/PagesAPIs.tsx
```diff
@@ -95,11 +95,27 @@ const ShareDeepLink = (): ReactElement =>
...
         },
         withCallback: (input, setResult) => {
-          shareDeepLink(input);
+          if (input.subEntityId && input.subEntityLabel) {
+            shareDeepLink(input);
+          } else {
+            shareDeepLink({
+              subEntityId: input.subPageId,
+              subEntityLabel: input.subPageLabel,
+              subEntityWebUrl: input.subPageWebUrl,
```
erinha, line 116: Is this intentionally reversed vs the previous change block?

packages/teams-js/src/public/interfaces.ts
```diff
@@ -513,6 +513,29 @@ export interface Context {
...
+  /**
+   * The fallback URL to which to navigate the user if the client cannot render the page.
+   * This URL should lead directly to the sub-entity.
+   */
+  subPageWebUrl?: string;
+}
+
+/**
+ * @deprecated
+ * As of 2.0.0-beta.1, please use {@link ShareDeepLinkParameters} instead.
```
erinha, line 537: I think this tag should say "As of 2.0.0-beta.5", right?

erinha (changes requested):
A couple small comments

apps/teams-test-app/src/components/PagesAPIs.tsx
MarcinJuraszek: This is calling into the back-compat function which takes old parameters. So yes, this is meant to be the other way around :)

jadahiya-MSFT (approved):

erinha (approved):
🚀

PR #1082, opened 2022-04-14
dialog.submit sends an empty array in case of no appIds

lakhveerk:
Sending an empty array is less error prone than sending `undefined` in an array.

### DIFF
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -79,7 +79,7 @@ export namespace dialog {
     ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
 
     // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
-    sendMessageToParent('tasks.completeTask', [result, Array.isArray(appIds) ? appIds : [appIds]]);
+    sendMessageToParent('tasks.completeTask', [result, appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : []]);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/public/dialog.ts
```diff
@@ -79,7 +79,7 @@ export namespace dialog {
     ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
 
     // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
-    sendMessageToParent('tasks.completeTask', [result, Array.isArray(appIds) ? appIds : [appIds]]);
+    sendMessageToParent('tasks.completeTask', [result, appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : []]);
```
lakhveerk, line 82: I cannot think of any scenarios, that can break v1 compatibility with this change. If you do, please let me know! :)

jadahiya-MSFT: This shouldn't break back compat, but I would check to see what is expected from the hosts perspective. Maybe they are expecting an `undefined` in an array? Maybe it's better to just send `undefined`?

jadahiya-MSFT (dismissed):

packages/teams-js/src/public/dialog.ts
MarcinJuraszek: Should we add a UT around this?

lakhveerk: Good idea! Added a unit test for this!

jadahiya-MSFT (approved):

MarcinJuraszek (approved):

PR #1083, opened 2022-04-14
Update CDN test name for clarity

AE-MS:
This build step only runs on release builds, so I'm updating the display name in case people get confused about why it was skipped for their build.


MarcinJuraszek (approved):

erinha (approved):
This is helpful, thank you!

PR #1086, opened 2022-04-15
shrshinde/isSupported-legacy-check

shrshindeMSFT:
- Adding isSupported check for legacy capability
- Adding unit test case for legacy capability and updating existing


lakhveerk (dismissed):

erinha (approved):
Looks good, thanks!

lakhveerk (approved):

PR #1087, opened 2022-04-15
shrshinde/isSupported-meeting-check

shrshindeMSFT:
- Adding isSupported check for meeting capability
- Adding and updating existing unit test cases


lakhveerk (approved):

packages/teams-js/src/public/meeting.ts
```diff
@@ -189,6 +189,9 @@ export namespace meeting {
 
   function getIncomingClientAudioStateHelper(): Promise<boolean> {
     return new Promise<boolean>(resolve => {
+      if (!isSupported()) {
```
lakhveerk, line 192: We are converting meeting.ts to v1 form. Do we still want these changes?

erinha: I think Lakhveer is right. We should close this review (sorry for not deciding this earlier!).

PR #1088, opened 2022-04-15
Added initialize function for dialogs when running in task FrameContext

jadahiya-MSFT:
- Added in an initialize function to dialog that will handle receiving messages prior to `registerOnMessageFromParent` has been called
- Modified DialogAPI.tsx to accommodate for these changes


packages/teams-js/src/public/dialog.ts
```diff
@@ -114,7 +123,13 @@ export namespace dialog {
    */
   export function registerOnMessageFromParent(listener: PostMessageChannel): void {
     ensureInitialized(FrameContexts.task);
+    removeHandler('messageForChild');
```
jadahiya-MSFT, line 126: This needs to be removed when registering to receive parent messages because we will now send messages.

packages/teams-js/src/public/dialog.ts
```diff
@@ -33,6 +33,15 @@ export namespace dialog {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   export type PostMessageChannel = (message: any) => void;
   export type DialogSubmitHandler = (result: ISdkResponse) => void;
+  const storedMessages: string[] = [];
+
+  export function initialize(): void {
```
TrevorJoelHarris, line 38: We should probably hide this from the documentation since I assume we don't want users to call this

jadahiya-MSFT: Good idea, I'll add in the correct tag

packages/teams-js/src/public/dialog.ts
lakhveerk: Can you write a comment line in code about un-registering please? It would help others to understand

packages/teams-js/src/public/dialog.ts
```diff
@@ -33,6 +33,23 @@ export namespace dialog {
...
   export type PostMessageChannel = (message: any) => void;
   export type DialogSubmitHandler = (result: ISdkResponse) => void;
+  const storedMessages: string[] = [];
+
+  /**
+   * @hidden
+   * Hide from docs because this function is only used during initialization
+   * ------------------
+   * Adds register handlers for messageForChild upon initialization and only in the tasks FrameContext. {@link FrameContexts.task}
+   * Function is called in {@link app.initializeHelper}
```
lakhveerk, line 43: Does this link actually go to app.initializeHelper? I thought, we would need to have 'app' imported in this file to make the link.

packages/teams-js/src/public/app.ts
```diff
@@ -518,6 +519,7 @@ export namespace app {
         authentication.initialize();
         menus.initialize();
         pages.config.initialize();
+        dialog.initialize();
```
lakhveerk, line 522: If a developer never uses a dialog, then the `dialogMessageHandler` will stay registered the whole time for no reason. 
Should we consider registering it only if a dialog is opened?

packages/teams-js/src/public/dialog.ts
jadahiya-MSFT: Good call, I'm going to remove the link because these docs are hidden anyways

packages/teams-js/src/public/app.ts
jadahiya-MSFT: There's no way to confirm that it is in a dialog or not, but the registered handler is non-invasive. Additionally it will only ever use `dialogMessageHandler` if we are in the `tasks` context.

packages/teams-js/src/public/dialog.ts
```diff
@@ -33,6 +33,23 @@ export namespace dialog {
...
+  /**
+   * @hidden
+   * Hide from docs because this function is only used during initialization
+   * ------------------
+   * Adds register handlers for messageForChild upon initialization and only in the tasks FrameContext. {@link FrameContexts.task}
+   * Function is called during app intitialization
+   * @internal
+   */
+  export function initialize(): void {
+    ensureInitialized(FrameContexts.task);
```
MarcinJuraszek, line 47: This will throw if framecontext is not `task`. That's not a good thing in this function, because `initialize` is called for every app, not just dialogs.

Looking at the other capabilities with `initialize` none of them uses `ensureInitialized`. So I think you shouldn't use it here either.

MarcinJuraszek (changes requested):

packages/teams-js/src/public/dialog.ts
```diff
@@ -131,6 +131,9 @@ export namespace dialog {
    */
   export function registerOnMessageFromParent(listener: PostMessageChannel): void {
     ensureInitialized(FrameContexts.task);
+    //We need to remove the original 'messageForChild'
+    //handler since the original does not allow for post messages.
+    //It is replaced by the user specified listener that is pased in.
```
lakhveerk, line 136: nit: passed

lakhveerk (approved):

packages/teams-js/src/public/dialog.ts
MarcinJuraszek: Well, sorry, I should have been more specific. I think we should keep a check and only register if frameContext is `task`, but it can't be done via `ensureInitialized` because that throws and that's not what we want.

jadahiya-MSFT: Oh don't worry, I picked that up on that. I'm just testing different checks right now. I tried doing `GlobalVars.frameContext === FrameContext.tasks`, unfortunately since GlobalVars isn't set at the time this leads to the handler not being set. My next approach is to do something similar to `menus.initialize` and have a helper function.

jadahiya-MSFT: I've added in the helper function

TrevorJoelHarris (dismissed):

lakhveerk (dismissed):

packages/teams-js/src/public/dialog.ts
erinha: Nit: please add a space" following "//".

MarcinJuraszek (approved):

TrevorJoelHarris (approved):

PR #1089, opened 2022-04-15
shrshinde/isSupported-meetingRoom-check

shrshindeMSFT:
- Adding isSupported check for meetingRoom capability
- Adding and updating unit test cases


packages/teams-js/test/private/meetingRoom.spec.ts
```diff
@@ -114,32 +121,38 @@ describe('meetingRoom', () => {
       );
     });
 
+    it('sendCommandToPairedMeetingRoom should throw error when meetingRoom is not supported.', async () => {
+      await framelessPlatformMock.initializeWithContext('content');
```
lakhveerk, line 125: Changes look good, but just noticing that we are not testing these unit tests for all frame contexts.

lakhveerk (approved):

packages/teams-js/test/private/meetingRoom.spec.ts
lakhveerk: I approved the changes, so you don't get blocked on this

erinha (approved):

packages/teams-js/test/private/meetingRoom.spec.ts
shrshindeMSFT: Thanks

PR #1090, opened 2022-04-15
shrshinde/isSupported-monetization-check

shrshindeMSFT:
- Adding isSupported check monetization capability
- Adding unit test cases


packages/teams-js/test/public/monetization.spec.ts
```diff
@@ -32,8 +35,19 @@ describe('monetization_v1', () => {
 
     Object.values(FrameContexts).forEach(context => {
       if (!allowedContexts.some(allowedContext => allowedContext == context)) {
+        it(`openPurchaseExperience should throw error when monetization is not supported. context: ${context}`, async () => {
+          await framelessPlatformMock.initializeWithContext('content');
+          expect.assertions(3);
```
lakhveerk, line 40: there are 3 assertions in initializeWithContext and 4th one here. So, I think it should be
`expect.assertions(4);`

shrshindeMSFT: We won't hit the 4th assertion as the function throws sdkError before the main logic

erinha: I'm a bit confused as well. We await the call to `initializeContext` so shouldn't that cause the first 3 assertions, then one more in the catch block. Which  particular assertion is missed because the function throws before the main logic?

erinha (changes requested):
Confusion over number of `expect.assertions` for `isSupported` unit test

packages/teams-js/test/public/monetization.spec.ts
lakhveerk: where does function throw sdkError? What `expect` statement won't be hit?

shrshindeMSFT: In frameless we need 3 assertions whereas in framed we need 1 assertion.

erinha (approved):

jadahiya-MSFT (approved):

PR #1091, opened 2022-04-15
Adding in runtime.isLegacyTeams handler to deep link capabilities

jadahiya-MSFT:


packages/teams-js/src/public/appInstallDialog.ts
```diff
@@ -24,8 +28,19 @@ export namespace appInstallDialog {
       if (!isSupported()) {
         throw new Error('Not supported');
       }
-      sendMessageToParent('appInstallDialog.openAppInstallDialog', [openAPPInstallDialogParams]);
-      resolve();
+      if (runtime.isLegacyTeams) {
```
TrevorJoelHarris, line 31: do we need unit tests for these changes?

jadahiya-MSFT: Unit tests have been added, I noticed there are some missing logic gaps in some of the tests (particularly calendar) next steps are to write new tests that will test functionality more extensively, but don't want to hold up this PR with those changes.

packages/teams-js/src/public/appInstallDialog.ts
```diff
@@ -24,8 +27,19 @@ export namespace appInstallDialog {
...
+        resolve(
+          sendAndHandleError(
+            'executeDeepLink',
+            `${teamsDeepLinkProtocol}://${teamsDeepLinkHost}${teamsDeepLinkUrlPathForAppInstall}${encodeURIComponent(
```
TrevorJoelHarris, line 34: Do you think it's worth putting this into deepLinkUtilities as well?

TrevorJoelHarris (approved):

packages/teams-js/src/public/appInstallDialog.ts
jadahiya-MSFT: Wasn't sure since it is just one component that is being translated. But I think in terms of future proofing it might be beneficial, so I will make that change.

TrevorJoelHarris (approved):

lakhveerk (approved):

PR #1097, opened 2022-04-17
Added in missing comments for authentication

jadahiya-MSFT:


packages/teams-js/src/public/authentication.ts
```diff
@@ -482,6 +484,9 @@ export namespace authentication {
     failureCallback?: (reason: string) => void;
   }
 
+  /**
+   * An interface that describes the Authentication pop up's parameters
```
erikadoyle, line 488: How about: 
Describes Authentication pop-up parameters

packages/teams-js/src/public/authentication.ts
```diff
@@ -507,6 +512,9 @@ export namespace authentication {
    */
   export type AuthenticateParameters = AuthenticatePopUpParameters & LegacyCallBacks;
 
+  /**
+   * An interface that describes the Authentication Token's Request Parameters
```
erikadoyle, line 516: Describes authentication token request parameters

erikadoyle (commented):
Looks good, just some minor things

erinha (approved):

packages/teams-js/src/public/authentication.ts
```diff
@@ -466,6 +466,8 @@ export namespace authentication {
   /**
    * @deprecated
    * As of 2.0.0-beta.1.
```
erikadoyle, line 468: Sorry! Just realized this--the @deprecated tag generates a full sentence at publish time: [This API is now deprecated](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/authentication.legacycallbacks?view=msteams-client-js-beta)
. . . so it kind of strands "As of 2.0.0.-beta.1." on its own line without much context (see link above for example). Could we rework each instance of "As of 2.0.0.-beta.1." in this file? Also @erinha correct me if I'm wrong, but I think we landed on dropping the "-beta.1" from the deprecated version note. 

How about something like: **As of 2.0.0, this [function|interface] has been deprecated in favor of a Promise-based model.**

jadahiya-MSFT: Got it, I'll make that change

erikadoyle: Awesome, thanks! Everything else looks good.

TrevorJoelHarris (approved):

packages/teams-js/src/public/authentication.ts
TrevorJoelHarris: @erikadoyle if you are able to approve this PR, please feel free to do that when you're happy with the updated comments!

shrshindeMSFT (approved):

PR #1098, opened 2022-04-18
Initiliazing for all FrameConexts for returnFocus api

Shyam-Gupta2011994:
Adding meetingStage FrameContext in returnFocus api

### DIFF
--- a/src/public/navigation.ts
+++ b/src/public/navigation.ts
@@ -13,10 +13,7 @@ import { sendMessageToParent } from '../internal/communication';
  * @param navigateForward Determines the direction to focus in teams app.
  */
 export function returnFocus(navigateForward?: boolean): void {
-  ensureInitialized(
-    FrameContexts.content,
-    FrameContexts.meetingStage
-  );
+  ensureInitialized();
 
   sendMessageToParent('returnFocus', [navigateForward]);
 }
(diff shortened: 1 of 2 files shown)


src/public/navigation.ts
```diff
@@ -13,7 +13,10 @@ import { sendMessageToParent } from '../internal/communication';
  * @param navigateForward Determines the direction to focus in teams app.
  */
 export function returnFocus(navigateForward?: boolean): void {
-  ensureInitialized(FrameContexts.content);
+  ensureInitialized(
```
ydogandjiev, line 16: Please change this to `ensureInitialized();`  since we will eventually want this API to work in all frame contexts.

ydogandjiev (changes requested):

src/public/navigation.ts
Shyam-Gupta2011994: Done

ydogandjiev (approved):

cloudtx (approved):

src/public/navigation.ts
```diff
@@ -13,7 +13,7 @@ import { sendMessageToParent } from '../internal/communication';
  * @param navigateForward Determines the direction to focus in teams app.
  */
 export function returnFocus(navigateForward?: boolean): void {
-  ensureInitialized(FrameContexts.content);
+  ensureInitialized();
```
ronakbakshi, line 16: Are you trying to add meeting stage context only? Or do you want this API to work in all contexts?

Shyam-Gupta2011994: Doing it for all now as per comments from Yuri

PR #1100, opened 2022-04-19
topic param name was wrong for group chat deep links

TrevorJoelHarris:
According to this, should be topicName: https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/build-and-test/deep-links


jadahiya-MSFT (approved):

AE-MS (approved):
Hmm wonder where I got the wrong one from.

jadahiya-MSFT:
> Hmm wonder where I got the wrong one from.

I think it might have been because I never explicitly exposed this, but upon switching to the execute deep link I should've double checked what the correct url format is. That's my bad.

AE-MS:
> > Hmm wonder where I got the wrong one from.
> 
> I think it might have been because I never explicitly exposed this, but upon switching to the execute deep link I should've double checked what the correct url format is. That's my bad.

Stuff happens. Glad we caught it!

jadahiya-MSFT (approved):

AE-MS (approved):

PR #1101, opened 2022-04-19
Revert media to v1 state

lakhveerk:
We're going to be treating media.ts as a back-compat ONLY file for GA. 
We'll be working overtime to expose functionality in media.ts in the right way. For now the namespace will be marked deprecated.


packages/teams-js/src/public/media.ts
```diff
@@ -25,17 +25,14 @@ import {
...
+import { generateGUID } from '../internal/utils';
+import { runtime } from '../public/runtime';
 import { FrameContexts, HostClientType } from './constants';
 import { ErrorCode, SdkError } from './interfaces';
-import { runtime } from './runtime';
 
 /**
+ * @deprecated
+ *
  * @alpha
```
lakhveerk, line 36: Do we want to keep the @alpha tag? It suggests that this namespace is intended to be used by third-party developers eventually but has not yet been released.

erinha: No, the @alpha tag is not supported in our documentation system. Please remove it. I'll be making a separate change to remove @alpha everywhere else.

packages/teams-js/src/public/media.ts
```diff
@@ -739,82 +689,46 @@ export namespace media {
...
+      const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
+      callback(invalidInput, null);
+      return;
+    }
 
-    return callCallbackWithErrorOrResultFromPromiseAndReturnPromise<string>(wrappedFunction, callback);
+    sendMessageToParent('media.scanBarCode', [config], callback);
   }
-
   export function isSupported(): boolean {
```
lakhveerk, line 732: v1 does not have isSupported function, but I think we would want to keep it?

erinha: No, we won't be adding isSupported infrastructure to namespaces that haven't been assessed for v2.0 alignment, so not adding any additional gates to the APIs under this namespace. Can you please remove this?

erinha: There will be changes necessary in the runtime.ts file as well, since this change should also remove `media` from the `IRuntime` interface.

packages/teams-js/src/public/media.ts
```diff
@@ -25,18 +25,13 @@ import {
...
-} from '../internal/utils';
+import { generateGUID } from '../internal/utils';
+import { runtime } from '../public/runtime';
 import { FrameContexts, HostClientType } from './constants';
 import { ErrorCode, SdkError } from './interfaces';
-import { runtime } from './runtime';
 
 /**
- * @alpha
+ * @deprecated
```
erinha, line 34: I generated the docs locally and I see that the deprecated tag only appears at the top of the page, and doesn't get applied to all the functions within the namespace. I think we want to add @deprecated to all the exported functions in `media`. 

@erikadoyle is there a recommended way to annotate the deprecated tag (or maybe a different approach to SDK reference documentation) to help developers understand that this file is deprecated as of v2, but we are still working on what to replace it with?

change/@microsoft-teams-js-2339868b-2814-483b-85e4-2d2075bd022e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Reverted media.ts to v1 state and made necessary changes in unit tests and test-app",
```
erinha, line 3: Can you re-word this to: "Updated `media` namespace to work as it did in v1 along with necessary changes to unit tests and teams-test-app" please?

packages/teams-js/src/public/media.ts
erikadoyle: Yes, each function needs its own @deprecated annotation in order to render the note. [appInitialization.ts](https://github.com/OfficeDev/microsoft-teams-library-js/blob/2.0-preview/packages/teams-js/src/public/appInitialization.ts) is a good example for this.

Don't think there's any MS Docs guidance for this specific scenario, but how about adding a note like this:

  /**
   * @deprecated
   * As of 2.0.0, use only for backwards compatibility of existing code.
   * Description goes here.
   */

erinha: Thanks @erikadoyle! That's super helpful. @lakhveerk could you please add the deprecated info as Erika described for each API in the namespace and the namespace itself? For right now, please use version 2.0.0-beta.6 in the comments. I'll be doing another pass before we release 2.0.0.

erinha (changes requested):
We're getting there, thanks! Mostly documentation updates left I think.

packages/teams-js/src/public/media.ts
lakhveerk: yeah thank you!

packages/teams-js/src/public/media.ts
```diff
@@ -739,82 +687,46 @@ export namespace media {
...
-   * @remarks
-   * Note: For desktop and web, this API is not supported. Callback will be resolved with ErrorCode.NotSupported.
-   *
-   * @deprecated
-   * As of 2.0.0-beta.1, please use {@link media.scanBarCode media.scanBarCode(config?: BarCodeConfig): Promise\<string\>} instead.
-   *
    * @param callback - callback to invoke after scanning the barcode
    * @param config - optional input configuration to customize the barcode scanning experience
    */
-  export function scanBarCode(callback: (error: SdkError, decodedText: string) => void, config?: BarCodeConfig);
```
TrevorJoelHarris, line 762: Not for this change, but probably we should update this documentation to point to updated barcode capability when that's ready.

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/media.ts
lakhveerk: Yeah, when we write the new bar code, we can update the documentation here.

erinha (approved):
Looks good, thank you!

TrevorJoelHarris (approved):

erinha (approved):

PR #1107, opened 2022-04-19
Add info to changelog about outputting ES6 modules from TS

erinha:
This change was introduced prior to Public Preview, however it was never documented in the changelog.

We recently discovered they may be other side effects of outputting ES6 modules from TypeScript instead of CommonJS, so this may be useful for developers.

### DIFF
--- a/packages/teams-js/CHANGELOG.md
+++ b/packages/teams-js/CHANGELOG.md
@@ -245,6 +245,7 @@ Tue, 01 Mar 2022 19:50:49 GMT
     - `applyRuntimeConfig` is added
 
 - Promises introduced
+
   - The following APIs that took in a callback function as a parameter now instead return a Promise.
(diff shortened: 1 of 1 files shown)


Jessica-Koch (approved):

packages/teams-js/CHANGELOG.md
```diff
@@ -267,6 +267,7 @@ Tue, 01 Mar 2022 19:50:49 GMT
     - `applyRuntimeConfig` is added
 
 - Promises introduced
+
```
erinha, line 270: FWIW I didn't intentionally put this newline here and I cannot remove it (in VS Code at least -- when I save the file, it auto-inserts the newline). Don't think it hurts leaving it in.

jadahiya-MSFT (approved):

PR #1108, opened 2022-04-19
Implemented private APIs needed for new Teams SDK

Stevenic:
Initial implementation of a private `interactive` namespace that's need for the new Teams SDK under development.

- Added new `interactive` namespace to list of private API's.
- Implemented unit tests for new API's.


packages/teams-js/src/private/interactive.ts
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * @hidden
+   * Hide from docs
+   * ------
+   * Returns the Fluid Tenant connection info for user's current context.
+   */
+  export function getFluidTenantInfo(): Promise<FluidTenantInfo> {
+    return new Promise<FluidTenantInfo>(resolve => {
+      ensureInitialized(FrameContexts.meetingStage, FrameContexts.sidePanel);
+      if (!isSupported()) {
```
MarcinJuraszek, line 79: you probably don't want these in your APIs. Teams does not send their 'support' matrix from existing clients, which means this would return false. I think for now it's best to keep this entire module out of the capabilities model.

packages/teams-js/src/private/interactive.ts
```diff
@@ -0,0 +1,207 @@
...
+        throw errorNotSupportedOnPlatform;
+      }
+
+      resolve(sendAndHandleSdkError('interactive.getClientRoles', clientId));
+    });
+  }
+
+  export function isSupported(): boolean {
```
MarcinJuraszek, line 204: See my other comment. I think for now it would be best to not include this in the capability model and keep it separate, without isSupported checks. Especially if it's not meant to be called by developers directly and if it's supposed to work in Teams (both angular and tmp) before they integrate the host sdk.

MarcinJuraszek: If you do want to be part of the capability model, this will need to be a new capability, and not part of `meeting`, as there are (or might be in the future) clients who will want support `meeting` but will not support this `interactive` stuff.

packages/teams-js/src/private/interactive.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   * ------
+   * The global time servers current timestamp.
+   */
+  export interface NtpTimeInfo {
+    ntpTime: string;
```
MarcinJuraszek, line 54: What format should this `string` representation be in?

MarcinJuraszek (changes requested):

Stevenic:
Closing and will re-open against different branch.

PR #1112, opened 2022-04-20
Add getDisplayMedia API

jinyu1122:
This is API is used to get user's selected display content(screen or window)
When it's called, Teams prompts the user to select and grant permission to capture the content of a display (screen or window) as MediaStream.

The SDK design has not been review yet, this PR is created to serve as the "Code Snippet" section of SDK design document,


TrevorJoelHarris:
This PR has seen no activity and been left in draft for 5 months. If you're still planning to check in, please re-open and remove the draft tag. A change like this would also have to go through a design review, please let me know if you need help setting that up.

PR #1113, opened 2022-04-20
Fixing dialog.open and dialog.bot.open function to send correct parameter type to the host

lakhveerk:


packages/teams-js/src/public/dialog.ts
```diff
@@ -225,4 +228,43 @@ export namespace dialog {
...
+
+  /**
+   * @hidden
+   * Hide from docs
+   * --------
+   * Convert UrlDialogInfo to DilogInfo to send the information to host in {@linkcode open} API.
+   *
+   * @internal
+   */
+  export function getDialogInfoFromUrlDialogInfo(urlDialogInfo: UrlDialogInfo): DialogInfo {
```
TrevorJoelHarris, line 240: Don't need to export this function

packages/teams-js/src/public/dialog.ts
```diff
@@ -225,4 +228,43 @@ export namespace dialog {
...
+
+  /**
+   * @hidden
+   * Hide from docs
+   * --------
+   * Convert BotUrlDialogInfo to DilogInfo to send the information to host in {@linkcode bot.open} API.
+   *
+   * @internal
+   */
+  export function getDialogInfoFromBotUrlDialogInfo(botUrlDialogInfo: BotUrlDialogInfo): DialogInfo {
```
TrevorJoelHarris, line 259: Don't need to export this function

packages/teams-js/src/public/dialog.ts
```diff
@@ -225,4 +228,43 @@ export namespace dialog {
       return runtime.supports.dialog ? (runtime.supports.dialog.bot ? true : false) : false;
     }
   }
+
+  /**
+   * @hidden
+   * Hide from docs
+   * --------
+   * Convert UrlDialogInfo to DilogInfo to send the information to host in {@linkcode open} API.
```
TrevorJoelHarris, line 236: "DialogInfo" instead of "DilogInfo"

packages/teams-js/src/public/dialog.ts
```diff
@@ -225,4 +228,43 @@ export namespace dialog {
...
+      fallbackUrl: urlDialogInfo.fallbackUrl,
+    };
+    return dialogInfo;
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   * --------
+   * Convert BotUrlDialogInfo to DilogInfo to send the information to host in {@linkcode bot.open} API.
```
TrevorJoelHarris, line 255: "DialogInfo" instead of "DilogInfo"

TrevorJoelHarris (dismissed):
If we have E2E tests that could have detected this, can you please also update those to check for this? (Either in this change or in a separate one)

change/@microsoft-teams-js-11fffd63-c4f2-4c8f-a18d-0e41115fb370.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "updated `dialog.open` and `dialog.bot.open` to send `DialogInfo` type over to the host instead of `UrlDialogInfo` or `BotUrlDialogInfo` type",
```
erinha, line 3: Can you please capitalize "Updated"?

lakhveerk:
> If we have E2E tests that could have detected this, can you please also update those to check for this? (Either in this change or in a separate one)

@jadahiya-MSFT Jay was adding e2e tests. I will follow up with him.

packages/teams-js/src/public/dialog.ts
lakhveerk: Thank you for catching it. I need to pay more attention to spellings and grammar 😣

erinha (changes requested):
Thanks for making these changes! Couple small comments to fix.

packages/teams-js/src/public/dialog.ts
```diff
@@ -233,15 +233,15 @@ export namespace dialog {
...
+   * Convert UrlDialogInfo to DialogInfo to send the information to host in {@linkcode open} API.
    *
    * @internal
    */
   export function getDialogInfoFromUrlDialogInfo(urlDialogInfo: UrlDialogInfo): DialogInfo {
     const dialogInfo: DialogInfo = {
       url: urlDialogInfo.url,
-      height: urlDialogInfo.size.height,
-      width: urlDialogInfo.size.width,
+      height: urlDialogInfo.size ? urlDialogInfo.size.height : DialogDimension.Small,
```
erinha, line 243: This might be a stupid question, but how did we decide to default to Small if no size is provided? Is this matching legacy behavior?

lakhveerk: We just picked "small" as default in our meeting with partners.
length/width are optional in tasks but, the "size" is a required parameter in dialogs. That means there should always be something in height/width. 
because of that I thought we would not need to check for urlDialogInfo.size ?, but seems like things still somehow can make through without providing size. Hence this extra checking.

erinha (approved):

TrevorJoelHarris (approved):

PR #1114, opened 2022-04-21
shrshinde/is supported pages check

shrshindeMSFT:


packages/teams-js/src/public/pages.ts
```diff
@@ -474,6 +523,9 @@ export namespace pages {
      * @param handler The handler to invoke when the user presses their Team client's back button.
      */
     export function registerBackButtonHandler(handler: () => boolean): void {
+      if (!isSupported()) {
```
erinha, line 526: Interesting that this function doesn't have `ensureInitialized` at the beginning. I think that is an oversight and it should be added.

erinha, line 526: Interesting that this function doesn't have `ensureInitialized` at the beginning. I think that is an oversight and it should be added.

erinha: I see that in publicAPIs.ts, the v1 version of `registerBackButtonHandler` does have `ensureInitialized()` in it, so we should definitely add it here too.

erinha: I see that in publicAPIs.ts, the v1 version of `registerBackButtonHandler` does have `ensureInitialized()` in it, so we should definitely add it here too.

shrshindeMSFT: Sure, I'll add it.

packages/teams-js/src/public/pages.ts
```diff
@@ -523,6 +523,7 @@ export namespace pages {
      * @param handler The handler to invoke when the user presses their Team client's back button.
      */
     export function registerBackButtonHandler(handler: () => boolean): void {
+      ensureInitialized();
```
erinha, line 526: Thanks for the quick fix! I'm a little surprised that a unit test didn't need to be updated because I would have expected a unit test to catch this. We should have unit tests for each API that fails if the library isn't initialized. Can you please add unit tests for this now that we have `ensureInitialized` added?

shrshindeMSFT: Added in the latest commit.

erinha: Thanks!

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -42,6 +42,21 @@ describe('Testing pages module', () => {
...
+            expect.assertions(1);
+            try {
+              pages.returnFocus();
+            } catch (e) {
+              expect(e).toEqual(errorNotSupportedOnPlatform);
+            }
+            utils.setRuntimeConfig({
+              apiVersion: 1,
+              supports: { pages: { appButton: {}, backStack: {}, config: {}, fullTrust: {}, tabs: {} } },
+            });
```
erinha, line 57: Why is this code necessary if we're resetting the runtime config in the `afterEach` code on line 31?

erinha: This comment applies to all the places we're doing this in the tests.

shrshindeMSFT: I'll update them in the next commit.

erinha: If you determine that it is required, can we add a helper method that resets the pages capabilities and subcapabilities?

erinha: Thinking more about this...If I understand this correctly, this seems like something that should be done as part of test setup for each test that requires that the pages specific capability be supported in order to test the behavior.

shrshindeMSFT: Sure, I'll work on it and get back to you.

erinha (changes requested):
Question about runtime resetting code in the unit tests

erinha (dismissed):
🚀  Thank you!

erinha (approved):

jadahiya-MSFT (approved):

PR #1116, opened 2022-04-21
shrshinde/isSupported-remoteCamera-check

shrshindeMSFT:
- Adding isSupported check to remoteCamera capability.
- Adding unit test cases.


packages/teams-js/test/private/remoteCamera.spec.ts
```diff
@@ -61,6 +62,23 @@ describe('remoteCamera', () => {
 
     Object.values(FrameContexts).forEach(context => {
       if (allowedContexts.some(allowedContexts => allowedContexts === context)) {
+        it(`remoteCamera.getCapableParticipants should throw error when remoteCamera is not supported. context : ${context}`, async () => {
```
erinha, line 65: I think there should be at least one test  case where isSupported returns true and we set the runtimeConfig explicitly. Can you please add one if it's not present?

packages/teams-js/test/private/remoteCamera.spec.ts
```diff
@@ -2,8 +2,9 @@ import { Utils } from '../utils';
 import { remoteCamera } from '../../src/private/remoteCamera';
 import { app } from '../../src/public/app';
 import { SdkError } from '../../src/public/interfaces';
-import { FrameContexts } from '../../src/public';
+import { call, FrameContexts } from '../../src/public';
```
erinha, line 5: Typo? Don't see  `call` being used anywhere here.

erinha (changes requested):
Looks good overall, one small comment.

shrshindeMSFT:
I'll make those changes and push it.

packages/teams-js/test/private/remoteCamera.spec.ts
```diff
@@ -375,6 +440,21 @@ describe('remoteCamera', () => {
 
     Object.values(FrameContexts).forEach(context => {
       if (allowedContexts.some(allowedContexts => allowedContexts === context)) {
+        it(`remoteCamera.registerOnCapableParticipantsChangeHandler should throw error when remoteCamera is not supported. context : ${context}`, async () => {
+          await utils.initializeWithContext(context);
+          utils.setRuntimeConfig({ apiVersion: 1, supports: {} });
+          let CapableParticipants: remoteCamera.Participant[];
```
erinha, line 446: nit: lower case "c" in "capableParticipants" for consistency.

erinha (changes requested):
Sorry  if I missed the small comment about variable naming in the tests before.

packages/teams-js/test/private/remoteCamera.spec.ts
shrshindeMSFT: Fixed in the latest commit.

erinha (approved):
Looks good, thanks!

aeunms (approved):
🚀

PR #1117, opened 2022-04-21
shrshinde/isSupported-logs-check

shrshindeMSFT:
- Adding isSupported check to logs capability.
- Adding unit test cases.
- Fixing breaking unit test cases.


packages/teams-js/src/public/constants.ts
```diff
@@ -82,3 +82,18 @@ export enum ChannelType {
 }
 
 export const errorNotSupportedOnPlatform: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
+
+export const minRuntimeConfigToUninitialize = {
```
erinha, line 86: Is this only used for unit tests? If so, I think we should prefix it with "_" (like we do for `app._initialize`, etc.) and add an @hidden tag, with a comment explaining it's only for unit tests, so it's not included in the public docs.

shrshindeMSFT: Sure, will make the changes accordingly.

erinha: Also please add an @internal tag. :)

erinha: I'm also wondering if this shouldn't go in the runtime.ts file, rather than constants.ts

packages/teams-js/src/public/runtime.ts
```diff
@@ -102,7 +102,7 @@ export const teamsRuntimeConfig: IRuntime = {
       update: {},
     },
     files: {},
-    logs: {},
+    log: {},
     media: {},
     meeting: {},
     meetingRoom: {},
```
erinha, line 108: Note: I don't think the `versionConsts` on line 147 should be exported. Can you fix this while you're touching this file please?

packages/teams-js/src/public/constants.ts
shrshindeMSFT: I thought of putting all the constants in contants.ts but will move to runtime.ts

packages/teams-js/src/public/runtime.ts
shrshindeMSFT: We need to export it as we are using it in `runtime.ts`

change/@microsoft-teams-js-690f07e9-cae4-40ab-89e5-b5700ba6ab91.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Initial commit to reset runtimeConfig to minimum for isSupported work",
```
erinha, line 3: Can you add another comment for calling out the "Adding isSupported check to logs capability"?

erinha: Reminder, you can run 'yarn beachball bump' to see how your changes will appear in the changelog. (I've found it very useful!)

shrshindeMSFT: Sure.

packages/teams-js/src/public/runtime.ts
erinha: This is in runtime.ts :) You can remove the export. I tested it and it works.

shrshindeMSFT: It'll work but when doing unit test cases, we need this function so if there is any other way, please let me know

erinha: Ahhh gotcha. Okay, fine to leave!

erinha (approved):
🚀

packages/teams-js/src/private/logs.ts
```diff
@@ -1,6 +1,7 @@
 import { sendMessageToParent } from '../internal/communication';
```
aeunms, line 1: Should we also change the name of the namespace and also file? If not, what are the guidelines for what should stay 'logs' and what should change to 'log'?

packages/teams-js/src/public/runtime.ts
```diff
@@ -205,3 +206,27 @@ export function generateBackCompatRuntimeConfig(highestSupportedVersion: string)
...
+/**
+ * @hidden
+ * Hide from docs.
+ * ------
+ * Constant used to set minimum runtime configuration
+ * while un-initializing an app in unit test case.
+ *
+ * @internal
+ */
+export const _minRuntimeConfigToUninitialize = {
```
aeunms, line 219: Seems like this PR is two changes- changing the logs/log and also changing the uninitialize structure. Maybe it would be good to split into two PRs? Or maybe that's already been discussed and decided against

packages/teams-js/src/private/logs.ts
erinha: We should not change the name of the namespace. There was a problem with the runtime being inconsistent with the host SDK and the decision was made to update the teams-js SDK to match what is in the host SDK. Although now that i think about it, I think this does set a confusing precedent.

packages/teams-js/src/public/runtime.ts
erinha: Great catch Angela. I agree that it would be cleaner to separate these into two separate PRs.

packages/teams-js/src/public/runtime.ts
```diff
@@ -100,7 +100,8 @@ export const teamsRuntimeConfig: IRuntime = {
       bot: {},
       update: {},
     },
-    logs: {},
+    // files: {},
```
shrshindeMSFT, line 103: Kept this for future reference when files will be supported.

packages/teams-js/src/private/logs.ts
aeunms: We should follow up on clarifying what should stay 'logs' vs 'log'.

packages/teams-js/src/public/runtime.ts
```diff
@@ -18,8 +18,9 @@ export interface IRuntime {
       readonly update?: {};
     };
     readonly location?: {};
-    readonly log?: {};
+    readonly logs?: {};
     readonly mail?: {};
+    readonly media?: {};
```
erinha, line 23: Merge conflict resolution gone wrong? I think that the change to remove `media` from the runtime happened since you last pulled, so there should not be references to `media` in the runtime.ts.

packages/teams-js/src/public/runtime.ts
erinha: I'm not sure we need this. We don't know what shape files will be in when it's ready for capability support, so I think we should just remove this.

packages/teams-js/test/private/files.spec.ts
```diff
@@ -1,14 +1,22 @@
-import { files } from '../../src/private/files';
```
erinha, line 1: Please make sure this file matches what's checked in to 2.0-preview. Lakhveer recently made changes and we don't want to undo them.

shrshindeMSFT:
Closing Reason - Too much confusion no clear goals in this pull request.

PR #1118, opened 2022-04-22
Reverting Files capability to v1 state

lakhveerk:
For GA release, we want to revert it back and going to treat it as only back-compat

### DIFF
--- a/packages/teams-js/src/private/files.ts
+++ b/packages/teams-js/src/private/files.ts
@@ -303,9 +303,6 @@ export namespace files {
    * @hidden
    * Hide from docs
    *
-   * @deprecated
-   * As of 2.0.0-beta.6, use only for backwards compatibility of existing code.
-   *
    * Gets a list of cloud storage folders added to the channel
    * @param channelId - ID of the channel whose cloud storage folders should be retrieved
    * @param callback - Callback that will be triggered post folders load
@@ -327,9 +324,6 @@ export namespace files {
   }
 
   /**
-   * @deprecated
-   * As of 2.0.0-beta.6, use only for backwards compatibility of existing code.
-   *
    * @hidden
    * Hide from docs
    * ------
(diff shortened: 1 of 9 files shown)


packages/teams-js/src/private/files.ts
```diff
@@ -445,58 +457,32 @@ export namespace files {
...
 
   /**
-   * @hidden
-   * Hide from docs.
-   * ------
-   * Opens a client-friendly preview of the specified file.
+   * @deprecated
+   * As of 2.0.0-beta.6, use only for backwards compatibility of existing code.
    *
-   * @param file - The file to preview.
```
TrevorJoelHarris, line 453: Was this something that was added specifically for 2.0? I don't see it anywhere in the updated code which was surprising.

packages/teams-js/test/private/files.spec.ts
```diff
@@ -18,59 +17,43 @@ describe('files', () => {
     utils.mockWindow.parent = utils.parentWindow;
 
     // Set a mock window for testing
-    app._initialize(utils.mockWindow);
```
TrevorJoelHarris, line 21: Fine to leave this as is but in case anyone else is looking and curious: app._initialize is called by _initialize so this change is actually the same

TrevorJoelHarris (approved):

packages/teams-js/src/private/files.ts
erinha: This should be added back to privateAPIs.ts

erinha: This particular API made me realize that we should not actually mark any of these APIs as deprecated. This is because the APIs in the private folder have not had the same  treatment of updating callbacks to return promises for v2. So they don't need to be marked deprecated because there isn't a new standard to use yet. (Meeting and media are exceptions because we did do a pass across all public APIs.)

erinha (changes requested):
Comments about openFilePreview and deprecated annotations.

erinha (approved):
Looks great! Thank you!

erinha (approved):

TrevorJoelHarris (approved):

PR #1126, opened 2022-04-25
shrshinde/isSupported-menus-check

shrshindeMSFT:
- Adding isSupported check to menus capability
- Adding unit test cases for menus capability

### DIFF
--- a/packages/teams-js/test/public/menus.spec.ts
+++ b/packages/teams-js/test/public/menus.spec.ts
@@ -1,0 +1,260 @@
+import { app, FrameContexts, menus } from '../../src/public';
+import { errorNotSupportedOnPlatform } from '../../src/public/constants';
+import { _minRuntimeConfigToUninitialize } from '../../src/public/runtime';
+import { FramelessPostMocks } from '../framelessPostMocks';
+import { Utils } from '../utils';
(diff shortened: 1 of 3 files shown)


packages/teams-js/test/public/menus.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+      };
+
+      it('should not allow calls before initialization', () => {
+        expect(() => menus.setUpViews([viewConfiguration], () => true)).toThrowError(
+          'The library has not yet been initialized',
+        );
+      });
+
+      Object.values(FrameContexts).forEach(frameContext => {
+        it(`menus.setUpViews should throw error when menus is not supported  when set to true in ${frameContext} context`, async () => {
```
erinha, line 39: nit: extra space between "not supported" and "when set"

erinha: Also I think this was copy/pasted elsewhere. :)

erinha (approved):
Nit of a comment. Looks good otherwise, thanks for adding unit tests!

aeunms (approved):
+1 to Erin's nit comments, but it's still approved!

PR #1131, opened 2022-04-25
shrshinde/isSupported-teams-check

shrshindeMSFT:
- Adding `isSupported` check to teams capability
- Adding `isSupported` unit test cases
- Cleaning up teams unit test cases.

### DIFF
--- a/packages/teams-js/test/private/teams.spec.ts
+++ b/packages/teams-js/test/private/teams.spec.ts
@@ -150,5 +150,16 @@ describe('Testing teams capabillity', () => {
         });
       });
     });
+
+    describe('Testing teams.isSupported function', () => {
+      it('teams.isSupported should return false if the runtime says teams is not supported', () => {
+        utils.setRuntimeConfig({ apiVersion: 1, supports: {} });
+        expect(teams.isSupported()).not.toBeTruthy();
+      });
+      it('teams.isSupported should return true if the runtime says teams is supported', () => {
+        utils.setRuntimeConfig({ apiVersion: 1, supports: { teams: {} } });
+        expect(teams.isSupported()).toBeTruthy();
+      });
+    });
   });
 });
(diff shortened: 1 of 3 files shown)


packages/teams-js/test/private/teams.spec.ts
```diff
@@ -1,90 +1,154 @@
...
-        'The library has not yet been initialized',
-      );
+          teams.refreshSiteUrl('threadId', callback);
+
+          const getCloudStorageFoldersMessage = utils.findMessageByFunc('teams.refreshSiteUrl');
+          expect(getCloudStorageFoldersMessage).not.toBeNull();
+          utils.respondToMessage(getCloudStorageFoldersMessage, false, 'Some Message');
+          expect(callback).toHaveBeenCalled();
+        });
+      });
```
erinha, line 151: Thanks for cleaning up the tests and making it clearer what work is left to do (e.g., add frameless scenarios, etc.). Would it be possible to add tests for `isSupported`?  Since this PR is technically about ensuring `isSupported` is called, it seems appropriate to have these tests as part of this PR.

erinha, line 151: Thanks for cleaning up the tests and making it clearer what work is left to do (e.g., add frameless scenarios, etc.). Would it be possible to add tests for `isSupported`?  Since this PR is technically about ensuring `isSupported` is called, it seems appropriate to have these tests as part of this PR.

erinha (changes requested):
Looks good overall, request to add test cases for `isSupported` while we're here.

packages/teams-js/test/private/teams.spec.ts
shrshindeMSFT: Sorry, I missed those. I have added them in the latest commit.

erinha (approved):
🚀  Looks great, thanks!

jadahiya-MSFT (approved):

PR #1134, opened 2022-04-26
shrshinde/isSupported-people-check

shrshindeMSFT:
- Adding isSupported check to people capability
- Adding isSupported unit test cases to people capability
- Adding isSupported function unit test cases to people capability

### DIFF
--- a/packages/teams-js/test/public/people.spec.ts
+++ b/packages/teams-js/test/public/people.spec.ts
@@ -1,28 +1,32 @@
 import { DOMMessageEvent } from '../../src/internal/interfaces';
 import { app } from '../../src/public/app';
-import { FrameContexts } from '../../src/public/constants';
+import { errorNotSupportedOnPlatform, FrameContexts } from '../../src/public/constants';
 import { ErrorCode, SdkError } from '../../src/public/interfaces';
 import { people } from '../../src/public/people';
+import { _minRuntimeConfigToUninitialize, v1HostClientTypes } from '../../src/public/runtime';
 import { FramelessPostMocks } from '../framelessPostMocks';
+import { Utils } from '../utils';
 
 /**
  * Test cases for selectPeople API
  */
 describe('people', () => {
-  const mobilePlatformMock = new FramelessPostMocks();
+  const framelessPlatformMock = new FramelessPostMocks();
+  const framedMock = new Utils();
   const minVersionForSelectPeople = '2.0.0';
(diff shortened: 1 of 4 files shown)


jadahiya-MSFT (approved):

aeunms (approved):

packages/teams-js/test/public/people.spec.ts
```diff
@@ -45,87 +49,96 @@ describe('people', () => {
...
       if (allowedContexts.some(allowedContext => allowedContext === context)) {
-        it(`should allow selectPeople calls with null peoplePickerInputs. context: ${context}`, async () => {
-          await mobilePlatformMock.initializeWithContext(context);
-          mobilePlatformMock.setClientSupportedSDKVersion(minVersionForSelectPeople);
-          people.selectPeople(null);
-          const selectPeopleMessage = mobilePlatformMock.findMessageByFunc('people.selectPeople');
-          expect(selectPeopleMessage).not.toBeNull();
-          expect(selectPeopleMessage.args[0]).toEqual(null);
-        });
+        Object.values(v1HostClientTypes).forEach(hostClientType => {
```
lakhveerk, line 55: Oh, did we decide to run the tests against all the hostCleintTypes (instead of framed/framless)? 

v1HostClientType contains desktop which can be frameless or framed, but we are always initializing it with framelessPlatformMock.

shrshindeMSFT: For People capability we have support in specific host client type and during unit testing we don't get runtime configuration. So, the work around is to provide host client type while initialization.

erinha: I think a comment in the tests would be super helpful!

lakhveerk: Right now, for web, we are initializing with frameless and passing web as a hostClientType when we know web is framed. 
Does it make sense to initialize using utils for web? and similar changes for others?

lakhveerk: I don't want to create unnecessary work for you. may be @erinha can suggest/comment?

shrshindeMSFT: For this PR we can keep it to isSupported work and later we can create another task for it.

lakhveerk: That is a good idea :)

lakhveerk (approved):

PR #1141, opened 2022-04-29
Add profile.showProfile API

grhewitt:
Adds an API for opening the [profile card](https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501). Apps can call this API when they want to show a profile card to the user, the hub fulfils this request by using it's existing profile card integration and displaying the card over the top of the hosted app.


packages/teams-js/src/public/profile.ts
```diff
@@ -0,0 +1,127 @@
...
+      if (!validateShowProfileRequest(showProfileRequest)) {
+        throw { errorCode: ErrorCode.INVALID_ARGUMENTS };
+      }
+
+      // Passing a DOMRect as part of the request breaks in Electron WebView with
+      // "Uncaught Error: An object could not be cloned.". To work around
+      // this we create a duplicate object of the values from the DOMRect.
+      const newShowProfileRequest = {
+        ...showProfileRequest,
+        targetElementBoundingRect: {
```
TrevorJoelHarris, line 27: Passing objects that contain functions out of this sdk has proven problematic in the past. You're working around one of the problems here by essentially making a new version that doesn't contain any functions, but we've seen that cause confusion in the past because another engineer will try to use the object and go "uh oh there are no functions what happened?". Might be better to just take the properties you need out of DomRect and put it into a sort of property bag. Will avoid confusion in the future.

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/profileUtil.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  return true;
+}
+
+function validatePersonaIdentifiers(identifiers: profile.PersonaIdentifiers): boolean {
+  if (!identifiers || typeof identifiers !== 'object') {
+    return false;
+  }
+
+  // Validate at least one identifier was passed.
```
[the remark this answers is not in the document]
TrevorJoelHarris: You mentioned trying to do some typescript magic to enforce the "at least one" requirement. Is this what you tried? https://stackoverflow.com/questions/48230773/how-to-create-a-partial-like-that-requires-a-single-property-to-be-set/48244432#48244432

packages/teams-js/src/public/profile.ts
grhewitt: Fixed by explicitly declaring an internal type that the hub can take a dependency on, should make it. clear to anyone on the hub side what values we have available.

packages/teams-js/src/public/profile.ts
```diff
@@ -0,0 +1,127 @@
...
+   *  - AppRequest: The show profile request is happening programmatically, without direct user interaction.
+   */
+  export type TriggerType = 'MouseHover' | 'MouseClick' | 'KeyboardPress' | 'AppRequest';
+
+  /**
+   * The set of identifiers that are supported for resolving the persona.
+   */
+  export type PersonaIdentifiers = {
+    /**
+     * The Teams messaging resource identifier.
```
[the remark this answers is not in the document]
grhewitt: Had a discussion with my team this week, it should be possible to resolve TeamsMRI on the Teams side, so I've removed it from this API

packages/teams-js/src/internal/profileUtil.ts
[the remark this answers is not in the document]
grhewitt: I added error messages clarifying why validation failed.

The one I tried was this one (
https://stackoverflow.com/questions/40510611/typescript-interface-require-one-of-two-properties-to-exist), though I guess the one you linked might have the same issue (hard to understand typescript errors when the user doesn't provide one).

packages/teams-js/src/internal/profileUtil.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  // Validate triggerType
+  if (!showProfileRequest.triggerType || typeof showProfileRequest.triggerType !== 'string') {
+    return [false, 'triggerType must be a valid string'];
+  }
+
+  return validatePersona(showProfileRequest.persona);
+}
+
+function validatePersona(persona: profile.Persona): [boolean, string | undefined] {
```
TrevorJoelHarris, line 39: Can we get a comment on this function as well?

change/@microsoft-teams-js-adac1a3d-5454-4c76-b1ef-d65e3f0c48c5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add showProfile API",
```
erinha, line 3: Please use past tense for change file comments and backticks for APIs. (I have a PR out for updating the CONTRIBUTING file including this info in it. :)).

In this case, I think the preferred message would be "Added \`profile.showProfile\` API and unit tests"

erinha (changes requested):
Comment on change file. Otherwise looks good!

packages/teams-js/src/internal/profileUtil.ts
grhewitt: Done!

erinha (approved):
🚀

PR #1142, opened 2022-04-29
Add to version checked APIs guidance

aeunms:
Adds detail about adding an isSupported() function for version checked APIs addition guidance. Also fixes broken link in CONTRIBUTING.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ e.g.
   ],
 ```
 
-If you're adding a capability onto an already existing version requirement, simply add your object into the existing array.
+If you're adding a capability to an already existing version requirement, simply add your object to the existing array.
 
 e.g.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,8 +42,8 @@ the PR.
 
 This option should only be used for work that meets ALL of the below requirements:
 
-- Work which has already been discussed with the TeamsJS devs as using this approach
-- Functionality implementations that may run in host clients that have not onboarded to the new declarative capabiity support architecture
+- Features which have already been discussed with the TeamsJS owners and for which approval to use this approach has been granted,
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -64,7 +73,7 @@ Exa)
 
 If you're adding a capability onto an already existing version requirement, simply add your object into the existing array.
```
erinha, line 74: nit: "If you're adding a capability **to** an already existing version requirement, simple add your object **to** the existing array."

(Bold to emphasize which words should be edited)

erinha (changes requested):
Thanks for making these fixes, it looks much better! Small comment since you're here fixing this file already. :)

CONTRIBUTING.md
aeunms: Sounds good, changes pushed!

erinha (dismissed):
🚀

CONTRIBUTING.md
```diff
@@ -40,12 +40,26 @@ the PR.
 
 ### Adding an API that utilizes version checks for compatibility
 
+This option should only be used for work that meets ALL of the below requirements:
+
+- Work which has already been discussed with the TeamsJS devs as using this approach
```
erinha, line 45: This is a great addition! Suggestion for some rewording that I think adds clarity:

- Features which have already been discussed with the TeamsJS owners and for which approval to use this approach has been granted,
- Feature implementation that has a requirement of running in host clients that have not onboarded to the new declarative capability support architecture.

Note: typo in existing text ("capabiity")

erinha (changes requested):
Thanks for adding the text we discussed offline. A small suggestion for rewording, but great overall!

CONTRIBUTING.md
aeunms: Great suggestions. Will make!

erinha (approved):

shrshindeMSFT (approved):

PR #1152, opened 2022-05-04
Added 3P cloud storage provider API support

HarshJain27:
As part of File app 2.0 work, exposing privately the 3P Cloud Storage related APIs. 
Following 7 actions are exposed through APIs:
1. Add 3P Cloud Storage provider
2. Sign out from logged in cloud storage provider
3. Add New File
4. Rename File
5. Upload File(s)
6. Download File(s)
7. Delete Files(s)

More details about individual APIs can be found in the design document shared below.

Design Doc Link : [Design Document: Exposing 3P Cloud Storage APIs via Teams SDK](https://microsoft.sharepoint.com/:w:/t/whatsupdoc/EbIW7S188cZOnznWfe1jZh0BWgamIyOztiXUC2VMDTrPyQ?e=DpxRym)

Relevant Task : [WorkItem](https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2106079)

It can be noted that while the APIs are being added as part of Files2.0 work, they are quite generic in terms of their usage and can have use cases in other places like Channels File Tab in Teams etc.


packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +296,70 @@ export namespace files {
     extension: string;
   }
 
+  export enum CloudStorageProviderFileAction {
```
TrevorJoelHarris, line 299: exported enums should have comments and be marked as @hidden since I believe they will show up in the documentation

packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +296,70 @@ export namespace files {
...
+    Upload = 'UPLOAD',
+    Delete = 'DELETE',
+    Logout = 'LOGOUT',
+  }
+
+  export interface I3PCloudStorageProviderRequest<T> {
+    content: T;
+  }
+
+  /**
```
TrevorJoelHarris, line 312: I don't think these comments get turned into documentation since they aren't exported, so @hidden shouldn't be necessary (and you have comments on the exported types using them below)

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+   *
+   * Initiates add 3P cloud storage provider flow
+   *
+   * @param callback Callback that will be triggered post add 3P cloud storage provider action
+   */
+  export function add3PCloudStorageProvider(callback: (error?: SdkError) => void): void {
+    ensureInitialized(FrameContexts.content);
+
+    if (!callback) {
+      throw new Error('[files.add3PCloudStorageProvider] callback cannot be null');
```
TrevorJoelHarris, line 609: I realize this is a pattern elsewhere in the file, but it would be really nice if functions used SdkError for everything instead of Error *and* SdkError

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+    }
+
+    sendMessageToParent('files.add3PCloudStorageProvider', [], callback);
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Initiates signout of 3P cloud storage provider flow
```
TrevorJoelHarris, line 619: Just assume I have left this comment on all exported functions: more explicit documentation about what specifically this function does would be hugely helpful both to callers of the function and to hosts other than Teams who may in the future want to support this capability as well

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+  export function remove3PCloudStorageProvider(
+    logoutRequest: I3PCloudStorageProviderRequest<I3PCloudStorageProviderLogoutRequestContentType>,
+    callback: (error?: SdkError) => void,
+  ): void {
+    ensureInitialized(FrameContexts.content);
+
+    if (!validateNonEmptyContent(logoutRequest, callback)) {
+      return;
+    }
+    if (!validateArgumentConsistency(logoutRequest, callback)) {
```
TrevorJoelHarris, line 632: Shouldn't we be throwing an SdkError rather than silently returning if no callback is specified?

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+   *
+   * @param callback Callback that will be triggered post signout of 3P cloud storage provider action
+   */
+  export function remove3PCloudStorageProvider(
+    logoutRequest: I3PCloudStorageProviderRequest<I3PCloudStorageProviderLogoutRequestContentType>,
+    callback: (error?: SdkError) => void,
+  ): void {
+    ensureInitialized(FrameContexts.content);
+
+    if (!validateNonEmptyContent(logoutRequest, callback)) {
```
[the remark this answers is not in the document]
TrevorJoelHarris: Either that or shouldn't we be throwing if there is no callback ahead of these two checks? Otherwise if you don't speciy a callback I think this will just silently do nothing

packages/teams-js/src/private/files.ts
TrevorJoelHarris: Either that or shouldn't we be throwing if there is no callback ahead of these two checks? Otherwise if you don't speciy a callback I think this will just silently do nothing

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+      return;
+    }
+    if (!validateArgumentConsistency(cloudStorageProviderFileActionRequest, callback)) {
+      return;
+    }
+    if (!validateContentForSupportedActions(cloudStorageProviderFileActionRequest, callback)) {
+      return;
+    }
+
+    if (!callback) {
```
TrevorJoelHarris, line 669: Same question about whether this check should go earlier or the validate calls should throw so that this doesn't silently return nothing when on no callback is specified and there's an error

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+    }
+
+    sendMessageToParent('files.remove3PCloudStorageProvider', [logoutRequest], callback);
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Initiates the 3P cloud storage provider file action (New / Rename / Upload / Download / Delete)
```
TrevorJoelHarris, line 646: More detail in this comment about what these actions actually do or should be used for would be very helpful

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+  function validateArgumentConsistency(
+    cloudStorageProviderFileActionRequest: I3PCloudStorageProviderRequest<
+      I3PCloudStorageProviderActionRequestContentType
+    >,
+    callback?: (err?: SdkError) => void,
+  ): boolean {
+    if (!cloudStorageProviderFileActionRequest.content.action) {
+      if (callback) {
+        callback({
+          errorCode: ErrorCode.INVALID_ARGUMENTS,
```
TrevorJoelHarris, line 707: This SdkError and the one on line 717 are the same, maybe construct it outside your if statements and use the same one in both places

packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +296,70 @@ export namespace files {
...
+    action: CloudStorageProviderFileAction.Logout;
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Interface for 3P cloud storage provider action specific request content
+   */
+  interface I3PCloudStorageProviderActionContent extends IBase3PCloudStorageProviderContent {
```
TrevorJoelHarris, line 339: It would be great to use the type system to make multiple types for each of the actions you want to support. Looking at the code below, it seems like depending on the action type (NEW, RENAME, etc.) you are expecting different combinations of these properties to be set, and returning errors otherwise. This will make it tough for callers to understand without going through trial and error. Making types like "I3PCloudStorageProviderActionContentForNew" (or whatever you want to name it) that requires an existingFileName and then either omits or makes optional the other parameters would make it a lot easier to use this for future callers.

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
...
+        });
+      }
+      return false;
+    }
+
+    return true;
+  }
+
+  function validateContentForSupportedActions(
+    cloudStorageProviderFileActionRequest: I3PCloudStorageProviderRequest<
```
TrevorJoelHarris, line 728: I mentioned this above, but being more restrictive about parameters developers are allowed to pass in would make it so you don't have to do complicated verification of parameters here.

packages/teams-js/src/private/interfaces.ts
```diff
@@ -63,6 +63,25 @@ export enum UserSettingTypes {
...
+/**
+ * @hidden
+ * Hide from docs.
+ * ------
+ * 3P cloud storage scenario changes that can be subscribed to
+ */
+export enum CloudStorageProvider3PScenario {
+  /**
+   * @hidden
+   * Use this key to subscribe to changes in user's 3P cloud storage provider list
```
TrevorJoelHarris, line 75: I assume this is for subscribing to when new cloud start providers are added or removed. For the second value (cloudStorageProviderContent) what does that specifically cover?

packages/teams-js/src/private/privateAPIs.ts
```diff
@@ -119,6 +119,24 @@ export function registerUserSettingsChangeHandler(
   registerHandler('userSettingsChange', handler, true, [settingTypes]);
 }
 
+/**
+ * @hidden
+ * register a handler to be called when a user's 3P cloud storage provider details changes. The changed setting type & value is provided in the callback.
+ *
+ * @param settingTypes - List of user setting changes to subscribe
```
TrevorJoelHarris, line 126: should be "scenarioTypes", and should probably @link to the CloudStorageProvider documentation

packages/teams-js/src/private/files.ts
HarshJain27: Removed the hidden marker, have retained the comments for clarity.

packages/teams-js/src/private/files.ts
HarshJain27: Splitted the common action method to multiple action methods, to be specific create action method for New and Rename.
The updated common method still represents 3 actions for file(s):
1. Upload 2. Download 3. Delete 
The reason of these 3 actions being represented by a common type is that they have exact same input parameters.

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +593,206 @@ export namespace files {
 
     sendMessageToParent('files.openDownloadFolder', [fileObjectId], callback);
   }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Initiates add 3P cloud storage provider flow
```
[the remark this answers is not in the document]
HarshJain27: Added more details for all 3P scenarios for better understanding of individual scenario.

packages/teams-js/src/private/files.ts
HarshJain27: Details added.

packages/teams-js/src/private/files.ts
HarshJain27: Following your suggestion changed the Errors to SdkErrors. Hope this deviation in consistency isn't a concern for anybody.

packages/teams-js/src/private/files.ts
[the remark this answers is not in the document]
HarshJain27: This piece of code has been removed since it no longer is required, hence this comment is no longer applicable.

packages/teams-js/src/private/files.ts
HarshJain27: This piece of code has been removed since it no longer is required, hence the comment is no longer applicable.

packages/teams-js/src/private/files.ts
HarshJain27: This piece of code has been removed since it no longer is required, hence the comment is no longer applicable.

packages/teams-js/src/private/files.ts
HarshJain27: After splitting action method for different scenarios, this point is addressed. Removed unnecessary verification code.

packages/teams-js/src/private/interfaces.ts
HarshJain27: Updated the code comments.
Sharing the same here as well : 
`Use this key to subscribe to cloud storage provider content changes i.e when file(s) is/are added / renamed / deleted / uploaded, the list of content is updated`

packages/teams-js/src/private/privateAPIs.ts
HarshJain27: That's correct. Updated the same.

packages/teams-js/src/private/privateAPIs.ts
```diff
@@ -119,6 +119,24 @@ export function registerUserSettingsChangeHandler(
...
+ *
+ * @internal
+ */
+export function register3PCloudStorageProviderChangeHandler(
+  scenarioTypes: CloudStorageProvider3PScenario[],
+  handler: (scenarioType: CloudStorageProvider3PScenario, value: any) => void,
+): void {
+  ensureInitialized();
+
+  registerHandler('cloudStorageProvider3PChange', handler, true, [scenarioTypes]);
```
ydogandjiev, line 137: Why do we need to introduce more complexity here with the scenarioTypes? I only see two scenario types and its unlikely that more will be added in the future. In either case, it would be better to have a separate handler registration method for each one, e.g.:
```
registerCloudStorageProvidersChanged
registerCloudStorageProviderContentChanged
```

packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +296,105 @@ export namespace files {
...
+   * Interface for 3P cloud storage provider request content type
+   */
+  export interface I3PCloudStorageProviderRequest<T> {
+    content: T;
+  }
+
+  /**
+   * Base interface for 3P cloud storage provider action request content
+   */
+  interface IBase3PCloudStorageProviderContent {
```
ydogandjiev, line 329: We generally don't prefix interfaces with "I" in the SDK and that's not really common in JavaScript. I would recommend omitting it. @erinha, @TrevorJoelHarris, thoughts? Do we have any formal guidance on this?

ydogandjiev (changes requested):

packages/teams-js/src/private/privateAPIs.ts
```diff
@@ -119,6 +119,24 @@ export function registerUserSettingsChangeHandler(
...
+/**
+ * @hidden
+ * register a handler to be called when a user's 3P cloud storage provider details changes. The changed setting type & value is provided in the callback.
+ *
+ * @param scenarioTypes - List of 3P cloud storage provider scenario changes to subscribe
+ * @param handler - When a subscribed setting is updated this handler is called
+ *
+ * @internal
+ */
+export function register3PCloudStorageProviderChangeHandler(
```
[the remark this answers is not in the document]
HarshJain27: The reason of keeping change handler here was to have common logic of change handlers together. After your comment realised that namespaces keep the respective change handers in their own files and not in a common place.
**Moved this to files.ts.**

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +628,166 @@ export namespace files {
...
+   * @hidden
+   * Hide from docs
+   *
+   * Initiates add 3P cloud storage provider flow, where a pop up window opens for user to select required
+   * 3P provider from the configured policy supported 3P provider list, following which user authentication
+   * for selected 3P provider is performed on success of which the selected 3P provider support is added for user
+   *
+   * @param callback Callback that will be triggered post add 3P cloud storage provider action
+   */
+  export function add3PCloudStorageProvider(callback: (error?: SdkError) => void): void {
```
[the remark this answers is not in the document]
HarshJain27: The reason of using 3P in all names was to make them verbose and explicit for others.
Keeping in mind consistency with the existing 3P methods, as suggested, removed 3P from all API names.

packages/teams-js/src/private/privateAPIs.ts
HarshJain27: Was trying to avoiding creating two separate APIs for similar use case.
As per your suggestion, splitted the single method in two.

packages/teams-js/src/private/files.ts
HarshJain27: Initially was referring to private **files namespace** and noticed the convention of prefixing interfaces with letter "I".
After your comment referred other namespaces and in most places the convention followed is to avoid prefix "I".
Removed the prefix "I" from interfaces i created.

erinha: I'm fine with no I*  prefix.

packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +297,105 @@ export namespace files {
...
   }
 
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Actions specific to 3P cloud storage provider file and / or account
+   */
+  export enum CloudStorageProviderFileAction {
+    New = 'NEW',
```
TrevorJoelHarris, line 307: I don't think you actually need any of these other than Download/Upload/Delete. Except for Logout, the other ones are extra parameters inside of interfaces that could still be typesafe without the extra hint. For Logout, I think if you removed action from BaseCloudStorageProviderContent and renamed that to just "CloudStorageProviderContent" you would also be able to just pass that type to the logout function.

If you do this, then your CloudStorageProviderFileAction enum will only have enum types in it that callers actually need to provide (Download/Upload/Delete for the function on 755) and you will still have a bunch of typesafe functions

packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +297,105 @@ export namespace files {
...
+    action: CloudStorageProviderFileAction.Logout;
+  }
+
+  /**
+   * Interface representing 3P cloud storage provider add new file action
+   */
+  interface CloudStorageProviderNewFileContent extends BaseCloudStorageProviderContent {
+    action: CloudStorageProviderFileAction.New;
+    newFileName: string;
+    newFileType: string;
```
TrevorJoelHarris, line 348: Is this file extension? If so might make more sense to name it something like "newFileExtension". Are they supposed to prepend it with a '.'? So, '.doc' vs. 'doc'? Might be worth documenting that

TrevorJoelHarris (changes requested):

ydogandjiev (changes requested):

packages/teams-js/src/private/files.ts
HarshJain27: Suggestion incorporated. Modified enum to only have types, which caller needs to provide actually. Consequently modified the required types and function argyments.

packages/teams-js/src/private/files.ts
HarshJain27: Yes it is file extension. Renamed it and added the relevant documentation.

packages/teams-js/src/private/files.ts
```diff
@@ -296,6 +297,105 @@ export namespace files {
...
+    itemList: CloudStorageFolderItem[] | ISharePointFile[];
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Interface representing 3P cloud storage provider logout action request content type
+   */
+  export type CloudStorageProviderLogoutRequestContentType = CloudStorageProviderLogoutContent;
```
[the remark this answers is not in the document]
HarshJain27: Suggestion incorporated. Removed the explicit types and using the exported interfaces instead.

ydogandjiev (dismissed):

packages/teams-js/src/private/files.ts
```diff
@@ -529,4 +596,201 @@ export namespace files {
...
+      callback,
+    );
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Register a handler to be called when a user's 3P cloud storage provider list changes i.e.
+   * post adding / reoving a 3P provider, list is updated
```
TrevorJoelHarris, line 754: "removing" instead of "reoving"

HarshJain27: Corrected!

TrevorJoelHarris (approved):

ydogandjiev (dismissed):

change/@microsoft-teams-js-87d62e41-c534-43da-9afb-5dda9cff9030.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added 3P cloud storage provider API support",
+  "packageName": "@microsoft/teams-js",
+  "email": "jainharsh@microsof.com",
```
TrevorJoelHarris, line 5: "microsoft" instead of "microsof"

change/@microsoft-teams-js-87d62e41-c534-43da-9afb-5dda9cff9030.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added 3P cloud storage provider API support",
```
TrevorJoelHarris, line 3: please append " to files.ts" to the end of your comment

TrevorJoelHarris (changes requested):

change/@microsoft-teams-js-87d62e41-c534-43da-9afb-5dda9cff9030.json
HarshJain27: Corrected typo.

change/@microsoft-teams-js-87d62e41-c534-43da-9afb-5dda9cff9030.json
HarshJain27: Done!

TrevorJoelHarris (approved):

PR #1154, opened 2022-05-04
Update all deprecated tags to reference version 2.0.0

erinha:
We received guidance that in preparation for release v2.0, all @deprecated tags that were introduced in 2.0 beta versions should now point to v2.0. This should be clearer for developers who upgrade from v1 to v2.


AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

aeunms (approved):
🚢

PR #1155, opened 2022-05-04
Update GitHub actions for archive/v1

erinha:
After `master` was renamed to `archive/v1`, the status checks that are triggered on the branch need to be updated to trigger on the new branch name.


AE-MS (approved):
:shipit:

PR #1156, opened 2022-05-04
Added a link to info about the updated Context

erinha:
Since the global `Context` interface is deprecated, this change provides a link to more information about the differences between the deprecated `Context` and `app.Context` interfaces.

### DIFF
--- a/packages/teams-js/src/public/interfaces.ts
+++ b/packages/teams-js/src/public/interfaces.ts
@@ -172,7 +172,12 @@ export enum FileOpenPreference {
 /**
  * @deprecated
  * As of 2.0.0, please use {@link app.Context} instead.
- * Represents structure of the received context message.
+ *
+ * @remarks
+ * For more details on the updated {@link app.Context} interface, visit
+ * {@link https://docs.microsoft.com/microsoftteams/platform/tabs/how-to/using-teams-client-sdk}.
+ *
+ * Represents the structure of the received context message.
  */
 export interface Context {
   /**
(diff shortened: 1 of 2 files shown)


AE-MS (dismissed):

AE-MS (approved):

packages/teams-js/src/public/interfaces.ts
```diff
@@ -172,7 +172,12 @@ export enum FileOpenPreference {
 /**
  * @deprecated
  * As of 2.0.0, please use {@link app.Context} instead.
- * Represents structure of the received context message.
+ *
+ * @remarks
+ * For more details on the updated {@link app.Context} interface, visit
+ * {@link https://docs.microsoft.com/microsoftteams/platform/tabs/how-to/using-teams-client-sdk}.
+ *
```
erikadoyle, line 179: Could we add this anchor to the link: **#updates-to-the-context-interface**:

https://docs.microsoft.com/microsoftteams/platform/tabs/how-to/using-teams-client-sdk#updates-to-the-context-interface

. . . it will not resolve yet, but will work upon docs publish. (And browsers will ignore the broken anchor until then anyway).

erikadoyle (commented):
Thanks for adding this @erinha!

packages/teams-js/src/public/interfaces.ts
erinha: Great suggestion, thanks for the updated link. Added!

AE-MS (approved):
:shipit:

shrshindeMSFT (approved):

aeunms (approved):
🚢

PR #1160, opened 2022-05-05
Adding frameless unit test cases for auth capability

shrshindeMSFT:


aeunms (approved):

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +511,744 @@ describe('Testing authentication module', () => {
...
   });
 
-  describe('Testing authentication.notifyFailure', () => {
-    const allowedContexts = [FrameContexts.authentication];
-    it('authentication.notifyFailure should not allow calls before initialization', () => {
-      expect(() => authentication.notifyFailure()).toThrowError('The library has not yet been initialized');
+  describe('FRAMELESS - authentication tests', () => {
+    beforeEach(() => {
+      // Use to send a mock message from the app.
+
```
erinha, line 900: nit: can you please remove this newline?

erinha: Or maybe remove the comment? I don't think it makes sense anymore?

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +511,744 @@ describe('Testing authentication module', () => {
...
-          expect(message.args[0]).toBe(errorMessage);
+            expect(() => authentication.authenticate(authenticationParams)).toThrowError(
+              `This call is only allowed in following contexts: ${JSON.stringify(
+                allowedContexts,
+              )}. Current context: "${context}".`,
+            );
+          });
+        }
+      });
+    });
```
erinha, line 946: Shouldn't we have positive unit tests for `authentication.registerAuthenticationHandlers`? I only see tests for error cases.

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +511,744 @@ describe('Testing authentication module', () => {
...
+
+    describe('Testing authentication.getAuthToken function', () => {
```
erinha, line 948: Maybe this is a stupid question, but is there a reason there are no frameless unit tests for `authentication.authenticate`?

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +511,744 @@ describe('Testing authentication module', () => {
...
+              data: {
+                id: message.id,
+                args: [false, errorMessage],
+              },
+            } as DOMMessageEvent);
           });
+        });
 
-          await utils.initializeWithContext(context);
+        it(`authentication.getAuthToken should successfully return getAuthToken in case of success from ${context} context`, async () => {
```
erinha, line 1025: nit: this test case should probably be called "authentication.getAuthToken should successfully return auth token ...". (Same for the framed version of it.)

erinha: Same for other test cases that return an auth token

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +511,744 @@ describe('Testing authentication module', () => {
...
+            await framelessPostMock.initializeWithContext(context);
+            expect(() => authentication.notifySuccess()).toThrow(
+              `This call is only allowed in following contexts: ${JSON.stringify(
+                allowedContexts,
+              )}. Current context: "${context}".`,
+            );
+          });
+        }
+      });
+    });
```
erinha, line 1232: Again, probably a stupid question, but why no succcessful frameless unit tests for `authentication.notifySuccess`?

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +511,744 @@ describe('Testing authentication module', () => {
...
+          it(`authentication.notifyFailure should not allow calls from ${context} context`, async () => {
+            await framelessPostMock.initializeWithContext(context);
+            expect(() => authentication.notifyFailure()).toThrow(
+              `This call is only allowed in following contexts: ${JSON.stringify(
+                allowedContexts,
+              )}. Current context: "${context}".`,
+            );
+          });
+        }
+      });
```
erinha, line 1251: Same here - why no successful frameless unit tests for `authentication.notifyFailure`?

erinha (changes requested):
Thank you for adding these! Mostly questions about the types of test cases being added or not.

packages/teams-js/test/public/authentication.spec.ts
erinha: Added successful unit tests for frameless authentication.authenticate and authentication.registerAuthenticationHandlers.

packages/teams-js/test/public/authentication.spec.ts
erinha: Added

packages/teams-js/test/public/authentication.spec.ts
erinha: I think the reason is that authentication.notifySuccess can't be called from a frameless window. But I'm following up.

packages/teams-js/test/public/authentication.spec.ts
erinha: I think the reason is that authentication.notifyFailure can't be called from a frameless window. But I'm following up.

packages/teams-js/test/public/authentication.spec.ts
erinha: I've added a work item to track confirming and adding test cases and will do that in a separate change.

erinha (dismissed):

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +499,857 @@ describe('Testing authentication module', () => {
...
+        app._uninitialize();
+      }
     });
 
-    Object.values(FrameContexts).forEach(context => {
-      if (allowedContexts.some(allowedContexts => allowedContexts === context)) {
-        it(`authentication.notifyFailure should successfully notify auth failure ${context} context`, async () => {
-          await utils.initializeWithContext('authentication');
+    describe('Testing authentication.initialize function', () => {
+      it('authentication.initialize should successfully register authentication.authenticate.success/failure handler', async () => {
```
TrevorJoelHarris, line 901: I think this is literally identical to the framed version of the test since you don't use the mock at all

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +499,857 @@ describe('Testing authentication module', () => {
...
+        }
+      });
     });
   });
 
-  describe('Testing authentication.notifyFailure', () => {
-    const allowedContexts = [FrameContexts.authentication];
-    it('authentication.notifyFailure should not allow calls before initialization', () => {
-      expect(() => authentication.notifyFailure()).toThrowError('The library has not yet been initialized');
+  describe('FRAMELESS - authentication tests', () => {
```
TrevorJoelHarris, line 885: Rather than duplicate all of these tests, should we just be looping through each one twice and providing it the framed and frameless mocks (kind of like how we iterate through framecontexts in tests in other places)?

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -461,431 +499,857 @@ describe('Testing authentication module', () => {
...
+    });
+
+    describe('Testing authentication.notifySuccess function', () => {
+      const allowedContexts = [FrameContexts.authentication];
+
+      it('authentication.notifySuccess should not allow calls before initialization', () => {
+        expect(() => authentication.notifySuccess()).toThrowError('The library has not yet been initialized');
+      });
+
+      Object.values(FrameContexts).forEach(context => {
```
TrevorJoelHarris, line 1321: Do you need some positive tests? Aka notify success works when using this set of frame contexts?

erinha: Yes. I opened a bug to track adding these after GA. I couldn't get the framelessPostMock working as per our earlier conversation. :)

erinha: Yes. I opened a bug to track adding these after GA. I couldn't get the framelessPostMock working as per our earlier conversation. :)

packages/teams-js/test/public/authentication.spec.ts
erinha: Nice catch!

erinha (approved):

packages/teams-js/test/public/authentication.spec.ts
erinha: I have opened a bug to track investigating this approach

TrevorJoelHarris (approved):

PR #1164, opened 2022-05-06
Proposal for sdk update to add search capability

TrevorJoelHarris:
This is a proposal for how apps could integrate with the host search experience.

Applications can register for 3 different events:
1) onChange: this will be called when search begins and every time the contents of the searchQuery change
2) onClosed: this will be called when search ends in the host
3) onExecute: <OPTIONAL> this will be called when search is executed by the user (pressing enter for example)
If an application registers for these events their application name will show up in the host as the default search scope *while their application is running*.

Applications can also unregister for search events, which will cause their application to stop showing up as a valid search scope in the host.

### DIFF
--- a/packages/teams-js/src/public/runtime.ts
+++ b/packages/teams-js/src/public/runtime.ts
@@ -33,6 +33,7 @@ export interface IRuntime {
     };
     readonly people?: {};
     readonly remoteCamera?: {};
+    readonly search?: {};
     readonly sharing?: {};
     readonly teams?: {
       readonly fullTrust?: {
@@ -72,6 +73,7 @@ export let runtime: IRuntime = {
     },
     people: undefined,
     remoteCamera: undefined,
+    search: undefined,
     sharing: undefined,
     teams: {
       fullTrust: {
@@ -109,6 +111,7 @@ export const teamsRuntimeConfig: IRuntime = {
       fullTrust: {},
     },
     remoteCamera: {},
+    search: undefined,
     sharing: {},
     teams: {
       fullTrust: {},
--- a/packages/teams-js/src/public/runtime.ts
+++ b/packages/teams-js/src/public/runtime.ts
@@ -111,7 +111,6 @@ export const teamsRuntimeConfig: IRuntime = {
       fullTrust: {},
     },
     remoteCamera: {},
-    search: undefined,
     sharing: {},
     teams: {
       fullTrust: {},
--- a/packages/teams-js/src/public/search.ts
+++ b/packages/teams-js/src/public/search.ts
@@ -1,5 +1,7 @@
 import { sendMessageToParent } from '../internal/communication';
 import { registerHandler, removeHandler } from '../internal/handlers';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
 import { runtime } from './runtime';
 
 /**
@@ -8,6 +10,10 @@ import { runtime } from './runtime';
  * @beta
  */
 export namespace search {
+  const onChangeHandlerName = 'search.queryChange';
+  const onClosedHandlerName = 'search.queryClosed';
+  const onExecutedHandlerName = 'search.queryExecute';
+
   /**
    * This interface contains information pertaining to the search term in the host search experience
    */
@@ -15,7 +21,7 @@ export namespace search {
     /** The current search term in the host search experience */
     searchTerm: string;
 
-    // may need some sort of timestamp or sequence value to
+    // TODO: may need some sort of timestamp or sequence value to
     // ensure messages are processed in correct order / combine them
     // having any sort of logic around combining messages or sorting them
     // would make sense to go into the teamsjs-sdk layer
@@ -28,6 +34,9 @@ export namespace search {
    * Allows the caller to register for various events fired by the host search experience.
    * Calling this function will cause the host search experience to set its default scope to
    * the name of your application.
+   * 
+   * Your application should *not* re-render inside of these callbacks, there may be a large number
+   * of onChangeHandler calls if the user is changing the searchQuery rapidly.
    *
    * @param onChangeHandler - This handler will be called when the user begins searching and every
    * time the user changes the contents of the query. The value of the query is the current term
@@ -46,26 +55,41 @@ export namespace search {
    * ``` ts
    * search.registerHandlers(
       query => {
-        alert(`Update your application to render a change to the search query: ${query.searchTerm}`);
+        console.log(`Update your application to render a change to the search query: ${query.searchTerm}`);
       },
       () => {
-        alert('Update your application to handle the search experience being closed');
+        console.log('Update your application to handle the search experience being closed');
       },
       query => {
-        alert(`Update your application to render an executed search result: ${query.searchTerm}`);
+        console.log(`Update your application to render an executed search result: ${query.searchTerm}`);
       },
      );
    * ```
    */
   export function registerHandlers(
     onChangeHandler: SearchQueryHandler,
-    onClosedHandler: () => SearchQueryHandler,
+    onClosedHandler: SearchQueryHandler,
     onExecuteHandler?: SearchQueryHandler,
   ): void {
-    registerHandler('searchQueryChange', onChangeHandler);
-    registerHandler('searchQueryClosed', onClosedHandler);
+    // TODO: figure out what frame contexts you want to support this in
+    // This is just a guess that I made and should be something you make
+    // an explicit decision about.
+    ensureInitialized(
+      FrameContexts.content,
+      FrameContexts.task,
+      FrameContexts.sidePanel,
+      FrameContexts.stage,
+      FrameContexts.meetingStage,
+    );
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    registerHandler(onChangeHandlerName, onChangeHandler);
+    registerHandler(onClosedHandlerName, onClosedHandler);
     if (onExecuteHandler) {
-      registerHandler('searchQueryExecute', onExecuteHandler);
+      registerHandler(onExecutedHandlerName, onExecuteHandler);
     }
   }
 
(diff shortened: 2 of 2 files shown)


packages/teams-js/src/public/runtime.ts
```diff
@@ -109,6 +111,7 @@ export const teamsRuntimeConfig: IRuntime = {
       fullTrust: {},
     },
     remoteCamera: {},
+    search: undefined,
```
TrevorJoelHarris, line 114: Should remove this to be consistent with mail and calendar

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,87 @@
...
+   * @param onExecuteHandler - This optional handler will be called whenever the user 'executes' the
+   * search (by pressing enter for example). The value of {@link SearchQuery.searchTerm} is the current 
+   * term the user is searching for. Should be used if your app wants to treat executing searches differently than responding
+   * to changes to the search query.
+   *
+   * @example
+   * ``` ts
+   * search.registerHandlers(
+      query => {
+        alert(`Update your application to render a change to the search query: ${query.searchTerm}`);
```
TrevorJoelHarris, line 49: console.log

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,87 @@
...
+      },
+      query => {
+        alert(`Update your application to render an executed search result: ${query.searchTerm}`);
+      },
+     );
+   * ```
+   */
+  export function registerHandlers(
+    onChangeHandler: SearchQueryHandler,
+    onClosedHandler: () => SearchQueryHandler,
```
TrevorJoelHarris, line 62: remove () =>

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,87 @@
...
+    // ensure messages are processed in correct order / combine them
+    // having any sort of logic around combining messages or sorting them
+    // would make sense to go into the teamsjs-sdk layer
+    // timestamp: number;
+  }
+
+  export type SearchQueryHandler = (query: SearchQuery) => void;
+
+  /**
+   * Allows the caller to register for various events fired by the host search experience.
```
TrevorJoelHarris, line 28: Add a comment about not rendering app in the callback for performance reasons

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,87 @@
...
+      },
+     );
+   * ```
+   */
+  export function registerHandlers(
+    onChangeHandler: SearchQueryHandler,
+    onClosedHandler: () => SearchQueryHandler,
+    onExecuteHandler?: SearchQueryHandler,
+  ): void {
+    registerHandler('searchQueryChange', onChangeHandler);
```
TrevorJoelHarris, line 65: search.

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,87 @@
...
+      () => {
+        alert('Update your application to handle the search experience being closed');
+      },
+      query => {
+        alert(`Update your application to render an executed search result: ${query.searchTerm}`);
+      },
+     );
+   * ```
+   */
+  export function registerHandlers(
```
TrevorJoelHarris, line 60: check isSupported.
ensureInitialized

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,87 @@
...
+   */
+  export function unregisterHandlers(): void {
+    // This should let the host know to stop making the app scope show up in the search experience
+    sendMessageToParent('search.unregister');
+    removeHandler('searchQueryChange');
+    removeHandler('searchQueryClosed');
+    removeHandler('searchQueryExecute');
+  }
+
+  export function isSupported(): boolean {
```
TrevorJoelHarris, line 84: documentation

packages/teams-js/src/public/search.ts
TrevorJoelHarris: make sure to camel case appropriately after adding the .

TrevorJoelHarris:
Unit tests would need to be written for this

erinha:
A change file would also need to be created for this via `yarn changefile`.

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+    // having any sort of logic around combining messages or sorting them
+    // would make sense to go into the teamsjs-sdk layer
+    // timestamp: number;
+  }
+
+  export type SearchQueryHandler = (query: SearchQuery) => void;
+
+  /**
+   * Allows the caller to register for various events fired by the host search experience.
+   * Calling this function will cause the host search experience to set its default scope to
```
MarcinJuraszek, line 35: Would it also 'steal focus' and set the cursor in the search box?

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+   * the name of your application.
+   * 
+   * Your application should *not* re-render inside of these callbacks, there may be a large number
+   * of onChangeHandler calls if the user is changing the searchQuery rapidly.
+   *
+   * @param onChangeHandler - This handler will be called when the user begins searching and every
+   * time the user changes the contents of the query. The value of the query is the current term
+   * the user is searching for. Should be used to put your application into whatever state is used
+   * to handle searching. This handler will be called once with an empty {@link SearchQuery.searchTerm}
+   * when search is beginning.
```
MarcinJuraszek, line 45: This "will be called once with an empty searchTerm" part is interesting. Does that mean it'll only ever be called with an empty search term once? Or that it'll only be called with an empty search term at the beginning of user interacting with the searchbox? What if I start typing, and then backspace back to an empty term? Would we call this again with an empty search box? That seems like more than "once" to me?

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+   * 
+   * Your application should *not* re-render inside of these callbacks, there may be a large number
+   * of onChangeHandler calls if the user is changing the searchQuery rapidly.
+   *
+   * @param onChangeHandler - This handler will be called when the user begins searching and every
+   * time the user changes the contents of the query. The value of the query is the current term
+   * the user is searching for. Should be used to put your application into whatever state is used
+   * to handle searching. This handler will be called once with an empty {@link SearchQuery.searchTerm}
+   * when search is beginning. 
+   * @param onClosedHandler - This handler will be called when the user finishes searching. Should be
```
MarcinJuraszek, line 46: What does this mean to "finish searching"? Focus away from the searchbox? Some explicit "X" UI on the host part to 'finish' a search? I think this could use some more detail.

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+   * @param onExecuteHandler - This optional handler will be called whenever the user 'executes' the
+   * search (by pressing enter for example). The value of {@link SearchQuery.searchTerm} is the current 
+   * term the user is searching for. Should be used if your app wants to treat executing searches differently than responding
+   * to changes to the search query.
+   *
+   * @example
+   * ``` ts
+   * search.registerHandlers(
+      query => {
+        console.log(`Update your application to render a change to the search query: ${query.searchTerm}`);
```
MarcinJuraszek, line 58: putting "render" here when we just said "you shouldn't re-render in any of these callbacks" seems a bit counterintuitive?

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+    registerHandler(onChangeHandlerName, onChangeHandler);
+    registerHandler(onClosedHandlerName, onClosedHandler);
+    if (onExecuteHandler) {
+      registerHandler(onExecutedHandlerName, onExecuteHandler);
+    }
+  }
+
+  /**
+   * Allows the caller to unregister for all events fired by the host search experience. Calling
+   * this function will cause your app to stop appearing in the set of search scopes in the host.s
```
MarcinJuraszek, line 98: nit: typo at the very end, should be `hosts.` instead of `host.s`

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+      FrameContexts.task,
+      FrameContexts.sidePanel,
+      FrameContexts.stage,
+      FrameContexts.meetingStage,
+    );
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+    // This should let the host know to stop making the app scope show up in the search experience
```
MarcinJuraszek, line 115: Does that mean the user will have no way to 'remove' that default 'app scope' from search UI once the app calls `registerHandlers`? And if there is a way for them 'escape' that default scope, will there also be a way to get back to it somehow?

packages/teams-js/src/public/search.ts
TrevorJoelHarris: Good question, and one I assume Outlook will be investigating as they work to figure out how users want this to work. I don't think there's anything inherit in the sdk design here that would mean it *has* to work one way or the other, but we should be clear with callers exactly what this does

packages/teams-js/src/public/search.ts
TrevorJoelHarris: Yes sorry I could/should have been more clear. Every time the contents of the searchbox change this function will get called with the current contents. *In addition* it will also be called once when search begins (in which case search should be empty). If you type something and then delete it all this will be called again with the empty string.

packages/teams-js/src/public/search.ts
TrevorJoelHarris: Agreed, but I assume to some extent this will be host-specific as the UI around search may vary from host to host. In Teams for example, you can either press the 'X' button in the search bar or press Esc a few times. Definitely agree that part of removing the beta tag here would be being as crisp as possible about exactly what this means

packages/teams-js/src/public/search.ts
TrevorJoelHarris: According to the spec, the user can change the search scope and this would only affect whatever the "default" current scope is while in the application

packages/teams-js/src/public/search.ts
DianaLicon-3: why do we need to know what the last query was before ending search?

TrevorJoelHarris: Ultimately, I think how that works is something I'm open to you and your team changing. My thinking was that apps may want to be able to access what the last term was in the search box for a few different reasons (their own telemetry, putting the term back into the search box in the future, etc.)

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Can also be used to clean up handlers on the host if desired
+    sendMessageToParent('search.unregister');
```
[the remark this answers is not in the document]
TrevorJoelHarris: Yes, that's the idea. If your app is ever not registered for search events the host should return to default global search scope.

packages/teams-js/src/public/search.ts
```diff
@@ -0,0 +1,131 @@
...
+    // TODO: figure out what frame contexts you want to support this in
+    // This is just a guess that I made and should be something you make
+    // an explicit decision about.
+    ensureInitialized(
+      FrameContexts.content,
+      FrameContexts.task,
+      FrameContexts.sidePanel,
+      FrameContexts.stage,
+      FrameContexts.meetingStage,
+    );
```
ruiconti, line 83: I believe that, for now, we only want to support `FrameContexts.content`, as it's the only place it would make sense for us to have a shared search experience with the hub.

TrevorJoelHarris:
Closing because @DianaLicon-3 has checked in a real version of this change

PR #1166, opened 2022-05-09
Attempt to workaround the failing build step on CI builds

erinha:
Testing out a workaround to see if the bundle size analysis will work while our CI builds have a step that is failing due to an external issue.


AE-MS (approved):
:shipit:

PR #1170, opened 2022-05-11
Add displayCapture.getSources private API

jinyu1122:
This PR is to add a private API: displayCapture.getSources. 
This API will return the sources of Electron.desktopCapturer.getSources. The returned result can be used for screen/window sharing purpose.


TrevorJoelHarris:
Apologies that no one responded to this earlier. It's because adding a new capability needs to go through an internal design review. If you need help figuring out how to do that, please let me know. Closing this PR until you've gone through a design review.

PR #1172, opened 2022-05-11
Remove callback version of stageView.open

erinha:
`stageView.open` is a relatively new API that is still under development. On the path to v2, we have converted APIs that take callbacks to instead return Promises. As such, the `open` function that takes a callback has already been deprecated, despite being under development still.

Would be a clearer message to developers if we just had the `open` function that returns a `Promise`.

### DIFF
--- a/packages/teams-js/test/public/stageView.spec.ts
+++ b/packages/teams-js/test/public/stageView.spec.ts
@@ -7,6 +7,8 @@ import { Utils } from '../utils';
 describe('stageView', () => {
   const utils = new Utils();
 
+  const allowedContexts = [FrameContexts.content];
+
   beforeEach(() => {
     utils.processMessage = null;
     utils.messages = [];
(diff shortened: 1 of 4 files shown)


packages/teams-js/test/public/stageView.spec.ts
```diff
@@ -48,69 +48,64 @@ describe('stageView', () => {
...
-      utils.initializeWithContext('content');
-
-      expect(() => stageView.open(null)).toThrowError('[stageView.open] Stage view params cannot be null');
+      expect.assertions(1);
+      utils.initializeWithContext('content').then(() => {
+        expect(() => stageView.open(null)).rejects.toThrowError('[stageView.open] Stage view params cannot be null');
+      });
     });
 
     describe('v1', () => {
```
erinha, line 64: I think the  v1 unit tests can just be removed actually since there is just version of the API now.

richardle798 (dismissed):

shrshindeMSFT (approved):

richardle798 (approved):

PR #1173, opened 2022-05-11
Update CHANGELOG and change files for consistency

erinha:
Cleaning up the CHANGELOG and change files that will end up the CHANGELOG for consistency before the next release.

### DIFF
--- a/change/@microsoft-teams-js-288b4aeb-d82e-48a5-9d09-73197b7f2265.json
+++ b/change/@microsoft-teams-js-288b4aeb-d82e-48a5-9d09-73197b7f2265.json
@@ -1,6 +1,6 @@
 {
   "type": "major",
-  "comment": "Moved subcapabilities and APIs within `legacy` namespace to `teams` namespace.",
+  "comment": "Moved sub-capabilities and APIs within `legacy` namespace to `teams` namespace",
   "packageName": "@microsoft/teams-js",
   "email": "jekoch@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-393fc2b1-d5c9-413a-98d5-b44bf8a88903.json
+++ b/change/@microsoft-teams-js-393fc2b1-d5c9-413a-98d5-b44bf8a88903.json
@@ -1,6 +1,6 @@
 {
-  "type": "none",
-  "comment": "Added missing documentation to the Pages capability",
+  "type": "patch",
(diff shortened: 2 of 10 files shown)


change/@microsoft-teams-js-288b4aeb-d82e-48a5-9d09-73197b7f2265.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "major",
-  "comment": "Moved subcapabilities and APIs within `legacy` namespace to `teams` namespace.",
```
TrevorJoelHarris, line 3: am I allowed to make fun of this

change/@microsoft-teams-js-393fc2b1-d5c9-413a-98d5-b44bf8a88903.json
```diff
@@ -1,6 +1,6 @@
 {
-  "type": "none",
-  "comment": "Added missing documentation to the Pages capability",
+  "type": "patch",
```
TrevorJoelHarris, line 2: For my own edification: why is this a patch? I would have assumed it was none because it had no effect on the shipping package we're delivering. Because it affects documentation?

TrevorJoelHarris (approved):

AE-MS (approved):
:shipit:

change/@microsoft-teams-js-393fc2b1-d5c9-413a-98d5-b44bf8a88903.json
erinha: Yes, I could see going either way, but I think since it's not obvious to me, it's better to have it in the changelog than not. I'm open to other perspectives!

change/@microsoft-teams-js-288b4aeb-d82e-48a5-9d09-73197b7f2265.json
erinha: No. 😝

change/@microsoft-teams-js-393fc2b1-d5c9-413a-98d5-b44bf8a88903.json
TrevorJoelHarris: My only perspective is that we should make sure we have a consistent rule about what all of the different types mean so we make sure the CHANGELOG gets updated deterministically.

PR #1190, opened 2022-05-24
Update build error when change file is not present in PR

erinha:
Add script to check for presence of change files in order to print useful error message.


tools/yaml-templates/build-test-publish.yml
```diff
@@ -17,15 +17,23 @@ steps:
...
+  #   condition: and(
+  #     eq(variables['Build.Reason'], 'PullRequest'),
+  #     not(startsWith(variables['System.PullRequest.SourceBranch'], 'release/')),
+  #     not(startsWith(variables['System.PullRequest.TargetBranch'], 'release/'))
+  #     )
+  #   inputs:
+  #     Arguments: 'beachball check'
+
+  - script: 'node enforceBeachball.js'
+    displayName: 'run enforceBeachball.js script'
```
erinha, line 31: Use "Check that changefile was created if needed" as name of task

aeunms (approved):
🚢

enforceBeachball.js
```diff
@@ -0,0 +1,10 @@
+const cp = require('child_process');
+
+cp.exec('yarn changefile', { timeout: 30000 }, (err, stdout, stderr) => {
+  if (!err && stdout.includes('No change files are needed')) {
+    console.log('Beachball guidelines were correctly followed. Continuing...');
+    return;
+  } else {
+    throw new Error("Change files are required before merging. Please run 'yarn changefile' from the monorepo root.");
```
AE-MS, line 8: >"Change files are required before merging. Please run 'yarn changefile' from the monorepo root." [](http://example.com/codeflow?start=20&length=96)

I don't know how fancy you want to make this but if someone sees this error, is it worth not just saying "run this" but also possibly "see this: [https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball)" ?

AE-MS (approved):
:shipit:

PR #1192, opened 2022-05-25
Adding barCode capability

lakhveerk:


apps/teams-test-app/src/components/BarCodeAPIs.tsx
```diff
@@ -0,0 +1,60 @@
...
+      submit: async input => {
+        const scannedCode = await barCode.scanBarCode(input);
+        return scannedCode;
+      },
+    },
+  });
+
+const HasPermission = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'HasPermission',
```
MarcinJuraszek, line 32: this should probably include the name of the permission, as we'll have plenty of these kind of "Has Permission" test app elements.

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  export function scanBarCode(barCodeConfig: BarCodeConfig): Promise<string> {
+    return new Promise<string>(resolve => {
+      ensureInitialized(FrameContexts.content, FrameContexts.task);
+
+      if (!validateScanBarCodeInput(barCodeConfig)) {
+        throw { errorCode: ErrorCode.INVALID_ARGUMENTS };
+      }
+
+      resolve(sendAndHandleSdkError('barcode.scan', barCodeConfig));
```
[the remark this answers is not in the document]
TrevorJoelHarris: (You'll also need to update runtime to take into account the version number from media.scanBarcode. Will point this out in a comment when I get to that file. Shreyas also helped do that work recently and could be a good resource)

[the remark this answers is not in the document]
TrevorJoelHarris: This function should also check isSupported(), which is a relatively new pattern that was introduced after the initial permissions doc was written.

packages/teams-js/src/public/runtime.ts
```diff
@@ -89,6 +91,7 @@ export const teamsRuntimeConfig: IRuntime = {
   supports: {
     appInstallDialog: {},
     appEntity: {},
+    barCode: {},
```
TrevorJoelHarris, line 94: versionConstants below is where we would need to add barCode support if we want to maintain backcompat

erinha:
You need to add a change file (`yarn changefile`) to make the pipeline happy. :)

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,44 @@
...
+import { runtime } from './runtime';
+
+export namespace barCode {
+  export interface BarCodeConfig {
+    timeOutIntervalInSec?: number;
+  }
+
+  export function scanBarCode(barCodeConfig: BarCodeConfig): Promise<string> {
+    return new Promise<string>(resolve => {
+      ensureInitialized(FrameContexts.content, FrameContexts.task);
```
TrevorJoelHarris, line 15: we should check isSupported here

lakhveerk: Should we add ensureInitialized and isSupported to other two fucntions - hasPermission, and requestPermission - as well?

lakhveerk: In our convo, we decided yes, we need to add both

change/@microsoft-teams-js-e08b2a2a-5a00-4e48-9366-4595b5a723d0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
TrevorJoelHarris, line 2: I don't think this is a major change because major means breaking change. This is probably a minor change because we're adding new functionality in a backwards compatible way

packages/teams-js/src/public/media.ts
```diff
@@ -691,38 +705,35 @@ export namespace media {
...
-      GlobalVars.hostClientType === HostClientType.teamsRoomsWindows ||
-      GlobalVars.hostClientType === HostClientType.teamsRoomsAndroid ||
-      GlobalVars.hostClientType === HostClientType.teamsPhones ||
-      GlobalVars.hostClientType === HostClientType.teamsDisplays
-    ) {
-      const notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
-      callback(notSupportedError, null);
-      return;
-    }
+  export function scanBarCode(callback: (error: SdkError, decodedText: string) => void, config?: BarCodeConfig): void;
```
TrevorJoelHarris, line 708: I think it would make more sense to just keep a single function signature here (the original one that returns void with a callback), mark it as deprecated, and then have it call the new promisified function in the barcode capability

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,89 @@
...
+     * Default value is 30 seconds and max allowed value is 60 seconds
+     */
+    timeOutIntervalInSec?: number;
+  }
+
+  /**
+   * Scan Barcode/QRcode using camera
+   *
+   * @remarks
+   * Note: For desktop and web, this API is not supported.
```
TrevorJoelHarris, line 29: I think we should remove this: we want developers to check isSupported to determine this and we want to make it so that desktop and web hosts can choose to implement this if they feel like it meets the needs of their users.

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * @return a scanned code
+   */
+  export function scanBarCode(barCodeConfig: BarCodeConfig): Promise<string> {
+    return new Promise<string>(resolve => {
+      ensureInitialized(FrameContexts.content, FrameContexts.task);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+      if (!isCurrentSDKVersionAtLeast(scanBarCodeAPIMobileSupportVersion)) {
```
TrevorJoelHarris, line 41: Is this still necessary? I assumed adding the versioning code to runtime would take care of this

packages/teams-js/src/public/media.ts
```diff
@@ -697,32 +700,12 @@ export namespace media {
     }
     ensureInitialized(FrameContexts.content, FrameContexts.task);
 
-    if (
```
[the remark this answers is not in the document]
TrevorJoelHarris: (In the future, it is probably worth making sure that functions like this have full sets of unit tests ahead of trying to make changes to them so we can ensure that the deprecated function hasn't changed behavior at all.)

packages/teams-js/test/public/barCode.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+              data: {
+                id: callbackId,
+                args: [undefined, response],
+              },
+            } as DOMMessageEvent);
+
+            await expect(promise).resolves.toBe(response);
+          });
+
+          it('scanBarCode calls with error', async () => {
```
TrevorJoelHarris, line 84: This test name should probably be a little more descriptive, I wasn't sure what "scanBarCode calls with error" was testing without reading through the test. Looks like something like "scanBarCode rejects promise with Error when error received from host"?

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
+import { sendAndHandleSdkError } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { validateScanBarCodeInput } from '../internal/mediaUtil';
+import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
+import { DevicePermission, ErrorCode } from './interfaces';
+import { runtime } from './runtime';
+
+/**
+ * Namespace to interact with the scan barCode module-specific part of the SDK.
```
erikadoyle, line 9: Suggestion: Namespace to interact with the barcode scanning-specific part of the SDK

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+export namespace barCode {
+  /**
+   * Barcode configuration supplied to scanBarCode API to customize barcode scanning experience in mobile
+   * All properties in BarCodeConfig are optional and have default values in the platform
+   */
```
erikadoyle, line 15: Could we punctuate these into sentences with periods and remove the line break between them?  The documentation will read easier, I think

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
...
+  export interface BarCodeConfig {
+    /**
+     * Optional; Lets the developer specify the scan timeout interval in seconds
+     * Default value is 30 seconds and max allowed value is 60 seconds
```
erikadoyle, line 19: Suggestion:

/**
* Optional; designates the scan timeout interval in seconds.
*  Default value is 30 seconds, max allowed value is 60 seconds.

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
...
+     */
+    timeOutIntervalInSec?: number;
+  }
+
+  /**
+   * Scan Barcode/QRcode using camera
```
erikadoyle, line 25: Suggestion: Scans barcode or QR code using device camera

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * @param barCodeConfig - input configuration to customize the barcode scanning experience
+   *
+   * @return a scanned code
```
erikadoyle, line 29: @return**s**

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+
+      resolve(sendAndHandleSdkError('media.scanBarCode', barCodeConfig));
+    });
+  }
+
+  /**
+   * Checks whether or not media has user permission
+   *
+   * @returns if the media has user permission
```
erikadoyle, line 48: how about:

 @returns `true` if the media has user permission

packages/teams-js/src/public/barCode.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+    const permissions: DevicePermission = DevicePermission.Media;
+
+    return new Promise<boolean>(resolve => {
+      resolve(sendAndHandleSdkError('permissions.has', permissions));
+    });
+  }
+
+  /**
+   * Request user permission for media
```
erikadoyle, line 63: How about: Requests user permission for media

erikadoyle (commented):
Added a few suggestions for documentation (consistent w/similar descriptions we've been iterating on w/geolocation)

packages/teams-js/src/public/media.ts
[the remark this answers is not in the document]
lakhveerk: We make the same check for barCode.scanBarCode API using versionConstants in runtime.ts. I thought that is cleaner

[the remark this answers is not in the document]
lakhveerk: I just read your below comment (in runtime.ts) and it makes sense to keep these. will make the change :)

packages/teams-js/test/public/barCode.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+            expect.assertions(4);
+            try {
+              barCode.hasPermission();
+            } catch (e) {
+              expect(e).toEqual(errorNotSupportedOnPlatform);
+            }
+          });
+
+          it('requestPermission call with successful result', async () => {
+            await mobilePlatformMock.initializeWithContext(context, HostClientType);
```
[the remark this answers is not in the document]
lakhveerk: great catch!! I was checking there is 1 argument (requested permission) but did not actually check it.

packages/teams-js/test/public/barCode.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+    timeOutIntervalInSec: 30,
+  };
+  describe('Testing scanBarCode API', () => {
+    it('should not allow scanBarCode calls before initialization', () => {
+      expect(() => barCode.scanBarCode(barCodeConfig)).rejects.toThrowError('The library has not yet been initialized');
+    });
+
+    Object.values(FrameContexts).forEach(context => {
+      if (allowedContexts.some(allowedContext => allowedContext === context)) {
+        Object.values(allowedHostClientTypes).forEach(hostClientType => {
```
TrevorJoelHarris, line 56: Because we don't need to check all of the hostClientTypes anymore (because barcode doesn't care about them, I think we can not loop through the whole big set of hostClientTypes. and just run this set of tests once.

packages/teams-js/src/public/media.ts
```diff
@@ -716,13 +719,7 @@ export namespace media {
...
-
-    if (!validateScanBarCodeInput(config)) {
-      const invalidInput: SdkError = { errorCode: ErrorCode.INVALID_ARGUMENTS };
-      callback(invalidInput, null);
-      return;
-    }
-
-    sendMessageToParent('media.scanBarCode', [config], callback);
+    config = config === undefined ? {} : config;
+    callCallbackWithErrorOrResultFromPromiseAndReturnPromise<string>(barCode.scanBarCode, callback, config);
```
TrevorJoelHarris, line 723: This function call should be reverted to just its old behavior and should no longer call the new barCode function. Barcode.isSupported will need to check both barcode and permissions, and it will be impossible to make this pattern work because permissions checking/requesting is totally new functionality.

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/media.ts
```diff
@@ -683,6 +683,9 @@ export namespace media {
   }
 
   /**
+   * @deprecated
+   * As of 2.0.1, please use {@link barCode.scanBarCode barCode.scanBarCode(config?: BarCodeConfig): Promise\<string\>} instead.
```
jadahiya-MSFT, line 687: @erinha Should we be using 2.0.1 here?

jadahiya-MSFT (dismissed):

packages/teams-js/src/public/media.ts
erinha: Good catch. No, we should use 2.1.0 for this. A good way to determine what the next version will be for deprecated tag purposes is by locally running `yarn beachball bump` from the root of the monorepo and seeing what the version is in packages/teams-js/package.json as a result. 

NOTE: Make sure you do that without any other changes in your local repo because it will delete a bunch of change files and modify the packages/teams-js/package.json and you'll want to undo those changes after you run the command.

lakhveerk: Thanks for catching this Jay and for the info Erin!

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

PR #1193, opened 2022-05-30
Renamed prepNextDevRelease to prepBetaRelease

jadahiya-MSFT:
With the release of 2.0.0, we would like to switch from next dev releases to beta releases. This PR aims to leverage how we bumped next dev releases to bump beta releases instead.

How it works:
`yarn beachball bump` returns the latest version bump into the package.json file, we then compare this to the npm's latest release to determine the correct beta bump version (e.g beachball bump returns 2.1.0 and npm returns 2.0.0 so our next beta release will be 2.1.0-beta.0)


erinha:
The build is failing complaining you need a change file. My bet would be this is legitimately change type "none".

packages/teams-js/prepBetaRelease.js
```diff
@@ -128,37 +128,46 @@ function getNewNextDevSuffixNum(currNextDevVer, currPkgJsonVer) {
...
 
   // get package version from package.json
   let currPkgJsonVer = getPkgJsonVersion(packageJson);
+  let currBetaPrefix = getPrefix(currBetaVer);
 
   console.log('package.json version: ' + currPkgJsonVer);
-  console.log('current next-dev tagged version: ' + currNextDevVer);
+  console.log('current beta tagged version: ' + currBetaVer);
 
   if (currPkgJsonVer.includes('dev')) {
```
erinha, line 154: I think this check should be changed to 'beta', right?

jadahiya-MSFT:
> The build is failing complaining you need a change file. My bet would be this is legitimately change type "none".

Was planning to add the changefile after review since I find it easier to formulate the changelog comment once all the changes are properly reflected, but I will add it now

packages/teams-js/prepBetaRelease.js
jadahiya-MSFT: Whoops good catch

change/@microsoft-teams-js-b469c20c-b2ae-4fcb-aa82-4d19b3292ae9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Replacing prepNextDevRelease.js with prepBetaRelease.js to handle beta version bumping and deprecating use of nextDev tags for releases",
```
erinha, line 3: Super-nit: I'd prefer change file updates to be in past tense. I know this won't be going in the change log because it's of type "none", but others might look at existing change files for examples and I'd like to be as consistent as possible.

jadahiya-MSFT: Thanks, this is good to keep in mind for the future :)

packages/teams-js/prepBetaRelease.js
```diff
@@ -167,13 +176,17 @@ function getNewPkgJsonContent(currNextDevVer) {
...
   const newPackageJson = getNewPkgJsonContent(devStdout);
   const newVersion = newPackageJson.version;
   saveJsonFile(newPackageJson);
   return newVersion;
 }
 
 (() => {
-  exec(`npm view @microsoft/teams-js version --tag next-dev`).then(({ stdout, stderr }) => prepNewDevRelease(stdout.trim()));
+  exec(`yarn beachball bump`).then(
+    exec(`npm view @microsoft/teams-js version --tag beta`).then(({ stdout, stderr }) =>
```
erinha, line 188: Okay, this is a little tricky. We will have to find a way to manually set the first 2.0 beta version. Since `npm view @microsoft-teams-js version --tag beta` currently returns "1.12.0-beta.5". The first time we do this, I think it needs to query for "--tag next-dev", which would return "2.0.0-beta.7-dev.14".

tools/yaml-templates/build-test-publish.yml
```diff
@@ -29,8 +29,8 @@ steps:
 
   # Temporarily disable this step until after v2.0.0 is released to avoid having dev releases with inaccurate
   # 2.0.0-beta.x naming
-  # - script: 'node prepNextDevRelease.js'
-  #   displayName: 'node prepNextDevRelease.js'
+  # - script: 'node prepBetaRelease.js'
```
erinha, line 32: I've been running the script and I think there is a problem. The `yarn beachball bump` command needs to be executed from the monorepo root. But the script is executed from the `packages/teams-js` directory, and when run from that directory, it seems like `yarn beachball bump` fails. 

Have you been able to run this script locally successfully?

packages/teams-js/prepBetaRelease.js
```diff
@@ -167,13 +176,17 @@ function getNewPkgJsonContent(currNextDevVer) {
...
+function prepBetaRelease(devStdout) {
   const newPackageJson = getNewPkgJsonContent(devStdout);
   const newVersion = newPackageJson.version;
   saveJsonFile(newPackageJson);
   return newVersion;
 }
 
 (() => {
-  exec(`npm view @microsoft/teams-js version --tag next-dev`).then(({ stdout, stderr }) => prepNewDevRelease(stdout.trim()));
+  exec(`yarn beachball bump`).then(
```
erinha, line 187: I think if you change this to the following the code works. But the comment about the first time we do this still applies.

```
`exec(`pushd ../.. && yarn beachball bump`).then(`
```

erinha: I should amend my previous comment. When I make the change above and run the script, I can see that `yarn beachball bump` no longer throws errors. But, it doesn't seem to be updating the package.json (though it's definitely running and updating the change log) before running the next command.

tools/yaml-templates/build-test-publish.yml
erinha: I think I found a way around this (https://github.com/OfficeDev/microsoft-teams-library-js/pull/1193/files#r886228992), but it still doesn't seem to be working properly for me locally.

packages/teams-js/prepBetaRelease.js
jadahiya-MSFT: So yarn beachball bump will return 2.1.0, and so I think this will be fine since it will automatically set it to 2.1.0-beta.0

tools/yaml-templates/build-test-publish.yml
jadahiya-MSFT: I have been able to run this script locally, but I will look into this more

packages/teams-js/prepBetaRelease.js
jadahiya-MSFT: > I think if you change this to the following the code works. But the comment about the first time we do this still applies.
> 
> ```
> `exec(`pushd ../.. && yarn beachball bump`).then(`
> ```

I've tried `pushd`, but for some reason it complains that my syntax isn't correct. Changing it to `cd` fixes it however.

packages/teams-js/prepBetaRelease.js
```diff
@@ -128,37 +128,46 @@ function getNewNextDevSuffixNum(currNextDevVer, currPkgJsonVer) {
     // If the next-dev beta suffix is not equal to or exactly 1 higher than the package.json beta suffix, throw an error since this does not follow our versioning structure.
     // e.g. package.json version: 2.0.0-beta.1 and current next-dev version: 2.0.0-beta.0-dev.0 should throw an error.
   } else if (betaSuffixNumInCurrNextDev !== betaSuffixNumInPkgJsonVer) {
-    throw new Error(`Invalid beta version suffix number ${betaSuffixNumInPkgJsonVer} in package.json version ${currPkgJsonVer}`);
```
erinha, line 131: Sorry I missed this earlier. Can you please update the `getNewNextDevSuffixNum` function to `getNewBetaSuffixNum` and any references to "*nextDev*"? E.g. `currNextDevVer` in the function signature. I also think that code should be simplified now that we're not looking for "*.beta.x-dev.y".

erinha: Ah upon further inspection, I see this function isn't being used at all. Which I think is a problem (see my comment on line 161).

packages/teams-js/prepBetaRelease.js
```diff
@@ -128,37 +128,46 @@ function getNewNextDevSuffixNum(currNextDevVer, currPkgJsonVer) {
...
-  if (currPkgJsonVer.includes('dev')) {
-    throw new Error(`The given package.json\'s version ${currPkgJsonVer} contains the substring \'dev\' which is reserved for non-prod versions. Please fix the package.json version first in order to allow for proper version incrementation.`);
+  if (currPkgJsonVer.includes('beta')) {
+    throw new Error(
+      `The given package.json\'s version ${currPkgJsonVer} contains the substring \'beta\' which is reserved for non-prod versions. Please fix the package.json version first in order to allow for proper version incrementation.`,
+    );
+  }
+  let betaVerNum = 0;
+  if (currBetaPrefix === currPkgJsonVer) {
+    betaVerNum = getSpecificVerSuffixNum('beta', currPkgJsonVer) + 1;
```
erinha, line 161: If you pass in "2.1.0-beta.7" as the input to `currBetaVer`, then this script will return `2.1.0-beta.0`, which is incorrect. It should be `2.1.0-beta.8`, based on the input. 

The script needs to take into account whether to increment the version that would come from `yarn beachball bump` OR increment the `currBetaVersion`. I think you need to update `getNewNextDevSuffixNumber` to take the versions into account to determine how to update the beta suffix number like the script originally had on old line 160. (This is related to my comment on old line 131.)

packages/teams-js/prepBetaRelease.js
erinha: I think it would be great if you could clean up the script and remove all unused functions/variables as part of this change.

erinha (changes requested):
Left comments about cleaning up dead code and making sure the package version is being set correctly. I think we're really close!

packages/teams-js/prepBetaRelease.js
```diff
@@ -86,6 +59,31 @@ function getPrefix(version) {
   }
 }
 
+/**
+ * Gets the prefix of the version that is higher amongst the inputs
+ * @param {string} currBetaVer The version currently tagged beta. (e.g. 2.0.0-beta.1)
+ * @param {string} currPkgJsonVer The version taken from the package.json. (e.g. 2.0.0)
+ * @returns The prefix of the version that is higher
+ */
+function getNextPrefix(currBetaVer, currPkgJsonVer) {
```
erinha, line 68: Should this be called `getBetaPrefix`?

We were using "next" because the tag was next and "next-dev" because that was part of the version suffix. But now we're using beta as the tag and the version suffix, so I don't think "next" makes as much sense.

packages/teams-js/prepBetaRelease.js
```diff
@@ -146,32 +139,22 @@ function getNewPkgJsonContent(currBetaVer) {
...
       `The given package.json\'s version ${currPkgJsonVer} has a non-integer beta version number. Please fix the package.json version first in order to allow for proper version incrementation.`,
     );
   }
-  const newVersion = currBetaPrefix + '-beta.' + betaVerNum;
+  const newVersion = newVersionPrefix + '-beta.' + betaVerNum;
 
   console.log('new version: ' + newVersion);
 
-  // update package.json with the new version
+  //update package.json with the new version
```
erinha, line 157: nit: keeping a space after the comments is more consistent with the rest of the code.

packages/teams-js/prepBetaRelease.js
```diff
@@ -102,37 +100,32 @@ function getSpecificVerSuffixNum(versionType, wholeVerNum) {
...
-    );
+  let newBetaSuffixNum = 0;
+
+  //If current beta versioning is higher or equal to currPkgJsonVer
+  if (getPrefix(currBetaVer) === nextPrefix) {
+    const suffixNumInCurrBeta = getSpecificVerSuffixNum('beta', currBetaVer);
+    if (suffixNumInCurrBeta < 0) {
+      throw new Error(`Invalid beta version suffix number ${suffixNumInCurrBeta} in current beta version`);
+    }
+    const suffixNumInPkgJsonVer = getSpecificVerSuffixNum('beta', currPkgJsonVer);
```
erinha, line 122: How would we ever get a "-beta.x" suffix in the package.json version number?

erinha (changes requested):
Thanks for doing the clean-up of unused functions!

This is looking great! There are a couple of comments/questions about some of the helper methods and error messages. We're nearly there. :)

packages/teams-js/prepBetaRelease.js
jadahiya-MSFT: I don't think `getBetaPrefix` is correct here. I think I should rename it to `getNewerPrefix` since this exists to determine which of the beta or package.json versions is newer. E.g if beta = 2.1.0 and package.json = 2.0.0. choose beta version. If package.json = 2.1.0 and beta = 2.0.0 then choose package.json.

packages/teams-js/prepBetaRelease.js
jadahiya-MSFT: I kept this in here since it was part of the logic for next-dev which made me think there are sometimes instances where the package.json version will have a beta version. I didn't think it was likely though due to the `yarn beachball bump` call. I will remove this

packages/teams-js/prepBetaRelease.js
erinha: I like `getNewerPrefix` and your reasoning makes sense!

erinha (dismissed):
🚀

erinha (approved):
Good catch!

shrshindeMSFT (approved):
LGTM!

PR #1200, opened 2022-06-06
Updating location capability

lakhveerk:
splitting up the location capability into a new set of functions and sub-capabilities. Also using permissions


packages/teams-js/src/public/location.ts
```diff
@@ -46,82 +44,107 @@ export namespace location {
...
+   * @returns User's current location
    */
-  export function getLocation(props: LocationProps): Promise<Location>;
-  /**
-   * @deprecated
-   * As of 2.0.0, please use {@link location.getLocation location.getLocation(props: LocationProps): Promise\<Location\>} instead.
-   * @param props {@link LocationProps} - Specifying how the location request is handled
-   * @param callback - Callback to invoke when current user location is fetched
-   */
-  export function getLocation(props: LocationProps, callback: (error: SdkError, location: Location) => void): void;
```
jadahiya-MSFT, line 61: Is this supposed to be deleted? This is the backwards compatible function with a required callback?

apps/teams-test-app/src/components/LocationAPIs.tsx
```diff
@@ -16,13 +16,13 @@ const GetLocation = (): React.ReactElement =>
     title: 'Get Location',
     onClick: {
       validateInput: input => {
-        if (input.allowChooseLocation === undefined) {
+        if (input && input.allowChooseLocation === undefined) {
```
jadahiya-MSFT, line 19: Is validating input necessary anymore if we're doing `location.getCurrentLocation()`? I think it might make more sense to validate `locationProps` in `withCallback` in this instance since input should be allowed to be undefined if we go down the `Promise` route

apps/teams-test-app/src/components/LocationAPIs.tsx
```diff
@@ -69,11 +69,33 @@ const ShowLocation = (): React.ReactElement =>
     },
   });
 
+const HasGeoLocationPermission = (): React.ReactElement =>
```
jadahiya-MSFT, line 72: Potential nit: I think most of our UI components put the Has[InsertNameHere]Permission as the last UI element in the section. I would have to confirm this, but maybe we want to keep that consistent pattern?

packages/teams-js/src/public/location.ts
jadahiya-MSFT: I see that it was actually moved down below, sorry about that

apps/teams-test-app/src/components/LocationAPIs.tsx
lakhveerk: Thank you for the quick review, Jay!!

packages/teams-js/src/public/location.ts
```diff
@@ -1,14 +1,12 @@
...
-  callCallbackWithErrorOrResultFromPromiseAndReturnPromise,
-} from '../internal/utils';
-import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
-import { ErrorCode, SdkError } from './interfaces';
+import { FrameContexts } from './constants';
+import { DevicePermission, ErrorCode, SdkError } from './interfaces';
 import { runtime } from './runtime';
-
+/**
+ * @alpha
```
TrevorJoelHarris, line 8: I seem to remember erin saying that @alpha doesn't actually *mean* anything in our documentation. Given that location is already a shipped capabiility that we are restructuring, we should probably just leave @alpha/@beta off of this

packages/teams-js/src/public/location.ts
```diff
@@ -46,82 +44,107 @@ export namespace location {
...
+   * - {@link location.map.chooseLocation location.map.chooseLocation(): Promise\<Location\>}
    *
-   * @param location {@link Location} - which needs to be shown on map
-   * @returns Promise that will be fulfilled when the operation has completed
+   * @param props {@link LocationProps} - Specifying how the location request is handled
+   * @param callback - Callback to invoke when current user location is fetched
    */
-  export function showLocation(location: Location): Promise<void>;
+  export function getLocation(props: LocationProps, callback: (error: SdkError, location: Location) => void): void {
+    if (!callback) {
```
TrevorJoelHarris, line 85: Is this preserving the old behavior? I don't see this error being called anywhere, and we should make sure that we aren't altering how this function works (from the caller's perspective) as part of this change.

packages/teams-js/src/public/location.ts
TrevorJoelHarris: (Expanding this comment to say what this capability is would be great as well)

packages/teams-js/src/public/location.ts
```diff
@@ -46,82 +44,107 @@ export namespace location {
...
       if (!location) {
         throw { errorCode: ErrorCode.INVALID_ARGUMENTS };
       }
-      if (!isSupported()) {
-        throw errorNotSupportedOnPlatform;
-      }
-      resolve(sendAndHandleError('location.showLocation', location));
-    });
-  }
+      return sendAndHandleError('location.showLocation', location);
```
TrevorJoelHarris, line 143: If we make this `sendAndHandleError('location.getLocation', { allowChooseLocation: false, showMap: true})` it would help make sure that teams continues to work with the new functions as well!

packages/teams-js/src/public/runtime.ts
```diff
@@ -97,6 +101,9 @@ export const teamsRuntimeConfig: IRuntime = {
       update: {},
     },
     logs: {},
+    location: {
```
TrevorJoelHarris, line 104: Do we need to make a change to versionContants below to take into account the addition of the 'map' subcapability?

packages/teams-js/src/public/tasks.ts
```diff
@@ -19,7 +19,8 @@ export namespace tasks {
   /**
    * @deprecated
    * As of 2.0.0, please use {@link dialog.open(urlDialogInfo: UrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} for url based dialogs
-   * and {@link dialog.bot.open(botUrlDialogInfo: BotUrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} for bot based dialogs.
+   * and {@link dialog.bot.open(botUrlDialogInfo: BotUrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} for bot based dialogs. In Teams,
+   * this function can be used for adaptive card based dialogs. Support for adaptive card based dialogs is coming to other hosts in the future.
```
TrevorJoelHarris, line 23: This maybe should be a separate change instead of being included with the location work. Good addition though

lakhveerk: lol, I have no idea how this got in here, but thank you for catching it

packages/teams-js/src/public/location.ts
lakhveerk: Seems like callback was required before and is required now. The function was overloaded to convert it to promises. I will dm you to understand what I am missing

packages/teams-js/src/public/location.ts
lakhveerk: Will do thank you!

packages/teams-js/src/public/location.ts
jadahiya-MSFT: @TrevorJoelHarris @lakhveerk I believe the confusion for this stems from the V1 `getLocation` and the old V2.0.0 `getLocation`.  V1's `getLocation` had a required `callback`, while V2.0.0 made this callback optional and would return a `Promise<void>`.  

I believe Trevor you are looking at the old code for `getLocation` that did overload the result into being a promise if a callback wasn't provided, but was effectively backwards compatible in functionality. i.e if callback was provided it would execute as the old `getLocation`.  Meanwhile Lakhveer converted the old backwards compatible function stub to being the implementation instead since it appears that `getLocation` is no longer needed and thus the overload logic wouldn't be needed anymore.

I believe for now it might be best to convert it back to the V2.0.0 behavior of overloading to promises with an optional callback handler for BC routing and then deprecate that functionality at a later time to what Lakhveer has here, but would be ok with discussing this later with you two. If we do proceed with making this specific change, I think it would be a good idea to communicate this out, in case someone is reliant on the current functionality.

packages/teams-js/src/public/location.ts
lakhveerk: per our convo, I will be keeping it the way it is.  showLocation is separate function than getLocation
`showLocation` takes only location parameter and show the location on map. By changing it to the new way, 

We will be breaking the back compatibility because in v1 (and in v2), previously, we send location.showLocation to the hubSDK.

showLocation takes co-ordinators in and show them on a map. If we don't send the location coordinators, host won't have anything to show on map

packages/teams-js/src/public/location.ts
```diff
@@ -46,82 +51,147 @@ export namespace location {
...
-        throw errorNotSupportedOnPlatform;
-      }
-      resolve(sendAndHandleError('location.getLocation', props));
+  /**
+   * Request user permission for location
+   *
+   * @returns if the user conseted permission for location
+   */
+  export function requestPermission(): Promise<boolean> {
+    ensureInitialized(FrameContexts.content, FrameContexts.task);
```
TrevorJoelHarris, line 88: Do you think it would make sense to put the implementations of hasPermission and requestPermission into a shared helper function so we don't have to repeat the logic in every capability that uses them?

lakhveerk: Like we talked, there is not much to share in these functions, it is good to keep them in separate files :)

packages/teams-js/src/public/location.ts
```diff
@@ -1,15 +1,17 @@
...
 import { ensureInitialized, isCurrentSDKVersionAtLeast } from '../internal/internalAPIs';
-import {
-  callCallbackWithErrorOrBooleanFromPromiseAndReturnPromise,
-  callCallbackWithErrorOrResultFromPromiseAndReturnPromise,
-} from '../internal/utils';
 import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
 import { ErrorCode, SdkError } from './interfaces';
 import { runtime } from './runtime';
 
+/**
```
TrevorJoelHarris, line 8: We should probably mark the namespace and interfaces as deprecated as well

packages/teams-js/src/public/location.ts
```diff
@@ -45,82 +50,61 @@ export namespace location {
...
-   *
-   * @param props {@link LocationProps} - Specifying how the location request is handled
-   * @returns Promise that will be fulfilled when the operation has completed
-   */
-  export function getLocation(props: LocationProps): Promise<Location>;
   /**
    * @deprecated
-   * As of 2.0.0, please use {@link location.getLocation location.getLocation(props: LocationProps): Promise\<Location\>} instead.
+   * As of 2.0.1, please use one of the following functions:
+   * - {@link geoLocation.getCurrentLocation geoLocation.getCurrentLocation(): Promise\<Location\>} to get the current locaiton.
```
TrevorJoelHarris, line 56: small typo, should be "location"

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
+import { sendAndHandleSdkError as sendAndHandleError } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
+import { DevicePermission, ErrorCode } from './interfaces';
+import { runtime } from './runtime';
+
+/**
+ * Namespace to interact with the geoLocation module-specific part of the SDK. This is the newwer version of location module.
```
TrevorJoelHarris, line 8: Small type: should be newer instead of newwer

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+      */
+    accuracy?: number;
+    /**
+      Time stamp when the location was captured
+      */
+    timestamp?: number;
+  }
+  /**
+   * Fetches current user coordinates
+   * @returns Promise of user's current {@linkcode Location}
```
erikadoyle, line 34: Did you mean `@link` here? (*@linkcode* doesn't seem to be a valid [typedoc tag](https://typedoc.org/guides/tags/)). The rendered doc will actually show the Promise return type and link to Location based on the function signature alone, so adding a link in the description isn't necessary (but you can if you like). For example, see [getContext()](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/app?view=msteams-client-js-latest#@microsoft-teams-js-app-getcontext) documentation (the broken link in the rendered doc is a known issue under investigation).

erikadoyle: Also for return description, could we modify slightly to match previous function descriptions that we promis-ified? For example, `@returns Promise that will resolve with the {@link location.Location} object` or `@returns Promise that will resolve with information about the user's location`.

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+    return sendAndHandleError('location.getLocation', { allowChooseLocation: false, showMap: false });
+  }
+
+  /**
+   * Checks whether or not location has user permission
+   *
+   * @returns if the location has user permission
```
erikadoyle, line 47: How about something like: 
@returns `true` if the user has granted the app permission to location information, `false` otherwise

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+    const permissions: DevicePermission = DevicePermission.GeoLocation;
+
+    return new Promise<boolean>(resolve => {
+      resolve(sendAndHandleError('permissions.has', permissions));
+    });
+  }
+
+  /**
+   * Request user permission for location
```
erikadoyle, line 62: How about: Request**s** user permission for location

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @returns if the user conseted permission for location
```
erikadoyle, line 64: How about: `@returns Promise that will resolve with state of location permission granted by the user`

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+   * Namespace to interact with the location on map module-specific part of the SDK.
+   */
+  export namespace map {
+    /**
+     * Allows user to choose location on map
+     *
+     * @returns The location chosen by the user after closing the map
```
erikadoyle, line 89: How about: `@returns Promise that will resolve with the location chosen by the user` (is the *after closing the map* necessary to the function of this API, or is it a UX detail that could change in the future?)

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+      return sendAndHandleError('location.getLocation', { allowChooseLocation: true, showMap: true });
+    }
+
+    /**
+     * Shows the location on map corresponding to the given coordinates
+     *
+     * @param location {@link Location} - which needs to be shown on map
```
erikadoyle, line 102: How about: `@param location to be shown on the map` (link to the Location paramater type will be generated automatically above this rendered description)

packages/teams-js/src/public/location.ts
```diff
@@ -45,82 +50,61 @@ export namespace location {
...
-   */
-  export function showLocation(location: Location): Promise<void>;
   /**
    * @deprecated
-   * As of 2.0.0, please use {@link location.showLocation location.showLocation(location: Location): Promise\<void\>} instead.
+   * As of 2.0.1, please use {@link geoLocation.map.showLocation geoLocation.map.showLocation(location: Location): Promise\<void\>} instead.
+   *
    * Shows the location on map corresponding to the given coordinates
+   *
    * @param location {@link Location} - which needs to be shown on map
```
erikadoyle, line 88: How about: `@param location - Location to be shown on the map` (Location type/link will be auto-rendered above this description)

erikadoyle (commented):
Looks good overall! Added a few suggestions~

packages/teams-js/src/public/location.ts
lakhveerk: Thank you, Erika!! These comments are super helpful!!

packages/teams-js/src/public/geoLocation.ts
lakhveerk: Just to match with the hasPermision API, what about this:
@returns true if the user consented permission for location, false otherwise

packages/teams-js/src/public/runtime.ts
```diff
@@ -154,6 +160,10 @@ export const versionConstants: Record<string, Array<ICapabilityReqs>> = {
     },
   ],
   '2.0.1': [
+    {
```
TrevorJoelHarris, line 163: I don't think we need this anymore because geoLocation isn't called by location at all we don't need a default isSupported value.

packages/teams-js/src/public/location.ts
```diff
@@ -45,80 +55,60 @@ export namespace location {
...
-      }
-      if (!isSupported()) {
-        throw errorNotSupportedOnPlatform;
-      }
-      resolve(sendAndHandleError('location.showLocation', location));
-    });
+    sendMessageToParent('location.showLocation', [location], callback);
   }
 
   export function isSupported(): boolean {
```
TrevorJoelHarris, line 114: should mark this as deprecated

TrevorJoelHarris (changes requested):

jadahiya-MSFT (approved):

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,132 @@
+import { sendAndHandleSdkError as sendAndHandleError } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
+import { DevicePermission, ErrorCode } from './interfaces';
+import { runtime } from './runtime';
+
+/**
+ * Namespace to interact with the geoLocation module-specific part of the SDK. This is the newer version of location module.
```
erinha, line 8: Just a question -- is this capability still under development and should all namespaces, functions, interfaces, enums, etc. in this file have a @beta tag?

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -0,0 +1,132 @@
...
+    accuracy?: number;
+    /**
+      Time stamp when the location was captured
+      */
+    timestamp?: number;
+  }
+  /**
+   * Fetches current user coordinates
+   * @returns Promise that will resolve with the {@link geoLocation.Location} object
+   */
```
erinha, line 35: Should we include in our doc comments that the function could reject with an error, and also throw a not supported on platform error?

erinha: This applies to all the functions in this file (and probably a bunch more files/capabilities that are definitely outside the scope of this PR).

packages/teams-js/src/public/geoLocation.ts
lakhveerk: How do we decide on that? It is done on teams-js side. On host side, if some host have not implemented the interface, that can be checked with isSupported. 
@TrevorJoelHarris

packages/teams-js/src/public/geoLocation.ts
lakhveerk: I have made the changes. Could you review again please?

packages/teams-js/src/public/geoLocation.ts
lakhveerk: Added @beta tags

jadahiya-MSFT (approved):

packages/teams-js/test/public/location.spec.ts
```diff
@@ -28,203 +29,301 @@ describe('location', () => {
...
-
-    await expect(promise).rejects.toEqual({ errorCode: ErrorCode.PERMISSION_DENIED });
-  });
-  it('Frameless - getLocation should throw error when not supported in the runtime config', async () => {
-    await framelessPlatform.initializeWithContext(FrameContexts.task);
-    framedPlatform.setRuntimeConfig({ apiVersion: 1, supports: {} });
-    const promise = location.getLocation(defaultLocationProps);
-    await expect(promise).rejects.toEqual(errorNotSupportedOnPlatform);
-  });
-  it('Frameless - showLocation should throw error when location is not supported', async () => {
```
erinha, line 210: Even though it's deprecated, we should still have tests for `location.isSupported`. I would have expected this file to look a little more like locationLegacy.spec.ts, which does have those tests I think. (I diff'ed location.spec.ts with your changes against locationLegacy.spec.ts which was helpful in reviewing :)).

erinha (changes requested):
Feedback on lack of location.isSupported unit tests.

packages/teams-js/test/public/location.spec.ts
```diff
@@ -326,4 +326,66 @@ describe('location', () => {
...
+      try {
+        location.showLocation(defaultLocation, emptyCallback);
+      } catch (e) {
+        expect(e).toEqual(errorNotSupportedOnPlatform);
+      }
+    });
+  });
+
+  it('Framed - getLocation should throw error when not supported in the runtime config', () => {
+    framelessPlatform.setRuntimeConfig({ apiVersion: 1, supports: {} });
```
erinha, line 367: This says Framed, but you're still using the framelessPlatform mock inside the test.

erinha (changes requested):
Thanks for adding the tests! Framed tests still using frameless mock, I think that's a bug.

packages/teams-js/test/public/location.spec.ts
lakhveerk: Thanks for catching it!

erinha (approved):
Looks good, thanks!

TrevorJoelHarris (approved):

PR #1203, opened 2022-06-07
[No Fork] Update build error when change file isn't present in PR

erinha:
Add script to check for presence of change files in order to print useful error message.

I created a PR for this from my fork but on that PR, attempts to trigger the ADO pipeline are failing and I'm not sure why. Trying a PR directly against the origin in case that is somehow the problem.


erinha:
Closing this in favor of PR #1190

PR #1206, opened 2022-06-09
Re-enabling prep beta script in pipeline

jadahiya-MSFT:


tools/yaml-templates/build-test-publish.yml
```diff
@@ -25,16 +25,15 @@ steps:
...
-  #     )
+  # Determines what the next beta version will be. E.g 2.0.0-beta.x
+   - script: 'node prepBetaRelease.js'
+     displayName: 'node prepBetaRelease.js'
+     workingDirectory: '$(System.DefaultWorkingDirectory)\packages\teams-js'
+     condition: and(
+       eq(variables['Build.SourceBranch'], 'refs/heads/main'),
+       in(variables['Agent.JobStatus'], 'Succeeded', 'SucceededWithIssues'),
+       ne(variables['Build.Reason'], 'PullRequest')
+       )
```
erinha, line 36: The OfficeDev.microsoft-teams-library-js pipeline check isn't getting queued. I saw this with my most recent PR -- calling a new script called `enforceBeachball.js` -- PR #1190. In my case, the whitespace around the closing parenthesis was off, which caused the pipeline to not get triggered. You might want to try fiddling with the whitespace to see if you can get the pipeline check to properly execute.

erinha (changes requested):
Need to get the pipeline to trigger, but the change looks good to me otherwise. See my comment about possible causes of the pipeline getting stuck in "Expected — Waiting for status to be reported" state.

tools/yaml-templates/build-test-publish.yml
jadahiya-MSFT: Weird, I will play around with the whitespace

PR #1213, opened 2022-06-15
Hiding unreleased public api's

sagarchkrbrty:
Added `@hidden` tag to unreleased public `video` APIs

### DIFF
--- a/packages/teams-js/src/public/video.ts
+++ b/packages/teams-js/src/public/video.ts
@@ -5,10 +5,12 @@ import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
 import { runtime } from './runtime';
 
 /**
+ * @hidden
  * Namespace to video extensibility of the SDK
  */
 export namespace video {
   /**
+   * @hidden
    * Represents a video frame
    */
   export interface VideoFrame {
@@ -39,6 +41,7 @@ export namespace video {
   }
 
   /**
+   * @hidden
    * Video frame format enum, currently only support NV12
    */
   export enum VideoFrameFormat {
@@ -46,6 +49,7 @@ export namespace video {
   }
 
   /**
+   * @hidden
    * Video frame configuration supplied to the host to customize the generated video frame parameters, like format
    */
   export interface VideoFrameConfig {
@@ -56,6 +60,7 @@ export namespace video {
   }
 
   /**
+   * @hidden
    *  Video effect change type enum
    */
   export enum EffectChangeType {
@@ -70,6 +75,7 @@ export namespace video {
   }
 
   /**
+   * @hidden
    *  Video frame call back function definition
    */
   export type VideoFrameCallback = (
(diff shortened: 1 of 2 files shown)


packages/teams-js/src/public/video.ts
```diff
@@ -5,10 +5,12 @@ import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
 import { runtime } from './runtime';
 
 /**
+ * @hidden
```
erinha, line 8: Are these APIs still under development? If so and there are plans to continue changing them before "releasing" them, consider adding a "@beta" tag to them as well, indicating to developers that they might change and they should not build production level apps using these APIs without understanding that they might break at any time.

AE-MS:
If these APIs are still under development, has there been a design review on this feature? I don't recall seeing it coming through office hours.
https://aka.ms/extendingteamsjssdk
Thanks!

erinha:
You need to provide a change file by running `yarn changefile` in the root of the monorepo. Suggested change type: patch. Suggested description of your change: "Added \`@hidden\` tag to unreleased public \`video\` APIs". (Note the backtick quotes and past tense used.)

erinha (changes requested):
See comments about change file and design review on APIs under development.

sagarchkrbrty:
> If these APIs are still under development, has there been a design review on this feature? I don't recall seeing it coming through office hours. https://aka.ms/extendingteamsjssdk Thanks!

These were part of V1 SDK

AE-MS:
> > If these APIs are still under development, has there been a design review on this feature? I don't recall seeing it coming through office hours. https://aka.ms/extendingteamsjssdk Thanks!
> 
> These were part of V1 SDK

That's good to know -- but if they are still under development, they are now being developed for the V2 SDK, correct?

packages/teams-js/src/public/video.ts
sagarchkrbrty: Good point. Added the tag.

sagarchkrbrty:
> You need to provide a change file by running `yarn changefile` in the root of the monorepo. Suggested change type: patch. Suggested description of your change: "Added `@hidden` tag to unreleased public `video` APIs". (Note the backtick quotes and past tense used.)
> 
> See the [CONTRIBUTING.md](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball) for more details.

Added the changefile

erinha:
Since these APIs have publicly shipped (both in v1 and in v2) they need to have public documentation. We can't actually go back and remove it since we're committed to supporting these APIs now that they've been released. I suggest you attend office hours to discuss your feature and its roadmap, given that it has already been publicly released and has not been previously reviewed. https://aka.ms/extendingteamsjssdk

packages/teams-js/src/public/video.ts
erinha: See my other comment after further thought about already releasing these APIs.

erinha:
Feel free to reopen after discussing the plan forward.

PR #1214, opened 2022-06-16
Get web storage capabilities

reenudeswal:
1.  The API  allows apps to discover if web storage is supported by the host client.
2. Currently, assignments app needs this API to identify if the Teams client (specifically Mobile) provides web storage support. Based on that Assignment can decide if they want to store the data or not. 
3. Newer Teams mobile client having support for this API will clear web storage on logout and older clients won’t.
4. Teams desktop always clears web storage on logout and will always return true in response for Api.
5. Teams web never clears web storage and will always return false.
6. The Api is marked at beta as currently there is no plan to support offline/storage capabilities in App SDK.
7. As the API is not a asynchronous function, hence promises are not used.


packages/teams-js/src/public/storage.ts
```diff
@@ -0,0 +1,24 @@
...
+import { HostClientType } from './constants';
+import { runtime } from './runtime';
+
+/**
+ * Contains functionality to allow web apps to store data in webview cache
+ *
+ * @beta
+ */
+export namespace storage {
+  export function isWebStorageClearedOnUserLogOut(): boolean {
```
erinha, line 12: Please add reference documentation comments indicating what this function does.

change/@microsoft-teams-js-4721f042-c277-4707-80ec-9a9dfe3ec273.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "minor",
+  "comment": {
```
erinha, line 3: Can you please update this so the comment field looks like the following:

"comment": "Added `storage` capability with APIs to allows apps to discover if web storage is supported by the host client",

(please make sure to use backticks around "storage" in the comment string)

packages/teams-js/src/public/app.ts
```diff
@@ -545,7 +545,6 @@ export namespace app {
             GlobalVars.frameContext = context;
             GlobalVars.hostClientType = clientType;
             GlobalVars.clientSupportedSDKVersion = clientSupportedSDKVersion;
-
```
erinha, line 548: nit: please undo this change. We'd like the newline in the file.

packages/teams-js/src/public/storage.ts
```diff
@@ -0,0 +1,24 @@
...
+    ensureInitialized();
+    // return true as storage is always cleared on desktop.
+    if (GlobalVars.hostClientType === HostClientType.desktop) {
+      return true;
+    }
+    return isSupported();
+  }
+
+  export function isSupported(): boolean {
```
erinha, line 21: Please add reference documentation comments (you can take a look at other files for isSupported comments as reference)

packages/teams-js/test/public/storage.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  describe('isWebStorageClearedOnUserLogOut', () => {
+    
+      it('should not allow calls before initialization', () => {
+      
+      expect(storage.isWebStorageClearedOnUserLogOut).toThrowError(
+        'The library has not yet been initialized',
+      );
+    });
+   it('should always allow isWebStorageClearedOnUserLogOut calls in desktop', async () => {
+      await desktopPlatformMock.initializeWithContext(FrameContexts.content,HostClientType.desktop);
```
shrshindeMSFT, line 36: Can you please add the unit test cases for other `FrameContexts` as well?

packages/teams-js/test/public/storage.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    }) 
+  });
+  describe('isSupported', () => {
+    it('storage.isSupported should return false if the runtime says storage is not supported', () => {
+      desktopPlatformMock.setRuntimeConfig({ apiVersion: 1, supports: {} });
+      expect(storage.isSupported()).not.toBeTruthy();
+    });
+
+    it('storage.isSupported should return true if the runtime says storage is supported', async () => {
+      desktopPlatformMock.setRuntimeConfig({ apiVersion: 1, supports: {storage} });
```
shrshindeMSFT, line 53: `{ apiVersion: 1, supports: {storage} }` should be `{ apiVersion: 1, supports: {storage: {}}}`

packages/teams-js/test/public/storage.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { ErrorCode } from '../../src/public';
+import { app } from '../../src/public/app';
+import { storage } from '../../src/public/storage';
+import { Utils } from '../utils';
+import {FrameContexts,HostClientType } from '../../src/public/constants';
+import { FramelessPostMocks } from '../framelessPostMocks';
+
+describe('storage', () => {
+  const mobilePlatformMock = new FramelessPostMocks();
```
shrshindeMSFT, line 9: See this is never used to initialize with context only used in `beforeEach` and `afterEach` section. There should be two sets of tests i.e., `Framed (Utils) and Frameless (FramelessPostMocks)` can you please add the remaining ones?

shrshindeMSFT (changes requested):

packages/teams-js/src/public/storage.ts
reenudeswal: done

packages/teams-js/test/public/storage.spec.ts
reenudeswal: added UT for both Framed and Frameless.

packages/teams-js/src/public/storage.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @beta
+ */
+export namespace storage {
+  /**
+   * Checks if web storage gets cleared when a user logs out from host client
+   *
+   * @returns true is web storage gets cleared on logout and false if it does not
+   */
+  export function isWebStorageClearedOnUserLogOut(): boolean {
```
erinha, line 17: Something that came up in a different PR -- can you please add the @beta tag to all functions/interfaces/enums, etc. defined in this file? This is so the purple beta text appears correctly in the public documentation.

erinha (changes requested):
Thanks for making changes based on the PR feedback! One more comment (this issue came up in a separate PR but is applicable here) about adding more @beta tags.

packages/teams-js/src/public/storage.ts
```diff
@@ -0,0 +1,34 @@
...
+import { ensureInitialized } from '../internal/internalAPIs';
+import { HostClientType } from './constants';
+import { runtime } from './runtime';
+
+/**
+ * Contains functionality to allow web apps to store data in webview cache
+ *
+ * @beta
+ */
+export namespace storage {
```
erinha, line 11: Upon further discussion of the feature, @MarcinJuraszek and I would prefer to call this `webStorage`. Can you also please confirm that when you refer to "web storage" in the comments and the function name, you are referring to this:  https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API?

When renaming the capability, please also rename the file (webStorage.ts), test file (webStorage.spec.ts) and teams-test-app files (WebStorageAPIs.tsx).

packages/teams-js/src/public/storage.ts
reenudeswal: sure,  is this needed for isSupported function also ?

erinha (approved):
Looks good! Thanks for making all the changes.

packages/teams-js/src/public/webStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return isSupported();
+  }
+
+  /**
+   * Checks if webStorage capability is supported by the host
+   * @returns true if the webStorage capability is enabled in runtime.supports.webStorage and
+   * false if it is disabled
+   *
+   * @beta
```
shrshindeMSFT, line 33: I don't think this should be `beta` tag as  this is just returning a `boolean` value. I don't know if I'm missing something here.

erinha: It's good to have because we haven't shipped these APIs yet. We can talk offline about the process for removing @beta tags, but until we have adequate testing and are ready to mark them as production ready, we should err on the side of caution and include them, especially for new capabilities.

packages/teams-js/test/public/webStorage.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    it('webStorage.isSupported should return true if the runtime says webStorage is supported', () => {
+      desktopPlatformMock.setRuntimeConfig({ apiVersion: 1, supports: {webStorage: {}}});
+      expect(webStorage.isSupported()).toBeTruthy();
+    });
+  });
+  describe('Frameless - isSupported', () => {
+    it('webStorage.isSupported should return false if the runtime says webStorage is not supported', async () => {
+      await mobilePlatformMock.initializeWithContext(FrameContexts.task, HostClientType.ios);
```
shrshindeMSFT, line 70: Why in one it is initialized with context and previous we are not? Plus if we are using context then we need it for all of the context not only `task` context. Same goes for `HostClientType.ios`

shrshindeMSFT (changes requested):

packages/teams-js/test/public/webStorage.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { app } from '../../src/public/app';
+import { FrameContexts, HostClientType } from '../../src/public/constants';
+import { _minRuntimeConfigToUninitialize } from '../../src/public/runtime';
+import { webStorage } from '../../src/public/webStorage';
+import { FramelessPostMocks } from '../framelessPostMocks';
+import { Utils } from '../utils';
+
+describe('webStorage', () => {
+  const mobilePlatformMock = new FramelessPostMocks();
```
erinha, line 9: Would be clearer to the reader if you call this `framelessPlatformMock` and `framedPlatformMock` as per Shreyas' comment below -- they do not map 1-1 to frameless ==  mobile and utils == desktop.

packages/teams-js/test/public/webStorage.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    Object.values(FrameContexts).forEach(context => {
+      it('should never allow isWebStorageClearedOnUserLogOut calls in web for all contexts', async () => {
+        await desktopPlatformMock.initializeWithContext(FrameContexts.content, HostClientType.web);
+        const result = webStorage.isWebStorageClearedOnUserLogOut();
+        expect(result).not.toBeTruthy();
+      });
+    });
+    Object.values(FrameContexts).forEach(context => {
+      it('should allow calls isWebStorageClearedOnUserLogOut calls in mobile for all contexts', async () => {
+          await mobilePlatformMock.initializeWithContext(context);
```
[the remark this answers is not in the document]
erinha: I think adding coverage that passes version, hostClientType, and context to `initializeWithContext` (for the appropriate framed or framelessPostMock) is necessary to fully test this code.

erinha (changes requested):
Test case feedback

packages/teams-js/test/public/webStorage.spec.ts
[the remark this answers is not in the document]
reenudeswal: Added coverage for all hostClientType, flrameContext covering differentversions for both framed and frameless windows.

packages/teams-js/test/public/webStorage.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  describe('isWebStorageClearedOnUserLogOut', () => {
+    it('should not allow calls before initialization', () => {
+      expect(webStorage.isWebStorageClearedOnUserLogOut).toThrowError('The library has not yet been initialized');
+    });
+    Object.values(HostClientType).forEach(clientType => {
+      Object.values(FrameContexts).forEach(context => {
+        Object.values(testVersions).forEach(version => {
+        it(`should allow Framed - isWebStorageClearedOnUserLogOut calls for version ${version} and context ${context} and hostClientType ${clientType}`, async () => {
```
erinha, line 41: The way these tests are structured does not make it easy to read, nor accurately conveys what each test is doing.

The goal: each test case should be printed to the console with the name of what the method is testing along with the interesting parameters and the expected output.
It's also helpful if all the framed scenarios are grouped together, followed by all the positive test cases and within the positive test cases, the varying parameter input values. Then all the framed scenarios with negative test cases and all the varying  parameter input values. Then all the frameless tests that are positive and all the frameless tests that are negative.

The way it's currently structured, it prints that it should allow `isWebStorageClearedOnUserLogOut` calls for every test, even the ones for which it should not.

It's also harder to parse the output when you're reading it.

Suggestion:
Break up this code into separate describe sections, e.g.,
describe("Framed - isWebStorageClearedOnUserLogOut")

and then within that, group the positive test cases first followed by the negative test cases.

I sent you examples offline.

Also, you can run the tests and see detailed output in packages/teams-js by using this command `yarn test tests/public/webStorage.spec.ts`

erinha (dismissed):
We're getting closer! Thanks for adding detailed test coverage! Feedback on organization of the test cases.

shrshindeMSFT (approved):
LGTM!!

TrevorJoelHarris (approved):

PR #1216, opened 2022-06-17
Add SDK reference documentation validation as part of build step

erinha:
This PR adds SDK reference validation as part of the teams-js build step.

It also adds validation of symbolic links and treats warnings as errors. Essentially if you add an {@link ...} to a reference doc comment that fails, this will fail the build.

I fixed several broken symbolic links as part of this change.

I also upgraded to the latest version of typedoc to be able to specify the `notDocumented` validation option (even though it's not enabled yet).

There are a couple of things to call out here.
1. I haven't turned on the `notDocumented` validation option in typedoc.json yet. I've opened a work item to track fixing the warnings that are produced if this is enabled. There are 246 and it is not obvious how to fix about 70 of them (e.g., some refer to API parameters defined as inline functions).
2. I had to explicitly add 7 entries to "intentionallyNotExported" in typedoc.json. I thought this was better than turning off the `notExported` validation option. This is a complicated setting and I don't fully understand how we would otherwise satisfy this warning. Here are two (closed) GitHub issues that talk about it:
https://github.com/TypeStrong/typedoc/issues/1687
https://github.com/TypeStrong/typedoc/issues/1739

If I didn't add those 7 entries, these warnings are produced (even with the changes I made in tasks.ts):
```
Warning: BotUrlDialogInfo, defined at src/public/interfaces.ts:613, is referenced by tasks.getBotUrlDialogInfoFromTaskInfo.getBotUrlDialogInfoFromTaskInfo but not included in the documentation.
Warning: TaskInfo, defined at src/public/interfaces.ts:672, is referenced by tasks.getDefaultSizeIfNotProvided.getDefaultSizeIfNotProvided but not included in the documentation.
Warning: UrlDialogInfo, defined at src/public/interfaces.ts:584, is referenced by tasks.getUrlDialogInfoFromTaskInfo.getUrlDialogInfoFromTaskInfo but not included in the documentation.
Warning: OpenSingleChatRequest, defined at src/public/chat.ts:26, is referenced by chat.openChat.openChat.openChatRequest but not included in the documentation.
Warning: OpenGroupChatRequest, defined at src/public/chat.ts:40, is referenced by chat.openGroupChat.openGroupChat.openChatRequest but not included in the documentation.
Warning: DialogSize, defined at src/public/interfaces.ts:569, is referenced by dialog.update.resize.resize.dimensions but not included in the documentation.
Warning: MediaControllerEvent, defined at src/public/media.ts:462, is referenced by media.VideoController.notifyEventToApp.notifyEventToApp.mediaEvent but not included in the documentation.
error Command failed with exit code 4.
```

Note that even though it is flagged in a warning, the documentation generated for `dialog.update.resize` (as an example) does produce a link to `DialogSize` that resolves in our public docs.

I believe this addition even with the caveats is better than what we currently have (relying on intrepid developers to check themselves that the docs are in good shape as part of their PRs).

### DIFF
--- a/packages/teams-js/src/public/tasks.ts
+++ b/packages/teams-js/src/public/tasks.ts
@@ -18,9 +18,8 @@ import { BotUrlDialogInfo, DialogInfo, DialogSize, TaskInfo, UrlDialogInfo } fro
 export namespace tasks {
   /**
    * @deprecated
-   * As of 2.0.0, please use {@link dialog.open(urlDialogInfo: UrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} for url based dialogs
-   * and {@link dialog.bot.open(botUrlDialogInfo: BotUrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): PostMessageChannel} for bot based dialogs.
-   *
+   * As of 2.0.0, please use {@link dialog.open dialog.open(urlDialogInfo: UrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): void} for url based dialogs
+   * and {@link dialog.bot.open dialog.bot.open(botUrlDialogInfo: BotUrlDialogInfo, submitHandler?: DialogSubmitHandler, messageFromChildHandler?: PostMessageChannel): void} for bot based dialogs.
    * Allows an app to open the task module.
    *
    * @param taskInfo - An object containing the parameters of the task module
--- a/packages/teams-js/src/public/tasks.ts
+++ b/packages/teams-js/src/public/tasks.ts
@@ -65,7 +65,7 @@ export namespace tasks {
 
   /**
    * @deprecated
-   * As of 2.0.0, please use {@link dialog.submit dialog.submit(result?: string | object, appIds?: string | string[]): void} instead.
+   * As of 2.0.0, please use {@link dialog.submit} instead.
    *
    * Submit the task module.
    *
@@ -76,6 +76,11 @@ export namespace tasks {
     dialog.submit(result, appIds);
   }
 
+  /**
+   * Converts TaskInfo to UrlDialogInfo
+   * @param taskInfo - TaskInfo object to convert
+   * @returns - Converted UrlDialogInfo object
+   */
   export function getUrlDialogInfoFromTaskInfo(taskInfo: TaskInfo): UrlDialogInfo {
     const urldialogInfo: UrlDialogInfo = {
       url: taskInfo.url,
@@ -89,6 +94,11 @@ export namespace tasks {
     return urldialogInfo;
   }
 
+  /**
+   * Converts TaskInfo to BotUrlDialogInfo
+   * @param taskInfo - TaskInfo object to convert
+   * @returns - converted BotUrlDialogInfo object
+   */
   export function getBotUrlDialogInfoFromTaskInfo(taskInfo: TaskInfo): BotUrlDialogInfo {
     const botUrldialogInfo: BotUrlDialogInfo = {
       url: taskInfo.url,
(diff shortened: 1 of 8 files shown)


AE-MS:
I love this change. Thank you for doing it.

packages/teams-js/src/public/tasks.ts
```diff
@@ -103,6 +112,13 @@ export namespace tasks {
     };
     return botUrldialogInfo;
   }
+
+  /**
+   * Sets the height and width of the TaskInfo object to the original height and width, if initially specified,
+   * otherwise uses the height and width values corresponding to TaskModuleDimension.Small
```
AE-MS, line 118: >TaskModuleDimension.Small [](http://example.com/codeflow?start=65&length=25)

Should this somehow be a hyperlinked reference to the definition of this?

AE-MS: (similar question about references to `TaskInfo` and `UrlDialogInfo` in new docs above this)

AE-MS:
> Note that even though it is flagged in a warning, the documentation generated for `dialog.update.resize` (as an example) does produce a link to `DialogSize` that resolves in our [public docs](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/dialog.update?view=msteams-client-js-latest#@microsoft-teams-js-dialog-update-resize).

Since you are turning on warnings as errors, should this cause an error and failure? I didn't see a reference to `dialog.update.resize` in the exclusion list in `typedoc.json`. Maybe I'm misunderstanding?

AE-MS (changes requested):
:clock1:

AE-MS (dismissed):
:shipit:

packages/teams-js/src/public/tasks.ts
```diff
@@ -66,7 +65,7 @@ export namespace tasks {
 
   /**
    * @deprecated
-   * As of 2.0.0, please use {@link dialog.submit dialog.submit(result?: string | object, appIds?: string | string[]): void} instead.
+   * As of 2.0.0, please use {@link dialog.submit} instead.
```
erinha, line 68: @erikadoyle I was not able to get this @link tag to resolve (and by "resolve" I mean stop producing a warning when I run "typedoc --emit none --validation.invalidLink src/index.ts") without this change. The function signature was copied exactly from the dialog.submit function so I can't figure out why this isn't working.

The [public documentation](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/tasks?view=msteams-client-js-latest#@microsoft-teams-js-tasks-submittask) doesn't resolve to dialog.submit even though the link doesn't appear obviously broken.
Do you know how to get this to work?

erinha:
> > Note that even though it is flagged in a warning, the documentation generated for `dialog.update.resize` (as an example) does produce a link to `DialogSize` that resolves in our [public docs](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/dialog.update?view=msteams-client-js-latest#@microsoft-teams-js-dialog-update-resize).
> 
> Since you are turning on warnings as errors, should this cause an error and failure? I didn't see a reference to `dialog.update.resize` in the exclusion list in `typedoc.json`. Maybe I'm misunderstanding?

The reference is to `DialogSize`, and `DialogSize` is the excluded type, e.g. 
```
Warning: DialogSize, defined at src/public/interfaces.ts:569, is referenced by dialog.update.resize.resize.dimensions but not included in the documentation.
```
And `DialogSize` does indeed have [reference documentation](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/dialogsize?view=msteams-client-js-latest).

packages/teams-js/src/public/tasks.ts
erinha: Made the change, appreciate the feedback!

AE-MS (approved):

packages/teams-js/src/public/tasks.ts
erikadoyle: Hmm, looks like none of the @deprecated note links in [tasks module](https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/tasks?view=msteams-client-js-latest#@microsoft-teams-js-tasks-submittask) are working. I took a look but couldn't discern why `{@link dialog.submit} isn't resolving to the intended target:

https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/dialog?view=msteams-client-js-latest#@microsoft-teams-js-dialog-submit

. . . and instead reloading the current page:

https://docs.microsoft.com/en-us/javascript/api/@microsoft/teams-js/tasks?view=msteams-client-js-latest

FWICT this is the correct syntax per [Typedoc link resolution reference](https://typedoc.org/guides/link-resolution/). I'll ask MS Docs engineering about this. Hoping this is a publishing pipeline fix; how about we move forward on that assumption, and file a new task if any link source updates are needed?

erinha: Thanks, that sounds good to me. Just so I'm being clear, looking at the source code, it looked like what was checked in _before_ my change should have worked properly. The function signature matched the destination of the link exactly, so I would have expected it was working, but as you mentioned it is not resolving.

jadahiya-MSFT (approved):

Issue #1210, opened 2022-06-13
optional submitHandler in dialog.open not checked before use 


antpowell:
Although submitHandler is optional in the signature of dialog.open, it doesn't get checked before trying to use it in sendMessageToParent function causing an error and fail out without an explanation why.

[dialog.ts](https://github.com/OfficeDev/microsoft-teams-library-js/blob/release/2.0.0/packages/teams-js/src/public/dialog.ts)

error message from console:
![image](https://user-images.githubusercontent.com/10569986/173418874-3cf4f561-e654-4b54-aa77-53fc55db5def.png)
![image](https://user-images.githubusercontent.com/10569986/173419483-ef6f1a76-7b40-4c41-9299-773bfc3edec4.png)

MarcinJuraszek:
@lakhveerk can you take a look?

lakhveerk:
Thank you for catching this Anthony! I am working to fix the bug :)

MarcinJuraszek:
Closing now that the fix is in. This will be released with next version.

PR #1217, opened 2022-06-20
check if submitHandler exists before calling it

lakhveerk:
It was a bug in dialog file, where an optional callback submitHadler was called without checking if it exists. The issue was reported on github: https://github.com/OfficeDev/microsoft-teams-library-js/issues/1210


packages/teams-js/src/public/dialog.ts
```diff
@@ -88,10 +88,13 @@ export namespace dialog {
...
     }
     const dialogInfo: DialogInfo = getDialogInfoFromUrlDialogInfo(urlDialogInfo);
-    sendMessageToParent('tasks.startTask', [dialogInfo], (err: string, result: string | object) => {
-      submitHandler({ err, result });
-      removeHandler('messageForParent');
-    });
+    const submitHandlerHelper = submitHandler
+      ? (err: string, result: string | object) => {
+          submitHandler({ err, result });
+          removeHandler('messageForParent');
```
MarcinJuraszek, line 94: Shouldn't we remove handler for `messageForParent` even if no submit Handler was provided?

packages/teams-js/src/public/dialog.ts
```diff
@@ -88,10 +88,13 @@ export namespace dialog {
...
+        }
+      : null;
```
MarcinJuraszek, line 96: Whenever I see `null` being used I really want to know if there's a reason for it to be `null` and not `undefined`.

MarcinJuraszek (changes requested):

packages/teams-js/src/public/dialog.ts
lakhveerk: Yes, that is true. Will update it

packages/teams-js/src/public/dialog.ts
```diff
@@ -88,10 +88,13 @@ export namespace dialog {
...
-    sendMessageToParent('tasks.startTask', [dialogInfo], (err: string, result: string | object) => {
-      submitHandler({ err, result });
-      removeHandler('messageForParent');
-    });
+    const submitHandlerHelper = submitHandler
+      ? (err: string, result: string | object) => {
+          submitHandler({ err, result });
+          removeHandler('messageForParent');
+        }
+      : () => removeHandler('messageForParent');
```
lakhveerk, line 96: Interestingly, In javascript/typescript we can call a function with more arguments than there are parameters. The extra arguments are simply ignored. 
https://www.typescriptlang.org/docs/handbook/2/functions.html

In case of no submitHandler, the callback will have zero parameters, but still be called with 2 arguments. And, it would work fine!

MarcinJuraszek (dismissed):

jadahiya-MSFT (approved):

MarcinJuraszek (dismissed):

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -132,6 +133,50 @@ describe('Dialog', () => {
...
+        it(`FRAMED: Should successfully call the callback when dialog.closed: ${context}`, async () => {
+          await framedMock.initializeWithContext(FrameContexts.content);
+          const urlDialogInfo: UrlDialogInfo = {
+            url: 'someUrl',
+            size: { height: DialogDimension.Large, width: DialogDimension.Large },
+            title: 'someTitle',
+            fallbackUrl: 'someFallbackUrl',
+          };
+          const submitString = 'succesfullySubmit';
+
```
shrshindeMSFT, line 146: Can you add assertions to make sure `try catch` is executed once?

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -132,6 +133,50 @@ describe('Dialog', () => {
...
+          dialog.open(urlDialogInfo);
+          const message = framelessMock.findMessageByFunc('tasks.startTask');
+
```
shrshindeMSFT, line 167: Same as above

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -492,6 +537,52 @@ describe('Dialog', () => {
...
+          await framedMock.initializeWithContext(FrameContexts.content);
+          const botUrlDialogInfo: BotUrlDialogInfo = {
+            url: 'someUrl',
+            size: { height: DialogDimension.Large, width: DialogDimension.Large },
+            title: 'someTitle',
+            fallbackUrl: 'someFallbackUrl',
+            completionBotId: 'botId',
+          };
+          const submitString = 'succesfullySubmit';
+
```
shrshindeMSFT, line 551: Here as well `expect.assertions`

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -492,6 +537,52 @@ describe('Dialog', () => {
...
+          dialog.open(botUrlDialogInfo);
+          const message = framelessMock.findMessageByFunc('tasks.startTask');
+
```
shrshindeMSFT, line 573: Same here as well.

shrshindeMSFT (changes requested):
Why to use `except.assertions` when using try catch block?

https://stackoverflow.com/questions/50816254/necessary-to-use-expect-assertions-if-youre-awaiting-any-async-function-calls#:~:text=The%20assertions%20in%20the%20catch%20block%20won%27t%20run,catch%20don%27t%20run%2C%20we%20get%20a%20failure.%20Share

jadahiya-MSFT (dismissed):

change/@microsoft-teams-js-0cceb483-48e4-4214-92f0-88ffa362f792.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Making sure submitHandler exists before calling it in dialog.open, dialog.bot.open and tasks.startTask API.",
```
erinha, line 3: Nit: please use past tense and surround API names with backticks ``

packages/teams-js/test/public/dialog.spec.ts
lakhveerk: I don't expect it to go to the catch block. So, it won't hit any expect statement. 
That is the reason, I think assertions would not work on these tests. 
Feel free to show me an example of how it can be done (if I am missing something).

lakhveerk: @erinha @shrshindeMSFT

jadahiya-MSFT (approved):

shrshindeMSFT (approved):
LGTM!!

PR #1220, opened 2022-06-22
Adding in View Published Packages Urls Powershell script

jadahiya-MSFT:
Currently in our ADO pipeline we are using this powershell script and needs to be changed into a published artifact instead. Adding in a powershell script and publishing it to a scripts folder during our build step will accomplish this.


erinha (approved):
Looks great, thanks!

TrevorJoelHarris (approved):

PR #1223, opened 2022-06-27
Upload logs to blob container directly when SAS is provided

shilcare:
Update the private logging API to support uploading app logs to the blob container directly when the SAS token is given instead of sending the logs back to the host.


TrevorJoelHarris:
I believe you are currently considering going through or currently going through security review for this change. Please re-open this PR once you have received approval from the security team for this approach. You should already have all the information you need to go through that process, but let me know if I can help.

PR #1227, opened 2022-06-29
Sample App

AnanyaBarthakur1:
Basic app with msal auth, msGraph, fluentUI


apps/sample-app/index.html
```diff
@@ -0,0 +1,32 @@
...
+    <meta
+      name="description"
+      content="Teams Perf Test App created using React"
+    />
+    <!--
+      manifest.json provides metadata used when your web app is installed on a
+      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
+    -->
+    <link rel="manifest" href="./public/manifest.json" />
+    <title>Teams Perf Test App</title>
```
MarcinJuraszek, line 16: Same here, can you make sure `title` matches what the app is meant to be?

apps/sample-app/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "sample-app",
+  "private": true,
+  "author": "Msft",
```
MarcinJuraszek, line 4: should be `Microsoft Teams`

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  React.useEffect(() => {
+    (async () => {
+      await app.initialize();
+      setIsInitialized(true);
+    })();
+  }, [setIsInitialized]);
+
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+  async function handle() {
```
MarcinJuraszek, line 28: should probably use `React.useCallback` instead.

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const result = await authentication.authenticate({
+      url: 'https://localhost:4003/?auth=1',
+    });
+    setAccessToken(result);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
MarcinJuraszek, line 35: This is not used, to can you just remove it instead of working around an ESLint error with a disable-next-line?

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const [context, setContext] = useState<app.Context>();
+  const [currTheme, setCurrTheme] = useState<Theme>(teamsLightTheme);
+
+  React.useEffect(() => {
+    (async () => {
+      try {
+        await app.initialize();
```
MarcinJuraszek, line 42: You have two separate places where `app.initialize()` is called. There should only be one.

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+  async function handle() {
+    const result = await authentication.authenticate({
+      url: 'https://localhost:4003/?auth=1',
+    });
+    setAccessToken(result);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [context, setContext] = useState<app.Context>();
+  const [currTheme, setCurrTheme] = useState<Theme>(teamsLightTheme);
```
MarcinJuraszek, line 37: the initial theme should be taken from `getContext` and later optionally updates via `registerOnThemeChangeHandler`

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+            case 'contrast':
+              setCurrTheme(teamsHighContrastTheme);
+              break;
+            case 'default':
+            default:
+              setCurrTheme(teamsLightTheme);
+          }
+        });
+      } catch (e) {
+        console.error(e);
```
MarcinJuraszek, line 61: If app initialization fails, you probably want to do something more than just print it to console.

apps/sample-app/src/AuthApp.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import TokenFetchComponent from './components/TokenFetch';
+import { PageLayout } from './PageLayout';
+
+const AuthApp: React.FC = () => {
+  const [isInitialized, setIsInitialized] = React.useState(false);
+
+  React.useEffect(() => {
+    (async () => {
+      await app.initialize();
```
MarcinJuraszek, line 16: Would be great to handle errors here too

apps/sample-app/src/PageLayout.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { SignInButton } from './components/Button';
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const PageLayout = props => {
+  const isAuthenticated = useIsAuthenticated();
+
+  return (
+    <>
+      {isAuthenticated ? <span>Signed In</span> : <SignInButton />}
```
MarcinJuraszek, line 13: you're using `SignInButton` as if it was a react component, but I don't think it is.

apps/sample-app/src/components/Profile.tsx
```diff
@@ -0,0 +1,53 @@
...
+      const client = Client.init(options);
+      const response = await client.api('/me/messages').get();
+      const messages = response.value as Message[];
+      setMessages(messages);
+    })();
+  }, [accessToken, setMessages]);
+
+  React.useEffect(() => {
+    (async () => {
+      const authProvider: AuthProvider = (callback: AuthProviderCallback) => {
```
MarcinJuraszek, line 34: Do we need to do this initialization twice, once for messages and once for user info?

package.json
```diff
@@ -103,5 +103,8 @@
     "url-parse": "^1.5.0",
     "y18n": "^4.0.1"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@fluentui/react": "^8.75.1",
```
MarcinJuraszek, line 107: These should be in the package.json for sample app, not the top-level one.

aeunms: I looked through the other apps and they did have their own dependencies and devDependencies, should I create a ticket to get that cleaned up for all apps?

aeunms: There were just few if that helps

MarcinJuraszek: hmm, I don't think it matters for `devDependencies`, and for `dependencies` it looks like it used to be an empty set `{}` and you're adding first items here. So not sure if there's work needed to clean this up for other apps. Just make sure you move the fluentui dependencies you're adding to the sample app's package.json.

apps/sample-app/src/components/Profile.tsx
```diff
@@ -0,0 +1,53 @@
...
+        callback(undefined, accessToken);
+      };
+      const options: Options = { authProvider };
+      const client = Client.init(options);
+      const response2 = await client.api('/me/').get();
```
TrevorJoelHarris, line 39: 2 as a suffix on response isn't necessary, this response is in a different scope than the one on line 26

apps/sample-app/src/components/TokenFetch.tsx
```diff
@@ -0,0 +1,32 @@
...
+    };
+    instance.acquireTokenSilent(request).then(response => {
+      setAccessToken(response.accessToken);
+      authentication.notifySuccess(response.accessToken);
+    });
+  }, [setAccessToken, accounts, instance]);
+
+  return (
+    <>
+      {!accessToken && <p>Fetching access token...</p> && <Spinner size={SpinnerSize.large} />}
```
TrevorJoelHarris, line 26: nit: Patterns like this where you check a boolean and then print UI and then on the next line check the opposite and print different UI can sometimes be a little easier to read and maintain if you write them like this:
```suggestion
      {!accessToken ? <><p>Fetching access token...</p><Spinner size={SpinnerSize.large} /></> : <p>...</p>}
```

apps/sample-app/src/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { PublicClientApplication } from '@azure/msal-browser';
+import { MsalProvider } from '@azure/msal-react';
+import { FluentProvider, teamsLightTheme } from '@fluentui/react-components';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import App from './App';
+import AuthApp from './AuthApp';
+import { msalConfig } from './components/authConfig';
+const msalInstance = new PublicClientApplication(msalConfig);
```
TrevorJoelHarris, line 12: We should construct msalInstance inside of the if block at line 18 so we only initialize this when we're actually going to use it

package.json
```diff
@@ -103,5 +103,8 @@
...
+    "@fluentui/react-northstar": "^0.63.1"
```
TrevorJoelHarris, line 108: Probably we shouldn't use northstar and "normal" react. We should just pick one flavor of react and use that (and I would vote for "not northstar")

TrevorJoelHarris (changes requested):

TrevorJoelHarris (commented):
Marking as comment and not Request changes because I will be out next week and I don't want to hold up your check in.

apps/sample-app/README.md
```diff
@@ -0,0 +1,9 @@
+# Sample App
```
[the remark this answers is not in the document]
MarcinJuraszek: Let's track that for future changes, and not block this PR on it

AE-MS:
>   "name": "teams-js-monorepo",

This is probably not correct.

---
Refers to: package.json:2 in 8ec63e2. [](commit_id = 8ec63e25abec1da39555c56281aa12c1a040e5f3, deletion_comment = False)

apps/sample-app/public/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
+  "$schema": "https://developer.microsoft.com/en-us/json-schemas/teams/v1.10/MicrosoftTeams.schema.json",
+  "manifestVersion": "1.10",
+  "version": "0.2",
+  "id": "8142d2d1-4113-411f-868f-8e6e606e8827",
+  "packageName": "sample app",
+  "developer": {
+    "name": "Ananya Barthakur",
```
AE-MS, line 8: >   "name": "Ananya Barthakur", [](http://example.com/codeflow?start=1&length=30)

This may not be the right name to put here. Consider "Microsoft"?

apps/sample-app/src/components/Button.tsx
```diff
@@ -0,0 +1,15 @@
+import { useMsal } from '@azure/msal-react';
+
+import { loginRequest } from './authConfig';
+
+function handleLogin(instance): void {
```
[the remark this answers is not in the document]
AnanyaBarthakur1: Yes, this "SignInButton" component serves as an automatic sign in for the user and is not meant to show an actual button. This is to improve the user experience by removing the need to press another button once the authentication page opens up

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const handle = React.useCallback(async () => {
+    const result = await authentication.authenticate({
+      url: 'https://localhost:4003/?auth=1',
+    });
+    setAccessToken(result);
+  }, [setAccessToken]);
+
+  const [currTheme, setCurrTheme] = useState<Theme>(teamsLightTheme);
+
+  React.useEffect(() => {
```
MarcinJuraszek, line 34: I don't think React guarantees the order in which `useEffect` callbacks are executed. So I'm wondering if there's a potential for this one to run before the one on line 14 is done, which would result in API calls being made before the app is properly initialized. 

Can you move all this to a single useEffect instead?

apps/sample-app/public/manifest.json
```diff
@@ -0,0 +1,52 @@
...
+        "meetingDetailsTab",
+        "meetingSidePanel",
+        "meetingStage"
+      ]
+    }
+  ],
+  "validDomains": ["localhost:4003"],
+  "permissions": ["identity", "messageTeamMembers"],
+  "devicePermissions": ["media", "geolocation"],
+  "showLoadingIndicator": false
```
MarcinJuraszek, line 51: Can you please track an item for the future to enable `showLoadingIndicator` for the sample app?

apps/sample-app/src/App.tsx
```diff
@@ -0,0 +1,68 @@
...
+    (async () => {
+      await app.initialize();
+      setIsInitialized(true);
+    })();
+  }, [setIsInitialized]);
+
+  const handle = React.useCallback(async () => {
+    const result = await authentication.authenticate({
+      url: 'https://localhost:4003/?auth=1',
```
MarcinJuraszek, line 27: hardcoding this to `https://localhost:4003` means if we were to deploy it somewhere (like Azure) it would stop working. Can you use relative paths instead? Using `URL` to get one might be a good idea: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL. I haven't tried it, but something like `(new URL('/?auth=1')).toString()` might work

apps/sample-app/src/components/TokenFetch.tsx
```diff
@@ -0,0 +1,41 @@
...
+  React.useEffect(() => {
+    const request = {
+      ...loginRequest,
+      account: accounts[0],
+    };
+    instance.acquireTokenSilent(request).then(response => {
+      setAccessToken(response.accessToken);
+      authentication.notifySuccess(response.accessToken);
+    });
+    instance.acquireTokenSilent(request).catch(response => {
```
MarcinJuraszek, line 22: You most likely don't want to call this again, instead you should attach the `catch()` to the initial `acquireTokenSilent` promise:

```
instance.acquireTokenSilent(request).then(response => {
      setAccessToken(response.accessToken);
      authentication.notifySuccess(response.accessToken);
    }).catch(response => { ... });
```

apps/sample-app/src/components/Profile.tsx
```diff
@@ -0,0 +1,60 @@
...
+      };
+      console.log('starting info');
+      const options: Options = { authProvider };
+      const client = Client.init(options);
+      // User Profile Info
+      const userResponse = await client.api('/me').get();
+      setUserInfo(userResponse);
+
+      // Calendar Info
+      /// get date and time (maybe something like this can go into utils)
```
MarcinJuraszek, line 32: yes, please move this to a utility function

apps/sample-app/src/components/utils.tsx
```diff
@@ -0,0 +1,28 @@
...
+  app.notifyFailure({
+    reason: app.FailedReason.Other,
+    message: 'App initialization failed',
+  });
+}
+
+export function getDates(): [string, string] {
+  const current = new Date();
+  const currDate = current.getFullYear() + '-' + (current.getMonth() + 1) + '-' + current.getDate();
+  const tomorrowDate = current.getFullYear() + '-' + (current.getMonth() + 1) + '-' + (current.getDate() + 1);
```
aeunms, line 26: What happens if currDate is the last day of the month? e.g. January 31st?

AnanyaBarthakur1: Oh right! Thanks for pointing this out to me. I have changed the code to use tomorrow.setDate(current.getDate() + 1); (getDate will get the right 'day number', and setDate will get the month and year accordingly)

aeunms (dismissed):
🚢  Great job getting so many great changes in in such a short time!! :)

apps/sample-app/src/components/Calendar.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { Button } from '@fluentui/react-components';
+import { calendar } from '@microsoft/teams-js';
+import React from 'react';
+
+export const Calendar: React.FunctionComponent = () => {
+  const handleCalendar = async (): Promise<void> => {
+    // create calendar paramaters
+    const calendarParams: calendar.ComposeMeetingParams = {
+      attendees: ['t-abarthakur@microsoft.com'],
```
TrevorJoelHarris, line 11: You may want to replace this with a placeholder value instead of using your real address.

TrevorJoelHarris: Something like 'emailaddress@company.com'

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    <div key={c}>
+                      {c === 'Call' && (
+                        <Menu>
+                          <MenuTrigger>
+                            <MenuItem>Call</MenuItem>
+                          </MenuTrigger>
+                          <MenuPopover>
+                            <MenuList>
+                              <MenuItem onClick={() => handleAudioCall(a)}>Audio {c}</MenuItem>
+                              <MenuItem onClick={() => handleVideoCall(a)}> Video {c}</MenuItem>
```
TrevorJoelHarris, line 80: Intentional space ahead of the word Video?

TrevorJoelHarris: Similar with the space ahead of {c} on line 85

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -0,0 +1,111 @@
...
+              </button>
+            </Text>
+          </Tooltip>
+        ))}
+      </div>
+    );
+  };
+
+  return (
+    <div>
```
TrevorJoelHarris, line 107: I don't think this div is needed, AvatarExample starts with a div

apps/sample-app/src/components/Profile.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { accessToken } = props;
+  const [userInfo, setUserInfo] = React.useState<User>();
+  const [calendar, setCalendar] = React.useState<Calendar>();
+
+  React.useEffect(() => {
+    (async () => {
+      const authProvider: AuthProvider = (callback: AuthProviderCallback) => {
+        callback(undefined, accessToken);
+      };
+      console.log('starting info');
```
TrevorJoelHarris, line 25: probably good to remove this console.log unless it's giving consumers of these samples some info they need to understand what's going on

apps/sample-app/src/components/TokenFetch.tsx
```diff
@@ -0,0 +1,43 @@
...
+      .catch(response => {
+        authentication.notifyFailure(response);
+      });
+  }, [setAccessToken, accounts, instance]);
+
+  return (
+    <>
+      {!accessToken ? (
+        <>
+          <p>Fetching access token...</p>
```
TrevorJoelHarris, line 33: <p> tags should be <Text> from fluent

apps/sample-app/src/components/utils.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+};
+export function appInitializationFailed(): void {
+  app.notifyFailure({
+    reason: app.FailedReason.Other,
+    message: 'App initialization failed',
+  });
+}
+
+export function getDates(): [string, string] {
```
TrevorJoelHarris, line 24: this is very minor, but having this return an interface like:

```
interface DateCollection {
  today: string;
  tomorrow: string;
}
```

Would be more user friendly and easier to read. I'll let you decide whether that's needed in this sample app.

TrevorJoelHarris (approved):

aeunms:
Merging to get the test app in and make deploying a lot easier. The changes suggested by @TrevorJoelHarris will be integrated in a future PR!

aeunms (approved):

aeunms (approved):

AE-MS (approved):

PR #1230, opened 2022-06-29
Test app: Fix the validation of getExternalProviders API

anmaur:
- Fix the validation of getExternalProviders API
- Extend the ApiWithCheckBoxInput to ingest setResult callback which allows callback based APIs to set the result of the text box.

### DIFF
--- a/apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
@@ -133,13 +133,13 @@ const GetExternalProviders = (): React.ReactElement =>
     name: 'getExternalProviders',
     title: 'Get External Providers',
     label: 'excludeAddedProviders',
-    onClick: async (excludeAddedProviders: boolean) => {
+    onClick: async (excludeAddedProviders: boolean, setResult: (result: string) => void) => {
       let result;
       const callback = (error: SdkError, providers: files.IExternalProvider[]): void => {
         if (error) {
-          result = JSON.stringify(error);
+          setResult(JSON.stringify(error));
         } else {
-          result = JSON.stringify(providers);
+          setResult(JSON.stringify(providers));
         }
       };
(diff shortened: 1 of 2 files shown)


apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
```diff
@@ -133,13 +133,13 @@ const GetExternalProviders = (): React.ReactElement =>
     name: 'getExternalProviders',
     title: 'Get External Providers',
     label: 'excludeAddedProviders',
-    onClick: async (excludeAddedProviders: boolean) => {
+    onClick: async (excludeAddedProviders: boolean, setResult: (result: string) => void) => {
       let result;
```
MarcinJuraszek, line 137: Is this really used anymore?

anmaur: Yes, For the promise based apis which still uses callback

MarcinJuraszek: I get that for the generic case, but in this particular one, you're doing `let result` and at the end `return result` without actually assigning anything to it.

apps/teams-test-app/src/components/privateApis/FilesAPIs.tsx
```diff
@@ -133,13 +133,13 @@ const GetExternalProviders = (): React.ReactElement =>
...
       const callback = (error: SdkError, providers: files.IExternalProvider[]): void => {
         if (error) {
-          result = JSON.stringify(error);
+          setResult(JSON.stringify(error));
         } else {
-          result = JSON.stringify(providers);
+          setResult(JSON.stringify(providers));
         }
       };
       await files.getExternalProviders(excludeAddedProviders, callback);
```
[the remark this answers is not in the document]
anmaur: makes sense. pushed a cleanup

jadahiya-MSFT (approved):

PR #1231, opened 2022-06-29
Updated view url powershell script to take in a parameter

jadahiya-MSFT:
$(System.DefaultWorkingDirectory) is not a recognized path when running in a powershell script/file. Changing this to be passed in as an argument instead.


shrshindeMSFT (approved):

packages/teams-js/view-published-packages-urls.ps1
```diff
@@ -1,4 +1,4 @@
-$version = Get-ChildItem -Path $(System.DefaultWorkingDirectory)/_OfficeDev.microsoft-teams-library-js/CDNFeed -Directory -Name
+$version = Get-ChildItem -Path ../CDNFeed -Directory -Name
```
anmaur, line 1: This relies on CDNFeed to be present at parent layer. Would it be possible to provide the parent directory as param?

erinha: I think Anupam makes a good suggestion. The pipeline originally used an inline script (not a file) and we were able to evaluate the variable that way. I'm curious if you can just pass the entire directory parent of CDNFeed to the script, as Anupam suggests, like:

"$(System.DefaultWorkingDirectory)/_OfficeDev.microsoft-teams-library-js"

(Maybe I'm missing something, but I don't follow why the script being published to the directory you mention would make passing a variable to the script not feasible. I'm happy to learn more!)

jadahiya-MSFT: Sorry it was me that made the mistake. I saw it as part of the script itself rather than as a param and I think that will work. 

The only thing that comes to mind of why it might not work is if the Powershell this script is being run in doesn't have the AAD specific environmental variables. Which seems to be the case with running the current instance of the compliant Powershell. This can be solved by rmaking a Get-Module request.

jadahiya-MSFT: Just ran this and sending the $(System.DefaultWorkingDirectory) as an argument/parameter into the script worked :)

erinha (approved):
🚀

MarcinJuraszek:
Fun fact, you should be able to read System.DefaultWorkingDirectory directly in PowerShell, but the syntax would be different. Sett https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.2

lakhveerk (approved):

PR #1232, opened 2022-06-29
update register on remove handler to be used in the hub sdk

Jessica-Koch:
PR to add elements for new e2e tests satisfying testing registerOnRemoveHandler failure


apps/teams-test-app/src/components/PagesConfigAPIs.tsx
```diff
@@ -106,6 +106,27 @@ const RegisterOnRemoveHandler = (): React.ReactElement =>
     },
   });
 
+const RegisterOnRemoveHandlerFailure = (): React.ReactElement =>
```
jadahiya-MSFT, line 109: Is there no way to fuse this with the above function? I.e if we are sent a failure notifyFailure and if sent a successful response notifySuccess?

Jessica-Koch: yes, was trying to avoid actually writing cypress tests :D but I should just write it correctly

jadahiya-MSFT (dismissed):

apps/teams-test-app/src/components/PagesConfigAPIs.tsx
```diff
@@ -86,27 +86,44 @@ const SetValidityState = (): React.ReactElement =>
...
-      },
-      withCallback: setResult => {
-        settings.registerOnRemoveHandler((removeEvent: settings.RemoveEvent): void => {
-          setResult('Remove event received.');
-          removeEvent.notifySuccess();
-        });
+      validateInput: () => undefined,
+      submit: {
+        withPromise: async (input, setResult) => {
+          if (input) {
```
erinha, line 96: How does this impact the existing e2e tests? Without changes to the tests that call this method, I would expect them not to work the same (and possibly fail?).

We have a test that uses this `RegisterOnRemoveHandler` element currently that is testing the `notifySuccess` path. But it doesn't expect to provide input, so we wouldn't end up calling `removeEvent.notifySuccess` like we currently do and would instead call `removeEvent.notifyFailure`. Unless I'm missing something (which is totally possible).

jadahiya-MSFT: This would definitely depend on the cypress test, which I believe Jess has mentioned she is making changes to currently.

Jessica-Koch: you know what, you're right, this will break all of them because they aren't looking for an input, originally I had these split into two different tests but jay suggested combining them.  I can see either being the way forward -

Option 1: I'll put the old test back so that the hub tests won't break, then make an update to the hub and finish this by making a PR to the teams sdk again. 

Option 2: split this test into a success case and a failure case then  open my PR for the hub sdk. 

How would you like to proceed?

jadahiya-MSFT: When I made my comment I originally thought the flow of execution would be as such:
TeamsJS SDK communicates desire to registerOnRemoveHandler. Upon removal the codepath would either pass or fail depending on if the remove is properly executed to where this SDK communicates to. From there based on the execution we would either `removeEvent.notifySuccess` or `removeEvent.notifyFailure` (i.e Remove Event has either passed or failed). 

After a discussion with Jess, I realized that this isn't what was needed and instead we would want to either force a failure or success scenario, as such I think it would be best to proceed with option 2 since they are two separate things we want to achieve rather than one.

apps/teams-test-app/src/components/PagesConfigAPIs.tsx
```diff
@@ -85,10 +85,10 @@ const SetValidityState = (): React.ReactElement =>
     },
   });
 
-const RegisterOnRemoveHandler = (): React.ReactElement =>
+const RegisterOnRemoveHandlerSuccess = (): React.ReactElement =>
```
erinha, line 88: We're going to need to keep the `RegisterOnRemoveHandler` element while we add the others or the tests looking for it will fail. :)

apps/teams-test-app/src/components/PagesConfigAPIs.tsx
```diff
@@ -106,7 +106,28 @@ const RegisterOnRemoveHandler = (): React.ReactElement =>
     },
   });
 
-const RegisterOChangeConfigHandler = (): React.ReactElement =>
```
erinha, line 109: Similarly, I love that you are fixing this typo, but you have to keep the old named element around until you update the cypress tests to look for the correctly spelled element.

jadahiya-MSFT: Correct me if I'm wrong, but I think fixing this typo should be fine since the cypress test actually looks for `config_registerChangeConfigsHandler` by examining the DOM

erinha: Oh right, good catch!

Jessica-Koch: this is a merge mistake, my git lens was showing that I added an 'O' so I thought maybe I was the one that introduced the typo.  My mistake

erinha (approved):
LGTM!

jadahiya-MSFT (approved):

PR #1238, opened 2022-07-08
Modified npmrc file to read the more secure secret for NPM-TOKEN

jadahiya-MSFT:
We can't dereference the old value with Lockbox requests without manually setting an environment variable with the value. Testing if dereferencing the generated secret variable instead.


lakhveerk (approved):

erinha (approved):
🚀

PR #1240, opened 2022-07-08
Modifying powershell scripts to generate npmrc file

jadahiya-MSFT:


packages/teams-js/scripts/add-npmrc-npmtoken.ps1
```diff
@@ -0,0 +1 @@
+Set-Content -Path $Env:SYSTEM_DEFAULTWORKINGDIRECTORY/_OfficeDev.microsoft-teams-library-js/NPMFeed/.npmrc -Value "//registry.npmjs.org/:_authToken=$NPM_TOKEN"
```
erinha, line 1: Does this change mean we don't need the .npmrc_publish file anymore? If so, can you please remove it?

jadahiya-MSFT: Aah, for some reason I thought `Set-Content` needed the file to exist in the first place, which it doesn't. I will remove .npmrc_publish and the corresponding yaml scripts

TrevorJoelHarris (approved):

erinha (approved):

PR #1244, opened 2022-07-11
external auth

lnnl:


ydogandjiev (dismissed):

apps/teams-test-app/src/components/AuthenticationAPIs.tsx
```diff
@@ -128,6 +128,37 @@ const Authenticate = (): React.ReactElement =>
     },
   });
 
+const externalAuthParams = {
+  url: 'auth_start.html?oauthRedirectMethod=deeplink&authId=1',
```
anmaur, line 132: Lets not hardcode these param, especially authId. Also, In android (including Teams), We use a URL which is placeholded at the Hub: like
https://teams-test-tab.azurewebsites.net/auth_start.html?authId={authId}&oauthRedirectMethod={oauthRedirectMethod}

lnnl: For the scenario test I have to hard code the authId because the scenario test cannot control the external browser, so after the external browser is opened, I have to manually raise the onExternalAuthDeeplink event to test the 2nd part of the process. But if we generate a new authId when we open the external browser, we won't be able to get that id when raising the event, so for the purpose of the scenario test, I need to hard code the authId

anmaur: Can you try looking into to mimic that in a reliable manner? For eg. updating a file from auth_end.html and listening/reading that from cypress

lnnl: updated, changed the url to [https://teams-test-tab.azurewebsites.net/auth_start.html?authId={authId}&oauthRedirectMethod={oauthRedirectMethod}](https://teams-test-tab.azurewebsites.net/auth_start.html?authId=%7BauthId%7D&oauthRedirectMethod=%7BoauthRedirectMethod%7D)

lnnl: I would say saving the id somewhere and retrieve it in some manner would require even more uncontrollable sources to work together correctly (network locations, access...), for this test a hardcode id in this case should be an acceptable compromise

ydogandjiev (dismissed):

apps/teams-test-app/src/components/AuthenticationAPIs.tsx
```diff
@@ -128,6 +128,37 @@ const Authenticate = (): React.ReactElement =>
     },
   });
 
+const externalAuthParams = {
+  url: 'https://teams-test-tab.azurewebsites.net/auth_start.html?authId=1&oauthRedirectMethod={oauthRedirectMethod}',
```
anmaur, line 132: pls change the authId to some static guid

apps/teams-test-app/src/components/AuthenticationAPIs.tsx
```diff
@@ -128,6 +128,38 @@ const Authenticate = (): React.ReactElement =>
     },
   });
 
+const externalAuthParams = {
```
MarcinJuraszek, line 131: Why do we need these to be hardcoded here? They seem relevant to the E2E scenarios tests you're writing, and should be typed into the input box by the test instead of the test app having them hardcoded.

apps/teams-test-app/src/components/AuthenticationAPIs.tsx
```diff
@@ -128,6 +128,38 @@ const Authenticate = (): React.ReactElement =>
...
+  url:
+    'https://teams-test-tab.azurewebsites.net/auth_start.html?authId=b9237908-2150-497a-8e47-8bf3ebf81b77&oauthRedirectMethod={oauthRedirectMethod}',
+  isExternal: true,
+};
+
+const AuthenticateExternal = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'authentication.authenticate3',
```
MarcinJuraszek, line 139: can this be `authentication.authenticate.external`?

MarcinJuraszek (changes requested):

apps/teams-test-app/src/components/AuthenticationAPIs.tsx
lnnl: done

apps/teams-test-app/src/components/AuthenticationAPIs.tsx
lnnl: updated

lnnl:
change no longer required because the existing test element can be used for the new external auth test

PR #1251, opened 2022-07-18
3p cloud storage provider action API improvements

HarshJain27:
3p cloud storage provider action API improvements

An improvement over changes of [This PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1152). 
Split a single a single action API into following 3 APIs:
1. Delete 2. Download 3. Upload

Corresponding Task : [Task Link](https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2519496)

### DIFF
--- a/packages/teams-js/src/private/files.ts
+++ b/packages/teams-js/src/private/files.ts
@@ -339,6 +339,7 @@ export namespace files {
   export interface CloudStorageProviderNewFileContent extends CloudStorageProviderContent {
     newFileName: string;
     newFileExtension: string;
+    destinationFolder: CloudStorageFolderItem | ISharePointFile;
   }
 
   /**
@@ -356,11 +357,31 @@ export namespace files {
    * @hidden
    * Hide from docs
    *
-   * Interface representing 3P cloud storage provider actions like Upload / Download / Delete file(s)
+   * Interface representing 3P cloud storage provider delete existing file(s) action
    */
-  export interface CloudStorageProviderActionContent extends CloudStorageProviderContent {
-    action: CloudStorageProviderFileAction;
+  export interface CloudStorageProviderDeleteFileContent extends CloudStorageProviderContent {
     itemList: CloudStorageFolderItem[] | ISharePointFile[];
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Interface representing 3P cloud storage provider download existing file(s) action
+   */
+  export interface CloudStorageProviderDownloadFileContent extends CloudStorageProviderContent {
+    itemList: CloudStorageFolderItem[] | ISharePointFile[];
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   *
+   * Interface representing 3P cloud storage provider upload existing file(s) action
+   */
+  export interface CloudStorageProviderUploadFileContent extends CloudStorageProviderContent {
+    itemList: CloudStorageFolderItem[] | ISharePointFile[];
+    destinationFolder: CloudStorageFolderItem | ISharePointFile;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/private/files.ts
```diff
@@ -710,40 +731,90 @@ export namespace files {
...
+    deleteFileRequest: CloudStorageProviderRequest<CloudStorageProviderDeleteFileContent>,
+    callback: (error?: SdkError) => void,
+  ): void {
+    ensureInitialized(FrameContexts.content);
+
+    if (!callback) {
+      throw getSdkError(ErrorCode.INVALID_ARGUMENTS, '[files.deleteCloudStorageProviderFile] callback cannot be null');
+    }
+
+    if (!(deleteFileRequest && deleteFileRequest.content)) {
```
TrevorJoelHarris, line 749: shouldn't you also be checking that itemList is present / non-null/non-empty/all CloudStorageFolderItems are files and not directories

packages/teams-js/src/private/files.ts
```diff
@@ -710,40 +731,90 @@ export namespace files {
...
+   * Initiates the download 3P cloud storage file(s) flow, which will download existing file(s) from the given 3P provider
    *
-   * @param cloudStorageProviderFileActionRequest 3P cloud storage provider file action (Upload / Download / Delete) request content
-   * @param callback Callback that will be triggered post 3P cloud storage action
+   * @param downloadFileRequest 3P cloud storage provider download file(s) action request content
+   * @param callback Callback that will be triggered post downloading existing file(s) flow is finished
    */
-  export function performCloudStorageProviderFileAction(
-    cloudStorageProviderFileActionRequest: CloudStorageProviderRequest<CloudStorageProviderActionContent>,
+  export function downloadCloudStorageProviderFile(
```
TrevorJoelHarris, line 768: How does download work exactly? I don't see the callback returning any file information

TrevorJoelHarris: Also, same comment from above about checking itemList

packages/teams-js/src/private/files.ts
```diff
@@ -710,40 +731,90 @@ export namespace files {
...
+    callback: (error?: SdkError) => void,
+  ): void {
+    ensureInitialized(FrameContexts.content);
+
+    if (!callback) {
+      throw getSdkError(ErrorCode.INVALID_ARGUMENTS, '[files.uploadCloudStorageProviderFile] callback cannot be null');
+    }
+
+    if (!(uploadFileRequest && uploadFileRequest.content)) {
+      throw getSdkError(
```
TrevorJoelHarris, line 811: validate itemList

TrevorJoelHarris: (also need to validate destination folder: make sure it's a directory, has valid params on it, etc.

packages/teams-js/src/private/files.ts
```diff
@@ -763,7 +834,7 @@ export namespace files {
       throw new Error('[registerCloudStorageProviderListChangeHandler] Handler cannot be null');
     }
 
-    registerHandler('files.cloudStorageProviderList', handler);
+    registerHandler('files.cloudStorageProviderListChange', handler);
```
TrevorJoelHarris, line 837: this is fine, but keep in mind you will also need to update the other sdk

TrevorJoelHarris (changes requested):

packages/teams-js/src/private/files.ts
HarshJain27: Done. Made validation and restriction on itemList and destination folder.

packages/teams-js/src/private/files.ts
HarshJain27: Sure.

packages/teams-js/src/private/files.ts
HarshJain27: Made validation on itemList. 

Download Working :
The consuming app (in our case from OD side) initiates the download of file(s) with required file metadata. SDK is just delegating the call and teams side takes care of download of the file(s). The reason callback has no information related to file is that the entire download process is taken care at teams side. Main reason behind this being 3P file access requiring authentication, infra for which is present at teams side only and not at OD side.

cloudtx (approved):

packages/teams-js/test/private/files.spec.ts
```diff
@@ -631,11 +631,21 @@ describe('files', () => {
   });
 
   describe('addCloudStorageProviderFile', () => {
+    const mockDestinationFolder: files.CloudStorageFolderItem = {
```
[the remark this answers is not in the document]
HarshJain27: Added the test cases for different validation applied.

packages/teams-js/test/private/files.spec.ts
```diff
@@ -891,6 +1008,43 @@ describe('files', () => {
...
+    });
+
+    it('should not allow calls with null destinationFolder request content', async () => {
+      await utils.initializeWithContext('content');
+      expect(() =>
+        files.uploadCloudStorageProviderFile(uploadFileRequestWithNullDestinationFolder, emptyCallback),
+      ).toThrowError('[files.uploadCloudStorageProviderFile] Invalid destination folder details');
+    });
+
+    it('should not allow upload calls for file as destination folder in request content', async () => {
```
TrevorJoelHarris, line 1041: I don't think you have a test for the CloudStorageProvider.Sharepoint case

HarshJain27: Yes i didn't. Added now.

cloudtx (approved):

TrevorJoelHarris (dismissed):

TrevorJoelHarris (dismissed):

cloudtx (dismissed):

packages/teams-js/src/private/files.ts
```diff
@@ -297,6 +297,28 @@ export namespace files {
...
+     */
+    name: string;
+    /**
+     * The file path.
+     */
+    webkitRelativePath: string;
+    /**
+     * Represents if file or a folder
+     */
+    isFolder?: boolean;
```
TrevorJoelHarris, line 319: I appreciate you making a well-documented interface for this, but we shouldn't have a function called "isFolder" on a file. Files are not folders.

packages/teams-js/src/private/files.ts
```diff
@@ -297,6 +297,28 @@ export namespace files {
...
+     */
+    lastModified: Date;
+    /**
+     * Name of the file
+     */
+    name: string;
+    /**
+     * The file path.
+     */
+    webkitRelativePath: string;
```
TrevorJoelHarris, line 315: Why is this called the webkitRelativePath? I think our goal should be to have a platform-agnostic interface.

TrevorJoelHarris (changes requested):

packages/teams-js/src/private/files.ts
HarshJain27: Removed the field **isFolder**, as it was added to make the interface common for file and folder. Removed it as per your comments. If needed in future, will look for generic alternatives.

packages/teams-js/src/private/files.ts
HarshJain27: Renamed the field to **filePath** and updated the comments to make it verbose.

TrevorJoelHarris (approved):

cloudtx (approved):

PR #1252, opened 2022-07-19
Added calendar and pages capability

AnanyaBarthakur1:


apps/sample-app/src/components/Calendar.tsx
```diff
@@ -1,21 +1,57 @@
...
+
+const teamsDeepLinkHost = 'teams.microsoft.com';
+const teamsDeepLinkProtocol = 'https';
+const teamsDeepLinkAttendeesUrlParameterName = 'attendees';
+const teamsDeepLinkUrlPathForCalendar = '/l/meeting/new';
+
+export const handleNewMail = async (): Promise<void> => {
+  if (!calendar.isSupported()) {
+    const context = await app.getContext();
+    if (context?.app?.host?.name === 'Teams') {
```
aeunms, line 17: Hmm I think we want to avoid checking for a specific host? Was this the workaround you mentioned today?

AnanyaBarthakur1: oh right! I just ran the code without any host check. It works perfectly okay, my initial reasoning for the check was 'what if its running on office and office doesn't support composing  a new meeting' but it turns out it redirects you to Teams anyway. Thanks!!

apps/sample-app/src/components/Calendar.tsx
```diff
@@ -1,21 +1,52 @@
...
-    // create calendar paramaters
+import { MessageListItem } from './Emails';
+
+const teamsDeepLinkHost = 'teams.microsoft.com';
+const teamsDeepLinkProtocol = 'https';
+const teamsDeepLinkAttendeesUrlParameterName = 'attendees';
+const teamsDeepLinkUrlPathForCalendar = '/l/meeting/new';
+
+export const handleNewMail = async (): Promise<void> => {
+  if (!calendar.isSupported()) {
```
aeunms, line 15: I think what you were saying was that using calendar here is a workaround for mail capability not being enabled, is that right?

AnanyaBarthakur1: I see why this is confusing, I should change "handleNewMail" to "handleNewMeeting". This functionality is independent of whether the "mail" capability is enabled. There are two different things happening - 1. "Mail" Capability from './utils' is to start a new email to one of the People Avatars (that is why it is meant to show up on the avatar hover menu alongside call, message functions). 2. "Add a new Meeting" comes from "Compose New Meeting" capability in "Calendar" to create a new meeting invite, without any specific person mentioned

apps/sample-app/src/components/Calendar.tsx
```diff
@@ -1,21 +1,52 @@
...
     };
-    // send to 'compose meeting'
     await calendar.composeMeeting(calendarParams);
-  };
+  }
+};
+
+export const handleOpenMail = async (e: MessageListItem): Promise<void> => {
+  if (!calendar.isSupported()) {
+    alert('open mail item is not supported');
```
aeunms, line 37: I think it could be a bit confusing for the devs looking to the app as the implementation example since it's saying 'opening mail item is not supported' when calendar is not rather than when mail is not supported- mail has an API called `openMailItem`. I think it would be clearer to say something like 'calendar is not supported- could not open this calendar item.'

apps/sample-app/src/components/Emails.tsx
```diff
@@ -0,0 +1,49 @@
...
+      subject: m.subject || '',
+      sender: m.sender?.emailAddress?.name || '',
+    };
+  });
+
+  const EmailExample: React.FunctionComponent = () => {
+    return (
+      <div>
+        {emailItems.map(e => (
+          <Text as="span" key={e.key}>
```
aeunms, line 30: Could we rename `e` to be a bit longer? It looks like it's suppose to be an email item, maybe rename it to be `emailItem` or something like that? I do notice it sounds a lot similar to `emailItems` but I think it's still more explanatory than just the single character.

aeunms: Applied to all the other places in the PR where it references `e`

AnanyaBarthakur1: Okay, also from an earlier PR I make a similar mapping function but only use 'c' (its in PeopleAvatars.tsx). I'll change those to be more descriptive as well

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -67,23 +61,23 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
...
-                              <MenuItem onClick={() => handleVideoCall(a)}> Video {c}</MenuItem>
+                              <MenuItem onClick={() => handleAudioCall(a)}>Audio {capability}</MenuItem>
+                              <MenuItem onClick={() => handleVideoCall(a)}> Video {capability}</MenuItem>
                             </MenuList>
                           </MenuPopover>
                         </Menu>
                       )}
-                      {c === 'Message' && <MenuItem onClick={() => handleMessage(a)}> {c}</MenuItem>}
-                      {c === 'Mail' && <MenuItem onClick={() => handleMail(a)}>{c}</MenuItem>}
+                      {capability === 'Message' && <MenuItem onClick={() => handleMessage(a)}> {capability}</MenuItem>}
```
aeunms, line 79: Perfecto

aeunms: There is still the `a` character but honestly I think it's fine since it's really just meant to be whatever input it's getting

aeunms (approved):
🚢

apps/sample-app/src/components/Calendar.tsx
```diff
@@ -1,21 +1,39 @@
...
-    // create calendar paramaters
+const teamsDeepLinkHost = 'teams.microsoft.com';
+const teamsDeepLinkProtocol = 'https';
+const teamsDeepLinkAttendeesUrlParameterName = 'attendees';
+const teamsDeepLinkUrlPathForCalendar = '/l/meeting/new';
+
+export const handleNewMeeting = async (): Promise<void> => {
+  if (!calendar.isSupported()) {
+    const calendarParams: calendar.ComposeMeetingParams = {
+      attendees: ['emailAddress@microsoft.com'],
```
AE-MS, line 15: Since you have this in two places (here and on line 26) you should just define it once as a constant outside and then reference it in both places.

apps/sample-app/src/components/Emails.tsx
```diff
@@ -0,0 +1,49 @@
...
+  sender?: string;
+}
+export const EmailList: React.FC<EmailProps> = (props: EmailProps) => {
+  const { messages } = props;
+  const emailItems = messages.map<MessageListItem>(m => {
+    return {
+      key: m.id,
+      subject: m.subject || '',
+      sender: m.sender?.emailAddress?.name || '',
+    };
```
AE-MS, line 23: If you're interested in being super-efficient without (I don't think) sacrificing readability here you can actually get rid of the return statement and just replace it with parentheses:
```
  const emailItems = messages.map<MessageListItem>(m => ({
      key: m.id,
      subject: m.subject || '',
      sender: m.sender?.emailAddress?.name || '',
    }));
```

apps/sample-app/src/components/Meetings.tsx
```diff
@@ -15,13 +15,13 @@ export const MeetingList: React.FC<MeetingProps> = (props: MeetingProps) => {
   // using a for loop instead of 'map' function to remove instances of 'OOF'
```
AE-MS, line 15: >// using a for loop instead of 'map' function to remove instances of 'OOF' [](http://example.com/codeflow?start=2&length=74)

This comment seems out of date.

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -32,17 +32,11 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
 
   for (let i = 0; i < messages.length; i++) {
     const message = messages[i];
-    if (
-      message['attendees'].length > 15 ||
-      message['subject']?.includes('OOF') ||
-      message['subject']?.includes('Canceled')
-    ) {
+    // 5 is an arbitrary number to show only relevant meetings
+    if (message['attendees'].length > 5 || message['isCancelled'] || message['showAs'] === 'free') {
```
AE-MS, line 36: >message['isCancelled'] || message['showAs'] === 'free' [](http://example.com/codeflow?start=43&length=54)

Since it seems like you do this two different places now (here and in Meetings.tsx) you should consider making a utility function like `ShouldShowMessage` or `IsMessageVisible` or something and then extract these two checks into that helper method and call it from the two places.

apps/sample-app/src/components/utils.tsx
```diff
@@ -76,13 +77,16 @@ export const handleMail = async (a: AvatarItem): Promise<void> => {
...
-export const handleCalendar = async (a: AvatarItem): Promise<void> => {
-  const calendarParams: calendar.ComposeMeetingParams = {
-    attendees: [`${a.id}`],
-  };
-  const result = await calendar.composeMeeting(calendarParams);
-  return alert(result);
+
+export const handleOpenMailItem = async (emailItem: MessageListItem): Promise<void> => {
+  if (!mail.isSupported()) {
+    alert('open mail item is not supported');
```
AE-MS, line 85: I don't know what the design for this sample is but you might consider not showing the UX that would trigger this handler if the mail functionality isn't supported, rather than teasing the button and then letting the user find out that it doesn't do anything only after its been clicked.

AE-MS (approved):
:shipit:

apps/sample-app/src/components/PeopleAvatars.tsx
AnanyaBarthakur1: I like this idea, will be a lot cleaner

PR #1265, opened 2022-07-22
Code cleanup from recent PR feedback

AnanyaBarthakur1:
(initializing calendar params just once, creating a util function for relevant messages, removing alert message if mail is not supported)

### DIFF
--- a/packages/teams-js/devdocs/logging.md
+++ b/packages/teams-js/devdocs/logging.md
@@ -16,6 +16,8 @@ TeamsJS logging leverages the popular [debug](https://www.npmjs.com/package/debu
 
 5. Click on the "Log Level" dropdown and ensure "Verbose" is checked (it will probably be set to "Default levels"). All TeamsJs log statements are currently written at Verbose.
 
+6. Refresh your app
+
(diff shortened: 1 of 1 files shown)


apps/sample-app/src/components/Emails.tsx
```diff
@@ -0,0 +1,47 @@
+import { DocumentCard, DocumentCardTitle } from '@fluentui/react';
+import { Text } from '@fluentui/react-components';
+import { Message } from '@microsoft/microsoft-graph-types';
+import React from 'react';
+
+//import { handleOpenMailItem } from './utils';
+//onClick={() => handleOpenMailItem(emailItem)}
```
AE-MS, line 7: Did you intend for this to be part of your check in?

AE-MS (changes requested):
:clock1:

apps/sample-app/src/components/Emails.tsx
AnanyaBarthakur1: Oops. nope. Thank you for pointing this out to me

AE-MS (approved):

PR #1266, opened 2022-07-22
In progress proposal for Adaptive Card Dialog support

TrevorJoelHarris:
I'm not sure this is ready for review yet which is why it is marked as draft.

A few points:

1. The Dialog capability has been restructured to better support both url and adaptive card-based dialogs. I feel ok doing this since the capability is still marked as @beta. You can now use dialog and its subcapabilities as follows:
     _dialog.url.open_ will open a url based dialog that does not interact with the bot framework
     _dialog.url.bot.open_ will open a url based dialog that **does** interact with the bot framework
     _dialog.adaptiveCard.open_ will open an adaptive card based dialog that does not interact with the bot framework
     _dialog.adaptiveCard.bot.open_ will open an adaptive card based dialog that **does** interact with the bot framework
2. dialog.adaptiveCard.getVersion() will return the current version (MAJOR.MINOR) of adaptive cards that the host supports for dialogs. Hosts will need to attest to the version of adaptive cards they are providing to app developers
3. The 2 open functions in the adaptiveCard subcapability will also verify that the version of adaptiveCards the host claims to provide meets a minimum version requirement. I have this set to 1.15 in this PR because that's the version the most recent JS adaptive card package supports.

### DIFF
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -194,7 +194,8 @@ export namespace dialog {
      * This function cannot be called from inside of a dialog
      *
      * @param adaptiveCardDialogInfo - An object containing the parameters of the dialog module.
-     * @param submitHandler - Handler that triggers when the adaptive card dialog executes the Action.Submit action from the adaptive card or when the user closes the dialog.
+     * @param submitHandler - Handler that triggers when the adaptive card dialog executes the Action.Submit action
+     * from the adaptive card or when the user closes the dialog.
      */
     export function open(adaptiveCardDialogInfo: AdaptiveCardDialogInfo, submitHandler?: DialogSubmitHandler): void {
       ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -230,6 +230,10 @@ export namespace dialog {
     }
 
     export function getVersion(): Promise<AdaptiveCardVersion> {
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+
       return sendAndHandleSdkError('adaptiveCard.version');
     }
 
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -9,12 +9,10 @@ import { ensureInitialized } from '../internal/internalAPIs';
 import { DialogDimension, errorNotSupportedOnPlatform, FrameContexts, minAdaptiveCardVersion } from './constants';
 import {
   AdaptiveCardDialogInfo,
-  AdaptiveCardVersion,
   BotAdaptiveCardDialogInfo,
   BotUrlDialogInfo,
   DialogInfo,
   DialogSize,
-  ErrorCode,
   UrlDialogInfo,
 } from './interfaces';
 import { runtime } from './runtime';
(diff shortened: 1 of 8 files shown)


packages/teams-js/src/public/dialog.ts
```diff
@@ -290,4 +242,227 @@ export namespace dialog {
...
+     * @param adaptiveCardDialogInfo - An object containing the parameters of the dialog module.
+     * @param submitHandler - Handler that triggers when a dialog calls the {@linkcode submit} function or when the user closes the dialog.
+     * @param messageFromChildHandler - Handler that triggers if dialog sends a message to the app.
+     *
+     * @returns a function that can be used to send messages to the dialog.
+     */
+    export function open(
+      adaptiveCardDialogInfo: AdaptiveCardDialogInfo,
+      submitHandler?: DialogSubmitHandler,
+      messageFromChildHandler?: PostMessageChannel,
```
MarcinJuraszek, line 283: For AC-based dialogs, AFAIU, no user-provided code runs. It's all data driven. How is the dialog supposed to send messages back to the launch page and/or respond to messages sent from the parent?

packages/teams-js/src/public/runtime.ts
```diff
@@ -7,6 +7,7 @@ export interface IRuntime {
   readonly apiVersion: number;
   readonly isLegacyTeams?: boolean;
   readonly supports: {
+    readonly adaptiveCard?: {};
```
MarcinJuraszek, line 10: curious - what is this representing?

TrevorJoelHarris: The intention is that it will just be something that hosts use to pass down the version of adaptive cards that they support. While I'd like for there to never be different versions, in practice (especially in the future) we'll have hosts who run different versions, even if it's just for A/B testing new versions of the dialog. Having this split apart from the main part of the adaptive card dialogs lets us keep supporting that flow for Teams while also enforcing something like a minimum viable AC version in the new capability

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: This is a good question, let me poke around a little (and possibly talk with teams about it if the documentation isn't clear). I suspect you're right, which would be great

TrevorJoelHarris: Got confirmation from teams: shouldn't be any reason to keep this for AC dialogs

TrevorJoelHarris: Updated this: removed the handler from here and moved all of the message handling code down into dialog.url

packages/teams-js/src/public/runtime.ts
TrevorJoelHarris: Sorry, I will be super clear just to make sure we're on the same page: this *is* a new capability, it's just that all it provides is the ability to check what version of adaptive cards is supported: 
image

TrevorJoelHarris: We're then saying that the new adaptive card support in dialog needs this to also be implemented, and will be using it to both let developers check what the currently supported version of adaptive cards is in the dialog on that host. We'll also use it internally to validate that the host is meeting a minimum acceptable version of the AC schema.

packages/teams-js/src/public/dialog.ts
```diff
@@ -111,183 +88,383 @@ export namespace dialog {
...
 
     /**
-     * Checks if dialog.update capability is supported by the host
+     * Query the host to see what version of the AdaptiveCard schema they support
      *
-     * @returns boolean to represent whether dialog.update is supported
+     * @returns a Promise containing the {@linkcode AdaptiveCardVersion}, describing what version of the AdaptiveCard
+     * schema the host supports
+     */
+    export function getVersion(): Promise<AdaptiveCardVersion> {
```
MarcinJuraszek, line 239: I believe this should be in a top-level `adaptiveCards` capability, unless there are scenarios where different capabilities leveraging ACs will somehow be backed by a different version.

packages/teams-js/src/public/dialog.ts
```diff
@@ -111,183 +88,383 @@ export namespace dialog {
...
     /**
-     * Update dimensions - height/width of a dialog.
+     * Allows app to open an adaptive card based dialog.
      *
-     * @param dimensions - An object containing width and height properties.
+     * @remarks
+     * This function cannot be called from inside of a dialog
+     *
+     * @param adaptiveCardDialogInfo - An object containing the parameters of the dialog module.
+     * @param submitHandler - Handler that triggers when a dialog calls the {@linkcode submit} function or when the user closes the dialog.
```
MarcinJuraszek, line 215: what does "when a dialog calls the submit function" mean here? Because for AC based dialogs, there is no dev-provided JS running at all, is there?

packages/teams-js/src/public/dialog.ts
```diff
@@ -111,183 +88,383 @@ export namespace dialog {
...
-    export function resize(dimensions: DialogSize): void {
-      ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
+    export function open(adaptiveCardDialogInfo: AdaptiveCardDialogInfo, submitHandler?: DialogSubmitHandler): void {
+      ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
       if (!isSupported()) {
         throw errorNotSupportedOnPlatform;
       }
-      sendMessageToParent('tasks.updateTask', [dimensions]);
+
+      getVersion().then(hostVersion => {
```
MarcinJuraszek, line 223: Does this mean a `NOT_SUPPORTED_ON_PLATFORM` kind of error can be thrown two different ways? If `isSupported` returns true, it'll be a simple throw (a synchronous one), but if it's the version that's not good enough, the function won't throw per se, and instead the promise we return will be rejected? Seems a bit messy?

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: For adaptive-card based dialogs, this would be called when your adaptive-card fires the 'Action.Submit' action https://adaptivecards.io/explorer/Action.Submit.html

Will update the documentation here to be more clear

TrevorJoelHarris: (Will also move dialog.submit to be dialog.url.submit)

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: Removing and replacing with adding to intialization flow

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: Agreed, I will clean this up and make them both synchronous

PR #1270, opened 2022-07-26
Added Env expansion to npmrc file

jadahiya-MSFT:
Adding in Env allows for the powershell script to access ADO specific Environment variables as well as ones that we add to our pipelines

### DIFF
--- a/change/@microsoft-teams-js-87caf6f6-44c9-4ff7-9cd4-e4a4c67e0ee9.json
+++ b/change/@microsoft-teams-js-87caf6f6-44c9-4ff7-9cd4-e4a4c67e0ee9.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "\"Testing npmrc auth token, will update description once tests are completed\"",
+  "packageName": "@microsoft/teams-js",
+  "email": "jadahiya@microsoft.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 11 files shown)


change/@microsoft-teams-js-87caf6f6-44c9-4ff7-9cd4-e4a4c67e0ee9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "\"Added in `Env` variable expansion to `NPM_TOKEN` variable in add-npmrc powershell script\"",
```
erinha, line 3: The escaped quotes look odd. Are they necessary?

erinha (changes requested):
Probably the change file shouldn't need escaped quotes. Otherwise looks good! Congratulations on figuring it out!

change/@microsoft-teams-js-87caf6f6-44c9-4ff7-9cd4-e4a4c67e0ee9.json
Jessica-Koch: Yeah, I agree, I understand that the backticks can't be used since that would cancel out those around 'env' and 'npm_token' but could single quotes be used instead?

Jessica-Koch: maybe they can't tho

erinha (approved):
🚀

TrevorJoelHarris (approved):

PR #1271, opened 2022-07-26
Adding in FrameContexts.task to openChat and openGroupChat

jadahiya-MSFT:


packages/teams-js/test/public/chat.spec.ts
```diff
@@ -56,7 +56,7 @@ describe('chat', () => {
         message: 'someMessage',
       };
       return expect(chat.openChat(chatRequest)).rejects.toThrowError(
-        'This call is only allowed in following contexts: ["content"]. Current context: "settings".',
```
erinha, line 59: I think we need to add additional unit tests that validate the API can be called from `FrameContexts.task` right?

jadahiya-MSFT: Good catch!

erinha: Thank you for adding them. I don't want to be too nit-picky, but I would prefer to see them follow the same code pattern established in other test files for validating allowed contexts and disallowed contexts. Example: See lines 49-63ish in dialog.spec.ts.

jadahiya-MSFT: I was following the approach in app.spec.ts, but the one in dialogs that you pointed out was better

AE-MS (approved):
:shipit:

packages/teams-js/test/public/chat.spec.ts
erinha: Thank you! And thanks for bringing my awareness to the fact that there is a different pattern in app.spec.ts. :)

erinha (approved):
🚀

Jessica-Koch (approved):

PR #1272, opened 2022-07-28
Added links and BoxButtons for browser API in test app

anmaur:
> Add a new file Links.tsx in test app which contains all the links and boxes for redirection APIs by the browser to help us test the behaviour.

### DIFF
--- a/apps/teams-test-app/src/App.tsx
+++ b/apps/teams-test-app/src/App.tsx
@@ -13,6 +13,7 @@ import CalendarAPIs from './components/CalendarAPIs';
 import CallAPIs from './components/CallAPIs';
 import DialogAPIs from './components/DialogAPIs';
 import GeoLocationAPIs from './components/GeoLocationAPIs';
+import Links from './components/Links';
 import LocationAPIs from './components/LocationAPIs';
 import LogAPIs from './components/LogsAPIs';
 import MailAPIs from './components/MailAPIs';
@@ -105,6 +106,7 @@ export const generateRegistrationMsg = (changeCause: string): string => {
 const App = (): ReactElement => {
   return (
     <>
+      <Links />
       <AppAPIs />
       <AppInitializationAPIs />
       <AppInstallDialogAPIs />
--- a/apps/teams-test-app/src/App.tsx
+++ b/apps/teams-test-app/src/App.tsx
@@ -106,7 +106,6 @@ export const generateRegistrationMsg = (changeCause: string): string => {
 const App = (): ReactElement => {
   return (
     <>
-      <Links />
       <AppAPIs />
       <AppInitializationAPIs />
       <AppInstallDialogAPIs />
@@ -120,6 +119,7 @@ const App = (): ReactElement => {
       <FilesAPIs />
       <FullTrustAPIs />
       <GeoLocationAPIs />
+      <Links />
       <LocationAPIs />
       <LogAPIs />
       <MailAPIs />
--- a/apps/teams-test-app/src/components/Links.tsx
+++ b/apps/teams-test-app/src/components/Links.tsx
@@ -1,0 +1,71 @@
+import React, { ReactElement } from 'react';
+
+import { ApiWithTextInput } from './utils';
+
+const MoveToLink = (): ReactElement =>
+  ApiWithTextInput<string>({
+    name: 'moveToLink',
+    title: 'Move to Link',
+    onClick: {
+      validateInput: input => {
+        if (typeof input !== 'string') {
+          throw new Error('Input should be a string');
+        }
+      },
+      submit: async input => {
+        window.location.href = input;
+        return Promise.resolve('Moved to new link');
+      },
+    },
+  });
+
+const OpenLinkInNewWindow = (): ReactElement =>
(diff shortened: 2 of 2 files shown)


apps/teams-test-app/src/App.tsx
```diff
@@ -105,6 +106,7 @@ export const generateRegistrationMsg = (changeCause: string): string => {
 const App = (): ReactElement => {
   return (
     <>
+      <Links />
```
AE-MS, line 109: There's something about this list that looks out of place now. :smile:

apps/teams-test-app/src/components/Links.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <br />
+    <a rel="noreferrer" href="https://teams-test-tab.azurewebsites.net/" target="_self">
+      Target Self
+    </a>
+    <br />
+    <a rel="noreferrer" href="https://teams-test-tab.azurewebsites.net/" target="_parent">
+      Target Parent
+    </a>
+    <br />
+    <a rel="noreferrer" href="https://teams-test-tab.azurewebsites.net/" target="_top">
```
AE-MS, line 58: Can these URLs be a constant that is referenced so they aren't repeated?

apps/teams-test-app/src/components/Links.tsx
```diff
@@ -0,0 +1,71 @@
...
+        window.open(input);
+        return Promise.resolve('Link opened');
+      },
+    },
+  });
+
+const Links = (): ReactElement => (
+  <>
+    <h1>Links</h1>
+    <a id="link_simple" href="https://teams-test-tab.azurewebsites.net/">
```
AE-MS, line 42: >https://teams-test-tab.azurewebsites.net/ [](http://example.com/codeflow?start=30&length=41)

Are you using this URL because it should point to the current app? Or just because you know it points to a valid instance of the test app? I worry about hardcoding a URL like this into the test app.

apps/teams-test-app/src/components/Links.tsx
```diff
@@ -0,0 +1,71 @@
...
+    title: 'Move to Link',
+    onClick: {
+      validateInput: input => {
+        if (typeof input !== 'string') {
+          throw new Error('Input should be a string');
+        }
+      },
+      submit: async input => {
+        window.location.href = input;
+        return Promise.resolve('Moved to new link');
```
AE-MS, line 17: Does this work? I mean I assume it must work since you wrote it, but it's fascinating that redirecting the browser to a new site lets the code keep running so that it resolves the promise?

AE-MS (changes requested):
:clock1:

apps/teams-test-app/src/components/Links.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { ApiWithTextInput } from './utils';
+
+const MoveToLink = (): ReactElement =>
+  ApiWithTextInput<string>({
+    name: 'moveToLink',
+    title: 'Move to Link',
+    onClick: {
+      validateInput: input => {
+        if (typeof input !== 'string') {
+          throw new Error('Input should be a string');
```
maglims, line 12: Do we mean to say input is a url string? I think string seems too generic and people might enter any string.

apps/teams-test-app/src/components/Links.tsx
anmaur: I was trying to point this to a URL which is different from test app but is still a URL which can communicate with the Hub. This would help us validate the redirect scenarios i.e. when an App redirect to a different valid URL, it should be able to use the APIs

apps/teams-test-app/src/components/Links.tsx
```diff
@@ -0,0 +1,71 @@
...
+      submit: async input => {
+        window.open(input);
+        return Promise.resolve('Link opened');
+      },
+    },
+  });
+
+const Links = (): ReactElement => (
+  <>
+    <h1>Links</h1>
```
maglims, line 41: Not sure if we should specially call out somewhere that this is not APIs  and instead they are windows interface related. Maybe people will know by reading app.tsx since it is different from other *callName*API.

apps/teams-test-app/src/components/Links.tsx
anmaur: haha, yes it does but its unpredictable. Till the browser does not move to new webpage. We would have short span of time to notice.

apps/teams-test-app/src/components/Links.tsx
maglims: oh, just to be more clear, I meant to say if it might be more clear if it says below :)
 throw new Error('Input should be a url string');

AE-MS (approved):
:shipit:

apps/teams-test-app/src/components/Links.tsx
anmaur: I agree and that behavious we get it from URL construct itself. It throws an exception with invalid URL message if it isn't able to form a URL. For eg. "Failed to construct 'URL': Invalid URL"

maglims (approved):

PR #1273, opened 2022-07-28
Elaborate on some of the context properties

AE-MS:
Elaborate on some of the context properties. The reference documentation seemed terse.

### DIFF
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -447,32 +447,32 @@ export namespace app {
    */
   export interface Context {
     /**
-     * Info of the app
+     * Properties about the current session for your app
      */
     app: AppInfo;
 
     /**
-     * Info of the current page of App
+     * Info about the current page context hosting your app
      */
     page: PageInfo;
 
     /**
-     * Info of the user
+     * Info about the current user
      */
     user?: UserInfo;
 
     /**
-     * Info of the Microsoft Teams channel
+     * When running in the context of a Teams channel, provides information about the channel, else `undefined`
      */
     channel?: ChannelInfo;
 
     /**
-     * Info of the Microsoft Teams chat
+     * When running the context of a Teams chat, provides information about the chat, else `undefined`
      */
     chat?: ChatInfo;
 
     /**
-     * Info of the Microsoft Teams meeting
+     * When running the context of a Teams meeting, provides information about the meeting, else `undefined`
      */
     meeting?: MeetingInfo;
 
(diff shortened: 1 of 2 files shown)


packages/teams-js/src/public/app.ts
```diff
@@ -447,32 +447,32 @@ export namespace app {
    */
   export interface Context {
     /**
-     * Info of the app
+     * Properties about the current session for your app
```
AE-MS, line 450: @erikadoyle is it okay to use pronouns like this in the documentation? I know it seems like it violates typical style but when I was writing this, it felt like using "your app" was the clearest way to differentiate between the app the developer reading this documentation is implement and the containing app (Teams, Outlook, etc.)

Happy to hear alternative suggestions that are also able to clearly differentiate between the developer's app and the hosting app.

erikadoyle: Yes! That's great actually; [MS Docs style](https://docs.microsoft.com/en-us/style-guide/top-10-tips-style-voice) recommends using an active voice (and "you"/"your"/"you'll" etc.) for current documentation because its more casual, friendly, and directed to customers.

erikadoyle (approved):

packages/teams-js/src/public/app.ts
```diff
@@ -453,32 +453,32 @@ export namespace app {
...
 
     /**
-     * Info of the Microsoft Teams channel
+     * When running in the context of a Teams channel, provides information about the channel, else `undefined`
      */
     channel?: ChannelInfo;
 
     /**
-     * Info of the Microsoft Teams chat
+     * When running the context of a Teams chat, provides information about the chat, else `undefined`
```
TrevorJoelHarris, line 476: probably should be "When running in _the_ context"

packages/teams-js/src/public/app.ts
```diff
@@ -453,32 +453,32 @@ export namespace app {
...
      */
     chat?: ChatInfo;
 
     /**
-     * Info of the Microsoft Teams meeting
+     * When running the context of a Teams meeting, provides information about the meeting, else `undefined`
```
TrevorJoelHarris, line 481: probably should be "When running in _the_ context"

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/app.ts
```diff
@@ -488,12 +488,12 @@ export namespace app {
     sharepoint?: any;
 
     /**
-     * Info of the sharePoint site associated with the team.
+     * When running in Teams for an organization with a tenant, provides information about the SharePoint site associated with the team.
```
aeunms, line 491: Interesting that some of these say "else `undefined`" and some don't, is it intentional?

AE-MS: Not intentional, and now fixed.

erikadoyle (dismissed):
These are such great improvements! Thank you @AE-MS

TrevorJoelHarris (approved):

TrevorJoelHarris (approved):

aeunms (approved):
🚢

PR #1274, opened 2022-07-28
add action items to teams-sdk

Jessica-Koch:
* propogate action items to app context 
* add action items to context test

### DIFF
--- a/packages/teams-js/src/public/interfaces.ts
+++ b/packages/teams-js/src/public/interfaces.ts
@@ -181,7 +181,7 @@ export enum FileOpenPreference {
  */
 
 export enum ActionObjectType {
-  M365Conontent = 'm365content',
+  M365Content = 'm365content',
 }
 
 export interface BaseActionObject<T extends ActionObjectType> {
--- a/packages/teams-js/src/public/interfaces.ts
+++ b/packages/teams-js/src/public/interfaces.ts
@@ -180,7 +180,34 @@ export enum FileOpenPreference {
  * Represents the structure of the received context message.
  */
 
+export enum ActionObjectType {
+  M365Conontent = 'm365content',
+}
+
+export interface BaseActionObject<T extends ActionObjectType> {
+  type: T;
+}
+
+/**
+ * Represents Action Information
+ */
+export interface ActionInfo {
+  /**
+   * Maps to the action id supplied inside the manifest
+   */
+  actionId: string;
+  actionObjects: BaseActionObject<ActionObjectType>[];
+}
+
 export interface Context {
+  /**
+   * @deprecated
+   * As of 2.0.0, please use {@link app.Context.actionInfo} instead
+   *
+   * Common information applicable to all content actions
+   */
+  actionInfo?: ActionInfo;
+
   /**
    * @deprecated
    * As of 2.0.0, please use {@link app.Context.team.groupId} instead
--- a/packages/teams-js/src/public/interfaces.ts
+++ b/packages/teams-js/src/public/interfaces.ts
@@ -170,7 +170,7 @@ export enum FileOpenPreference {
 }
 
 /**
- * Action Types
+ * Possible Action Types
  */
 export enum ActionObjectType {
   M365Content = 'm365content',
@@ -178,21 +178,11 @@ export enum ActionObjectType {
 
 /**
  * Data pertaining to object(s) the action is being performed on
+ *
+ * @param T The type of action being implemented
  */
 export interface BaseActionObject<T extends ActionObjectType> {
   type: T;
-}
-
-/**
- * Common information which applies to all actions and an array of
- * corresponding action objects
- */
-export interface ActionInfo {
-  /**
-   * Maps to the action id supplied inside the manifest
-   */
-  actionId: string;
-  actionObjects: BaseActionObject<ActionObjectType>[];
 }
 
 /**
@@ -232,6 +222,20 @@ export enum SecondaryM365ContentIdName {
 export function isM365ContentType(actionItem: unknown): actionItem is M365ContentAction {
   // eslint-disable-next-line no-prototype-builtins
   return actionItem && Object.prototype.hasOwnProperty.call(actionItem, 'secondaryId');
+}
+
+/**
+ * Information common to all actions
+ */
+export interface ActionInfo {
+  /**
+   * Maps to the action id supplied inside the manifest
+   */
+  actionId: string;
+  /**
+   * Array of corresponding action objects
+   */
+  actionObjects: BaseActionObject<ActionObjectType>[];
 }
 
 /**
--- a/packages/teams-js/src/public/interfaces.ts
+++ b/packages/teams-js/src/public/interfaces.ts
@@ -217,14 +217,6 @@ export enum SecondaryM365ContentIdName {
 }
 
 /**
- * Type guard to determine if an action item is of M365Content Type
- */
-export function isM365ContentType(actionItem: unknown): actionItem is M365ContentAction {
-  // eslint-disable-next-line no-prototype-builtins
-  return actionItem && Object.prototype.hasOwnProperty.call(actionItem, 'secondaryId');
-}
-
-/**
  * Information common to all actions
  */
 export interface ActionInfo {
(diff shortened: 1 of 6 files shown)


packages/teams-js/src/public/interfaces.ts
```diff
@@ -180,7 +180,34 @@ export enum FileOpenPreference {
  * Represents the structure of the received context message.
  */
 
+export enum ActionObjectType {
+  M365Conontent = 'm365content',
```
TrevorJoelHarris, line 184: Got a couple extra letters here.

shrshindeMSFT: You beat me to it

shrshindeMSFT (approved):
LGTM!!

packages/teams-js/src/public/interfaces.ts
```diff
@@ -180,7 +180,34 @@ export enum FileOpenPreference {
...
+}
+
+export interface BaseActionObject<T extends ActionObjectType> {
```
TrevorJoelHarris, line 187: I think we need the M365Content type to be defined here, and then we should be testing it in the E2E test to make sure the data makes it down to the application from the host

TrevorJoelHarris (changes requested):
Left a couple of comments

packages/teams-js/src/public/interfaces.ts
Jessica-Koch: isn't the M365 content type above this interface?  Also,  you're exactly right, I plan on adding the e2e tests after this PR is merged!

Jessica-Koch: do you mean that we should not have the generic and we should _only_ have the M365Content type?  I wanted to keep these as flexible as possible since we don't know the types of objects we could add in the future

TrevorJoelHarris: Followed up with you offline. E2E tests after this merge sounds perfect, I mis-spoke and meant unit tests (app.spec.ts). You'll also need to add the M365ContentAction interface to this file.

Jessica-Koch: ok after speaking I realized I missed some items from the other PR.  I needed to specify thee actual content type.  I added these missing interfaces and a typeguard so that we can identify these from each other in the future

packages/teams-js/test/public/app.spec.ts
```diff
@@ -496,9 +543,17 @@ describe('Testing app capability', () => {
...
+          const secondActionItem = actualContext.actionInfo?.actionObjects[1];
+
           expect(actualContext).toEqual(expectedContext);
           expect(actualContext.page.frameContext).toBe(context);
-          expect(actualContext.meeting.id).toBe('dummyMeetingId');
+          expect(actualContext.meeting?.id).toBe('dummyMeetingId');
+          expect(actualContext.actionInfo?.actionId).toBe('actionId');
+          expect(actualContext.actionInfo?.actionObjects.length).toBe(5);
+          expect(firstActionItem.secondaryId?.name).toEqual(SecondaryM365ContentIdName.DriveId);
+          expect(isM365ContentType(secondActionItem)).toBe(false);
```
Jessica-Koch, line 556: I wanted to test the typeguard and also add all of the different objects.  I didn't think it was necessary to actually iterate through the 5 objects to test for equality but maybe that is an oversight on my part.  I would be more than happy to add that if necessary

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,54 @@ export enum FileOpenPreference {
   Web = 'web',
 }
 
+export enum ActionObjectType {
```
erinha, line 172: This needs documentation comments to satisfy the build failures.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,54 @@ export enum FileOpenPreference {
...
+  M365Content = 'm365content',
+}
+
+export interface BaseActionObject<T extends ActionObjectType> {
```
erinha, line 176: This also needs documentation comments. If it's appropriate to keep it hidden from public docs, please add the @hidden tag.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,54 @@ export enum FileOpenPreference {
...
+ */
+export interface ActionInfo {
+  /**
+   * Maps to the action id supplied inside the manifest
+   */
+  actionId: string;
+  actionObjects: BaseActionObject<ActionObjectType>[];
+}
+
+export interface M365ContentAction extends BaseActionObject<ActionObjectType.M365Content> {
```
erinha, line 191: I think these types also need reference doc comments

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,71 @@ export enum FileOpenPreference {
   Web = 'web',
 }
 
+/**
+ * Action Types
+ */
+export enum ActionObjectType {
```
erinha, line 175: Okay, so I figured out what the build error is about. All these types/enums/interfaces/etc. that you have added need to be exported from the interfaces file in the public/index.ts file. If they aren't exported from the SDK via the barrel file (the index.ts), then I think no documentation will get generated for them. 

When I synced your branch and explicitly exported them from the index.ts file, the errors got resolved.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,71 @@ export enum FileOpenPreference {
...
+  M365Content = 'm365content',
+}
+
+/**
+ * Data pertaining to object(s) the action is being performed on
+ */
+export interface BaseActionObject<T extends ActionObjectType> {
+  type: T;
```
erinha, line 183: I believe this `type: T` property may also need its own doc comment, not 100% sure though.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,71 @@ export enum FileOpenPreference {
...
+/**
+ * Common information which applies to all actions and an array of
+ * corresponding action objects
+ */
+export interface ActionInfo {
+  /**
+   * Maps to the action id supplied inside the manifest
+   */
+  actionId: string;
+  actionObjects: BaseActionObject<ActionObjectType>[];
```
erinha, line 195: `actionObjects` may need its own doc comment

packages/teams-js/src/public/interfaces.ts
TrevorJoelHarris: this is an interesting issue!

Jessica-Koch: this is extremely interesting, we should post this somewhere because I don't think I would have solved this on my own

packages/teams-js/src/public/interfaces.ts
```diff
@@ -169,6 +169,75 @@ export enum FileOpenPreference {
...
+  DriveId = 'driveId',
+  GroupId = 'groupId',
+  SiteId = 'siteId',
+  UserId = 'userId',
+}
+
+/**
+ * Type guard to determine if an action item is of M365Content Type
+ */
+export function isM365ContentType(actionItem: unknown): actionItem is M365ContentAction {
```
TrevorJoelHarris, line 222: This probably isn't the right place to put this since this file is for interfaces/enums and not functions. It also looks like this is only used in the test, so maybe more appropriate to put in the test file

TrevorJoelHarris: (We also wouldn't add public facing documentation for test only stuff no matter what file it's in)

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/interfaces.ts
Jessica-Koch: sure I will add it to the test, I put it there because I didn't know if we would be expanding this later.  Since we only have one type of content action I figured that this could be used later

TrevorJoelHarris (approved):

shrshindeMSFT (approved):
LGTM!

PR #1277, opened 2022-07-29
Remove intentionallyNotExported exceptions from documentation validation

erinha:
I finally figured out how to resolve errors that were previously worked around in the typedoc.json file. We need to explicitly export interfaces (and types, etc.) defined in the global namespace in order to have the referring documentation show up correctly.

Note my change in the media.ts file in order to make sure that `MediaControllerEvent` appears in the docs, since it's already referenced by other public documentation. The alternative would be to hide the documentation that refers to it.


TrevorJoelHarris:
It would be great to document the compiler errors not exporting documented types causes and how to fix it.

TrevorJoelHarris (approved):

aeunms (approved):
🚢

shrshindeMSFT (approved):

PR #1284, opened 2022-08-02
improve UI, add appID for 'Pages' & update manifest

AnanyaBarthakur1:


MarcinJuraszek (approved):

aeunms (approved):
I do see some of the same ' ' vs '' stuff as from the other PR, but I don't think it's a huge deal. And it's already being fixed in the other PR I think.

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -59,48 +69,55 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
...
+        <div>
+          {AvatarItemList.map(a => (
+            <Tooltip
+              content={
+                <>
+                  <Text weight="semibold" as="span">
+                    {a.name}
+                  </Text>
+                  <MenuList>
+                    {capabilities.map(capability => (
```
MarcinJuraszek, line 85: I think it's OK to leave it as is, but I'm curious why you think this is better than having the `isSupported` check here, something like:

```
{mail.isSupported() && <MenuItem onClick={() => handleMail(a)}>Mail</MenuItem>
```

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -59,48 +69,55 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
...
-            </Text>
-          </Tooltip>
-        ))}
-      </div>
+      <>
+        <Title3 block className="paddingClass">
+          People to Meet Today
+        </Title3>
+        <div>
+          {AvatarItemList.map(a => (
```
MarcinJuraszek, line 77: nit: use `avatar` instead of `a`, to make the loop body easier to read.

MarcinJuraszek (dismissed):

apps/sample-app/src/components/PeopleAvatars.tsx
AnanyaBarthakur1: I agree that this is more readable and simpler to understand. I'll try incorporating it this way and see if the UI remains the same (which seems like it should!) and push it in a future PR?

MarcinJuraszek: sounds good!

MarcinJuraszek (dismissed):

MarcinJuraszek (approved):

aeunms (approved):

PR #1289, opened 2022-08-03
Updated ReadMe + SSL certificate

AnanyaBarthakur1:
Instructions to run the app, verified browser certificate and key, better UI


apps/sample-app/src/App.tsx
```diff
@@ -20,12 +29,19 @@ const App: React.FC = () => {
         app.notifyAppLoaded();
         app.notifySuccess();
         const context = await app.getContext();
-        const themeNow = getTheme(context?.app?.theme);
-        setCurrTheme(themeNow);
-        app.registerOnThemeChangeHandler(function(theme) {
-          setCurrTheme(getTheme(theme));
-        });
+        if (context?.app?.host?.name === 'Teams') {
```
aeunms, line 32: I think we still want to avoid checking for specific hosts, right?

apps/sample-app/src/AuthApp.tsx
```diff
@@ -18,6 +18,7 @@ const AuthApp: React.FC = () => {
         await app.initialize();
         setIsInitialized(true);
       } catch (e) {
+        alert('initalization error catch');
```
aeunms, line 21: Nit: I think it might be more accurate from the user's perspective to see something like just 'initialization error'.

apps/sample-app/src/components/Emails.tsx
```diff
@@ -20,25 +21,27 @@ export const EmailList: React.FC<EmailProps> = (props: EmailProps) => {
...
+      <>
+        <Title3 className="paddingClass"> Recent Emails</Title3>
+        <div className="column">
+          {emailItems.map(emailItem => (
+            <Text as="span" key={emailItem.key}>
+              <DocumentCard key={emailItem.key} onClick={() => handleOpenMailItem(emailItem)}>
+                <DocumentCardTitle title={emailItem.subject || ''} shouldTruncate />
+                <DocumentCardActivity
+                  activity={emailItem.importance || ''}
+                  people={[{ name: emailItem.sender || ' ', profileImageSrc: '' }]}
```
aeunms, line 38: I've noticed some ' 's sprinkled in among ''s, maybe make them all '' for consistency? (Unless you specifically need them to be the ' ')

apps/sample-app/README.md
```diff
@@ -1,11 +1,36 @@
 # Sample App
 
-Sample app leveraging teams js v2 for multi host experience
+The sample app leverages teams js v2 SDK for creating a multi host app experience. The sample app displays the usage of the following teams js v2 capabilties:
```
aeunms, line 3: Nit: teams js -> TeamsJS

aeunms, line 3: Capabilities are the namespaces the APIs are in, so I think these would be described more accurately as "APIs"

apps/sample-app/README.md
```diff
@@ -1,11 +1,36 @@
...
 
-(I will be working on this section below in the coming few weeks)
+- Get context (app.getContext)
```
aeunms, line 5: I think it's a bit confusing to have a mix of actual APIs (e.g. isSupported()) and just describing (e.g. Opening Mail Item)

AnanyaBarthakur1: That makes sense! Would it be better to replace those with openMailItem(), startCallParams() etc, or remove those from the description itself?

apps/sample-app/src/App.tsx
AnanyaBarthakur1: I implement this so we can have the right host theme for the app. I think if we want the app to adjust to its host's theme we would have to check the host.name?

apps/sample-app/README.md
```diff
@@ -1,11 +1,61 @@
...
 
-(I will be working on this section below in the coming few weeks)
+- core.openLink
+- authentication.authenticate
+- isSupported
+- app.getContext
+- app.initialize
+- app.registerOnThemeChangeHandler
+- call.startCall
+- chat.OpenSingleChatRequest
```
aeunms, line 12: Nit: Capitalized OpenSingleChatRequest

apps/sample-app/README.md
```diff
@@ -1,11 +1,61 @@
...
+2. Edit 'devServer' in the webpack.config.js folder inside sample-app to point at the following files this way -
+   `devServer: {`
+   `// ...`
+   `https: {`
+   `key: fs.readFileSync("cert.key"),`
+   `cert: fs.readFileSync("cert.crt"),`
+   `ca: fs.readFileSync("ca.crt"),`
+   `},`
+   `// ....`
+   `},`
```
aeunms, line 59: You can do multi-line code snippets with triple backticks, like this:

\`\`\`
// code snippet
\`\`\`

apps/sample-app/src/App.tsx
```diff
@@ -20,12 +28,17 @@ const App: React.FC = () => {
...
```
aeunms, line 31: @MarcinJuraszek What do you think about a comment saying a disclaimer where it is ok to check for specific host name for details like theme but that we would not recommend checking the name to see if a capability is supported, and to simply use isSupported() calls in that case?

MarcinJuraszek: How about a link to https://docs.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/using-teams-client-sdk?tabs=javascript%2Cmanifest-teams-toolkit#differentiate-your-app-experience in a comment?

apps/sample-app/src/App.tsx
```diff
@@ -20,12 +28,17 @@ const App: React.FC = () => {
...
+          const themeNow = getTheme(context?.app?.theme);
+          setCurrTheme(themeNow);
+          app.registerOnThemeChangeHandler(function(theme) {
```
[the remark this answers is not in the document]
AnanyaBarthakur1: Yes I will be changing this part of the code to use getContext and adding a util function to get themes that aren't Teams

PR #1292, opened 2022-08-05
Added in electron testing to TeamsJS pipelines

jadahiya-MSFT:


MarcinJuraszek (approved):

TrevorJoelHarris (approved):

erinha:
(Sorry I'm late to the party.) Should these electron tests be added as separate jobs so they can be run in parallel with the web tests?

erinha:
> (Sorry I'm late to the party.) Should these electron tests be added as separate jobs so they can be run in parallel with the web tests?

Discussed offline, the overhead for running these tests is about < 1 minute, so I'm fine with this approach.

erinha (approved):
Thanks for doing this!

PR #1302, opened 2022-08-12
Trharris/internalcommentfixes

TrevorJoelHarris:
Functions marked with the @internal tag in comments are intended for Microsoft use only, and in general were being hidden from dev documentation. Unfortunately, the @internal and @hidden tags don't prevent exported functions from showing up in intellisense for developers. To help make it clear that @internal functions are only for internal use I have added the phrase "Limited to Microsoft-internal use" to *all* @internal functions in teamsjs-sdk.

As I went through, I also fixed a few other small problems:

- Some exported items in the private folder weren't marked as hidden, which made them show up in the dev documentation. @hidden tag has been added to all of these that I found
- the `initalizePrivateApis` function has been removed from privateAPIs.ts. It was not being exported in the index file and was a no-op, so there's no need to keep it in the package. Also removed associated calls to it and unit tests validating it was being called.

### DIFF
--- a/change/@microsoft-teams-js-8f021e4d-e427-4bb6-8811-25e6f8a2f811.json
+++ b/change/@microsoft-teams-js-8f021e4d-e427-4bb6-8811-25e6f8a2f811.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Updated comments on items marked with the `@internal` tag to make it clear they are intended for Microsoft use only. Removed app.initializePrivateApis, an unexported no-op function.",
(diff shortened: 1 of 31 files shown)


MarcinJuraszek:
Any chance you could share a screenshot of how this looks in VSCode when somebody is trying to use these methods now?

jadahiya-MSFT (approved):

TrevorJoelHarris:
> Any chance you could share a screenshot of how this looks in VSCode when somebody is trying to use these methods now?

image

MarcinJuraszek (approved):

erinha (approved):

PR #1306, opened 2022-08-16
optional: better accessible app

AnanyaBarthakur1:
- using aria labels 
- changed PeopleAvatar.tsx to use "Popover" instead of "MenuList" so that it can be tabbed through


apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -63,43 +62,48 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
   const AvatarExample: React.FunctionComponent = () => {
     return (
       <>
-        <Title3 block className="paddingClass">
+        <Title3 block className="paddingClass" id="People to Meet Today">
```
MarcinJuraszek, line 65: from developer.mozilla.org:

> id's value must not contain [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) (spaces, tabs etc.). Browsers treat non-conforming IDs that contain whitespace as if the whitespace is part of the ID. In contrast to the [class](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-class) attribute, which allows space-separated values, elements can only have one single ID value.

AnanyaBarthakur1: Thank you! I am going to go ahead and remove those ids, also realized we do not need it as the title text itself mentions what that section is about

MarcinJuraszek (approved):

MarcinJuraszek (approved):

aeunms (approved):
🚢

PR #1310, opened 2022-08-17
Added new liveShare hub

Stevenic:
- Added public `liveShare` namespace.
- Added internal `LiveShareHost` class that calls clients "interactive" hub.
- Added logic to dynamically import `@microsoft/live-share` package if installed as an app dependency.
- TODO: add unit tests...

To support this I had to make the following changes to the project as a whole:
- Upgraded prettier from 1.18.2 to 2.7.1. Needed to support importing only "types" from a package.
- Changed `module` property in "tsconfig.json" from "es6" to "es2020".  needed to support dynamic `import()` function.
- Added Fluid packages to `devDependencies` list. Needed to support importing Fluid types.

The above project changes were needed to avoid having to take an actual dependency on Live Share or Fluid.  The new `liveShare` namespace exposes all of the type information from Fluid without having to bloat the SDK's dependency tree.  Fluid itself is over 10 times larger then the SDK and pulls in nearly 1000 dependencies.  Developers should only have to pay the cost of Fluid if they're actually going to use Fluid.

The new `liveShare` namespace attempts to dynamically import the `@microsoft/live-share` package and will simply throw an error if the package can't be found. Developers can then simply add Live Share to their dependencies list to resolve the error.


MarcinJuraszek:
Would you mind rebasing on top of main? We made a separate change to update prettier and it should make it much easier to see what you're actually changing and what is just noise from that prettier update.

packages/teams-js/src/public/liveShare.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /**
+   * Initializes the Live Share client.
+   * @param options Optional. Configuration options passed to the Live Share client.
+   *
+   * @beta
+   */
+  export async function initialize(options?: LiveShareOptions): Promise<void> {
+    if (client) {
```
MarcinJuraszek, line 73: should there be some frame context check here to only allow these in a meeting context?

Stevenic:
> Would you mind rebasing on top of main? We made a separate change to update prettier and it should make it much easier to see what you're actually changing and what is just noise from that prettier update.

I had originally based my branch off main but looks like it was probably from before the prettier changes. I updated from main and all the prettier related stuff dropped out.

erinha:
This change requires a change file. See the section "CHANGE LOG Using Beachball" in the contribution guidelines for details: https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball.

Once the change file has been created, the Build Test Publish check will be able to run. Part of that check includes validating the size of the bundle created, which I'd like to see before approving this change.

Stevenic:
> This change requires a change file. See the section "CHANGE LOG Using Beachball" in the contribution guidelines for details: https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball.
> 
> Once the change file has been created, the Build Test Publish check will be able to run. Part of that check includes validating the size of the bundle created, which I'd like to see before approving this change.

I added the changelog entry and test are running now

erinha:
I synced your branch as part of reviewing these changes and it looks like the yarn.lock file needs to be included in this PR.

packages/teams-js/src/internal/liveShareHost.ts
```diff
@@ -0,0 +1,238 @@
...
+ */
+export enum UserMeetingRole {
+  guest = 'Guest',
+  attendee = 'Attendee',
+  presenter = 'Presenter',
+  organizer = 'Organizer',
+}
+
+/**
+ * State of the current Live Share sessions backing fluid container.
```
erinha, line 19: Should this also have a `@hidden` tag on it? Generally all the code that lives in the internal directory is hidden from public docs so I would expect to see `@hidden` throughout this file.

packages/teams-js/src/public/liveShare.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  /**
+   * @hidden
+   * Hide from docs
+   * ------
+   * Returns the LiveShareHost object. Called by existing apps that use the TeamsFluidClient
+   * directly. This prevents existing apps from breaking and will be removed when Live Share
+   * goes GA.
```
erinha, line 121: Probably good to add `@beta` to this even though it's hidden. Especially since the comment indicates it's going to be removed in the near future. (Can't hurt to add `@beta` in any case!)

packages/teams-js/src/public/liveShare.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Namespace to interact with the Live Share module-specific part of the SDK.
+ *
+ * @beta
+ */
+export namespace liveShare {
+  const LIVE_SHARE_PACKAGE = '@microsoft/live-share';
+  const LIVE_SHARE_HOST = new LiveShareHost();
+  let client: LiveShareClient;
+
+  interface LiveShareClient {
```
erinha, line 16: Can you add documentation comments for this interface please?

packages/teams-js/src/internal/liveShareHost.ts
erinha: If these are public interfaces that app developers are expected to interact with, I think they should be defined in a file in the public folder.

packages/teams-js/test/public/liveShare.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { LiveShareHost } from '../../src/internal/liveShareHost';
+import { liveShare } from '../../src/public/liveShare';
+
+describe('MicrosoftTeams-LiveShare', () => {
+  describe('initialize', () => {
+    it('should throw error due to package not being found', async () => {
```
erinha, line 6: Can you please add positive test cases for the functions defined in liveShare.ts?

change/@microsoft-teams-js-3d42bacc-ebe5-46c9-82ff-61a56c6863e5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added new liveShare hub",
```
erinha, line 3: Can you please add a little more detail to this (and if referencing a namespace or type in the code, please surround with backtick quotes, i.e., '\`')? Also "hub" and "host" tend to have a distinct meaning that I don't think this aligns with, so would prefer to use different wording.

Something like "Added \`liveShare\` capability to interact with @microsoft/live-share package, enabling the Live Share feature" or something?

packages/teams-js/src/internal/liveShareHost.ts
erinha: I think if they aren't meant to be exposed to app developers, then in addition to adding the `@hidden` tag, you might also want to add the following comments as well (this helps with VS Code intellisense, letting devs know they shouldn't call the functions in their apps):

```
 * @internal
 ```

**Review by erinha — CHANGES_REQUESTED** [review]
Mostly documentation comments

File: packages/teams-js/test/public/liveShare.spec.ts (hunk 5 above)
Comment (Stevenic) [reply packages/teams-js/test/public/liveShare.spec.ts:6]: That's really difficult to do because of the fact the code is trying to dynamically load a package dependency. We we need two separate version of the unit tests, one with the package dependency and one without. We'd need to brainstorm how that would work.

File: change/@microsoft-teams-js-3d42bacc-ebe5-46c9-82ff-61a56c6863e5.json (hunk 6 above)
Comment (Stevenic) [reply change/@microsoft-teams-js-3d42bacc-ebe5-46c9-82ff-61a56c6863e5.json]: Updated

File: packages/teams-js/src/internal/liveShareHost.ts (hunk 2 above)
Comment (Stevenic) [reply packages/teams-js/src/internal/liveShareHost.ts:24]: The public interfaces are exposed by `@microsoft/live-share`. There's a bit of a chicken and an egg here which is why I'm having to redefine things. Once everything is published I can go back in and remove some of these type definitions.

The `LiveShareHost` class needs to be internal as it's implementing the callbacks needed to host Live Share within an app like the Teams client.

I've updated everything to be marked as both `@hidden` and `@internal` although looking through otehr files in this folder there's a lot of inconsistency with this.

File: packages/teams-js/src/public/liveShare.ts (hunk 3 above)
Comment (Stevenic) [reply packages/teams-js/src/public/liveShare.ts:121]: Good catch... Updated

**Comment by Stevenic** [comment]
The build errors don't look to be related to anything in my check-in

File: packages/teams-js/test/public/liveShare.spec.ts (hunk 5 above)
Comment (erinha) [reply packages/teams-js/test/public/liveShare.spec.ts:6]: Discussed further offline. We'll work together post merging this PR to see how best to set up the tests so we can validate the positive scenarios.

**Review by erinha — APPROVED** [review]

File: packages/teams-js/src/public/liveShare.ts [hunk 7]
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Initializes the Live Share client.
+   * @param options Optional. Configuration options passed to the Live Share client.
+   *
+   * @beta
+   */
+  export async function initialize(options?: LiveShareOptions): Promise<void> {
+    if (client) {
```
MarcinJuraszek, line 73: Because you're calling into an async thing between the `if(client)` and when client is actually checked, there's potential for a race condition, even in single-threaded JavaScript context. Is that OK?

Stevenic: Good observation. I can protect against that

Stevenic: I added logic to protect against multiple calls to initialize()

erinha (approved):

MarcinJuraszek (approved):

PR #1318, opened 2022-08-19
Fix issue where prettier 2.x upgrade caused line ending weirdness in windows

TrevorJoelHarris:

## Description

Prettier upgrade caused EOL default to change from 'auto' to 'lf'. Windows default is CLRF and our .gitattributes file didn't specify a specific line ending for everyone

This will switch everyone to LF. I *think* this will work for everyone but would be good to get a mac user to pull this branch and verify

### Main changes in the PR:

1. Update .gitattributes file to say that repo should use lf line endings
2. Add LF explicitly to prettier options in .prettierrc.js
3. Fix two small prettier issues that cropped up due to someone else checking in during the original prettier upgrade

### Validation performed:

1. Did a full clean/install/build and verified that every file in the repo wasn't marked as changed with updated line endings

### Unit Tests added:

No tests needed, only updating repo settings and prettier fixes

### End-to-end tests added:

No

### Change file added:

No

I'm actually kind of surprised, I would have thought the prettier fixes would need one.


TrevorJoelHarris:
For more details on how I figured out how to fix this, see here: https://prettier.io/docs/en/options.html#end-of-line

erinha (approved):
🚀

erinha:
> ### Change file added:
> No
> 
> I'm actually kind of surprised, I would have thought the prettier fixes would need one.

Change files are only required for changes under the packages/teams-js directory. Since none of yours affect files in that location, you're good!

If you're concerned that there are changes being made that should produce a change file in our change log for consumers of the teams-js package, we can discuss that further (I don't think these qualify personally).

erinha:
> This will switch everyone to LF. I _think_ this will work for everyone but would be good to get a mac user to pull this branch and verify

Verified in my local repo that after cleaning, installing, and building, I see no changes in the repo.

MarcinJuraszek (approved):

PR #1324, opened 2022-08-24
fix wrongly returned values

aeunms:
## Description

The test app's chat capability currently has bugs not being caught by the e2e testing. openChat(), openGroupChat(), and openConversation() show that there was "no response from the hub sdk" regardless of whether the promise was successfully resolved or not. This PR adds the elements that have the fix while also preserving the current content while the e2e test is updated. 

### Main changes in the PR:

1. Add elements with fix to accurately reflect API call result

### Validation performed:

1. Launch Orange and run the branch's test app in it. 

### Unit Tests added:

No unit tests for test app

### End-to-end tests added:

Will add e2e tests once it can expect main to contain the new elements. Please let me know if I should do it earlier and if I am missing documentation for the correct steps. 

### Change file added:
`yarn changefile` was run and no changefile was needed.

### Next/remaining steps:

> Remove this section if n/a

- [ ] Need to switch the e2e test onto the new elements
- [ ] After above step, delete deprecated elements

### Screenshots:

Screen Shot 2022-08-24 at 4 32 54 PM

### DIFF
--- a/apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
@@ -18,10 +18,10 @@ const CheckChatCapability = (): React.ReactElement =>
     onClick: async () => `Chat module ${chat.isSupported() ? 'is' : 'is not'} supported`,
   });
 
-const OpenChat = (): React.ReactElement =>
+const DeprecatedOpenChat = (): React.ReactElement =>
   ApiWithTextInput<OpenSingleChatRequest>({
     name: 'openChat',
-    title: 'Open Chat',
+    title: '[Deprecated] Open Chat',
     onClick: {
       validateInput: (input) => {
         if (!input.user) {
@@ -35,10 +35,33 @@ const OpenChat = (): React.ReactElement =>
     },
   });
 
-const OpenGroupChat = (): React.ReactElement =>
+const OpenChat = (): React.ReactElement =>
+  ApiWithTextInput<OpenSingleChatRequest>({
+    name: 'openChat2',
+    title: 'Open Chat',
+    onClick: {
+      validateInput: (input) => {
+        if (!input.user || !input.message) {
+          throw new Error('Both user and message are required on the input');
+        }
+      },
+      submit: async (input) => {
+        return await chat
+          .openChat(input)
+          .then(() => {
+            return 'openChat was called';
+          })
+          .catch(() => {
+            return 'openChat' + noHostSdkMsg;
+          });
+      },
+    },
+  });
+
+const DeprecatedOpenGroupChat = (): React.ReactElement =>
   ApiWithTextInput<OpenGroupChatRequest>({
     name: 'openGroupChat',
-    title: 'Open Group Chat',
+    title: '[Deprecated] Open Group Chat',
     onClick: {
       validateInput: (input) => {
         if (!input.users) {
@@ -52,7 +75,30 @@ const OpenGroupChat = (): React.ReactElement =>
     },
   });
 
-const OpenConversation = (): React.ReactElement =>
+const OpenGroupChat = (): React.ReactElement =>
+  ApiWithTextInput<OpenGroupChatRequest>({
+    name: 'openGroupChat2',
+    title: 'Open Group Chat',
+    onClick: {
+      validateInput: (input) => {
+        if (!input.users) {
+          throw new Error('users is required on the input');
+        }
+      },
+      submit: async (input) => {
+        return await chat
+          .openGroupChat(input)
+          .then(() => {
+            return 'openGroupChat was called';
+          })
+          .catch(() => {
+            return 'chat.openGroupChat()' + noHostSdkMsg;
+          });
+      },
+    },
+  });
+
+const DeprecatedOpenConversation = (): React.ReactElement =>
   ApiWithTextInput<OpenConversationRequest>({
     name: 'openConversation2',
     title: 'Open Conversation',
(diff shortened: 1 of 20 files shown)


apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
```diff
@@ -35,10 +35,33 @@ const OpenChat = (): React.ReactElement =>
...
+    name: 'openChat2',
+    title: 'Open Chat',
+    onClick: {
+      validateInput: (input) => {
+        if (!input.user || !input.message) {
+          throw new Error('Both user and message are required on the input');
+        }
+      },
+      submit: async (input) => {
+        return await chat
```
MarcinJuraszek, line 49: Using `async`/`await` AND `.then().catch()` seems wrong. I think you could just remove the `async`/`await` part here, or if you want to continue using it, it should look more like this:

```
try {
    await chat.openChat(input);
    return 'openChat was called';
} catch () {
    return 'openChat - error';
}
```

MarcinJuraszek: Same for the other new elements you're adding.

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
```diff
@@ -35,10 +35,33 @@ const OpenChat = (): React.ReactElement =>
...
+          .openChat(input)
+          .then(() => {
+            return 'openChat was called';
+          })
+          .catch(() => {
+            return 'openChat' + noHostSdkMsg;
```
MarcinJuraszek, line 55: It's not really that there wasn't any response, it's that the promise was rejected which means something went wrong. Right?

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
aeunms: I tried doing this but got some errors saying it's not returning in all cases, could you help navigate the error? I tried adding another return after the try catch but for some reason it would return the success case and then return the last return as well. (How does that work anyways?? Shouldn't it exit the function as soon as it returns from the try catch? Weird)

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
aeunms: True. My understanding was that we use the noHostSdkMsg for both cases but I will see how other APIs are handled.

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
aeunms: Oh wait, my bad- it was another case where I would see both the returns happen. In this case, I don't see the "you're not returning error happening", let me try it all the way through.

aeunms: I now can remember that we wanted to stay away from using both async/await and .then/.catch at the same time, but I am not sure that if `chat.openChat(input)` rejects, that it'll go to the catch block. I think it's possible that it expects the reject to be caught only through `.catch(() => {})` right on the promise, but I'm having trouble googling the answer. I can see that if the API is not supported by the hub sdk that it would go into the catch block however.

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
MarcinJuraszek: What did you find when you looked at other APIs?

MarcinJuraszek: To me it feels like there are three possible scenarios:
1. We made the call to the host, and didn't hear back yet
2. We heard back and it's a success (yay!)
3. We heard back and got some errors back (sad :().

It feel like having 1. and 3. result in the same thing printed on the screen makes the tests not 100% reliable, because we can't really say if the error made it back to the app itself.

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
aeunms: Chatted offline, it did indeed turn out to be a trivial thing to test for, thanks!

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
aeunms: I agree, I think we need to differentiate between all 1, 2, and 3. From looking at and testing various APIs like `setConfig` and `getAuthToken` as well as the structure in the test app components like `ApiWithTextInput` I think I better understand how the rejections and manually throwing errors affect the test app result. Could you confirm my following understanding?

```
// I think this is a correct way to write a test app element
await <some function that returns a promise>
setResult('complete')
```
I think: 
* If the above line 1 succeeds -> line 2 setResult('complete') is called
* If line 1 fails -> ApiWithTextInput catches it and calls line 59 in `apps/teams-test-app/src/components/utils/ApiWithTextInput.tsx`, which is `setResult('Error: ' + err);`. Hence the error is displayed

```js
// and so something like below is a bug, since noHostSdkMsg would be displayed only when `chat.openGroupChat(input)` call succeeds?
const OpenGroupChat = (): React.ReactElement =>
  ApiWithTextInput<OpenGroupChatRequest>({
    name: 'openGroupChat',
    title: 'Open Group Chat',
    onClick: {
      validateInput: (input) => {
        if (!input.users) {
          throw new Error('users is required on the input');
        }
      },
      submit: async (input) => {
        await chat.openGroupChat(input);
        return 'chat.openChat()' + noHostSdkMsg;
      },
    },
  });
```

aeunms: And with the first "correct" example I'm thinking the noHostSdkMsg is first displayed by the ApiWithTextInput and stays that way unless we have a new result from the call, so we usually wouldn't need to manually show the noHostSdkMsg, maybe not ever?

MarcinJuraszek: That matches my understanding.

MarcinJuraszek (approved):

AE-MS (approved):
:shipit:

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
aeunms: I thought I left a couple comments after this but I guess it didn't actually get posted 🤦🏼‍♀️  thank you for the knowledge, I basically had just said I agree all 1 2 3 should be covered with appropriate responses for each and listed a few examples I found to understand how to do so correctly. Since the PR was approved I'm assuming I did it correctly 😋

PR #1333, opened 2022-09-01
3p cloud storage provider upload action API fixes

HarshJain27:
## Description
This PR has following three changes and corresponding test code update
1. Renamed filePath field to webkitRelativePath : This is required because the Blob object has specific fields which needs to be retained for passing the blob object from consuming app (One Drive File Browser in this case) to the SDK successfully without any data loss, without which the feature doesn't work.
2. Removed two validation checks for destinationFolder fields : This change is required for cases when file is uploaded to root folder of a 3P provider. In that scenario both **objectUrl** and **isSubdirectory** fields are not applicable and doesn't make sense to perform these validation.
3. Added an optional field provider in callback for add cloud storage API: This change is done to ensure that callers of add 3P provider gets the provider value in valid scenarios. 

Below PRs contain the original 3P API changes.
https://github.com/OfficeDev/microsoft-teams-library-js/pull/1152
https://github.com/OfficeDev/microsoft-teams-library-js/pull/1251

File uploads to 3P providers should work as expected after this change.

### Unit Tests added:

No, Modified the unit tests as applicable.

### Change file added:
Yes

### DIFF
--- a/change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
+++ b/change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "1. Renamed `filePath` field to `webkitRelativePath` 2. Removed two validation checks for `destinationFolder` fields",
+  "comment": "1. Renamed `filePath` field to `webkitRelativePath` 2. Removed two validation checks for `destinationFolder` fields 3. Added an optional field `provider` in callback of `addCloudStorageProvider` API",
   "packageName": "@microsoft/teams-js",
   "email": "jainharsh@microsof.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
+++ b/change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "1. Renamed filePath field to webkitRelativePath 2. Removed two validation checks for destinationFolder fields",
+  "comment": "1. Renamed `filePath` field to `webkitRelativePath` 2. Removed two validation checks for `destinationFolder` fields",
   "packageName": "@microsoft/teams-js",
   "email": "jainharsh@microsof.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
+++ b/change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "1. Renamed filePath field to webkitRelativePath 2. Removed two validation checks for destinationFolder fields",
+  "packageName": "@microsoft/teams-js",
+  "email": "jainharsh@microsof.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/private/files.ts
+++ b/packages/teams-js/src/private/files.ts
@@ -816,7 +816,9 @@ export namespace files {
    * for selected 3P provider is performed on success of which the selected 3P provider support is added for user
    * @beta
    *
-   * @param callback Callback that will be triggered post add 3P cloud storage provider action
+   * @param callback Callback that will be triggered post add 3P cloud storage provider action.
+   * If the error is encountered (and hence passed back), no provider value is sent back.
+   * For success scenarios, error value will be passed as null and a valid provider value is sent.
    *
    * @internal
    * Limited to Microsoft-internal use
--- a/packages/teams-js/src/private/files.ts
+++ b/packages/teams-js/src/private/files.ts
@@ -431,7 +431,7 @@ export namespace files {
     /**
      * The file path to uniquely identify it within the file hierarchy
      */
-    filePath: string;
+    webkitRelativePath?: string;
   }
 
   /**
@@ -1046,15 +1046,7 @@ export namespace files {
       );
     }
 
-    if (
-      !(
-        uploadFileRequest.content.destinationFolder &&
-        (uploadFileRequest.content.providerCode === CloudStorageProvider.SharePoint
-          ? (<ISharePointFile>uploadFileRequest.content.destinationFolder).isFolder
-          : (<CloudStorageFolderItem>uploadFileRequest.content.destinationFolder).isSubdirectory) &&
-        uploadFileRequest.content.destinationFolder.objectUrl
-      )
-    ) {
+    if (!uploadFileRequest.content.destinationFolder) {
       throw getSdkError(
         ErrorCode.INVALID_ARGUMENTS,
         '[files.uploadCloudStorageProviderFile] Invalid destination folder details',
--- a/packages/teams-js/src/private/files.ts
+++ b/packages/teams-js/src/private/files.ts
@@ -534,6 +534,7 @@ export namespace files {
   /**
    * @hidden
    * Hide from docs
+   * @beta
    *
    * Interface representing 3P cloud storage provider upload existing file(s) action
    *
--- a/packages/teams-js/src/private/files.ts
+++ b/packages/teams-js/src/private/files.ts
@@ -1015,6 +1015,7 @@ export namespace files {
    * Hide from docs
    *
    * Initiates the upload 3P cloud storage file(s) flow, which will upload file(s) to the given 3P provider
+   * @beta
    *
    * @param uploadFileRequest 3P cloud storage provider upload file(s) action request content
    * @param callback Callback that will be triggered post uploading file(s) flow is finished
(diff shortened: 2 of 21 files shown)


packages/teams-js/test/private/files.spec.ts
```diff
@@ -576,7 +576,7 @@ describe('files', () => {
     it('should send the message to parent correctly', () => {
       utils.initializeWithContext('content');
 
-      const callback = jest.fn((err) => {
```
MarcinJuraszek, line 579: please make sure to sync your main branch to match what's top of the repo, and run yarn. We've update prettier recently and looks like you don't have these changes because you're re-formatting the files with old patterns.

HarshJain27: I had synced the code in my local repo with the forked repo. Furthermore, i have ran **yarn install** followed by **yarn build** which caused the changes in the test file which is being pointed out here. As per offline conversation with you resolving this.

erinha: To add a little more guidance here, once you sync the latest changes, `yarn install`, and build the repo, you may see a ton of files modified after the build (likely with no actual content changes). (This seems to be more prevalent on Windows machines than Macs.)
 
There are 2 options to fix this in your local repo.
 
Option 1:	
Delete your local repo and re-clone.

Option 2:
OR, from the root:
	

1. `yarn clean:full`
2. `yarn install`
3. `yarn build`
4. `git restore .` // This will undo all the changes that the build created.
5. Repeat steps 1-3.
6. No changes should exist.

change/@microsoft-teams-js-f9182c21-8a37-405a-9fa4-9d394fece0dd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "1. Renamed filePath field to webkitRelativePath 2. Removed two validation checks for destinationFolder fields",
```
TrevorJoelHarris, line 3: References to code (like filePath, webkitRelativePath, etc.) should be surrounded by ``

packages/teams-js/test/private/files.spec.ts
```diff
@@ -1098,38 +1032,38 @@ describe('files', () => {
...
-      ).toThrowError('[files.uploadCloudStorageProviderFile] Invalid destination folder details');
-    });
-
-    it('should not allow upload calls for file as destination folder in request content', async () => {
-      await utils.initializeWithContext('content');
-      expect(() =>
-        files.uploadCloudStorageProviderFile(uploadFileRequestWithFileAsDestinationFolder, emptyCallback),
-      ).toThrowError('[files.uploadCloudStorageProviderFile] Invalid destination folder details');
-    });
+    // it('should not allow upload calls for non sharepoint file as destination folder in request content', async () => {
```
TrevorJoelHarris, line 1035: Commented out unit tests shouldn't be checked in

packages/teams-js/src/private/files.ts
```diff
@@ -431,7 +431,7 @@ export namespace files {
     /**
      * The file path to uniquely identify it within the file hierarchy
      */
-    filePath: string;
+    webkitRelativePath?: string;
```
TrevorJoelHarris, line 434: Given that this is renaming a field of an exported type that isn't marked with beta, how are you planning on making this not break existing users of this interface?

TrevorJoelHarris (changes requested):

packages/teams-js/test/private/files.spec.ts
HarshJain27: Removed commented code.

packages/teams-js/src/private/files.ts
HarshJain27: The feature is in development phase and have not been released for anyone, so no one is using this **File** interface and setting **filePath** field right now. In the process of consuming these changes we realised that **filePath** field will be problematic and hence came back to address it here in SDK.

erinha: I think that given the offline conversation about the usage of this interface and the recency and the fact that it's private, this change should be okay. However, we should add @beta tag to this interface and any other interfaces or APIs that are in active development to ensure you can continue to adjust the API until it's customer-ready.

erinha: @TrevorJoelHarris please correct me if you disagree.

packages/teams-js/test/private/files.spec.ts
```diff
@@ -442,7 +442,7 @@ describe('files', () => {
     it('should trigger callback correctly', async () => {
       await utils.initializeWithContext('content');
 
-      const callback = jest.fn((err) => {
```
erinha, line 445: These changes look like the issue Marcin pointed out about pulling from `main` and re-running `yarn` haven't been resolved yet.

packages/teams-js/src/private/files.ts
HarshJain27: Applied @beta tag to the upload API. From now onwards we will definitely follow the practice of marking the API / interfaces beta until they become fully customer ready.

packages/teams-js/test/private/files.spec.ts
HarshJain27: Thanks for adding detailed guidance Erin. Both the workarounds didn't help me. I tried in both Mac and Windows machines. Hence going ahead with these re-formatted changes to the files.spec.ts file with old patterns. Let me know if i am still missing something here.

TrevorJoelHarris (dismissed):

erinha (approved):

MarcinJuraszek (approved):

packages/teams-js/test/private/files.spec.ts
```diff
@@ -576,15 +576,16 @@ describe('files', () => {
...
         expect(err).toBeFalsy();
+        expect(provider).toEqual(files.CloudStorageProvider.Dropbox);
       });
 
       files.addCloudStorageProvider(callback);
 
       const addCloudStorageProviderMessage = utils.findMessageByFunc('files.addCloudStorageProvider');
       expect(addCloudStorageProviderMessage).not.toBeNull();
-      utils.respondToMessage(addCloudStorageProviderMessage, false);
+      utils.respondToMessage(addCloudStorageProviderMessage, false, files.CloudStorageProvider.Dropbox);
```
TrevorJoelHarris, line 588: What does the false do here? I don't see any bool params in your function or callback

packages/teams-js/test/private/files.spec.ts
```diff
@@ -576,15 +576,16 @@ describe('files', () => {
     it('should send the message to parent correctly', () => {
       utils.initializeWithContext('content');
 
-      const callback = jest.fn((err) => {
+      const callback = jest.fn((err, provider) => {
         expect(err).toBeFalsy();
+        expect(provider).toEqual(files.CloudStorageProvider.Dropbox);
```
TrevorJoelHarris, line 581: I would expect to see some unit tests for the case when an error is returned but no provider

packages/teams-js/src/private/files.ts
```diff
@@ -813,13 +814,14 @@ export namespace files {
...
    * for selected 3P provider is performed on success of which the selected 3P provider support is added for user
+   * @beta
    *
    * @param callback Callback that will be triggered post add 3P cloud storage provider action
    *
    * @internal
    * Limited to Microsoft-internal use
    */
-  export function addCloudStorageProvider(callback: (error?: SdkError) => void): void {
+  export function addCloudStorageProvider(callback: (error?: SdkError, provider?: CloudStorageProvider) => void): void {
```
TrevorJoelHarris, line 824: Is it valid for the callback to have both a provider error at the same time? You should have some guidance in comments to developers about what to expect from this callback now that it's more complex

TrevorJoelHarris (changes requested):

packages/teams-js/test/private/files.spec.ts
HarshJain27: image

As mentioned in doc comments of **toBeFalsy** method, in success scenario, we are not caring for error value and hence passed it as false instead of null for the test case.

packages/teams-js/src/private/files.ts
HarshJain27: Updated the documentation. Thanks for pointing out, this will helps folks.

packages/teams-js/test/private/files.spec.ts
HarshJain27: Added the failure test case as well.

packages/teams-js/test/private/files.spec.ts
```diff
@@ -576,15 +576,36 @@ describe('files', () => {
...
+        expect(err).toEqual(sdkError);
       });
 
       files.addCloudStorageProvider(callback);
 
       const addCloudStorageProviderMessage = utils.findMessageByFunc('files.addCloudStorageProvider');
       expect(addCloudStorageProviderMessage).not.toBeNull();
-      utils.respondToMessage(addCloudStorageProviderMessage, false);
+      utils.respondToMessage(addCloudStorageProviderMessage, sdkError);
       expect(callback).toHaveBeenCalled();
```
erinha, line 609: Can you add verification that the `provider` is not sent back in the error case?

erinha (changes requested):
Unit test request

packages/teams-js/test/private/files.spec.ts
HarshJain27: Added the verification.

erinha (approved):

TrevorJoelHarris (approved):

PR #1335, opened 2022-09-07
Register handler fix

### LINKED ISSUE #1331: register*Handler back compat APIs are throwing errors in v2, but not v1
The back compatibility layer seems to have a bug for many (all?) `register*Handler` APIs. In v1 application developer was able to call them even if in a given context particular `register*Handler` API made no sense - the callback would simply never get invoked. With v2 these `register*Handler` APIs have been moved to appropriate capabilities, a set of `isSupported` checks has been added and the old functions were preserved as back compatibility layer. However, the back compatibility layer calls into the new function in v2 capabilities, which means they run the `isSupported` check and ultimately fail. This is a breaking change, as a v1 developer would expect that using the back compatibility layer 'just works' without change in behavior for these APIs.

An example:

In v1 `registerChangeSettingsHandler` works just fine even in a context that's not related to Configurable Tabs. App can register for it, and will simply never be called. With v2 the same function throws `{ errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM }` because the `isSupported` check in `pages.config` fails.

erinha:

## Description

Previously, the v1 versions of the `register*Handler` functions would throw if the v2 version of the API's capability was not supported, breaking backwards compatibility.

E.g.,:
If `pages.config` was not supported, calling `registerChangeSettingsHandler` would throw `ErrorCode.NOT_SUPPORTED_ON_PLATFORM`. 

This changes ensures that the v1 versions do not throw if the v2 capability is not supported, while keeping the v2 version's `isSupported` check present.

Fixes #1331 

### Main changes in the PR:

1. Added a register*HandlerHelper function in the v2 version of the API's capability. This helper takes an additional optional closure parameter that represents the v2-specific functionality that should still be executed. 
2. Updated the v2 register*Handler function to call the helper and pass in the `isSupported` check as the additional parameter.
3. Updated the v1 register*Handler function to call the helper without passing any additional parameter.
4. Note that I had to export these helper functions in order to call them in the v1 versions, but I added documentation comments to mark them hidden and internal and effectively hide them from the public docs.

### Validation performed:

1. Added unit tests prior to implementing the fix to ensure existing (incorrect) behavior was correctly failing the tests (yay TDD!).

### Unit Tests added:

Yes

### End-to-end tests added:

Will add separately. Don't want to block this PR as it needs to get into the next release.

### Change file added:

Yes

### Next/remaining steps:

- [ ] Add e2e tests

### DIFF
--- a/apps/sample-app/src/components/PeopleAvatars.tsx
+++ b/apps/sample-app/src/components/PeopleAvatars.tsx
@@ -66,8 +66,7 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
           People to Meet Today
         </Title3>
         <div>
-
-          {AvatarItemList.map(avatar => (
+          {AvatarItemList.map((avatar) => (
             <>
               <Popover trapFocus openOnHover={true} size="medium">
                 <PopoverTrigger>
--- a/packages/teams-js/src/public/pages.ts
+++ b/packages/teams-js/src/public/pages.ts
@@ -50,6 +50,27 @@ export namespace pages {
   }
 
   /**
+   * @hidden
+   * Undocumented helper function with shared code between deprecated version and current version of the registerFocusEnterHandler API.
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @param handler - The handler for placing focus within the application.
+   * @param versionSpecificHelper - The helper function containing logic pertaining to a specific version of the API.
+   */
+  export function registerFocusEnterHandlerHelper(
+    handler: (navigateForward: boolean) => void,
+    versionSpecificHelper?: () => void,
+  ): void {
+    ensureInitialized();
+    if (versionSpecificHelper) {
+      versionSpecificHelper();
+    }
+    registerHandler('focusEnter', handler);
+  }
+
+  /**
    * Sets/Updates the current frame with new information
    *
    * @param frameInfo - Frame information containing the URL used in the iframe on reload and the URL for when the
@@ -212,6 +233,27 @@ export namespace pages {
   }
 
   /**
+   * @hidden
+   * Undocumented helper function with shared code between deprecated version and current version of the registerFullScreenHandler API.
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @param handler - The handler to invoke when the user toggles full-screen view for a tab.
+   * @param versionSpecificHelper - The helper function containing logic pertaining to a specific version of the API.
+   */
+  export function registerFullScreenHandlerHelper(
+    handler: (isFullScreen: boolean) => void,
+    versionSpecificHelper?: () => void,
+  ): void {
+    ensureInitialized();
+    if (versionSpecificHelper) {
+      versionSpecificHelper();
+    }
+    registerHandler('fullScreenChange', handler);
+  }
+
+  /**
    * Checks if the pages capability is supported by the host
    * @returns true if the pages capability is enabled in runtime.supports.pages and
    * false if it is disabled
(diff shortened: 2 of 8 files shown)


packages/teams-js/src/public/pages.ts
```diff
@@ -567,9 +671,27 @@ export namespace pages {
...
+     * @param versionSpecificHelper - The helper function containing logic pertaining to a specific version of the API.
+     */
+    export function registerBackButtonHandlerHelper(handler: () => boolean, versionSpecificHelper?: () => void): void {
       ensureInitialized();
-      if (!isSupported()) {
-        throw errorNotSupportedOnPlatform;
+      if (versionSpecificHelper) {
+        versionSpecificHelper();
       }
       backButtonPressHandler = handler;
```
TrevorJoelHarris, line 696: I don't think I understand what this handler is for. We're re-assigning it to a new variable, and then ignoring backButtonPressHandler. We then check handler but don't register handler or send it to the host. I don't *think* this is a product of your change, but is it something we need to fix?

packages/teams-js/src/public/pages.ts
```diff
@@ -674,9 +832,27 @@ export namespace pages {
...
+     * @hidden
+     * Undocumented helper function with shared code between deprecated version and current version of the onHoverLeave API.
+     *
+     * @internal
+     * Limited to Microsoft-internal use
+     *
+     * @param handler - The handler to invoke when existing hover of the personal app button in the app bar.
+     * @param versionSpecificHelper - The helper function containing logic pertaining to a specific version of the API.
+     */
+    export function onHoverLeaveHelper(handler: () => void, versionSpecificHelper?: () => void): void {
```
TrevorJoelHarris, line 852: All of these helpers look very similar. Is there some way we can have a single shared helper function for this case? Maybe taking in the message name, a generic handler object, and the versionSpecificHelper closure?

apps/sample-app/src/components/PeopleAvatars.tsx
```diff
@@ -66,8 +66,7 @@ export const PeopleAvatarList: React.FC<AvatarProps> = (props: AvatarProps) => {
           People to Meet Today
         </Title3>
         <div>
-
-          {AvatarItemList.map(avatar => (
+          {AvatarItemList.map((avatar) => (
```
TrevorJoelHarris, line 69: ?

jadahiya-MSFT (approved):

packages/teams-js/src/public/pages.ts
jadahiya-MSFT: Hmm the `backButtonPressedHandler` in general seems to follow different conventions. We might want to table this for now and investigate why it follows this format for later.

apps/sample-app/src/components/PeopleAvatars.tsx
erinha: I think this is a leftover Prettier change? Not sure, but I didn't actually edit this file.

packages/teams-js/src/public/pages.ts
erinha: I have gone over this. There isn't a way to do this without leaving _some_ duplicated code present in the library. Either we have duplicate handler helpers (as it is currently) or we have one shared helper function and duplicated code in both v1 and v2 versions of each handler that sets up the `ensureInitialized` call (since that is different across the register*Handler calls), and duplicated code to pass the name of the handler around. 

My preference given that this needs to get into the next release is to leave it as-is. If anyone feels strongly that they would prefer the duplicated code in each of the v1/v2 versions of the register*Handler calls, I will open a work item to track fixing it post-next-release.

packages/teams-js/src/public/pages.ts
erinha: So we do register the handler on line 697. We also assign it to a variable that is used in the `handleBackButtonPress` function. Maybe I'm misunderstanding the concern here?

erinha: This is very interesting indeed. We should follow up on this separate from this change, since I didn't modify it. I don't think this warrants a fix at this time.

packages/teams-js/src/public/pages.ts
TrevorJoelHarris: I think it's fine to leave as-is: if it becomes a problem we can revisit in the future. We've talked about it offline and I recognize that this is a slightly trickier problem than we thought.

fwiw, I don't think having duplicated code between v1 and v2 is bad: I think actually pulling apart v1 and v2 functions more over time might help not introduce bugs like this

packages/teams-js/src/public/pages.ts
TrevorJoelHarris: @erinha Please open a bug or task to look into this. Fine with leaving it since you haven't added to or changed the problem

TrevorJoelHarris (approved):

PR #1340, opened 2022-09-08
Adding logging for runtime initialization and pages.config functions

TrevorJoelHarris:
Using the https://www.npmjs.com/package/debug package. Strings use printf formatting (%i and %d are interchangeable)

## Description

Added logging using the https://www.npmjs.com/package/debug package. Strings use printf formatting (%i and %d are interchangeable). This logging should help track down issues around app initialization, setting the runtime object, and use of pages.config.setValidityState.

### Main changes in the PR:

1. Added telemetry to runtime.ts
2. Added telemetry to app.ts
3. Added telemetry to pages.ts

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

No, only telemetry added, no functional changes

### End-to-end tests added:

No

### Change file added:

Yes

### Next/remaining steps:

> Need to verify that the telemetry works as expected

### DIFF
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -591,7 +591,6 @@ export namespace app {
                   clientSupportedSDKVersion && applyRuntimeConfig(givenRuntimeConfig);
                 } catch (e) {
                   if (e instanceof SyntaxError) {
-                    initializeHelperLogger('Generating back compat runtime config');
                     applyRuntimeConfig(generateBackCompatRuntimeConfig(GlobalVars.clientSupportedSDKVersion));
                   } else {
                     throw e;
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -14,6 +14,7 @@ import { defaultSDKVersionForCompatCheck, version } from '../internal/constants'
 import { GlobalVars } from '../internal/globalVars';
 import * as Handlers from '../internal/handlers'; // Conflict with some names
 import { ensureInitialized, processAdditionalValidOrigins } from '../internal/internalAPIs';
+import { getLogger } from '../internal/telemetry';
 import { compareSDKVersions, runWithTimeout } from '../internal/utils';
 import { logs } from '../private/logs';
 import { authentication } from './authentication';
@@ -31,6 +32,8 @@ import { teamsCore } from './teamsAPIs';
  * @beta
  */
 export namespace app {
+  const appLogger = getLogger('app');
+
   // ::::::::::::::::::::::: MicrosoftTeams client SDK public API ::::::::::::::::::::
 
   export const Messages = {
@@ -540,6 +543,7 @@ export namespace app {
     );
   }
 
+  const initializeHelperLogger = appLogger.extend('initializeHelper');
   function initializeHelper(validMessageOrigins?: string[]): Promise<void> {
     return new Promise<void>((resolve) => {
       // Independent components might not know whether the SDK is initialized so might call it to be safe.
@@ -550,6 +554,11 @@ export namespace app {
         Handlers.initializeHandlers();
         GlobalVars.initializePromise = initializeCommunication(validMessageOrigins).then(
           ({ context, clientType, runtimeConfig, clientSupportedSDKVersion = defaultSDKVersionForCompatCheck }) => {
+            initializeHelperLogger('context: %s', context);
+            initializeHelperLogger('clientType: %s', clientType);
+            initializeHelperLogger('runtimeConfig: %s', runtimeConfig);
+            initializeHelperLogger('clientSupportedSDKVersion: %s', clientSupportedSDKVersion);
+
             GlobalVars.frameContext = context;
             GlobalVars.hostClientType = clientType;
             GlobalVars.clientSupportedSDKVersion = clientSupportedSDKVersion;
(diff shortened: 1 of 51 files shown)


packages/teams-js/src/public/runtime.ts
```diff
@@ -206,8 +211,15 @@ export const versionConstants: Record<string, Array<ICapabilityReqs>> = {
...
  */
 export function generateBackCompatRuntimeConfig(highestSupportedVersion: string): IRuntime {
+  generateBackCompatRuntimeConfigLogger('generating back compat runtime config for %s', highestSupportedVersion);
+
   let newSupports = { ...teamsRuntimeConfig.supports };
 
+  generateBackCompatRuntimeConfigLogger(
+    'Supported capabilities in config before updating based on highestSupportedVersion: %o',
+    JSON.stringify(newSupports),
+  );
```
AE-MS, line 221: Out of curiosity, do you know what this output looks like when run from minimized code? Is it something we will still be able to interpret?

packages/teams-js/src/public/pages.ts
```diff
@@ -374,12 +377,14 @@ export namespace pages {
...
     /**
      * Sets the validity state for the configuration.
      * The initial value is false, so the user cannot save the configuration until this is called with true.
      * @param validityState - Indicates whether the save or remove button is enabled for the user.
      */
     export function setValidityState(validityState: boolean): void {
+      setValidityStateLogger('setting validity state to %i', validityState);
       ensureInitialized(FrameContexts.settings, FrameContexts.remove);
       if (!isSupported()) {
         throw errorNotSupportedOnPlatform;
```
AE-MS, line 390: >throw errorNotSupportedOnPlatform; [](http://example.com/codeflow?start=8&length=34)

Based on the logging you added, I think that we could infer what happens when looking at logs but maybe since we are interested in this very specific issue it would make sense to also add a statement here that just says something like "pages.config not supported, setValiditityState throwing errorNotSupportedOnPlatform" to make it very obvious?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/runtime.ts
TrevorJoelHarris: It appears to be stable in minified code based off what I've seen from the bug I'm looking at. %i and %o both seem to work as expected

packages/teams-js/src/public/pages.ts
TrevorJoelHarris: no prob

packages/teams-js/src/public/app.ts
```diff
@@ -550,6 +554,11 @@ export namespace app {
         Handlers.initializeHandlers();
         GlobalVars.initializePromise = initializeCommunication(validMessageOrigins).then(
           ({ context, clientType, runtimeConfig, clientSupportedSDKVersion = defaultSDKVersionForCompatCheck }) => {
+            initializeHelperLogger('context: %s', context);
```
MarcinJuraszek, line 557: This should already be printed as part of the communication layer logging?

packages/teams-js/src/public/pages.ts
```diff
@@ -374,14 +377,17 @@ export namespace pages {
...
     /**
      * Sets the validity state for the configuration.
      * The initial value is false, so the user cannot save the configuration until this is called with true.
      * @param validityState - Indicates whether the save or remove button is enabled for the user.
      */
     export function setValidityState(validityState: boolean): void {
+      setValidityStateLogger('setting validity state to %i', validityState);
       ensureInitialized(FrameContexts.settings, FrameContexts.remove);
       if (!isSupported()) {
+        setValidityStateLogger('pages.config not supported, setValiditityState throwing errorNotSupportedOnPlatform');
```
MarcinJuraszek, line 390: I understand you're adding it here because you're tracking a particular issue. Do we expect every single API to have this sort of logging around `ensureIinitialized` and `if(!isSupported())` at some point? If so, should we at least log an issue to track that? If not, should we log an issue to track removing this log statement once we're done with the investigations?

AE-MS (dismissed):
:shipit:

MarcinJuraszek:
Let's wait for Trevor to come back to address my comments, and probably only keep the generic logs and not things very specific to this particular investigation.

MarcinJuraszek (changes requested):
Let's discuss it next week. I think most of these are very specific to the particular investigation, and I don't think they should be merged to the office release.

packages/teams-js/src/public/pages.ts
TrevorJoelHarris: Since you've tracked down the root cause, I'll remove logging from the specific functions in this file (but leave some of the more generic logging in place).

I do think it would be worth thinking of how to add logging for things like "isSupported" and "ensureInitialized" in a way that doesn't make the bundle double in size. I'll open a task to track that

packages/teams-js/src/public/app.ts
TrevorJoelHarris: You are correct, communication.handleParentMessage. Was a little hard to correlate because the message name wasn't being logged

packages/teams-js/src/public/runtime.ts
```diff
@@ -206,8 +211,15 @@ export const versionConstants: Record<string, Array<ICapabilityReqs>> = {
  * @returns runtime which describes the APIs supported by the legacy host client.
  */
 export function generateBackCompatRuntimeConfig(highestSupportedVersion: string): IRuntime {
+  generateBackCompatRuntimeConfigLogger('generating back compat runtime config for %s', highestSupportedVersion);
+
   let newSupports = { ...teamsRuntimeConfig.supports };
 
+  generateBackCompatRuntimeConfigLogger(
+    'Supported capabilities in config before updating based on highestSupportedVersion: %o',
+    JSON.stringify(newSupports),
```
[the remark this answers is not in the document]
TrevorJoelHarris: it is because I didn't realize debug would do it for me

AE-MS (approved):
:shipit:

PR #1345, opened 2022-09-12
Update dialog.submit and dialog.sendMessageToParentFromDialog

Jessica-Koch:

## Description

> Dialog.submit now throws an error if it is called when contentActions within appContext is defined and arguments are passed to the submit function since they will not do what is expected in this context. 

**NEED INPUT**
Since content actions are new, there are no built in sdk errors that truly explain what is going on in these cases, is there a new SDK error I can add to the constants file that would fit better?  Or is there an existing one someone is partial to? 

### Main changes in the PR:

1. change dialog.submit to throw error 
2. change sendMessageToParentFromDialog to throw error 

### Validation performed:

1. ran unit tests

### Unit Tests added:

yes
### End-to-end tests added:

no
### Change file added:

<Yes/No>

### Next/remaining steps:

- [ ] add e2e tests

### DIFF
--- a/change/@microsoft-teams-js-87e67164-063e-4f04-81b1-03d17d698b48.json
+++ b/change/@microsoft-teams-js-87e67164-063e-4f04-81b1-03d17d698b48.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "update dialog.submit and dialog.sendMessageToParentFromDialog to handle situation when called and contentActions are present and arguments are passed to functions",
+  "packageName": "@microsoft/teams-js",
+  "email": "jekoch@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-87e67164-063e-4f04-81b1-03d17d698b48.json
+++ b/change/@microsoft-teams-js-87e67164-063e-4f04-81b1-03d17d698b48.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "update dialog.submit and dialog.sendMessageToParentFromDialog to handle situation when called and contentActions are present and arguments are passed to functions",
+  "comment": "updated dialog.submit and dialog.sendMessageToParentFromDialog to handle situation when called and contentActions are present and arguments are passed to functions",
   "packageName": "@microsoft/teams-js",
   "email": "jekoch@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -108,32 +108,24 @@ export namespace dialog {
    * @param result - The result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
    * @param appIds - Helps to validate that the call originates from the same appId as the one that invoked the task module
    */
-  export function submit(result?: string | object, appIds?: string | string[]): void {
+  export function submit(result?: string | object, appIds?: string | string[]): Promise<void> {
     ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
 
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
 
-    app
-      .getContext()
-      .then((appContext) => {
-        // if actionInfo exists in appContext, the dialog was loaded from a 3rd party app,
-        // and if submit was called with a result, throw an error because the result cannot be
-        // passed along to the desired recipient
-        if (result && appContext.actionInfo) {
-          throw errorInvalidArguments;
-        } else {
-          // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
-          sendMessageToParent('tasks.completeTask', [
-            result,
-            appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : [],
-          ]);
-        }
-      })
-      .catch((error) => {
-        throw error;
-      });
+    return app.getContext().then((appContext) => {
+      // if actionInfo exists in appContext, the dialog was loaded from a 3rd party app,
+      // and if submit was called with a result, throw an error because the result cannot be
+      // passed along to the desired recipient
+      if (result && appContext.actionInfo) {
+        throw errorInvalidArguments;
+      } else {
+        // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
+        sendMessageToParent('tasks.completeTask', [result, appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : []]);
+      }
+    });
   }
 
   /**
@@ -147,7 +139,7 @@ export namespace dialog {
   export function sendMessageToParentFromDialog(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     message: any,
-  ): void {
+  ): Promise<void> {
     ensureInitialized(FrameContexts.task);
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
@@ -156,18 +148,13 @@ export namespace dialog {
     // if actionInfo exists in appContext, the dialog was loaded from a 3rd party app,
     // and if sendMessageToParentFromDialog was called with a message,
     // throw an error because the message cannot be passed along to the desired recipient
-    app
-      .getContext()
-      .then((appContext) => {
-        if (appContext.actionInfo) {
-          throw errorNotSupportedInCurrentContext;
-        } else {
-          sendMessageToParent('messageForParent', [message]);
-        }
-      })
-      .catch((error) => {
-        throw error;
-      });
+    app.getContext().then((appContext) => {
+      if (appContext.actionInfo) {
+        throw errorNotSupportedInCurrentContext;
+      } else {
+        sendMessageToParent('messageForParent', [message]);
+      }
+    });
   }
 
   /**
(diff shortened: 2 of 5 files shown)


packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -458,6 +463,32 @@ describe('Dialog', () => {
           }
         });
 
+        it('FRAMED: should throw an error if actionInfo is present in appContext and submit is called with arguments', async () => {
```
shrshindeMSFT, line 466: add `${context}` in the test name.

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -458,6 +463,32 @@ describe('Dialog', () => {
...
+
+          expect.assertions(1);
+          try {
+            await dialog.submit('someResult');
+          } catch (e) {
+            expect(e).toEqual(errorInvalidArguments);
+          }
+        });
+
+        it('FRAMELESS: should throw an error if actionInfo is present in appContext and submit is called with arguments', async () => {
```
shrshindeMSFT, line 479: Same as above

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -905,6 +936,31 @@ describe('Dialog', () => {
...
+
+            expect.assertions(1);
+            try {
+              await dialog.sendMessageToParentFromDialog('someMessage');
+            } catch (e) {
+              expect(e).toEqual(errorInvalidArguments);
+            }
+          });
+
+          it('FRAMELESS: should throw an error if actionInfo is present in appContext and sendMessageToParentFromDialog is called with arguments', async () => {
```
shrshindeMSFT, line 952: Same as above. Include the context in the test name.

shrshindeMSFT (changes requested):

packages/teams-js/src/public/dialog.ts
```diff
@@ -122,13 +133,22 @@ export namespace dialog {
...
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
 
-    sendMessageToParent('messageForParent', [message]);
+    // if actionInfo exists in appContext, the dialog was loaded from a 3rd party app,
+    // and if sendMessageToParentFromDialog was called with a message,
+    // throw an error because the message cannot be passed along to the desired recipient
+    return app.getContext().then((appContext) => {
+      if (message && appContext.actionInfo) {
```
lakhveerk, line 146: If I am understanding it correctly, we would not want to allow `sendMessageToParentFromDialog` at all if `actionInfo` is provided. No need to check for message

packages/teams-js/src/public/dialog.ts
```diff
@@ -122,13 +133,22 @@ export namespace dialog {
...
+        throw errorInvalidArguments;
```
lakhveerk, line 147: If my above comment is true, we would want a different error message here

packages/teams-js/src/public/dialog.ts
```diff
@@ -101,14 +102,24 @@ export namespace dialog {
...
+  export function submit(result?: string | object, appIds?: string | string[]): Promise<void> {
     ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
+
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
 
-    // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
-    sendMessageToParent('tasks.completeTask', [result, appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : []]);
+    return app.getContext().then((appContext) => {
```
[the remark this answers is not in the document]
Jessica-Koch: this was originally just for error handling, but I could just add a `.catch` to the end and throw the error again

packages/teams-js/src/public/dialog.ts
Jessica-Koch: tbh, I don't like this error message at all, but since content actions is new, there is no sdk error that really makes sense.  I would love input on this...adding to the description

change/@microsoft-teams-js-87e67164-063e-4f04-81b1-03d17d698b48.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "update dialog.submit and dialog.sendMessageToParentFromDialog to handle situation when called and contentActions are present and arguments are passed to functions",
```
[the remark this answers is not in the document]
Jessica-Koch: changed, thank you!

packages/teams-js/src/public/dialog.ts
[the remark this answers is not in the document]
Jessica-Koch: I know we return promise<void> a lot throughout this codebase but to make the return the same as before, I have changed it

packages/teams-js/src/public/dialog.ts
Jessica-Koch: you are correct

packages/teams-js/test/public/dialog.spec.ts
```diff
@@ -452,12 +458,38 @@ describe('Dialog', () => {
           framedMock.setRuntimeConfig({ apiVersion: 1, supports: {} });
           expect.assertions(4);
           try {
-            dialog.submit('someResult', ['someAppId', 'someOtherAppId']);
+            return await dialog.submit('someResult', ['someAppId', 'someOtherAppId']);
```
shrshindeMSFT, line 461: Can't we just call `dialog.submit`?

packages/teams-js/src/public/dialog.ts
```diff
@@ -103,12 +110,30 @@ export namespace dialog {
...
-    // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
-    sendMessageToParent('tasks.completeTask', [result, appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : []]);
+    app
+      .getContext()
+      .then((appContext) => {
+        // if actionInfo exists in appContext, the dialog was loaded from a 3rd party app,
+        // and if submit was called with a result, throw an error because the result cannot be
+        // passed along to the desired recipient
+        if (result && appContext.actionInfo) {
+          throw errorInvalidArguments;
```
ydogandjiev, line 125: Throwing exceptions inside of a promise chain like this is not a good idea since there is no good way for developers to catch and handle them.

packages/teams-js/src/public/dialog.ts
```diff
@@ -103,12 +110,30 @@ export namespace dialog {
...
 
-    // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
-    sendMessageToParent('tasks.completeTask', [result, appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : []]);
+    app
+      .getContext()
+      .then((appContext) => {
+        // if actionInfo exists in appContext, the dialog was loaded from a 3rd party app,
+        // and if submit was called with a result, throw an error because the result cannot be
+        // passed along to the desired recipient
+        if (result && appContext.actionInfo) {
```
ydogandjiev, line 124: This code is making non-obvious assumptions that could change in the future. It's totally not obvious why actionInfo in the context would change the behavior of this API. I think it would be good to add a separate API here specific for these actions and a new FrameContext for this flow.

packages/teams-js/src/public/dialog.ts
```diff
@@ -103,12 +110,30 @@ export namespace dialog {
...
+        } else {
+          // Send tasks.completeTask instead of tasks.submitTask message for backward compatibility with Mobile clients
+          sendMessageToParent('tasks.completeTask', [
+            result,
+            appIds ? (Array.isArray(appIds) ? appIds : [appIds]) : [],
+          ]);
+        }
+      })
+      .catch((error) => {
+        throw error;
```
ydogandjiev, line 135: Same as above; if this API needs to return errors asynchronously it should return a promise chain

ydogandjiev (changes requested):

packages/teams-js/src/public/dialog.ts
MarcinJuraszek: We would need to change the return type of this function to `Promise<void>` to let the caller catch the error.

packages/teams-js/src/public/dialog.ts
Jessica-Koch: that's why I put a code comment, after a conversation offline with @MarcinJuraszek and @ydogandjiev, we decided to talk about this when @TrevorJoelHarris is back.

packages/teams-js/src/public/dialog.ts
Jessica-Koch: this was what was originally in the code, but I was asked to switch.  I assumed that this was fine since it could be wrapped in a try catch, but doing it the way you suggest is actually more consistent with the rest of the sdk

packages/teams-js/src/public/dialog.ts
Jessica-Koch: When I first opened the PR it was returning `Promise<void>` and the caller was catching the error but I was asked to change it.  I assumed that rethrowing the error would allow it to be wrapped in a try/catch.  I will change it back to be more consistent with what is in the rest of the codebase.

Jessica-Koch: Thinking about it now, if I was using the api as is (not returning a promise), it wouldn't be immediately obvious to me to wrap this in a try catch, so I definitely agree

packages/teams-js/test/public/dialog.spec.ts
Jessica-Koch: we don't, they are redundant, it was an oversight when I changed the code

PR #1348, opened 2022-09-14
shrshinde/dialogsAdaptiveCards

shrshindeMSFT:

## Description

### Main changes in the PR:

1. Added new sub-capability `adaptive cards` to `dialogs`

### Validation performed:

1. Added unit test cases

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### Next/remaining steps:

> Add e2e test case.


packages/teams-js/src/public/dialog.ts
```diff
@@ -210,6 +217,121 @@ export namespace dialog {
...
 
+  /**
+   * @hidden
+   * Hide from docs
+   * --------
+   * Convert AdaptiveCardDialogInfo to DialogInfo to send the information to host in {@linkcode adaptiveCard.open} API.
+   *
+   * @internal
+   */
+  export function getDialogInfoFromAdaptiveCardDialogInfo(adaptiveCardDialogInfo: AdaptiveCardDialogInfo): DialogInfo {
```
lakhveerk, line 228: we avoid exporting internal functions that users don't need.

packages/teams-js/src/public/dialog.ts
```diff
@@ -210,6 +217,121 @@ export namespace dialog {
...
+      ): void {
+        ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
+        if (!isSupported()) {
+          throw errorNotSupportedOnPlatform;
+        }
+
+        const dialogInfo: DialogInfo = getDialogInfoFromBotAdaptiveCardDialogInfo(botAdaptiveCardDialogInfo);
+
+        sendMessageToParent('tasks.startTask', [dialogInfo], (err: string, result: string | object) => {
+          submitHandler?.({ err, result });
```
lakhveerk, line 316: why does this not have removeHandler call?

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -64,7 +64,7 @@ export namespace geoLocation {
     }
     const permissions: DevicePermission = DevicePermission.GeoLocation;
 
-    return new Promise<boolean>(resolve => {
+    return new Promise<boolean>((resolve) => {
```
lakhveerk, line 67: These prettier fixes annoy me. You did not touch this file and it still make changes here. I hope somebody else's setting don't put this parenthesis back in next PR

shrshindeMSFT: I can understand this trying to fix it

packages/teams-js/src/public/dialog.ts
shrshindeMSFT: We need to export as we are using them in unit test cases. If we don't export then we cannot test them :)

packages/teams-js/src/public/dialog.ts
shrshindeMSFT: There is no handler registered in the code.

packages/teams-js/src/public/dialog.ts
lakhveerk: Me and Aaron talked about this (not on your PR, but the same topic). Typically, we would not want to expose some code just for testing purposes. Here is online doc to read more about this. 
**https://enterprisecraftsmanship.com/posts/unit-testing-private-methods/**

packages/teams-js/src/public/dialog.ts
```diff
@@ -210,6 +217,121 @@ export namespace dialog {
...
+     */
+    export function open(adaptiveCardDialogInfo: AdaptiveCardDialogInfo, submitHandler?: DialogSubmitHandler): void {
+      ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+      const dialogInfo: DialogInfo = getDialogInfoFromAdaptiveCardDialogInfo(adaptiveCardDialogInfo);
+      sendMessageToParent('tasks.startTask', [dialogInfo], (err: string, result: string | object) => {
+        submitHandler?.({ err, result });
+        removeHandler('messageForParent');
```
lakhveerk, line 279: We are not registering any handler here either. Should we delete this removeHandler?

packages/teams-js/src/public/interfaces.ts
```diff
@@ -749,6 +749,43 @@ export interface DeepLinkParameters {
...
+}
+
+/**
+ * Data structure to describe dialog information needed to open a adaptive card based dialog.
+ */
+export interface AdaptiveCardDialogInfo extends BaseDialogInfo {
+  /**
+   * JSON defining an adaptive card.
+   */
+  card?: string;
```
lakhveerk, line 774: Why is card optional here?

packages/teams-js/src/public/interfaces.ts
```diff
@@ -782,11 +819,6 @@ export interface UrlDialogInfo {
    */
   size: DialogSize;
```
lakhveerk, line 820: base dialog has size in it, that means we can delete size from here

packages/teams-js/src/public/interfaces.ts
```diff
@@ -749,6 +749,43 @@ export interface DeepLinkParameters {
   subEntityWebUrl?: string;
 }
 
+/**
+ * Shared Dialog Properties
+ */
+export interface BaseDialogInfo {
```
lakhveerk, line 755: Since we are adding support for Card and Bot, does it make sense to add depreciate tag to the original dialogInfo interface?

packages/teams-js/src/public/dialog.ts
```diff
@@ -210,6 +217,121 @@ export namespace dialog {
...
+      width: botAdaptiveCardDialogInfo.size ? botAdaptiveCardDialogInfo.size.width : DialogDimension.Small,
+      title: botAdaptiveCardDialogInfo.title,
+      completionBotId: botAdaptiveCardDialogInfo.completionBotId,
+    };
+    return dialogInfo;
+  }
+  /**
+   * Subcapability for interacting with adaptive card dialogs
+   */
+  export namespace adaptiveCard {
```
lakhveerk, line 261: should we add beta tag to this?

packages/teams-js/src/public/interfaces.ts
shrshindeMSFT: Deleted in latest commit

packages/teams-js/src/public/interfaces.ts
shrshindeMSFT: It depends on do we want to update it across the platform, and if so, why? Also, my point of view is we need to add all these interfaces back up when are removing the support for `DialogInfo`

packages/teams-js/src/public/dialog.ts
shrshindeMSFT: Just wondering why, we want to state it as `beta`?

packages/teams-js/src/public/interfaces.ts
shrshindeMSFT: Fixed in latest commit

packages/teams-js/src/public/dialog.ts
lakhveerk: In case you had to tweak it a little in future (based on customer feedback or you find a bug), you will be able to change it without having to bump the major version of library

packages/teams-js/src/public/interfaces.ts
lakhveerk: What scenario are you thinking? May be that back compat issue can affect the url dialog since we made the url required in UrlDialogInfo interface

packages/teams-js/src/public/interfaces.ts
lakhveerk: Because the only place where `DialogInfo` interface is being used is in tasks module and that is deprecated. 

What do you mean by this: "we need to add all these interfaces back up when are removing the support for DialogInfo"

packages/teams-js/src/public/tasks.ts
```diff
@@ -36,8 +41,11 @@ export namespace tasks {
         (sdkResponse: dialog.ISdkResponse) => submitHandler(sdkResponse.err, sdkResponse.result)
       : undefined;
     if (taskInfo.card !== undefined || taskInfo.url === undefined) {
-      ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
```
lakhveerk, line 39: Not a comment on your change, but I am wondering why we did not throw an error here when both card and url and provided, but instead waited until it reaches to the host SDK?

packages/teams-js/src/public/tasks.ts
```diff
@@ -36,8 +41,11 @@ export namespace tasks {
...
-      sendMessageToParent('tasks.startTask', [taskInfo as DialogInfo], submitHandler);
+      if (taskInfo.completionBotId) {
+        dialog.adaptiveCard.bot.open(getBotAdaptiveCardDialogInfoFromTaskInfo(taskInfo), dialogSubmitHandler);
```
lakhveerk, line 45: Since adaptive card is a sub capability and hosts can choose to not implement it, should we make a check here before we call dialog.adaptiveCard.xxx?

lakhveerk (commented):
PR looks good. Left some questions/comments :)

packages/teams-js/src/public/interfaces.ts
shrshindeMSFT: I think we do use it in `dialog.adaptiveCards.open` 
`const dialogInfo: DialogInfo = getDialogInfoFromAdaptiveCardDialogInfo(adaptiveCardDialogInfo);`

packages/teams-js/src/public/tasks.ts
shrshindeMSFT: AFAIK we wanted the host to resolve it. Don't know if this the exact reason.

packages/teams-js/src/public/interfaces.ts
shrshindeMSFT: Fixed in latest commit

packages/teams-js/src/public/interfaces.ts
lakhveerk: I will get @TrevorJoelHarris 's view on this as well :)

packages/teams-js/src/public/tasks.ts
lakhveerk: If something is going to fail, we want to fail it sooner. I learned it from @TrevorJoelHarris. Leaving this open in case Trevor wants to comment on it.

packages/teams-js/src/public/tasks.ts
lakhveerk: I think that makes sense.

packages/teams-js/src/public/tasks.ts
TrevorJoelHarris: Failing fast is always good. In this specific case, this function was originally written by Teams and in order to maintain backwards compatibility with their existing function we don't want to make it fail either at a different *time* than it used to or in a different *way* than it used to. Apps that calls tasks.startTask should see an error come from teams if they specify both card and url, but since we don't know what that error is we'll continue to let Teams handle it. We should *definitely* fail fast in our new dialog capability though (or even better, do what Shreyas is doing in this change and make failure not possible by making different function signatures that don't allow for invalid parameter combinations)

packages/teams-js/src/public/tasks.ts
```diff
@@ -36,12 +41,15 @@ export namespace tasks {
         (sdkResponse: dialog.ISdkResponse) => submitHandler(sdkResponse.err, sdkResponse.result)
       : undefined;
     if (taskInfo.card !== undefined || taskInfo.url === undefined) {
```
TrevorJoelHarris, line 43: Now that we're moving all of these calls to dialog, we should make sure and be extra careful to not miss any cases. I think that before this change, if someone didn't specify card OR url, we would fall into line 39 and call sendMessageToParent which would let the host decide what to do with the incorrect set of parameters. After this change, we would be calling the dialog.adaptiveCard.open or dialog.adaptiveCard.bot.open functions, which I don't think will throw errors in the same way. We want to make sure that this function continues to fail the same way in that case.

packages/teams-js/src/public/tasks.ts
```diff
@@ -128,4 +136,45 @@ export namespace tasks {
...
     return taskInfo;
   }
+
+  /**
+   * @hidden
+   * Converts {@link TaskInfo} to {@link AdaptiveCardDialogInfo}
+   * @param taskInfo - TaskInfo object to convert
+   * @returns - converted AdaptiveCardDialogInfo
+   */
+  function getAdaptiveCardDialogInfoFromTaskInfo(taskInfo: TaskInfo): AdaptiveCardDialogInfo {
```
TrevorJoelHarris, line 146: these two functions should be exported from dialog.ts and used here instead of duplicated

packages/teams-js/src/public/interfaces.ts
TrevorJoelHarris: (We should also update the comment on TaskInfo to point to UrlDialogInfo and the other ones now

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/interfaces.ts
lakhveerk: hidden sounds good! If we ever plan to change the hubSDK interfaces to match the interfaces in appSDK (for simplicity?), we can come back to this :)

packages/teams-js/src/public/tasks.ts
lakhveerk: It is always backwards compatibility :/
Thank you for commenting! :)

change/@microsoft-teams-js-f6e6a65b-3151-423e-909d-9b396b402601.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Added adaptive cards sub capability to dialogs",
```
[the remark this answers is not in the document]
shrshindeMSFT: Resolved in the latest commit

TrevorJoelHarris:
I am closing this PR since you opened a new one for this work

PR #1350, opened 2022-09-15
Fixes Search Api initialization in tests

jobutle:
## Description

> Fixes search tests in Teams tests app. The Search Service's parameters function is being called with the wrong order of parameters


jobutle:
Closing so we can reopen with branch instead of main from fork

PR #1351, opened 2022-09-15
 Fixes Search Api initialization in tests

jobutle:
## Description
Fixes search tests in Teams tests app. The Search Service's parameters function is being called with the wrong order of parameters

See registerHandlers [here](https://github.com/OfficeDev/microsoft-teams-library-js/blob/b3b61ca9d4911c899f25ee3a0ae6059d859ea328/packages/teams-js/src/public/search.ts)

### DIFF
--- a/apps/teams-test-app/src/components/SearchAPIs.tsx
+++ b/apps/teams-test-app/src/components/SearchAPIs.tsx
@@ -27,7 +27,7 @@ const RegisterHandlers = (): React.ReactElement =>
       };
       setResult('register handlers');
 
-      search.registerHandlers(onChange, onClosed, onExecute);
+      search.registerHandlers(onClosed, onExecute, onChange);
       return 'recieved';
     },
   });
--- a/apps/teams-test-app/src/components/SearchAPIs.tsx
+++ b/apps/teams-test-app/src/components/SearchAPIs.tsx
@@ -28,7 +28,7 @@ const RegisterHandlers = (): React.ReactElement =>
       setResult('register handlers');
 
       search.registerHandlers(onClosed, onExecute, onChange);
-      return 'recieved';
+      return 'received';
     },
   });
 
--- a/packages/teams-js/src/public/search.ts
+++ b/packages/teams-js/src/public/search.ts
@@ -47,13 +47,6 @@ export namespace search {
    * Your application should *not* re-render inside of these callbacks, there may be a large number
    * of onChangeHandler calls if the user is typing rapidly in the search box.
    *
-   * @param onChangeHandler - This optional handler will be called when the user first starts using the
-   * host's search box and as the user types their query. Can be used to put your application into a 
-   * word-wheeling state or to display suggestions as the user is typing. 
-   * 
-   * This handler will be called with an empty {@link SearchQuery.searchTerm} when search is beginning, and subsequently,
-   * with the current contents of the search box.
-   * 
    * @param onClosedHandler - This handler will be called when the user exits or cancels their search.
    * Should be used to return your application to its most recent, non-search state. The value of {@link SearchQuery.searchTerm} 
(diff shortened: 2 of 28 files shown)


erinha:
Thanks for catching and fixing this! I took a look at the link you provided to the implementation of `search.registerHandlers`. Your fix for the test app is correct, but I see that the documentation for the `search.registerHandlers` function is very misleading as it lists the parameters (in @param comments and in the example provided) in incorrect order. Would you be willing to fix the documentation in search.ts as part of this PR as well?

jobutle:
> Thanks for catching and fixing this! I took a look at the link you provided to the implementation of `search.registerHandlers`. Your fix for the test app is correct, but I see that the documentation for the `search.registerHandlers` function is very misleading as it lists the parameters (in @param comments and in the example provided) in incorrect order. Would you be willing to fix the documentation in search.ts as part of this PR as well?

jobutle:
> Thanks for catching and fixing this! I took a look at the link you provided to the implementation of `search.registerHandlers`. Your fix for the test app is correct, but I see that the documentation for the `search.registerHandlers` function is very misleading as it lists the parameters (in @param comments and in the example provided) in incorrect order. Would you be willing to fix the documentation in search.ts as part of this PR as well?

Good catch! Done

MarcinJuraszek:
I think you'll need a changelog added via beachball.

apps/teams-test-app/src/components/SearchAPIs.tsx
```diff
@@ -27,7 +27,7 @@ const RegisterHandlers = (): React.ReactElement =>
       };
       setResult('register handlers');
 
-      search.registerHandlers(onChange, onClosed, onExecute);
+      search.registerHandlers(onClosed, onExecute, onChange);
       return 'recieved';
```
anmaur, line 31: nit: spelling

erinha: Oh good catch! Since you're here @jobutle,  would you mind please correcting line 31 to  'received'?

packages/teams-js/src/public/search.ts
```diff
@@ -62,6 +55,12 @@ export namespace search {
...
    * The value of {@link SearchQuery.searchTerm} is the complete query the user entered in the search box.
    *
+   * @param onChangeHandler - This optional handler will be called when the user first starts using the
+   * host's search box and as the user types their query. Can be used to put your application into a 
+   * word-wheeling state or to display suggestions as the user is typing. 
+   * 
+   * This handler will be called with an empty {@link SearchQuery.searchTerm} when search is beginning, and subsequently,
+   * with the current contents of the search box.
    * @example
    * ``` ts
```
erinha, line 65: Can you please update the example and move the code on lines 67-69 to after lines 73-75?

packages/teams-js/src/public/search.ts
```diff
@@ -62,6 +55,12 @@ export namespace search {
    * search (by pressing Enter for example). Should be used to display the full list of search results. 
    * The value of {@link SearchQuery.searchTerm} is the complete query the user entered in the search box.
    *
+   * @param onChangeHandler - This optional handler will be called when the user first starts using the
```
erinha, line 58: Beautiful, thanks!

erinha:
> I think you'll need a changelog added via beachball.

Yep, you'll need to run `yarn changefile` and then add a description (please use past tense and put any type/interface/enum values in  backtick quotes \`)

erinha (changes requested):
Appreciate you helping clean up this code. A couple minor requests plus info for adding a change file.

TrevorJoelHarris (approved):

erinha (approved):
Thank you again for fixing this!

PR #1353, opened 2022-09-19
Enable eslint null check rule to work both in the build and VSCode

AE-MS:
## Description

VSCode looks for the tsconfig file in a different place than the build. The build was working, VSCode was not. I put the config file in two locations. Maybe I might have been able to figure out how to get them to both look in the same place but since this is only temporary as the `strictNullCheck` changes are being incrementally implemented, it didn't feel like it was worth the time to figure that out.

### Main changes in the PR:

Copy `tsconfig.strictNullChecks.json` to root, which is where VSCode looks for it

### Validation performed:

Ensure `strictNullChecks` was being successfully applied both by the build and VSCode

### Unit Tests added:

Nope.

### End-to-end tests added:

Nope.

### Change file added:

Not needed for this change.


tsconfig.strictNullChecks.json
```diff
@@ -0,0 +1,5 @@
+{
```
erinha, line 1: Does the build not work if this is only present at the root?

erinha: I see from your commit history that you tried that. :)

erinha (approved):

tsconfig.strictNullChecks.json
AE-MS: It needs to be in both to satisfy both requirements. If the file is only present in the root then the build fails because it looks for it under packages\teams-js.

TrevorJoelHarris (approved):

PR #1360, opened 2022-09-22
Fix more violations of strictNullChecks

AE-MS:
## Description

- Fix some places where `undefined` / `null` was being handled but it just wasn't part of the type definition (by adding it to the type definition)
- Fix the code which assesses the `runtime` object to detect `undefined` cases and throw in those.

### Validation performed:

Run unit tests and e2e tests as part of this PR

### Unit Tests added:

Yes, added a unit test to cover the one new behavior: if "null" is passed as the runtime config stored in `clientSupportedSDKVersion` (the flipped order scenario). This shouldn't ever happen in real life but the type system allows it, and now we protect against it.

### End-to-end tests added:

No, there are no e2e tests that would exercise functionality not already covered by unit tests.

### Change file added:

Yes

### Related PRs:

https://github.com/OfficeDev/microsoft-teams-library-js/pull/1356
https://github.com/OfficeDev/microsoft-teams-library-js/pull/1352

### Next/remaining steps:

Fix more violations

### DIFF
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -592,7 +592,9 @@ export namespace app {
                   initializeHelperLogger('givenRuntimeConfig parsed to %o', givenRuntimeConfig ?? 'null');
 
                   if (!givenRuntimeConfig) {
-                    throw new Error('givenRuntimeConfig successfully parsed to null');
+                    throw new Error(
+                      'givenRuntimeConfig string was successfully parsed. However, it parsed to value of null',
+                    );
                   } else {
                     applyRuntimeConfig(givenRuntimeConfig);
                   }
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -591,7 +591,11 @@ export namespace app {
                   const givenRuntimeConfig: IRuntime | null = JSON.parse(clientSupportedSDKVersion);
                   initializeHelperLogger('givenRuntimeConfig parsed to %o', givenRuntimeConfig ?? 'null');
 
-                  clientSupportedSDKVersion && givenRuntimeConfig && applyRuntimeConfig(givenRuntimeConfig);
+                  if (!givenRuntimeConfig) {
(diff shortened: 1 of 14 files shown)


packages/teams-js/src/public/app.ts
```diff
@@ -589,9 +588,14 @@ export namespace app {
...
                     GlobalVars.clientSupportedSDKVersion = runtimeConfig;
                   }
-                  /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-                  const givenRuntimeConfig: IRuntime = JSON.parse(clientSupportedSDKVersion);
-                  clientSupportedSDKVersion && applyRuntimeConfig(givenRuntimeConfig);
+                  const givenRuntimeConfig: IRuntime | null = JSON.parse(clientSupportedSDKVersion);
+                  initializeHelperLogger('givenRuntimeConfig parsed to %o', givenRuntimeConfig ?? 'null');
+
+                  if (!givenRuntimeConfig) {
+                    throw new Error('givenRuntimeConfig successfully parsed to null');
```
erinha, line 595: This seems like a slightly weird error message containing the word "successfully". Suggest removing the word "successfully".

erinha (changes requested):
Comment about error message. Let me know if you disagree!

packages/teams-js/src/public/app.ts
AE-MS: Interesting; I actually intentionally used that word to indicate that the parsing was successful and it was not the result of a parsing error. 🤔

erinha: I wondered about that perspective and since I don't have a better suggestion, I guess it's fine.

erinha (dismissed):

packages/teams-js/src/public/app.ts
AE-MS: Maybe this new string is clearer?

erinha (approved):
Thank you, this error message is definitely an improvement IMO!

TrevorJoelHarris (approved):

PR #1361, opened 2022-09-22
Remove Latest BC run from build

MarcinJuraszek:
## Description

> With recent changes to how E2E tests are run, there is no need for an explicit back-compat targeted run.

### Main changes in the PR:

1. Remove a Job that uses `latestBackcompat` sdk checkpoint.

### Validation performed:

N/A

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

No - "No change files are needed" from `yarn changelog`

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -106,35 +106,6 @@ jobs:
           mergeTestResults: true
         condition: succeededOrFailed()
 
-  - job: E2ETest4
-    displayName: 'E2E Test - BC Latest'
-    pool:
-      vmImage: 'ubuntu-latest'
-    steps:
-      - template: tools/yaml-templates/build-app-host.yml
-        parameters:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -106,35 +106,6 @@ jobs:
...
-
-      - task: PublishTestResults@2
-        inputs:
-          testResultsFormat: 'JUnit'
-          testResultsFiles: '**/e2e-tests-report*.xml'
-          testRunTitle: 'E2E Tests - BC Latest'
-          mergeTestResults: true
-        condition: succeededOrFailed()
-
   - job: E2ETest5
```
maglims, line 109: should this be E2ETest4 now Since we are removing that?

maglims (approved):

AE-MS (approved):
:shipit:

PR #1367, opened 2022-09-28
Fixed profile.isSupported and showProfile TriggerType

grhewitt:
### Main changes in the PR:

1. Update `generateBackCompatRuntimeConfig` to set `profile.isSupported` to true for Teams Desktop and Web clients.
2. Update the values of the `TriggerType` parameter passed to `showProfile` to better support its use case - `showProfile` does not need to distinguish `"MousePress"` and `"KeyboardPress"` so we unify them into a single "Press" value.

### Validation performed:

1. Built local version of SDK and copied into a Teams test app.
2. Loaded test app and validated that `profile.isSupported` returns true in Web and Desktop, and that `showProfile` API works as expected.
3. Validated that API handler on Teams side is forward compatible with new type changes.

### Unit Tests added:

None added because it's a typing only change in profile and runtime code already covers it with tests.

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/src/public/profile.ts
```diff
@@ -52,11 +52,10 @@ export namespace profile {
   /**
    * The type of the profile trigger.
    *  - MouseHover: The user hovered a target.
-   *  - MouseClick: The user clicked a target.
-   *  - KeyboardPress: The user initiated the show profile request with their keyboard.
+   *  - Press: The target was pressed with either a mouse click or keyboard key press.
```
TrevorJoelHarris, line 55: Should this also include touch events?

grhewitt: Currently this API is only supported on Teams web and desktop, the plan is to keep this type minimal and only add new values when we see a need for them (such as when we start testing out integration in our first mobile hub)

TrevorJoelHarris: Right, but I assume this same value would also cover touch events in the future? Just trying to save you a future checkin where you have to update the documentation to say "and touch events", but touch events can't possibly trigger this yet then that's fine

TrevorJoelHarris (approved):

packages/teams-js/src/public/profile.ts
grhewitt: Ah sorry! I misunderstood, I thought you meant to add a new value for touch events. I think you're right that this value would cover touch events when we eventually add mobile support. I'll opt to update this doc header later, just in case I'm wrong and we actually have to do further changes to the API to support touch.

noahdarveau-MSFT (approved):

PR #1369, opened 2022-09-28
Document appIds info in dialog file

lakhveerk:

## Description

>Added comments that clarify what is an array of appIds for in dialog.submit and tasks.submitTask API

### Unit Tests added:

<Yes/No>

Not required. This is a change in comment and not in actual API

### End-to-end tests added:

<Yes/No>

Not required. This is a change in comment and not in actual API

### Change file added:

<Yes/No>

Yes


lakhveerk:
@erikadoyle  @JoshuaPartlow 

Tagging Erika and Joshua, to make sure the comment I added makes sense, and we can update the public doc: https://learn.microsoft.com/en-us/microsoftteams/platform/task-modules-and-cards/task-modules/task-modules-bots?tabs=nodejs#submit-the-result-of-a-task-module

cc: @TrevorJoelHarris

packages/teams-js/src/public/dialog.ts
```diff
@@ -99,7 +99,9 @@ export namespace dialog {
    * Submit the dialog module and close the dialog
    *
    * @param result - The result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
-   * @param appIds - Helps to validate that the call originates from the same appId as the one that invoked the task module
+   * @param appIds - Helps to validate that the call originates from the same appId as the one that invoked the task module.
```
TrevorJoelHarris, line 102: I think that this isn't *quite* accurate. I believe these are used by Teams or other hosts to confirm that the value of result is being sent back to an application with id matching one of the appids here.

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/dialog.ts
```diff
@@ -99,7 +99,9 @@ export namespace dialog {
    * Submit the dialog module and close the dialog
    *
    * @param result - The result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
-   * @param appIds - Helps to validate that the call originates from the same appId as the one that invoked the task module
+   * @param appIds - Valid application/s that the dailog can send result back to once dialog is submitted. It pervents malicious apps to get the dialog result.
+   * Since web apps can power multiple Teams apps across different environments or different branding schemas with the same domain,
+   * an array of valid appIDs corresponded to the different Teams apps can be provided.
```
erikadoyle, line 104: Taking a stab at paraphrase/suggestion to also check my understanding here:

Valid Teams application(s) that can receive the result of the submitted dialogs. Specifying this parameter helps prevent malicious apps from retrieving the dialog result. Multiple app IDs can be specified because a web app from a single underlying domain can power multiple Teams apps across different environments and branding schemes.

packages/teams-js/src/public/dialog.ts
lakhveerk: does this look good now?

packages/teams-js/src/public/dialog.ts
lakhveerk: When submitting a dialog, the result goes back to whoever opened it. However, if a malicious app somehow figures the dialog-url of some valid app and opens a dialog with it, we wouldn't want to send the result back to that malicious app. To prevent it, hosts check, during submit call, that app that invoked the dialog have an appid that matches one of the ones in the list (assuming the list was specified). If not, the host should throw out the submit content and not pass it back to the app

My wording in Line 102 makes it appear like any app that has been listed in the appIds array can receive the result, but that is not true. The one that opened the dialog will receive the result, but the condition is its appId has to match with the provided appId string (if a string is specified) or has to be in the array list (if an array of appIds is provided) in submit call.

Is my understanding correct @TrevorJoelHarris

TrevorJoelHarris: (But I would remove the reference to Teams and just say "Valid application(s)")

packages/teams-js/src/public/tasks.ts
```diff
@@ -72,7 +72,9 @@ export namespace tasks {
    * Submit the task module.
    *
    * @param result - Contains the result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
-   * @param appIds - Helps to validate that the call originates from the same appId as the one that invoked the task module
+   * @param appIds - Valid application/s that the dailog can send result back to once dialog is submitted. It pervents malicious apps to get the dialog result.
```
TrevorJoelHarris, line 75: Make sure you update this to be the same as the comment in dialog once you change the wording to match what erika suggested

TrevorJoelHarris (changes requested):

TrevorJoelHarris (approved):

shrshindeMSFT (dismissed):

jadahiya-MSFT (dismissed):

shrshindeMSFT (approved):

jadahiya-MSFT (approved):

PR #1373, opened 2022-09-30
User/grahamc/220930 fix issue6761620

GrahamMcMynn:

## Description

This change fixes this issue:
Task 6761620: Events not firing in child iFrames

### Main changes in the PR:

1. Events are proxied to child frames.

### Validation performed:

1. I ran a SharePoint app which IFrames a child frame and ensured that the back button/beforeUnload/apphover/appclick events were all proxied to the child Iframe.

### Unit Tests added:

Added a unit test that checked that an event was in both parent and child frames.

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/change/@microsoft-teams-js-7e2d7681-be7b-4c3c-b196-470aa65092c4.json
+++ b/change/@microsoft-teams-js-7e2d7681-be7b-4c3c-b196-470aa65092c4.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Currently if a child IFrame is opened and the TeamsJs SDK is used it will not receive all the events that are sent to the parent IFrame.  This change allows the proxying of events if they are not handled in the parent frame.",
+  "comment": "Enabled proxying of window events to child frames if unhandled by current frame",
   "packageName": "@microsoft/teams-js",
   "email": "grahamc@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-7e2d7681-be7b-4c3c-b196-470aa65092c4.json
+++ b/change/@microsoft-teams-js-7e2d7681-be7b-4c3c-b196-470aa65092c4.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Enabled proxying of window events to child frames if unhandled by current frame",
+  "comment": "Enabled proxying of window events to child frames if they are unhandled by current frame",
   "packageName": "@microsoft/teams-js",
   "email": "grahamc@microsoft.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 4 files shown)


packages/teams-js/src/internal/handlers.ts
```diff
@@ -44,6 +44,9 @@ export function callHandler(name: string, args?: unknown[]): [true, unknown] | [
     callHandlerLogger('Invoking the registered handler for message %s with arguments %o', name, args);
     const result = handler.apply(this, args);
     return [true, result];
+  } else if (Communication.childWindow) {
+    sendMessageEventToChild(name, [args]);
+    return [true, undefined];
```
TrevorJoelHarris, line 49: I think this should be false instead of true, since the boolean signifies whether or not the handler was called and in this case it wasn't (it was proxied to the child window in case the child window has a handler for it)

packages/teams-js/src/public/pages.ts
```diff
@@ -640,7 +642,12 @@ export namespace pages {
 
     function handleBackButtonPress(): void {
       if (!backButtonPressHandler || !backButtonPressHandler()) {
-        navigateBack();
+        if (Communication.childWindow) {
+          // If no parent window function handled it let the child window
```
TrevorJoelHarris, line 646: I think this should be current window, right? I would interpret "Parent" as "parent of the current window", which I don't think is what you mean

GrahamMcMynn: Yes, I will correct that comment.  I was thinking pare of the child window that I am sending it too.

packages/teams-js/src/internal/handlers.ts
GrahamMcMynn: That seems reasonable.  I wasn't really sure what I should return, I will change it to false.

change/@microsoft-teams-js-7e2d7681-be7b-4c3c-b196-470aa65092c4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Currently if a child IFrame is opened and the TeamsJs SDK is used it will not receive all the events that are sent to the parent IFrame.  This change allows the proxying of events if they are not handled in the parent frame.",
```
AE-MS, line 3: >Currently if a child IFrame is opened and the TeamsJs SDK is used it will not receive all the events that are sent to the parent IFrame.  This change allows the proxying of events if they are not handled in the parent frame. [](http://example.com/codeflow?start=14&length=224)

For consistency among changelog messages, can you rephrase to something like this (feel free to copy if you want):

"Enabled proxying of window events to child frames if unhandled by current frame."

AE-MS (changes requested):
:clock1:

AE-MS (approved):

TrevorJoelHarris (approved):

TrevorJoelHarris (approved):

AE-MS (approved):

PR #1381, opened 2022-10-06
Stevenic/live share fix

Stevenic:
### Main changes in the PR:

Switched from using dynamic `import()` to find `LiveShareClient` class to using a global window variable.  Dynamic import works for a single level of webpack but once you get multiple levels of webpack going on it stops being able to find the module you're trying to import.

The Live Share module simply needs to dynamically find the `LiveShareClient` class it needs to create when `liveShare.initialize()` is called. I switched to having `@microsoft/live-share` take the much simpler approach of registering a reference to this class as a global window variable. A `GUID` was used to avoid potential variable name collisions. 

**UPDATE:** This PR also addresses an issue related to #1382 . The use of `import type` to conditionally import Fluid types doesn't seem to always work as expected and customers started encountering TS compile issues with 2.4.0.  I have removed the use of `import type` and instead now directly include a subset types defining the various Fluid data structures Live Share accepts as either input or returns as output.

**UPDATE:** The PR now adds in core files and unit tests that were removed as part of reverting [This PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1310)

**UPDATE:** Per discussions with the team I removed the `liveShare` namespace and now I'm directly exposing the `LiveShareHost` class from the SDK.  A new instance of the class will be passed in as a parameter to the constructor of Live Shares `LiveShareClient` class.

**UPDATE:** Added static `LiveShareHost.create()` method which performs app initialization checks.

### Validation performed:

Wrote a test application that successfully loaded the live-share dice roller app into teams.

### Unit Tests added:

Added unit tests for `LiveShareHost` class.

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@microsoft/teams-js",
   "author": "Microsoft Teams",
-  "version": "2.3.0",
+  "version": "2.5.0",
```
jadahiya-MSFT, line 4: Please don't manually change the version of the Teams JS package

packages/teams-js/package.json
```diff
@@ -24,8 +24,8 @@
   },
   "devDependencies": {
     "@types/debug": "4.1.7",
-    "@fluidframework/azure-client": "~1.0.1",
-    "fluid-framework": "~1.2.2"
+    "@fluidframework/azure-client": "~1.0.2",
```
jadahiya-MSFT, line 27: I believe these two need to move to being `dependencies` instead `devDependencies` so that they are picked up during install calls by anyone consuming the package

packages/teams-js/package.json
Stevenic: Sorry that was so I can test locally... It's reverted now

packages/teams-js/package.json
Stevenic: I was able to remove them all together. We don't want them under `dependencies`  as they would increase the size of the teams-js SDK by a factor of 10.

packages/teams-js/src/public/liveShare.ts
```diff
@@ -0,0 +1,404 @@
...
+     * When loading an existing container, it will already be attached.
+     */
+    readonly attachState: string;
+
+    /**
+     * A newly created container starts detached from the collaborative service.
+     * Calling `attach()` uploads the new container to the service and connects to the collaborative service.
+     *
+     * @remarks This should only be called when the container is in the
+     * {@link @fluidframework/container-definitions#AttachState.Detatched} state.
```
TrevorJoelHarris, line 293: Does this @link actually resolve now that no types are being included from this package?

packages/teams-js/src/public/liveShare.ts
```diff
@@ -0,0 +1,404 @@
...
+   * @beta
+   */
+  export async function initialize(options?: LiveShareOptions): Promise<void> {
+    if (initializing || client) {
+      throw new Error('Live Share has already been initialized.');
+    }
+
+    try {
+      initializing = true;
+      const LiveShareClient = window['53de46f8-db62-4b8d-ae81-330f828ac86c'] as LiveShareClient;
```
TrevorJoelHarris, line 352: Please add a comment above the use of this guid explaining how it works and what it is a reference to.

packages/teams-js/src/public/liveShare.ts
```diff
@@ -0,0 +1,404 @@
...
+
+  /**
+   * Initializes the Live Share client.
+   * @param options Optional. Configuration options passed to the Live Share client.
+   *
+   * @beta
+   */
+  export async function initialize(options?: LiveShareOptions): Promise<void> {
+    if (initializing || client) {
+      throw new Error('Live Share has already been initialized.');
```
TrevorJoelHarris, line 347: This error message should probably say that Live Share has either been initialized or is already in the process of initializing.

packages/teams-js/test/public/liveShare.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { LiveShareHost } from '../../src/internal/liveShareHost';
+import { liveShare } from '../../src/public/liveShare';
+
+describe('MicrosoftTeams-LiveShare', () => {
+  describe('initialize', () => {
```
TrevorJoelHarris, line 5: Since you have some logic guarding against calling initialize incorrectly (can't call while it's already initializing or after it's finished initializing) it would be good to have unit tests around that.

TrevorJoelHarris, line 5: You need some unit tests that verify that the "correct" path works (Calling initialize doesn't throw any errors)

packages/teams-js/test/public/liveShare.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    it('should throw error due to package not being found', async () => {
+      await expect(liveShare.initialize()).rejects.toThrowError(
+        'Unable to initialize Live Share client. Ensure that your project includes "@microsoft/live-share"',
+      );
+    });
+  });
+
+  describe('joinContainer', () => {
```
TrevorJoelHarris, line 13: You need some unit tests that verify the "correct" path works (the right client function is called)

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/liveShareHost.ts
```diff
@@ -0,0 +1,267 @@
...
+export interface IFluidTenantInfo {
+  /**
+   * The Fluid Tenant ID Live Share should use.
+   */
+  tenantId: string;
+
+  /**
+   * The Fluid service endpoint Live Share should use.
+   */
+  serviceEndpoint?: string;
```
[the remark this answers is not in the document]
Stevenic: In general, end users will never see the LiveShareHost directly. It's an implementation detail for the application that's hosting live share.

The `serviceEndpoint` field of the tenant info structure is a new field that fluid added when they went GA. It replaces the other fields. We have a weird transition period (which is almost over) where we have to support both GA and non-GA fluid so LiveShare has logic to automatically figure out the `serviceEndpoint` if its missing.

packages/teams-js/src/public/liveShare.ts
Stevenic: probably not... good catch. I'll look those over.

packages/teams-js/src/public/liveShare.ts
Stevenic: No problem

packages/teams-js/src/public/liveShare.ts
Stevenic: Good tweak

packages/teams-js/test/public/liveShare.spec.ts
Stevenic: We have an issue right now that we can't even test the positive test case where Live Share initializes. To do that a couple of things would need to happen.

- teams-js would need to take a `devDependency` on live-share but given we can't publish a version of live-share that works with this code until teams-js has been updated, we have a chicken & egg problem. Someone has to publish first...

Stevenic: See the previous comment about `live-share` would need to be published which hasn't happened yet

packages/teams-js/test/public/liveShare.spec.ts
Stevenic: Same as other comment

packages/teams-js/src/internal/liveShareHost.ts
```diff
@@ -0,0 +1,267 @@
...
+   * ID to the mapping service.
+   */
+  shouldCreate: boolean;
+
+  /**
+   * If `containerId` is undefined and `shouldCreate` is false, the container isn't ready but
+   * another client is creating it. The local client should wait the specified amount of time and
+   * then ask for the container info again.
+   */
+  retryAfter: number;
```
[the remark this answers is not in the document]
Stevenic: There's documentation in development on the Live Share side of things that will describe how to build new hosts. End users/developers will never see this. This value is currently set by the live-share service based off heuristics for how slow or fast container creation is currently taking.

packages/teams-js/test/public/liveShare.spec.ts
TrevorJoelHarris: I wrote this elsewhere, but I think using dependency injection or something similar would let you isolate the part of your code that is creating the LiveShareClient so that you could test the rest of the logic.

packages/teams-js/src/public/liveShare.ts
Stevenic: Removed namespace per discussion

packages/teams-js/src/public/liveShare.ts
Stevenic: Removed namespace per discussion

packages/teams-js/src/public/liveShare.ts
Stevenic: Removed namespace per discussion

packages/teams-js/test/public/liveShare.spec.ts
Stevenic: Removed namespace per discussion

packages/teams-js/test/public/liveShare.spec.ts
Stevenic: Removed namespace per discussion

packages/teams-js/test/public/liveShare.spec.ts
Stevenic: Removed namespace per discussion

packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * @deprecated
+   * As of Fluid 1.0 this configuration information has been deprecated in favor of
+   * `serviceEndpoint`.
+   */
+  storageEndpoint: string;
+}
+
+/**
```
TrevorJoelHarris, line 134: Probably want @hidden on the class as well

packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -0,0 +1,251 @@
...
+ * Live Share host implementation for O365 and Teams clients.
+ * 
+ * @beta
+ */
+export class LiveShareHost {
+  /**
+   * @hidden
```
TrevorJoelHarris, line 141: This class and every public function in this class should be marked with the @hidden and @beta. tags The @beta tag doesn't cascade into the class, so just putting it at the top here isn't enough to get each functions documentation page to be correctly marked with the @beta tag. RIght now the LiveShareHost class is missing the @hidden tag so it will get a doc page generated with pointers to any functions that aren't marked as hidden (create).

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/liveShareHost.ts
TrevorJoelHarris: Talked with Steve, the class and the create function actually do want to show in the docs for developers.

TrevorJoelHarris (approved):

packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -0,0 +1,267 @@
...
+   */
+  storageEndpoint: string;
+}
+
+/**
+ * Live Share host implementation for O365 and Teams clients.
+ * 
+ * @beta
+ */
+export class LiveShareHost {
```
TrevorJoelHarris, line 139: LiveShare will add an isSupported check as part of the 2.6 release. @ashleyrocha will be the person doing the work on the host side.

AE-MS (approved):
:shipit:

PR #1410, opened 2022-10-20
Make sure initialize completes before most API calls are allowed

MarcinJuraszek:
## Description

Make sure `initialize` call finishes and runtime information is correctly set up before other APIs can be used. There are exception for APIs I think people are most likely calling right after calling `initialize`, potentially before it finishes. I also had to allow `register*Handler` calls with `null`/`undefined` handler to go through, to keep UTs working (we clean registrations between many tests, some of which didn't run `initialize` at all (by design).

### Main changes in the PR:

1. Update `ensureInitialized` to check that `initialize` finished, not just got called.
2. Add `ensureInitializeCalled` to be used by APIs that we want to allow to be used before `initialize` finishes.
3. Update said APIs to use `ensureInitializeCalled` instead of `ensureInitialized`
4. Fix UTs

### Validation performed:

1. UTs updated

### Unit Tests added:

Yes, added tests to validate that the APIs we do want to be callable before initialize finishes can actually be called after initialize called but before it completes.

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/change/@microsoft-teams-js-51980b95-c86a-4b3c-a2d0-1b9e475014b3.json
+++ b/change/@microsoft-teams-js-51980b95-c86a-4b3c-a2d0-1b9e475014b3.json
@@ -1,5 +1,5 @@
 {
-  "type": "major",
+  "type": "minor",
   "comment": "Updated most APIs to require initialization to be fully finished before they are allowed to be called.",
   "packageName": "@microsoft/teams-js",
   "email": "email not defined",
--- a/change/@microsoft-teams-js-51980b95-c86a-4b3c-a2d0-1b9e475014b3.json
+++ b/change/@microsoft-teams-js-51980b95-c86a-4b3c-a2d0-1b9e475014b3.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Updated most APIs to require initialization to be fully finished before they are allowed to be called.",
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/internal/handlers.ts
+++ b/packages/teams-js/src/internal/handlers.ts
@@ -100,7 +100,8 @@ export function registerHandlerHelper(
   contexts: FrameContexts[],
   registrationHelper?: () => void,
 ): void {
-  ensureInitialized(...contexts);
+  // allow for registration cleanup even when not finished initializing
+  handler && ensureInitialized(...contexts);
   if (registrationHelper) {
     registrationHelper();
   }
--- a/packages/teams-js/src/internal/internalAPIs.ts
+++ b/packages/teams-js/src/internal/internalAPIs.ts
@@ -2,7 +2,12 @@ import { HostClientType } from '../public/constants';
 import { ErrorCode, SdkError } from '../public/interfaces';
 import { defaultSDKVersionForCompatCheck, userOriginUrlValidationRegExp } from './constants';
 import { GlobalVars } from './globalVars';
+import { getLogger } from './telemetry';
 import { compareSDKVersions } from './utils';
+
+const internalLogger = getLogger('internal');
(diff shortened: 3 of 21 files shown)


MarcinJuraszek:
Open questions:
1. Should `app.isInitialized()` be updated to use `GlobalVars.initializeCompleted`?
2. I marked is as *major* change with beachball. Is that how we want to mark it?

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -8,12 +8,22 @@ import { compareSDKVersions } from './utils';
...
     throw new Error('The library has not yet been initialized');
   }
+}
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function ensureInitialized(...expectedFrameContexts: string[]): void {
+  if (!GlobalVars.initializeCompleted || !GlobalVars.frameContext) {
```
TrevorJoelHarris, line 22: Why was it necessary to also check that frameContext was set? is there a case where that might be an issue when initializeCompleted is true, or just trying to be careful?

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -8,12 +8,22 @@ import { compareSDKVersions } from './utils';
...
+    throw new Error('The library has not yet been initialized');
```
TrevorJoelHarris, line 23: It might be nice to throw two different errors from this function. One for "Initialization hasn't begun yet" and another for "initialization has begun but hasn't finished yet"

change/@microsoft-teams-js-51980b95-c86a-4b3c-a2d0-1b9e475014b3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
TrevorJoelHarris: Because this change is fixing a bug and making the sdk work the way we already said it did when 2.* was released, I think this change should be considered a minor change instead of a major one.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -100,7 +100,8 @@ export function registerHandlerHelper(
   contexts: FrameContexts[],
   registrationHelper?: () => void,
 ): void {
-  ensureInitialized(...contexts);
+  // allow for registration cleanup even when not finished initializing
+  handler && ensureInitialized(...contexts);
```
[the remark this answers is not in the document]
MarcinJuraszek: if `handler` is null/undefined (which is the case when you want to clean up previous registration), `ensureInitialized` won't actually be called (because of `&&`).

packages/teams-js/src/internal/internalAPIs.ts
MarcinJuraszek: I tried that, but the amount of downstream issues that caused with UTs and E2E tests makes me think that's the kind of 'breaking change' we should avoid.

packages/teams-js/src/internal/internalAPIs.ts
MarcinJuraszek: I removed it. It was mostly being careful.

packages/teams-js/src/internal/handlers.ts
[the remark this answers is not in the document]
TrevorJoelHarris: Ah, got it that makes sense

TrevorJoelHarris (dismissed):
### LOOKS GOOD TO ME

packages/teams-js/test/public/appInstallDialog.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { isConstructorDeclaration } from 'typescript';
```
erinha, line 1: Is this used anywhere?

erinha (changes requested):
Unnecessary import in appInstallDialog.spec.ts?

erinha (approved):
🚀

TrevorJoelHarris (approved):

PR #1411, opened 2022-10-20
Making our use of the @beta tag consistent

TrevorJoelHarris:
## Description

Our use of the @beta tag in this repo was not consistent for a couple of reasons. We initially thought that the @beta tag would cascade down and apply to things contained within something marked with @beta, but it does not.

After this change, I believe we now correctly have applied the beta tag everywhere. This means:

- If a namespace has a @beta tag on it, all children also have the @beta tag
- We no longer mark certain namespaces with @beta because we feel that making breaking changes there would now require bumping the major version of the package
- @beta tags indicate to readers of documentation that the item in question could change shape or disappear at any time and we will not bump the major version if something like that happens. When we feel that new things are sufficiently stable the @beta tag will be removed, at which point making breaking changes there would constitute a major version change.

### Main changes in the PR:

1. Removed beta tags from app, authentication, pages capabilities.
2. All functions in chat, dialog, profile capabilities are now marked as @beta

### Validation performed:

None, comment update only

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

No, documentation only

### DIFF
--- a/change/@microsoft-teams-js-8835237b-4a82-4082-b65a-5db37e32c03e.json
+++ b/change/@microsoft-teams-js-8835237b-4a82-4082-b65a-5db37e32c03e.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Removed the @beta tag from `pages`, `app`, and `authentication`. Added @beta tags to `chat`, `dialog`, and `profile` capabilities.",
+  "packageName": "@microsoft/teams-js",
+  "email": "trharris@microsoft.com",
+  "dependentChangeType": "none"
+}
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -44,7 +44,7 @@ export namespace dialog {
   export type PostMessageChannel = (message: any) => void;
 
   /**
-   * Handler used for receiving results when a dialog closes, either the value passed to {@linkcode submit}
+   * Handler used for receiving results when a dialog closes, either the value passed by {@linkcode submit}
    * or an error if the dialog was closed by the user.
    * @beta
    */
--- a/packages/teams-js/src/public/dialog.ts
+++ b/packages/teams-js/src/public/dialog.ts
@@ -18,6 +18,8 @@ import { runtime } from './runtime';
 export namespace dialog {
   /**
    * Data Structure to represent the SDK response when dialog closes
+   *
+   * @beta
    */
   export interface ISdkResponse {
     /**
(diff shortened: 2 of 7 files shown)


packages/teams-js/src/public/dialog.ts
```diff
@@ -33,8 +35,19 @@ export namespace dialog {
...
+
+  /**
+   * Handler used to receive and process messages sent between a dialog and the app that launched it
+   * @beta
+   */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   export type PostMessageChannel = (message: any) => void;
+
+  /**
+   * Handler used for receiving results when a dialog closes, either the value passed to {@linkcode submit}
```
lakhveerk, line 47: value passed by `submit`?

packages/teams-js/src/public/dialog.ts
```diff
@@ -18,6 +18,8 @@ import { runtime } from './runtime';
 export namespace dialog {
   /**
    * Data Structure to represent the SDK response when dialog closes
+   *
+   * @beta
```
lakhveerk, line 22: Since the branch was published, I thought we could not edit the file (or add beta tag) without bumping the version.  
Is this possible because @beta was added to the top of the file?

TrevorJoelHarris: Yes, I'm (very slightly) bending the rules here in the name of getting us into a consistent state. We already had the entire namespace marked with the beta tag, but that doesn't cause all of the functions within the namespace to get decorated correctly which could lead to some developer confusion. All of the contents of @dialog are in beta, so doing this will make it more likely for developers reading our documentation to get the right message

packages/teams-js/src/public/dialog.ts
TrevorJoelHarris: no prob

lakhveerk (approved):

AE-MS:
Consider filing a follow-up work item for you/PM to have some process/timeline for monitoring and removing beta tag usage. E.g. when do we decide to remove @beta from the code? What workflow keeps them from just sticking around forever besides "Trevor notices and decides to do something."

change/@microsoft-teams-js-8835237b-4a82-4082-b65a-5db37e32c03e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Removed the @beta tag from `pages`, `app`, and `authentication`. Added @beta tags to `chat`, `dialog`, and `profile` capabilities.",
```
AE-MS, line 3: >Added @beta tags [](http://example.com/codeflow?start=79&length=16)

Maybe this should be "cascaded existing @beta tags in..." to make it clear that we aren't undoing previously solidified capabilities.

AE-MS (approved):
:shipit:

PR #1413, opened 2022-10-21
Trharris/newdomains

TrevorJoelHarris:
## Description

Added two new `validOrigins` so that apps can begin to be used in those domains. The two new origins are:

1. www.microsoft365.com
2. *.www.microsoft365.com (for pre-production environments)

### Main changes in the PR:

1. Added two new origins to `validOrigins` list
2. Updated the list of origins being tested in unit tests

### Validation performed:

1. Updated unit tests with missing origins and verified they were passing

### Unit Tests added:

Yes

Added new origins to list of `validOrigins`, and each entry in the list adds one new unit test.

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/src/internal/constants.ts
```diff
@@ -143,6 +143,8 @@ export const validOrigins = [
   'powerpoint.office.com',
   'www.officeppe.com',
   '*.www.office.com',
```
maglims, line 145: not related to this PR but shall this line be put before line 140?

packages/teams-js/test/private/privateAPIs.spec.ts
```diff
@@ -90,15 +90,28 @@ describe('AppSDK-privateAPIs', () => {
...
     'https://retailservices.teams.microsoft.com',
+    'https://www.office.com',
+    'https://word.office.com',
+    'https://excel.office.com',
+    'https://powerpoint.office.com',
+    'https://www.officeppe.com',
+    'https://test.www.office.com',
+    'https://test.www.microsoft365.com',
+    'https://www.microsoft365.com',
     'https://tasks.office.com',
```
maglims, line 115: similar here... not sure if it should be pullled up a bit along with other *.office.com

maglims (approved):

packages/teams-js/test/private/privateAPIs.spec.ts
TrevorJoelHarris: This one I actually want to keep where it is. tasks.office.com isn't actually in the list of validOrigins. It's only allowed here because it's being passed in to the intializeWithContext call as an additional validOrigin by the test (along with https://www.example.com)

packages/teams-js/src/internal/constants.ts
TrevorJoelHarris: Sure thing

erinha (approved):

packages/teams-js/src/internal/constants.ts
```diff
@@ -137,12 +137,14 @@ export const validOrigins = [
...
   '*.teams.microsoft.com',
+  '*.www.office.com',
   'www.office.com',
   'word.office.com',
   'excel.office.com',
   'powerpoint.office.com',
   'www.officeppe.com',
-  '*.www.office.com',
+  '*.www.microsoft365.com',
+  'www.microsoft365.com',
```
AE-MS, line 147: Are there really going to be subdomains that include the www part? E.g. foobar.www.microsoft.365.com instead of just foobar.microsoft365.com?

AE-MS (changes requested):
:clock1:

AE-MS:
>   const supportedDomains = [

I know you didn't put this list here in the first place, so maybe this is a comment for a future change.

Do we really need to have a second copy of this list for testing? Can't it just directly reference the one in constants.ts?

---
Refers to: packages/teams-js/test/private/privateAPIs.spec.ts:87 in 6aaca2f. [](commit_id = 6aaca2f8fa354afc04b9b93e6cca5988ba42557e, deletion_comment = False)

packages/teams-js/src/internal/constants.ts
maglims: I had the same question..

alexcotsalas: Yes, our test environments will live at blah.www.microsoft365.com (not at blah.microsoft365.com). Thanks!

TrevorJoelHarris:
I'm open to suggestions, but I couldn't think of a good way to share them since once list is "regular expressions we use to match valid domains" and the other list is "actual domains that had better match at least one regular expression"

AE-MS (approved):

jadahiya-MSFT (approved):

PR #1414, opened 2022-10-21
add code comment for dialog.submit

Jessica-Koch:

## Description
add a code comment for dialog.submit to warn developers of ramifications when passing parameters to the submit function


packages/teams-js/src/public/dialog.ts
```diff
@@ -102,6 +102,8 @@ export namespace dialog {
    *
    * @param result - The result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
    * @param appIds - Valid application(s) that can receive the result of the submitted dialogs. Specifying this parameter helps prevent malicious apps from retrieving the dialog result. Multiple app IDs can be specified because a web app from a single underlying domain can power multiple apps across different environments and branding schemes.
+   *
+   * MPORTANT -  When in a ContentActions scenario, the passed in result will not go anywhere as there is no  registered "parent" to send the result to
```
TrevorJoelHarris, line 106: I might remove the "IMPORTANT" and add this to the description on the result parameter.

2 other thoughts:

1) It would be good to be more specific about what a "ContentActions" scenario is. Possibly by linking to the part of the context that would be set?
2) Is the value of submit always ignored? I don't remember content actions supporting bot ids, but if they *do* support opening dialogs with completionBotIds then we should make sure to be specific about that as well. May be a good thing to confirm with PM if we don't already know.

Jessica-Koch: thank you for the feedback! after following up, I will follow up with PM about the bot scenario and update the comment as needed

TrevorJoelHarris (dismissed):

change/@microsoft-teams-js-db4c1ae0-ce9d-45a5-a3f1-2f9c8d825f1d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "add code comment for dialog.submit",
```
TrevorJoelHarris, line 3: Changefile formatting should match how we describe it in Contributing.md. Comments should be in past tense and code references should be surrounded by back ticks. Something like:

"Added code comment for `dialog.submit`"

This format matches how we want the patch notes to read and makes less work for people making the releases.

change/@microsoft-teams-js-db4c1ae0-ce9d-45a5-a3f1-2f9c8d825f1d.json
```diff
@@ -0,0 +1,7 @@
...
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
```
TrevorJoelHarris, line 5: Your email address associated with your github account should probably be here? Not sure why it isn't

TrevorJoelHarris (changes requested):

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

packages/teams-js/src/public/dialog.ts
```diff
@@ -117,7 +117,9 @@ export namespace dialog {
   /**
    * Submit the dialog module and close the dialog
    *
-   * @param result - The result to be sent to the bot or the app. Typically a JSON object or a serialized version of it
+   * @param result - The result to be sent to the bot or the app. Typically a JSON object or a serialized version of it,
+   *  If this function is called from a dialog while {@link M365ContentAction} is set in the context object by the host, result will be ignored
```
lakhveerk, line 121: Is this how it supposed to work? or is it a bug and we are writing comment to let customers know the result will be ignored? I am also not sure what M365ContentAction is.

nit: the @link is not linking anything.
image

Jessica-Koch: this is a limitation of this new feature, currently, it isn't expected that this will be a problem, but we want to make developers aware of this.  Also, how do @links work then?  I actually am not sure

lakhveerk: I see.

@link works just fine since you have imported the M365ContentAction. Sorry for the confusion. I was checking the wrong thing

PR #1416, opened 2022-10-21
Enable running eslint on teams-js unit test files

AE-MS:
## Description

This change enables eslint to run on the teams-js unit test files. It also disables eslint for the test files that have many errors (which is basically all of them). This strategy:
1. Will allow us to fix the errors incrementally in each file, re-enabling eslint on a per-file basis as they are fixed.
2. (Most importantly) Ensures that any NEW test files get eslint "for free" and stop making the problem worse.

Note: the only changes I intentionally made are outlined above. Some changes in the test files were made automagically by the prettier/eslint plugin for VSCode when I saved the files. There's nothing wrong with these changes but they aren't specifically about enabling eslint.

### Validation performed:

- Ensured that lint errors introduced into test files without a disable statement show up in the `yarn lint` step
- Ensured that new lint errors in product files (which were already being linted) continue to break the lint step.
- Ensured that all builds, lintings (lints? lint steps?), and tests pass.

### Unit Tests added:

No.

### End-to-end tests added:

No.

### Change file added:

Yes.

### Next/remaining steps:

Start fixing errors over time. There are a lot of them so it will take a long time across many people. But, importantly, no new errors can be introduced in new test files.


packages/teams-js/test/private/logs.spec.ts
```diff
@@ -65,7 +69,7 @@ describe('logs', () => {
         await utils.initializeWithContext(context);
 
         let handlerInvoked = false;
-        const log: string = '1/1/2019 Info - App initialized';
```
AE-MS, line 68: >: string [](http://example.com/codeflow?start=17&length=8)

This change was made automatically by prettier

lakhveerk (approved):
LGTM

erinha (approved):

PR #1422, opened 2022-10-25
Fix lint errors in Utils.ts (used by unit tests)

AE-MS:
## Description

The lint errors in Utils.ts have been fixed. Any required downstream changes from fixes those errors have also been addressed in this PR.

### Validation performed:

Unit tests pass.

### Unit Tests added:

No; just maintaining the current behavior.

### End-to-end tests added:

No; just maintaining the current behavior.

### Change file added:

Yes.


packages/teams-js/test/utils.ts
```diff
@@ -58,20 +55,20 @@ export class Utils {
       outerHeight: 768,
       screenLeft: 0,
       screenTop: 0,
-      addEventListener: function (type: string, listener: (ev: MessageEvent) => void, useCapture?: boolean): void {
```
aeunms, line 61: Asking because I genuinely don't know and want to learn, not being sassy: How is it not breaking anything when the functions in the file no longer take in such arguments? Is it because the rest of the code base never called it or never called it with the arguments? What about removeEventListener() where listener was a required argument?

AE-MS: This is a great question and something I had to reason about myself before deciding it was safe to remove the unused arguments. Feel free to poke holes in my justification!

I think there are two key parts to this:
### The typing of the function. Will TypeScript yell at you if you try to pass extra arguments to a function?
TypeScript can only yell in this case based on the arguments it knows the function accepts based on the declared type. This code you see here is essentially changing which function on the `Window` object the `addEventListener` corresponds to. The actual type of the object is `Window` and the actual type of the `addEventListener` function on `Window` is declared as such (there are two overloads):
```
addEventListener<K extends keyof WindowEventMap>(type: K, listener: (this: Window, ev: WindowEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
```
When code TypeScript code outside of this class tries to call this function, it's going to check it against the `addEventListener` type on the `Window` object; it's not going to check it against the actual arguments of the function underneath because all the type checking is based on compile-time types and not runtime realities. Based on this, it should not break the build since all the code that uses `addEventListener` does it based on the type declared on `Window`.
(you'll also perhaps note that even the existing code doesn't use the right type)

### At runtime in JavaScript, what happens if you pass extra arguments to a function?
They just get ignored. Since the implementation provided here doesn't use those other parameters, they were already effectively being ignored. You can see this if you run [this code](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcAUwYCcDOUBciYIAtgEYCmWAlIQG5wwAmiA3gFCJeIQI5wAbCgDoBcAOZoABgDFseQgBJWmXFAC+U6gG5269uxToAjABpEAJnMBmHUA) in the TS playground. When you hit the "Run" button you'll see that TypeScript yells at you (in this case it knows the function only accepts one parameter and you're passing three) but if you look at the output in the right pane, the code ran just fine and ignored the extra two parameters that were passed.

MarcinJuraszek (approved):

TrevorJoelHarris (approved):

PR #1423, opened 2022-10-25
Adds Outlook consumer domain to domains' whitelist

ruiconti:
## Description

The `app.initialize` call is failing for consumer mailboxes in Outlook because `outlook.live.com` (a sometimes origin used for consumer endpoints) is not listed as a valid domain.

### DIFF
--- a/packages/teams-js/src/internal/constants.ts
+++ b/packages/teams-js/src/internal/constants.ts
@@ -136,6 +136,7 @@ export const validOrigins = [
   'outlook-sdf.office.com',
   'outlook.office365.com',
   'outlook-sdf.office365.com',
+  'outlook.live.com',
   '*.teams.microsoft.com',
   '*.www.office.com',
   'www.office.com',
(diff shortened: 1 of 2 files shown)


packages/teams-js/src/internal/constants.ts
```diff
@@ -136,6 +136,7 @@ export const validOrigins = [
   'outlook-sdf.office.com',
   'outlook.office365.com',
   'outlook-sdf.office365.com',
+  'outlook.live.com',
```
TrevorJoelHarris, line 139: Do you have any test environment URLs you also need to match? For example, *.teams.microsoft.com below is used to match some of the teams test environments.

TrevorJoelHarris:
If you take a look at this recent PR, can you add a URL to the same unit test file adding a URL that matches your new domain? 

https://github.com/OfficeDev/microsoft-teams-library-js/pull/1413/files

I'm actually not convinced that unit test is terribly useful, but since it's currently tracking entries for every domain it would be nice to keep it current until we decide to fix it.

TrevorJoelHarris (changes requested):

TrevorJoelHarris (approved):

AE-MS (approved):

TrevorJoelHarris (approved):

PR #1430, opened 2022-10-28
Make sure isSupported functions throw if checked before initialization is complete

TrevorJoelHarris:
## Description

Our runtime object is used to help determine which capabilities are supported in the current host. We intended to start off the runtime object with all capabilities marked as unsupported and then update them to whatever the host is actually supporting as a result of the app.initialize() call. Unfortunately, there was a bug with the way we did this. In the initial runtime object, if we define a capability like this:

```
capability: {
  subCapability: undefined,
}
```

Then if an app developer checked if a capability with subcapabilities was supported before calling app.Initialize we would return true.

The fix here is just to not put anything inside the `supports` part of runtime. We will initialize supports correctly when we hear from the host, and if an app developer checks any capabilities to see if they're supported ahead of calling isSupported the function will throw an error saying that initialize needs to be called first

### Main changes in the PR:

1. Remove all values from the `supports` section of runtime and wait for values to come from the host during initialize.
2. Added unit tests for every capability and subcapability that has an isSupported function to make sure that it throws when the runtime is in an uninitialized state.
3. Updated all isSupported functions to call ensureInitalized

### Validation performed:

1. Ran all existing unit tests

### Unit Tests added:

Added tests for verifying that every isSupported function correctly returns false if checked before runtime has been initialized

### End-to-end tests added:

No

### Change file added:

Yes


TrevorJoelHarris:
As I was filling out the PR template, I realized that I might want to add a unit test around app.initialize for this. But currently there are none! Let me see what I can do.

packages/teams-js/src/public/barCode.ts
```diff
@@ -94,6 +94,7 @@ export namespace barCode {
    * @beta
    */
   export function isSupported(): boolean {
```
AE-MS, line 96: >isSupported [](http://example.com/codeflow?start=18&length=11)

Should the docs for all these `isSupported` methods be updated to indicate that they [throw](https://typedoc.org/tags/throws/) when called before initialization is completed?

packages/teams-js/src/public/call.ts
```diff
@@ -64,6 +64,7 @@ export namespace call {
   }
 
   export function isSupported(): boolean {
```
AE-MS, line 66: >isSupported [](http://example.com/codeflow?start=18&length=11)

I know that you aren't making this any worse and that generally I don't see the need to fix up other code in the area that you're making a change, but I wonder if now would be the time to put documentation on these methods, since they are simple and you are in here already?

If you accept my previous "@throws" suggestion then that might be motivation to do it in this change.

packages/teams-js/src/public/runtime.ts
```diff
@@ -58,55 +58,21 @@ export interface IRuntime {
...
-        joinedTeams: undefined,
-      },
-    },
-    teamsCore: undefined,
-    video: undefined,
-  },
+  supports: {},
 };
 
+export let runtime: IRuntime = _uninitializedRuntime;
```
AE-MS, line 74: >_uninitializedRuntime [](http://example.com/codeflow?start=31&length=21)

Is there still value in having an uninitializedRuntime? Or should this now be declared as
```
export let runtime: IRuntime | null = null
```

?

packages/teams-js/src/public/call.ts
TrevorJoelHarris: I can do that, it might slow down when this change gets checked in though. It will take a little while to add and check comments on however many functions that is. We can talk about it tomorrow if I haven't done it by then

packages/teams-js/test/private/logs.spec.ts
```diff
@@ -29,6 +29,13 @@ describe('logs', () => {
...
 
+  describe('Testings logs.isSupported', () => {
+    it('should throw if called before initialization', () => {
+      utils.setRuntimeConfig(_uninitializedRuntime);
+      expect(() => logs.isSupported()).toThrowError('The library has not yet been initialized');
+    });
+  });
+
   describe('Testing logs.registerGetLogHandler function', () => {
     it('logs.registerGetLogHandler should not allow calls before initialization', () => {
```
AE-MS, line 40: >logs.registerGetLogHandler should not allow calls before initialization [](http://example.com/codeflow?start=8&length=71)

Should there be a test case that validates calling `registerGetLogHandler` before initialize when not passing a handler does NOT throw?

AE-MS: Same question for the three other ones in `pages` that also intentionally don't do stuff when no `handler` is passed.

packages/teams-js/test/public/calendar.spec.ts
```diff
@@ -301,14 +301,21 @@ describe('calendar', () => {
...
-    it('should return true if the runtime says calendar is supported', () => {
+    it('should return true if the runtime says calendar is supported', async () => {
+      await utils.initializeWithContext(FrameContexts.content);
       utils.setRuntimeConfig({ apiVersion: 1, supports: { calendar: {} } });
       expect(calendar.isSupported()).toBeTruthy();
     });
+
+    it('should throw if called before initialization', () => {
+      utils.setRuntimeConfig(_uninitializedRuntime);
+      expect(() => calendar.isSupported()).toThrowError('The library has not yet been initialized');
```
AE-MS, line 318: >'The library has not yet been initialized' [](http://example.com/codeflow?start=56&length=42)

Should this string be a constant that multiple places, including the product code, references?

packages/teams-js/test/private/logs.spec.ts
TrevorJoelHarris: Probably

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/runtime.ts
TrevorJoelHarris: Seems reasonable. I think I will leave _uninitializedRuntime in place and just change it to null. That way if we come across some reason to make uninitializedRuntime something OTHER than null in the future we won't need to go back in and change a billion files again.

TrevorJoelHarris: Okay, I went a bit down this parth and I think I'm going to leave it as-is for now. Pretty much every isSupported function in our codebase is written like this:

`return runtime.supports.dialog ? (runtime.supports.dialog.bot ? true : false) : false`

If we make runtime nullable every function gets more annoying to write, and making this nullable is (I think) duplicative with the change I've already got to add ensureInitialized to every isSupported function. Since I think ensureInitialized gives more correct errors than failing isSupported and saying the capability isn't supported because runtime doesn't exist / doesn't have a supports member, going to leave it.

If you disagree let me know!

packages/teams-js/test/public/calendar.spec.ts
TrevorJoelHarris: Done! Wow there are a lot

packages/teams-js/src/public/barCode.ts
TrevorJoelHarris: done

packages/teams-js/src/public/call.ts
TrevorJoelHarris: Done

packages/teams-js/test/private/logs.spec.ts
TrevorJoelHarris: Done

packages/teams-js/src/internal/constants.ts
```diff
@@ -175,3 +175,12 @@ export const teamsDeepLinkProtocol = 'https';
...
 export const teamsDeepLinkHost = 'teams.microsoft.com';
+
+/**
+ * @hidden
+ * Error thrown when the {@linkcode app.initialize} has not yet successfully completed
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export const errorLibraryNotInitialized: Error = new Error('The library has not yet been initialized');
```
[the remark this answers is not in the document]
TrevorJoelHarris: I hope not because we do something very similar for errorNotSupportedOnPlatform, but I will verify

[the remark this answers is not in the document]
AE-MS: If it is a problem maybe you could just make this a function that throws and then it can be called in the context of the callstack that needs to trigger it?

packages/teams-js/src/private/appEntity.ts
```diff
@@ -108,14 +108,16 @@ export namespace appEntity {
   /**
    * @hidden
    *
-   * Checks if appEntity capability is supported by the host
-   * @returns true if the appEntity capability is enabled in runtime.supports.appEntity and
-   * false if it is disabled
+   * Checks if the appEntity capability is supported by the host
+   * @returns boolean to represent whether the appEntity capability is supported
+   *
+   * @throws if {@linkcode app.initialize} has not successfully completed
```
AE-MS, line 114: >@throws if [](http://example.com/codeflow?start=5&length=10)

Consider adding the type of the error being throw, e.g.,
```
@throws {@linkcode type} blah blah
```

erikadoyle (commented):
LGTM (in terms of reference docs consistency)

packages/teams-js/src/public/location.ts
```diff
@@ -120,6 +120,7 @@ export namespace location {
    * @returns boolean to represent whether Location is supported
```
AE-MS, line 120: I know this is deprecated but should it still have the @throws comment in it?

packages/teams-js/src/public/pages.ts
```diff
@@ -213,18 +213,20 @@ export namespace pages {
    */
```
AE-MS, line 213: Should this also document that it can @throws ?

packages/teams-js/src/public/pages.ts
```diff
@@ -380,7 +384,7 @@ export namespace pages {
      */
```
AE-MS, line 384: Should this also document that it can @throws?

packages/teams-js/src/public/pages.ts
```diff
@@ -418,7 +422,7 @@ export namespace pages {
...
```
AE-MS, line 422: Same

packages/teams-js/test/private/logs.spec.ts
AE-MS: I think maybe my suggestion wasn't clear -- from my reading of the `registerGetLogHandler` code, it intentionally does NOT throw, even if `initialize` hasn't finished, if no `handler` was passed.  Is there a unit test case for that? I see a test case here for ensuring that it throws when calling `registerGetLogHandler` when given a handler has a parameter, but I'm talking about the different behavior when no handler parameter is passed.

packages/teams-js/src/internal/constants.ts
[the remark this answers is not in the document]
TrevorJoelHarris: Alright, after playing with this for a while marcin was right that this messed up our callstacks. It seemed like the best way to fix it was to make the error string a constant and then have everyone who wants to use it make a new Error and pass this string to it. That makes sure all of the strings are using the same shared constant.

Note: I think this isn't an issue for errorNotSupportedOnPlatform because that's an SdkError instead of an Error, which doesn't store callstack information

AE-MS (dismissed):
:shipit:

packages/teams-js/src/private/videoEx.ts
```diff
@@ -212,16 +212,17 @@ export namespace videoEx {
 
   /**
    * @hidden
+   *
    * Checks if video capability is supported by the host
    * @beta
    *
-   * @returns true if the video capability is enabled in runtime.supports.video and
-   * false if it is disabled
+   * @returns boolean to represent whether the video capability is supported
```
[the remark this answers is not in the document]
TrevorJoelHarris: No, you should probably just interpret this as "there are 72 files and it's easy to miss stuff". Fixed this one

packages/teams-js/src/private/logs.ts
```diff
@@ -26,7 +26,7 @@ export namespace logs {
   export function registerGetLogHandler(handler: () => string): void {
     // allow for registration cleanup even when not finished initializing
     handler && ensureInitialized();
-    if (!isSupported()) {
+    if (handler && !isSupported()) {
```
jadahiya-MSFT, line 29: If `handler` is undefined for whatever reason, wouldn't this skip throwing that it is not supported?

jadahiya-MSFT: actually this should be fine

TrevorJoelHarris: Yes, it would! We've already been following that path in these functions (see line 28), specifically because we have some unit tests that don't call initialize (for whatever reason) and then part of cleaning up between tests is coming and in and calling functions like this with null handlers. I suspect there might be a better way to structure this long term, but outside the scope of this already huge PR

packages/teams-js/src/public/profile.ts
```diff
@@ -164,6 +161,11 @@ export type Rectangle = {
   height: number;
 };
 
+/**
+ * @hidden
+ * An internal representation of the showProfile parameters suitable for sending via postMessage.
+ * The hub expects to receive an object of this type.
+ */
```
erinha, line 168: Does this need a beta tag? (Likely outside the scope of your PR, but just noticed it)

packages/teams-js/test/private/files.spec.ts
```diff
@@ -36,7 +37,7 @@ describe('files', () => {
   describe('getCloudStorageFolders', () => {
     it('should not allow calls before initialization', async () => {
       await expect(() => files.getCloudStorageFolders('channelId', emptyCallback)).toThrowError(
-        'The library has not yet been initialized',
+        new Error(errorLibraryNotInitialized),
```
erinha, line 40: Just out of curiosity - it's interesting to me that these tests were previously calling `toThrowError` and passing a string vs an `Error`. Does Jest internally do the same sort of comparison? Or did we used to throw a `string` and now we are throwing an `Error`?

packages/teams-js/src/public/profile.ts
TrevorJoelHarris: Since it's hidden I don't think it' s a big deal (no dev docs will get generated for this) but I will add it for correctness!

packages/teams-js/test/private/files.spec.ts
TrevorJoelHarris: toThrowError does some magic under the covers looking for Errors that match the supplied Error OR errors with the same message as the provided string

erinha (approved):
🚀

change/@microsoft-teams-js-cc4b3c12-cbf1-4531-a77a-fe0f5c58b446.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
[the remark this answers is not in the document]
TrevorJoelHarris: Good question! I wanted to try to communicate that this bug fix does carry a possibility of breaking existing applications that were using the isSupported functions incorrectly before this.

PR #1431, opened 2022-10-28
Create a common request message validator and use it in the pages test

AE-MS:
## Description

As part of fixing the lint errors in the `pages` unit tests, I moved many repeated blocks of code which validated `MessageRequest` objects into a single, shared, helper. These tests now consistently validate their `MessageRequest` objects and have many of their lint errors fixed/removed.

Some of the validation places in pages still remain because they use a different type, also called `MessageRequest`. This type is only used by the "frameless" tests. Rather than also write another validator for these, it seemed like a future change would be to understand if these two types could be shared/reused since they are identically named (but not actually defined identically). I left those tests alone for the purposes of this change.

### Main changes in the PR:

1. Create request validation helper methods
2. Make the pages tests use them

### Validation performed:

Unit tests were run and passed.

### Unit Tests added:

No new unit tests were added for this change.

### End-to-end tests added:

No.

### Change file added:

It told me no changefile was needed (probably because none of the changes are in then teams-js package directory).

### Next/remaining steps:

Try to fix the frameless tests, then expand usage of this new helper to other unit tests outside of `pages`.

### DIFF
--- a/packages/teams-js/test/public/pages.spec.ts
+++ b/packages/teams-js/test/public/pages.spec.ts
@@ -1,3 +1,4 @@
+import { validOrigins } from '../../src/internal/constants';
 import { DOMMessageEvent } from '../../src/internal/interfaces';
 import { getGenericOnCompleteHandler } from '../../src/internal/utils';
 import { app } from '../../src/public/app';
@@ -7,6 +8,7 @@ import { pages } from '../../src/public/pages';
 import { _minRuntimeConfigToUninitialize } from '../../src/public/runtime';
 import { version } from '../../src/public/version';
 import { FramelessPostMocks } from '../framelessPostMocks';
+import { MatcherType, validateExpectedValuesInRequest, validateRequestWithoutValues } from '../resultValidation';
 import { MessageResponse, Utils } from '../utils';
 
 /* eslint-disable */
@@ -61,9 +63,7 @@ describe('Testing pages module', () => {
           pages.returnFocus(true);
 
           const returnFocusMessage = utils.findMessageByFunc('returnFocus');
-          expect(returnFocusMessage).not.toBeNull();
-          expect(returnFocusMessage.args.length).toBe(1);
-          expect(returnFocusMessage.args[0]).toBe(true);
+          validateExpectedValuesInRequest(returnFocusMessage, MatcherType.ToBe, true);
         });
 
         it(`pages.returnFocus should not successfully returnFocus when set to false and initialized with ${context} context`, async () => {
@@ -72,9 +72,7 @@ describe('Testing pages module', () => {
           pages.returnFocus(false);
 
           const returnFocusMessage = utils.findMessageByFunc('returnFocus');
-          expect(returnFocusMessage).not.toBeNull();
-          expect(returnFocusMessage.args.length).toBe(1);
-          expect(returnFocusMessage.args[0]).toBe(false);
+          validateExpectedValuesInRequest(returnFocusMessage, MatcherType.ToBe, false);
         });
       });
     });
@@ -103,9 +101,7 @@ describe('Testing pages module', () => {
             return true;
           });
           const messageForRegister = utils.findMessageByFunc('registerHandler');
-          expect(messageForRegister).not.toBeNull();
-          expect(messageForRegister.args.length).toBe(1);
-          expect(messageForRegister.args[0]).toBe('focusEnter');
+          validateExpectedValuesInRequest(messageForRegister, MatcherType.ToBe, 'focusEnter');
         });
 
         it(`pages.registerFocusEnterHandler should successfully invoke focus enter handler when set to true and initialized with ${context} context`, async () => {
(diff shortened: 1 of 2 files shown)


packages/teams-js/test/resultValidation.ts
```diff
@@ -0,0 +1,52 @@
...
+export enum MatcherType {
+  ToBe,
+  ToStrictEqual,
+}
+
+// This function will return the number of assertions used to validate a request object using
+// validateRequestWithoutArguments or validateExpectedArgumentsInRequest.
+// The number of assertions used changes depending on how many arguments are being validated, so
+// you have to pass in the number of arguments being validated.
+// The value returned from this function can be used to in calls to expect.assertions() if you have them.
```
aeunms, line 12: Nit: extra "to"

aeunms (approved):
This is pretty awesome!

packages/teams-js/test/resultValidation.ts
```diff
@@ -0,0 +1,52 @@
...
+  matcher: MatcherType,
+  ...expectedArgs: unknown[]
+): void {
+  expect(request).not.toBeNull();
+  expect(request?.func).toEqual(expectedFunctionName);
+  if (expectedArgs.length > 0) {
+    expect(request!.args).toBeDefined();
+    expect(request!.args!.length).toBe(expectedArgs.length);
+
+    for (let i = 0; i < expectedArgs.length; ++i) {
```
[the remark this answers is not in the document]
AE-MS: I would argue that `++i` is the usual, but I guess it depends on what circles you run in.  😁

If you don't need the behavior of post-increment, you should always use pre-increment because it _can_ be more efficient. This can certainly vary across languages, technologies, runtimes, etc. However, AFAIK it's not ever the case that pre-increment is less efficient, it just might be as efficient. Even if efficiency were not a factor, I would just go back to "if you aren't depending on the side-effects of post-increment then don't use it."

https://stackoverflow.com/questions/24901/is-there-a-performance-difference-between-i-and-i-in-c

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -460,13 +468,15 @@ describe('Testing pages module', () => {
             const promise = pages.navigateToApp(navigateToAppParams);
 
             const executeDeepLinkMessage = utils.findMessageByFunc('executeDeepLink');
-            utils.respondToMessage(executeDeepLinkMessage, true);
-            await promise;
-
-            expect(executeDeepLinkMessage).not.toBeNull();
-            expect(executeDeepLinkMessage.args[0]).toBe(
+            validateExpectedArgumentsInRequest(
```
TrevorJoelHarris, line 471: Looks like the order of this was switched (used to happen after the respond / await, and now we validate before). Why the change?

TrevorJoelHarris: (Assume I wrote this for every place where it is switched around)

packages/teams-js/test/resultValidation.ts
```diff
@@ -0,0 +1,52 @@
...
+  ToStrictEqual,
+}
+
+// This function will return the number of assertions used to validate a request object using
+// validateRequestWithoutArguments or validateExpectedArgumentsInRequest.
+// The number of assertions used changes depending on how many arguments are being validated, so
+// you have to pass in the number of arguments being validated.
+// The value returned from this function can be used in calls to expect.assertions() if you have them.
+export function getNumberOfAssertionsUsedToValidateRequest(numberOfArgumentsBeingValidated: number): number {
+  return 2 + (numberOfArgumentsBeingValidated === 0 ? 0 : 2 + numberOfArgumentsBeingValidated);
```
TrevorJoelHarris, line 14: 2?

TrevorJoelHarris (changes requested):

packages/teams-js/test/public/pages.spec.ts
AE-MS: In the original ordering, we were using the `executeDeepLinkMessage` object in the call to `utils.respondToMessage` and THEN checking if it was not null.

With this reordering, we first check if it's null and then we use it.

Functionally this may not be that different -- prior to my reordering, the test would fail (by crashing) when `utils.respondToMessage` tried to use the null `executeDeepLinkMessage` object. The check that `executeDeepLinkMessage` was null was essentially pointless. Now with my reordering, the test will still fail, but it will fail with an error message telling you that `executeDeepLinkMessage` was null as opposed to just crashing somewhere inside of `utils.respondToMessage` where it assumes the message isn't null.

packages/teams-js/test/resultValidation.ts
AE-MS: I can add a comment if you think that would make it clearer. It's literally the number of test assertions in these methods.

AE-MS: See if it's any clearer now, I tried to replace some of the numbers with words -- but there's no way to automatically calculate these unfortunately.

aeunms (approved):
🚀

PR #1434, opened 2022-11-03
[Don't Review] iOS Pipeline Connection Settings

KangxuanYe:
**[This PR is still on testing]**

## Description

azure-pipelines.yml file is edited and a new YAML file, which is called iOS-test.yml file is added. Those files will connect iOS side with teams-library-js so that each PR will also run iOS E2E Test in the future when PR checks in. iOS-test.yml file set main UI/E2E test working flow and will publish result at the end of test. It helps teams-library-js make sure each change will not break something in iOS SDK. 

### Main changes in the PR:

1. Pipeline YAML files are added/edited

### Validation performed:

1. Check pipeline result

### Next/remaining steps:

- [ ] check in PR of [Task 7049402 Preparation for adding iOS E2E test pipeline to App SDK side]( https://office.visualstudio.com/MetaOS/_workitems/edit/7049402/) when this is ready.

### DIFF
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -1,0 +1,67 @@
+resources:
+  repositories:
+  - repository: ${{ parameters.iOSAppHostingSdkGitPath }}
+    type: git
+    name: 'iOS SDK'  # repository in Azure DevOps
+    ref: 'refs/heads/main'   
+
+steps:
+  - checkout: self
+  - checkout: ${{ parameters.iOSAppHostingSdkGitPath }}
+    persistCredentials: true
+  # - checkout: ${{ parameters.iOSAppHostingSdkGitPath }}
+  #   path: iOSHost
+  #   persistCredentials: true
+
+  - task: InstallSSHKey@0
+    displayName: 'Install an SSH key for OneDSTelemetry'
+    inputs:
+      knownHostsEntry: 'github.com ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBEmKSENjQEezOmxkZMy7opKgwFB9nkt5YRrYMjNuG5N87uRgg6CLrbo5wAdT/y6v0mKV0U2w0WZ2YB/++Tpockg='
+      sshPublicKey: '$(OneDSTelemetryKey)'
+      sshKeySecureFile: 'ios_sdk_dev_1ds_key'
+
+  - bash: |
+      sudo gem install cocoapods
+    displayName: Install Cocoapods
+    workingDirectory: '$(Agent.BuildDirectory)/iOSHost'
+
+  - bash: |
+      pod install
+    displayName: Run Cocoapods - pod install
+    workingDirectory: '$(Agent.BuildDirectory)/iOSHost'
+
+  # TODO: build iOS SDK
+
+  - task: NodeTool@0
+    inputs:
+      versionSpec: "14.x"
+    displayName: "Install Node.js"
+
+  - task: YarnInstaller@3
+    inputs:
+      versionSpec: "1.x"
+
+  - task: Yarn@2
+    inputs:
+      Arguments: 'install'
+      ProjectDirectory: '$(ClientSdkProjectDirectory)'
+    displayName: 'Run yarn install on TeamsJS using task instead of bash'
+    retryCountOnTaskFailure: 2
+
+  - task: Yarn@2
+    inputs:
+      Arguments: 'build'
+      ProjectDirectory: '$(ClientSdkProjectDirectory)'
+    displayName: 'Build TeamsJS'
+
+  - bash: 'nohup yarn start-test-app &'
+    displayName: "Run sample test app in background"
+    workingDirectory: '$(ClientSdkProjectDirectory)'
+
+  - bash: 'sh disable_slow_animation.sh'
+    displayName: 'Disable Slow Animation'
+    workingDirectory: '$(Agent.BuildDirectory)/iOSHost/jobs/ciDevTools'
+
+  - bash: 'sh parallel_ui_testing.sh'
+    displayName: "iOS UI/E2E Tests"
+    workingDirectory: '$(Agent.BuildDirectory)/iOSHost/jobs/ciDevTools'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -150,3 +150,12 @@ jobs:
       - template: tools/yaml-templates/android-test.yml
         parameters:
           androidAppHostingSdkGitPath: git://$(AndroidAppHostingSdkGitPath)
+
+  - job: E2ETestIOS
+    displayName: 'E2E Test - IOS'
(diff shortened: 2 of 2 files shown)


TrevorJoelHarris:
FYI when you are opening a PR you can mark it as a Draft to let people know it's not ready for review yet

KangxuanYe:
Should be able to review now. Branch would be replaced as PR in iOS SDK side hasn't checked in yet as well. I would like to carefully have these two PRs come in at the same time with full test.

KangxuanYe:
> FYI when you are opening a PR you can mark it as a Draft to let people know it's not ready for review yet

Will definitely take a try next time if I have the chance to write another PR.

tools/yaml-templates/ios-test.yml
```diff
@@ -0,0 +1,91 @@
+parameters:
+  - name: 'iOSAppHostingSdkGitPath'
+    default: none
+    type: string
+
+steps:
+  - checkout: self
+  # branch should be changed after test
```
AE-MS, line 8: Can you say more about what these means?

tools/yaml-templates/ios-test.yml
```diff
@@ -0,0 +1,91 @@
...
+  - checkout: ${{ parameters.iOSAppHostingSdkGitPath }}@refs/heads/kx/pipeline_connection_preparation
+    path: iOSHost
+    persistCredentials: true
+
+  - task: InstallSSHKey@0
+    displayName: 'Install an SSH key for OneDSTelemetry'
```
AE-MS, line 14: >'Install an SSH key for OneDSTelemetry' [](http://example.com/codeflow?start=17&length=39)

Can you say more about why this is needed?

AE-MS:
Can you share a screenshot of what it looks like when a test fails?

AE-MS (changes requested):
:clock1:

tools/yaml-templates/ios-test.yml
KangxuanYe: This is a ssh key required by iOS side because E2E Test would mock the full situations. For that, telemetry stuff would be used and this key is required. I do not know every detail on that but probably @baljesingh could help answering why we need that. Also, based on the discussion between me and Baljeet, I think he is working on the process to get rid of this SSH key or something like that.

baljesingh: Yes, OneDSPackage is required for end-to-end testing of the telemetry system. We need this package at Orange side (testing app) not at HubSDK. 

https://github.com/microsoft/cpp_client_telemetry

To install this package, we use CocoaPods, which further requires access to the GitHub repository mentioned above. For accessing the repository, we need an SSH key. This is a temporary solution. We'll ship it with the package later to eliminate the need to download it every time.

tools/yaml-templates/ios-test.yml
KangxuanYe: I am not sure what exactly "these" refers to but I would try my best to explain what these codes are doing.

Checkout is a term in yml file to tell pipeline which repo should be checked out. For iOS-related job in the pipeline, we need to check out the Teams-js library repository resource and iOS host SDK repository resource. Self refers to Teams-js library repo.

As for the branch in line 8&9 (including comments), It is a branch in iOS host SDK. Currently there is another PR in SDK side. That is a PR providing some bash scripts  which need to be used for this pipeline. Once we feel this PR is good for testing and publishing results, I would first check in the PR in iOS host SDK side and then change the branch in line 9 to main in this yml file. After that, as this PR be checked in, we then could announce we have a full-connected pipeline.

Hope that makes sense. Please feel free to leave any questions if I didn't explain them clearly.

Reference: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines

KangxuanYe:
> Can you share a screenshot of what it looks like when a test fails?

Definitely I could. But my concern is that test cases have some sensitive terms. Do you want me to put a screenshot here or share the published result URL link here?

tools/yaml-templates/ios-test.yml
AE-MS: So this key is required to install the cocoapod containing the package?

Is it a concern that the key is just publicly visible in this repository now?

AE-MS:
> > Can you share a screenshot of what it looks like when a test fails?
> 
> Definitely I could. But my concern is that test cases have some sensitive terms. Do you want me to put a screenshot here or share the published result URL link here?

Certainly, if there are non-public terms then I would recommend sharing the screenshot using a private medium and not this pull request. Those sensitive terms wouldn't appear publicly when a GitHub PR that runs these tests fails, right?

KangxuanYe:
> > > Can you share a screenshot of what it looks like when a test fails?
> > 
> > 
> > Definitely I could. But my concern is that test cases have some sensitive terms. Do you want me to put a screenshot here or share the published result URL link here?
> 
> Certainly, if there are non-public terms then I would recommend sharing the screenshot using a private medium and not this pull request. Those sensitive terms wouldn't appear publicly when a GitHub PR that runs these tests fails, right?

Yes. You could also check that in pipeline by reviewing the failed attempt before. E2E test would only display the prompt which is either "E2E Test Passes successfully" or "E2E Test failed". Nothing else would be shown in GitHub. Also, Github would show something like "Bash Script exit code 1" when pipeline fails.

I would send screenshots with the failed pipeline attempt URL link in the private medium.

tools/yaml-templates/ios-test.yml
KangxuanYe: Will think about how to explain it more before I ask Trevor for review.

tools/yaml-templates/ios-test.yml
baljesingh: @AE-MS we used the secure method to install the SSH key. 
https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/install-ssh-key?view=azure-devops
This key is required so that cocoapod can access the MS Internal repo to install this package.

AE-MS (approved):

TrevorJoelHarris (changes requested):

TrevorJoelHarris (approved):

tools/yaml-templates/ios-test.yml
baljesingh: @TrevorJoelHarris  As of now, we saved the public key in the pipeline as a variable(marked as secret). We followed the "Usage and best practices" section of the link mentioned above.
"_**We recommend that you don't pass in your public key as plain text to the task configuration. Instead, [set a secret variable](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?tabs=yaml%2cbatch&view=azure-devops&preserve-view=true#secret-variables) in your pipeline for the contents of your mykey.pub file**_"

jadahiya-MSFT (approved):

KangxuanYe:
@AE-MS @TrevorJoelHarris Branch connected to iOS host SDK has been reverted to main branch in iOS-test.yml. Please take a look for final review. I believe connect is ready.

TrevorJoelHarris (approved):

PR #1436, opened 2022-11-04
App audio handling with Mic mute status API

kpyapali:

## Description

> In meetings, Extensible App will handle the audio channel (Mic and Speaker) instead of Teams. 
> Also, the Ubar Mic button will be used by Extensible App to propagate the mute status.

### Main changes in the PR:

1. New API, requestAppAudioHandling() - informs Teams to go Audioless (no audio) and request the App to handle the audio channel
2. New API, sendMicMuteStatusResponse() - Send Mic mute status response acknowledgement for callbackMicMuteStateChangedHandler

### Validation performed:

1. Verified E2E with 3P app and Teams client with the new API's
2. requestAppAudioHandling will make the Teams meeting to go "Audioless" no-audio mode.
3. Mic mute state propagation when Mic status changed
4. sendMicMuteStatusResponse will send the acknowledgement for mute handler.

### Unit Tests added:

> Unit tests added appropriately

Yes

### End-to-end tests added:

Yes

### Change file added: PENDING

<Yes/No>


packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,31 @@ export namespace meeting {
...
+   * @beta
+   */
+  export interface IMicState {
+    /** Indicates the mute status of the mic*/
+    isMicMuted?: boolean;
+
+    /**
+     * error object in case there is a failure
+     */
+    error?: SdkError;
```
ydogandjiev, line 235: Why is this needed? What's the scenario that would cause an error to be sent for a mic state update and what would the expectation be for the app that receives the error?

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +586,60 @@ export namespace meeting {
...
+   * result can either contain a true boolean value (successful termination), or null (unsuccessful fetch)
+   * @beta
+   */
+  export function requestAppAudioHandling(
+    callback: (
+      error: SdkError | null,
+      requestAppAudioHandlingSdkResponse: IRequestAppAudioHandlingSdkResponse | null,
+    ) => void,
+    isAppHandlingAudio: boolean,
+    callbackMicMuteStateChangedHandler: (micState: IMicState) => void,
```
ydogandjiev, line 608: Can you please combine the isAppHandlingAudio and callbackMicMuteStateChangedHandler into a RequestAppAudioHandlingParams object so that it's easy to add more properties in the future. Also, move this parameter ahead of the callback (which should come last).

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +586,60 @@ export namespace meeting {
...
+      callback(error, requestAppAudioHandlingSdkResponse);
+      if (!error) {
+        registerHandler('meeting.micStateChanged', callbackMicMuteStateChangedHandler);
+      }
+    };
+
+    if (isAppHandlingAudio) {
+      sendMessageToParent('meeting.requestAppAudioHandling', [isAppHandlingAudio], callbackInternal);
+    } else {
+      sendMessageToParent('meeting.requestAppAudioHandling', [isAppHandlingAudio]);
```
ydogandjiev, line 631: The callback provided by the developer should always be invoked which won't happen in the case where isAppHandlingAudio is specified as false. Please make sure to always invoke the callback so that the app developers knows when the API has completed and we have a way to return errors back to them.

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +586,60 @@ export namespace meeting {
...
+    }
+    ensureInitialized(runtime, FrameContexts.sidePanel, FrameContexts.meetingStage);
+
+    const callbackInternal = (
+      error: SdkError | null,
+      requestAppAudioHandlingSdkResponse: IRequestAppAudioHandlingSdkResponse | null,
+    ): void => {
+      callback(error, requestAppAudioHandlingSdkResponse);
+      if (!error) {
+        registerHandler('meeting.micStateChanged', callbackMicMuteStateChangedHandler);
```
ydogandjiev, line 624: I think it would be better to do this before invoking the callback. Also, you need to clear the handler when a app stops handling the audio.

ydogandjiev (changes requested):

packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,36 @@ export namespace meeting {
...
+
+    /**
+     * error object in case there is a failure
+     */
+    error?: SdkError;
+  }
+
+  export interface RequestAppAudioHandlingParams {
+    isAppHandlingAudio: boolean;
+    callbackMicMuteStateChangedHandler: (micState: IMicState) => void;
```
TrevorJoelHarris, line 240: Rather than having this interface return a nullable boolean and a nullable error to the user, it would probably be better to just have it always return isMicMuted to the user and then throw any error it gets from the host

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +591,78 @@ export namespace meeting {
       sendMessageToParent('meeting.appShareButton.setOptions', [shareInformation]);
     }
   }
+
+  /**
+   * Ext App will handle the Audio(Mic & Speaker). Teams to go Audio less (No Audio)
```
TrevorJoelHarris, line 596: I'm not sure how to interpret this sentence. I think it's supposed to describe what this function does, and if so it should be more explicitly described. Something like "Have the external application handle audio (mic & speaker) and turn off host audio". It would also be good to describe when host audio will resume

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +591,78 @@ export namespace meeting {
...
+      requestAppAudioHandlingSdkResponse: IRequestAppAudioHandlingSdkResponse | null,
+    ): void => {
+      if (doesHandlerExist('meeting.micStateChanged')) {
+        removeHandler('meeting.micStateChanged');
+      }
+      callback(error, requestAppAudioHandlingSdkResponse);
+    };
+
+    if (requestAppAudioHandlingParams.isAppHandlingAudio) {
+      sendMessageToParent(
```
TrevorJoelHarris, line 644: It would make sense to move the callback definition of callbackInternalRequest into this check, and probably to just make a helper function for this called something like startAppAudioHandling

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +591,78 @@ export namespace meeting {
...
+        'meeting.requestAppAudioHandling',
+        [requestAppAudioHandlingParams.isAppHandlingAudio],
+        callbackInternalRequest,
+      );
+    } else {
+      sendMessageToParent(
```
TrevorJoelHarris, line 650: It would make sense to move the callback definition of callbackInternalStop into this check, and probably to just make a helper function for this called something like stopAppAudioHandling

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +591,78 @@ export namespace meeting {
...
+        [requestAppAudioHandlingParams.isAppHandlingAudio],
+        callbackInternalStop,
+      );
+    }
+  }
+
+  /**
+   * Send Mic mute status response acknowledgement for callbackMicMuteStateChangedHandler
+   *
+   * @param micState - Mic mute ack
```
TrevorJoelHarris, line 661: "Mic mute ack" is not a descriptive enough comment for public facing documentation. Please use @link to reference where a developer should use this

TrevorJoelHarris: Also, why do you have the user pass this callback into requestAppAudioHandling? Are there cases where an app developer would want to specify a custom handler for this? If not maybe requestAppAudioHandling should just register this callback internally so that app developers don't have to know about it

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +585,94 @@ export namespace meeting {
...
+   *     false - Audio transferred to Teams and Teams to restore the audio.
+   *   callbackMicMuteStateChangedHandler - Callback contains mic status param.
+   * @param callback - Callback contains RequestAppAudioHandlingSdkResponse param, error or operation result.
+   * error can either contain an error of type SdkError (error indication), or null (non-error indication)
+   * result can either contain a true boolean value (successful termination), or null (unsuccessful fetch)
+   * @beta
+   */
+  export function requestAppAudioHandling(
+    requestAppAudioHandlingParams: RequestAppAudioHandlingParams,
+    callback: (error: SdkError | null, isHostAudioless: boolean | null) => void,
```
TrevorJoelHarris, line 608: I think I mentioned this last week, but having a callback that returns a nullable error *and* a nullable success value is pretty difficult for consumers to use correctly. Would be better to just always only return the boolean and throw some sort of well formed error (possibly just the one you're getting from the host) rather than returning both. You'll also need some client side error handling here to make sure that we never receive both an error and a boolean at the same time since the host should never send that combination

apps/teams-test-app/src/components/MeetingAPIs.tsx
```diff
@@ -327,6 +327,52 @@ const SetOptions = (): React.ReactElement =>
...
+      ): void => {
+        if (error) {
+          setResult(JSON.stringify(error));
+        } else {
+          setResult('requestAppAudioHandling() succeeded: ' + JSON.stringify(requestAppAudioHandlingSdkResponse));
+        }
+      };
+      const callbackMicMuteStateChangedHandler = (micState: meeting.MicState): void => {
+        if (!micState) {
+          throw new Error('micStatus should be MicState');
```
TrevorJoelHarris, line 347: This error message seems wrong now, should probably be "micState should not be null"

packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,30 @@ export namespace meeting {
     error?: SdkError;
   }
 
+  /**
+   * Interface for mic state change
+   *
+   * @beta
+   */
+  export interface MicState {
```
TrevorJoelHarris, line 218: As I mentioned below, it would be better to have an interface that only returns the correct values to the app developer and throws SdkErrors if something exceptional is received from the host.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1341,4 +1342,157 @@ describe('meeting', () => {
...
+            `This call is only allowed in following contexts: ${JSON.stringify(
+              allowedContexts,
+            )}. Current context: "${context}".`,
+          );
+        });
+      }
+    });
+  });
+
+  describe('sendMicMuteStatusResponse', () => {
```
TrevorJoelHarris, line 1473: Given the current state of the MicState interface I would assume you would need unit tests making sure a user can't pass in an SdkError AND a mic status value

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1341,4 +1342,157 @@ describe('meeting', () => {
...
+              args: [null, requestAppAudioHandlingSdkResponse],
+            },
+          } as DOMMessageEvent);
+          expect(callbackCalled).toBe(true);
+          expect(returnedSdkError).toBeNull();
+          expect(returnedRequestAppAudioHandlingSdkResponse).not.toBeNull();
+          expect(returnedRequestAppAudioHandlingSdkResponse).toBe(requestAppAudioHandlingSdkResponse);
+        });
+
+        it('should throw if the requestAppAudioHandling message sends and fails', async () => {
```
TrevorJoelHarris, line 1414: I don't see any logic in here checking for throws. I do see you looking at what error was returned on line 1450 though.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1341,4 +1342,157 @@ describe('meeting', () => {
       }
     });
   });
+
+  describe('requestAppAudioHandling', () => {
```
TrevorJoelHarris, line 1346: I don't see any unit tests around setting isAppHandlingAudio to false, and that is a significantly different code path then when it is true. Please make sure you have a complete set of unit tests for all functionality. For example, you should make sure callbacks aren't called if the app stops handling audio first.

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: Not sure why you resolved this one: you still have a nullable error and nullable boolean value being returned to developers in the same function

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,39 @@ export namespace meeting {
...
+  }
+
+  export interface MeetingFailedResponse {
+    id: number;
+    /**
+     * The reason for the failure
+     */
+    reason: string;
+    /**
+     * This property is currently unused.
```
TrevorJoelHarris, line 233: Why do you need an unused optional message property? That seems like it would overlap with the required reason property.

packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,39 @@ export namespace meeting {
...
+  export interface MicState {
+    id: number;
+    /**
+     * Indicates the mute status of the mic.
+     */
+    isMicMuted: boolean;
+  }
+
+  export interface MeetingFailedResponse {
+    id: number;
```
TrevorJoelHarris, line 227: id should have a user facing comment. I assume that this id is unique among MeetingFailedResponses and monotonically increasing?

packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,39 @@ export namespace meeting {
     error?: SdkError;
   }
 
+  /**
+   * Interface for mic state change
+   *
+   * @beta
+   */
+  export interface MicState {
+    id: number;
```
TrevorJoelHarris, line 219: Same comment here as on MeetingFailedResponse.id

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: @kpyapali *Please* stop resolving this comment without making any code change or any explanation in a response. You still have a nullable sdk error and a nullable boolean being passed to this callback. When you return an SdkError to the caller this way from a function that already throws errors, you've created two different paths that errors can come out of the function. Is there some reason you can't have startAppAudioHandling and stopAppAudioHandling throw the SdkErrors if they exist? That would simplify a lot of this code from the consumer's point of view

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris, line 219: How is this id being set? My understanding was that the id needed to be set by the teamsjs-sdk so that the host could use them to make sure they were handling the requests in the right order

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +594,93 @@ export namespace meeting {
...
+    }
+  }
+
+  function startAppAudioHandling(
+    requestAppAudioHandlingParams: RequestAppAudioHandlingParams,
+    callback: (error: SdkError | null, isHostAudioless: boolean | null) => void,
+  ): void {
+    const callbackInternalRequest = (error: SdkError | null, isHostAudioless: boolean | null): void => {
+      if (error && isHostAudioless != null) {
+        throw new Error('[requestAppAudioHandling] Callback response - both parameters cannot be set');
```
thomasnakagawa, line 640: @TrevorJoelHarris Question about throwing errors here. Since this is in a `sendMessageToParent` callback, would the app still be able to catch these errors? If not, then it seems like using the optional error parameter in the callback is the only way to tell the app about these errors?

packages/teams-js/src/public/meeting.ts
```diff
@@ -210,6 +210,39 @@ export namespace meeting {
...
+     * This property is currently unused.
+     */
+    message?: string;
+  }
+
+  export interface RequestAppAudioHandlingParams {
+    isAppHandlingAudio: boolean;
+    callbackMicMuteStateChangedHandler: (
+      micState: MicState,
+      callbackResponse: (error?: MeetingFailedResponse, micState?: MicState) => void,
```
thomasnakagawa, line 242: I remember we said we should remove this `callbackResponse` parameter because it's a non-established pattern. Without it though, I don't think we can do the thing where the mute response is automatically sent (line 645 - 649 in this file), since muting/unmuting on the app side would be an async operation. Instead, the app would have to call the other API `updateMicState` to notify the host that the mute operation is done.

Unless there's some way to use async for this, I think we need to make it so app _have_ to call the other API to confirm the mute/unmute.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: No, the app can't really catch these errors, and in this case I think that's correct. These errors are truly exceptional: the host is passing the app malformed data in response to a request. I'd much rather have this function fail immediately so we get a bug opened against us if something goes wrong here. Good question!

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +601,121 @@ export namespace meeting {
...
+      }
+      if (typeof isHostAudioless !== 'boolean') {
+        throw new Error('[requestAppAudioHandling] Callback response - isHostAudioless must be a boolean');
+      }
+
+      const micStateChangedCallback = async (micState: MicState): Promise<void> => {
+        const newMicState = await requestAppAudioHandlingParams.callbackMicMuteStateChangedHandler(micState);
+
+        const micStateDidUpdate = newMicState.isMicMuted === micState.isMicMuted;
+        setMicStateWithReason(
```
kpyapali, line 661: How to send the error code to Host?

packages/teams-js/src/public/meeting.ts
```diff
@@ -560,4 +601,121 @@ export namespace meeting {
...
+   * @param micState - The new state that the microphone is in
+   *   isMicMuted - Boolean to indicate the current mute status of the mic.
+   *
+   * @beta
+   */
+  export function updateMicState(micState: MicState): void {
+    setMicStateWithReason(micState, MicStateChangeReason.AppInitiated);
+  }
+
+  function setMicStateWithReason(micState: MicState, reason: MicStateChangeReason): void {
```
kpyapali, line 717: func name and message name being sent are not aligned.
Better to have aligned name?

packages/teams-js/src/public/meeting.ts
thomasnakagawa: Would it make sense to have a try catch here and then use a new MicStateChangeReason `MicStateChangeReason.AppFailedToChange`?

packages/teams-js/src/public/meeting.ts
thomasnakagawa: Makes sense, how does `sendMicMuteStatusResponseWithReason` for the function name sound? Or should we change the message name?

TrevorJoelHarris:
This change will require a changefile. Please see [here](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball) for details on how to do this. Pay special attention to formatting requirements.

packages/teams-js/src/public/meeting.ts
thomasnakagawa: Or how about changing the message name to `meeting.updateMicState`, since it's not necessarily sent as a response

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: Are you trying to handle cases where requestAppAudioHandlingParams.callbackMicMuteStateChangedHandler throws an error? If so you could do something like that, but you'll need to be careful about what data about the error you pass up to the host (You can't pass the message for example because it may contain information we aren't allowed to log).

packages/teams-js/src/public/meeting.ts
```diff
@@ -613,4 +655,125 @@ export namespace meeting {
...
+      }
+
+      const micStateChangedCallback = async (micState: MicState): Promise<void> => {
+        try {
+          const newMicState = await requestAppAudioHandlingParams.callbackMicMuteStateChangedHandler(micState);
+
+          const micStateDidUpdate = newMicState.isMicMuted === micState.isMicMuted;
+          setMicStateWithReason(
+            newMicState,
+            micStateDidUpdate ? MicStateChangeReason.HostInitiated : MicStateChangeReason.AppDeclinedToChange,
```
TrevorJoelHarris, line 718: I think you have this backwards, right now micStateDidUpdate is true if the value of isMicMuted didn't change. (see line 715). Probably you want the opposite of that for this code to make sense.

packages/teams-js/src/public/meeting.ts
```diff
@@ -613,4 +655,125 @@ export namespace meeting {
...
+
+    sendMessageToParent(
+      'meeting.requestAppAudioHandling',
+      [requestAppAudioHandlingParams.isAppHandlingAudio],
+      callbackInternalStop,
+    );
+  }
+
+  /**
+   * Notifies the host that the microphone state has changed in the app.
```
TrevorJoelHarris, line 765: If a developer calls this directly without having called requestAppAudioHandling first, is that a way for an app to ask the host to mute the microphone? Or should this function only be called after requestAppAudioHandling?

packages/teams-js/src/public/meeting.ts
thomasnakagawa: I changed it, lmk if you think of another name you'd rather use

packages/teams-js/src/public/meeting.ts
thomasnakagawa: Add the change for this

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: (Assuming I'm correct about this, I would have also expected a unit test to catch this and fail. Not sure if that means you didn't write a unit test for this case or if something else is missing, but would be good to look into)

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/meeting.ts
thomasnakagawa: I believe this is correct because when `micState` and `newMicState` have the same value, that means the app has applied the mic state that was requested from the host. If it returned a different value, that means the value the host requested is not reflected in the app

packages/teams-js/src/public/meeting.ts
thomasnakagawa: From my understanding, yes apps can tell the host about its mute state without first having a request to mute or unmute. The `MicStateChangeReason.AppInitiated` reason is how the host can know that its a mic update that it didnt request

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: Ah okay gotcha, that particular detail I missed but makes sense

TrevorJoelHarris (approved):

packages/teams-js/src/public/meeting.ts
```diff
@@ -245,6 +245,47 @@ export namespace meeting {
...
+     * Indicates whether the app is requesting to start handling audio, or if
+     * it's giving audio back to the host
+     */
+    isAppHandlingAudio: boolean;
+    /**
+     * Callback for the host to tell the app to change its microphone state
+     * @param micState The microphone state for the app to use
+     * @returns A promise with the updated microphone state
+     */
+    callbackMicMuteStateChangedHandler: (micState: MicState) => Promise<MicState>;
```
ydogandjiev, line 286: "Callback" and "Handler" are repetitive and mean essentially the same thing. If you look at other similar interfaces (e.g. AuthenticateParameters), you will see that the name of these callbacks is usually something like:
successCallback
failureCallback

I think it would be much better if this was just called `micMuteStateChangedCallback`.

ydogandjiev (approved):

TrevorJoelHarris (approved):

ydogandjiev (approved):

PR #1440, opened 2022-11-04
Fix all lint errors in framelessPostMocks, remove unnecessary that=this

AE-MS:
## Description

I was trying to understand how the framelessPostMock stuff in the unit tests work and one thing I kept tripping over was the whole `that = this` thing it was doing (with no explanation). After some more poking around I decided it was being done because the original author wanted to properly capture a `this` pointer to use in the callback methods. In this change, I have updated the callback methods to use arrow functions notation instead, which automatically captures the `this` pointer correctly and doesn't require any this/that stuff.

It's possible that really the whole thing needs to be reworked but, if that's true, at least it's something I can do iteratively by making the code less and less confusing until I understand better what it's doing and why.

I also fixed the other lint errors in `framelessPostMocks.ts`.

### Main changes in the PR:

- Use arrow functions to capture `this` pointer where it was previously using `that`
- Fix remaining lint errors in `framelessPostMocks.ts`

### Validation performed:

Ensure all tests pass

### Unit Tests added:

No.

### End-to-end tests added:

No.

### Change file added:

Beachball tells me it's not needed.

### DIFF
--- a/packages/teams-js/test/framelessPostMocks.ts
+++ b/packages/teams-js/test/framelessPostMocks.ts
@@ -4,10 +4,6 @@ import { DOMMessageEvent, ExtendedWindow, MessageRequest, MessageResponse } from
 import { app } from '../src/public/app';
 import { applyRuntimeConfig, IRuntime } from '../src/public/runtime';
 
-/* eslint-disable */
-/* As part of enabling eslint on test files, we need to disable eslint checking on the specific files with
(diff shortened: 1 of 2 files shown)


erinha (approved):
Woohoo no more `that = this;`!!

jekloudaMSFT (approved):

PR #1441, opened 2022-11-04
Post Android E2E Test Results on Cancellation and Remove Retry

jekloudaMSFT:

## Description
Post Android E2E test results on cancellation so that we can see test results after timeout. Remove retry from Android E2E tests since the second run may not have enough time to finish before timing out

### Main changes in the PR:

1. Remove retry from Android E2E tests
2. Add on cancelled condition to test results publishing

### Validation performed:

1. Run Android E2E tests, tests should complete and post results on success or failure.

### Unit Tests added:

No unit tests necessary: YAML change

### End-to-end tests added:

No

### Change file added:

No

### Next/remaining steps:

Going to try implementing per-test retries on the Android side for more efficient flakey test mitigation

### DIFF
--- a/tools/yaml-templates/android-test.yml
+++ b/tools/yaml-templates/android-test.yml
@@ -42,7 +42,6 @@ steps:
   - bash: 'chmod u+x ./gradlew && chmod u+x e2eTest.sh && ./e2eTest.sh'
     displayName: 'Run Android E2E Tests'
     workingDirectory: '$(Agent.BuildDirectory)/androidHost/apps/orangeandroid'
-    retryCountOnTaskFailure: 1
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results'
@@ -52,4 +51,4 @@ steps:
       testRunTitle: 'E2E Tests - Android'
       searchFolder: '$(Agent.BuildDirectory)/androidHost'
       mergeTestResults: true
-    condition: succeededOrFailed()
+    condition: or(succeededOrFailed(), eq(variables['Agent.JobStatus'], 'Canceled'))


jekloudaMSFT:
Successful Android test run with changes: https://office.visualstudio.com/ISS/_build/results?buildId=16623045&view=logs&j=2893a480-1613-5722-9236-ab558bdf2fb7

tools/yaml-templates/android-test.yml
```diff
@@ -52,4 +51,4 @@ steps:
       testRunTitle: 'E2E Tests - Android'
       searchFolder: '$(Agent.BuildDirectory)/androidHost'
       mergeTestResults: true
-    condition: succeededOrFailed()
+    condition: or(succeededOrFailed(), eq(variables['Agent.JobStatus'], 'Canceled'))
```
KangxuanYe, line 54: Is the change gonna publish result anyway even if E2E task is cancelled?

jekloudaMSFT: Yes, that's the goal, succeededOrFailed() resolves to false when a task is cancelled due to timeout

AE-MS (approved):

tools/yaml-templates/android-test.yml
KangxuanYe: > Yes, that's the goal, succeededOrFailed() resolves to false when a task is cancelled due to timeout

Great. Another lesson learned as iOS could do the same thing leveraging on what you found. I would do some tests later.

KangxuanYe (approved):

PR #1446, opened 2022-11-10
Export notifyError function

shilcare:
## Description
Add a `notifyFatalError` function in videoEx to enable video apps to notify the host of fatal errors.

### Main changes in the PR:

1. Add `notifyFatalError` function in videoEx

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:
Yes

### End-to-end tests added:
No

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/private/videoEx.ts
+++ b/packages/teams-js/src/private/videoEx.ts
@@ -278,6 +278,10 @@ export namespace videoEx {
    * Limited to Microsoft-internal use
    */
   export function notifyFatalError(errorMessage: string): void {
+    ensureInitialized();
+    if (!video.isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
     notifyError(errorMessage, ErrorLevel.Fatal);
   }
 }
--- a/packages/teams-js/src/private/videoEx.ts
+++ b/packages/teams-js/src/private/videoEx.ts
@@ -269,9 +269,7 @@ export namespace videoEx {
 
   /**
    * @hidden
-   * Sending fatal error notification to host
-   *
-   * **Note**: Call this function only when your app meets fatal error and can't continue.
+   * Sending fatal error notification to host. Call this function only when your app meets fatal error and can't continue.
    * The host will stop the video pipeline and terminate this session, and optionally, show an error message to the user.
    * @beta
    * @param errorMessage - The error message that will be sent to the host
--- a/packages/teams-js/src/private/videoEx.ts
+++ b/packages/teams-js/src/private/videoEx.ts
@@ -15,6 +15,18 @@ import { video } from '../public/video';
 export namespace videoEx {
   /**
    * @hidden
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/private/videoEx.ts
```diff
@@ -104,7 +104,9 @@ export namespace videoEx {
             () => {
               notifyVideoFrameProcessed(timestamp);
             },
-            notifyError,
+            (errorMessage: string) => {
+              notifyError(`[VideoFrameProcessingFailed]: ${errorMessage}`);
+            },
```
nandin-borjigin, line 109: If the expected outcome of calling `notifyError` is to stop the video pipeline, it might be risky to use it here as it hadn't been that way from a dev's perspective.

Rather, we could expose `notifyError` as a standalone API, which accepts different levels of errors (e.g., Fatal, Alert, Minor, etc.) and the Teams client can decide on what to do based on the error level.

For frame processing failure, we may use `Alert` or `Minor` levels for backward compatibility

packages/teams-js/test/private/videoEx.spec.ts
```diff
@@ -634,4 +634,24 @@ describe('videoEx', () => {
...
   });
+
+  describe('notifyFatalError', () => {
+    it('FRAMED - should send error to host successfully', async () => {
+      await framedPlatformMock.initializeWithContext('sidePanel');
+      const fakeErrorMsg = 'fake error';
+      videoEx.notifyFatalError(fakeErrorMsg);
+      const message = framedPlatformMock.findMessageByFunc('video.notifyError');
+      expect(message.args.length).toBe(2);
+      expect(message.args[0]).toEqual(fakeErrorMsg);
```
TrevorJoelHarris, line 645: these tests should probably be validating errorLevel as well, yes?

packages/teams-js/src/private/videoEx.ts
```diff
@@ -246,11 +258,26 @@ export namespace videoEx {
...
+   * @beta
+   * @param errorMessage - The error message that will be sent to the host
    *
    * @internal
    * Limited to Microsoft-internal use
    */
-  function notifyError(errorMessage: string): void {
-    sendMessageToParent('video.notifyError', [errorMessage]);
+  export function notifyFatalError(errorMessage: string): void {
+    notifyError(errorMessage, ErrorLevel.Fatal);
```
TrevorJoelHarris, line 281: Your exported functions should be calling ensureInitialized() and checking video.isSupported() (see other functions in this file)

packages/teams-js/test/private/videoEx.spec.ts
shilcare: updated

packages/teams-js/src/private/videoEx.ts
shilcare: added, thanks

TrevorJoelHarris (approved):

maglims (approved):

PR #1455, opened 2022-11-16
Wenghe/register for video effect returns promise

WengHe:
## Description

There is an issue that on switching to a new video effect sometimes some unprocessed frames are shown to the users, or the video is frozen until the new effect is ready, this is because the video effect is not ready on switching. To mitigate this, we would like to make a change to the API video.registerForVideoEffect(callback) to let the callback return a Promise, which will be resolved when the video effect is prepared. 

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### Next/remaining steps:

The feature is 100% completed after this


packages/teams-js/src/public/video.ts
```diff
@@ -154,7 +184,24 @@ export namespace video {
...
-    registerHandler('video.effectParameterChange', callback, false);
+
+    const callbackHandler = (effectId: string | undefined) => {
+      const result = callback(effectId);
+      if (result) {
+        result
+          .then((effectRequirement) => {
+            sendMessageToParent('videoEffectReadiness', [true, effectId, effectRequirement]);
+          })
+          .catch((reason) => {
```
shilcare, line 195:
```js
Promise.resolve(callback(effectId))
    .then((result) => {
        // 
    })
```

shilcare (commented):
use `Promise.resolve(promiseOrUndefinedResult)` so that you can handle it as a promise no matter the result is a promise or not.

packages/teams-js/src/public/video.ts
```diff
@@ -154,7 +170,19 @@ export namespace video {
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
-    registerHandler('video.effectParameterChange', callback, false);
+
+    const callbackHandler = (effectId: string | undefined): void => {
```
nandin-borjigin, line 174: `callback` or `handler`, they are basically same. but stacking both of them seems weird.

packages/teams-js/src/public/video.ts
```diff
@@ -154,7 +170,19 @@ export namespace video {
...
+      Promise.resolve(callback(effectId))
```
nandin-borjigin, line 175: There are two options for backward compatibility (for those video apps which don't return a promise)

1. Treat them as a normal case and send `videoEffectReadiness/true` back to the client.
2. Treat them as a special case and send `videoEffectReadiness/false` back to the client with a specific reason.

I personally prefer #2 because it makes the situation explicit and leave a room for client to decide on whether the case should be treated as normal or not.

WengHe: After the offline discussion, I will go with sending `videoEffectReadiness/true + "additional info"` back to the client, so that the client doesn't need to treat them as a special failure case while having a chance to identify if it is from an old video app which returns the callback directly which may cause perf alerts.

packages/teams-js/src/public/video.ts
```diff
@@ -88,11 +88,33 @@ export namespace video {
...
+    /**
+     * The effect can't be initialezed
+     */
+    InitializationFailure = 'InitializationFailure',
+  }
+
+  enum EffectSuccessReason {
+    EffectPrepared = 'EffectPrepared',
+
+    DirectReturned = 'DirectReturned',
```
nandin-borjigin, line 109: From grammar point of view: `DirectlyReturned` is more accurate.

From semantic point of view: `ReturnedWithoutPromise` or `NoPromiseReturned` is more declarative.

WengHe: updated to "NoPromiseReturned "

packages/teams-js/src/public/video.ts
```diff
@@ -88,11 +88,33 @@ export namespace video {
...
+   * Predefined failure reasons for preparing the selected video effect
+   * @beta
+   */
+  export enum EffectFailureReason {
+    /**
+     * A wrong effect id is provided
+     */
+    InvalidEffectId = 'InvalidEffectId',
+    /**
+     * The effect can't be initialezed
```
TrevorJoelHarris, line 101: Should be 'initialized'

packages/teams-js/src/public/video.ts
```diff
@@ -88,11 +88,33 @@ export namespace video {
...
+    NoPromiseReturned = 'NoPromiseReturned',
+  }
+
   /**
    * Video effect change call back function definition
+   * Return a Promise which will be resolved when the effect is prepared, or reject with a failure reason
    * @beta
    */
-  export type VideoEffectCallBack = (effectId: string | undefined) => void;
+  export type VideoEffectCallBack = (effectId: string | undefined) => Promise<void> | void;
```
TrevorJoelHarris, line 117: this is also used in video.registerForVideoFrame(...). Am I correct in assuming that for that function, the callback should be called every time a frame is ready?

I don't think I understand why this is returning either void or Promise<void>. Shouldn't it just return a Promise in both cases? It will be very confusing for developers trying to figure out whether they should return a promise or not.

TrevorJoelHarris: In your design doc you have this set to just return a Promise. I don't think there's any need to maintain backwards compatibility here, this capability and it's functions are already marked as @beta.

packages/teams-js/src/public/video.ts
```diff
@@ -154,7 +176,24 @@ export namespace video {
...
+
+    const effectParameterChangeHandler = (effectId: string | undefined): void => {
+      const promise = callback(effectId);
+      if (promise) {
+        promise
+          .then(() => {
+            sendMessageToParent('video.videoEffectReadiness', [true, effectId, EffectSuccessReason.EffectPrepared]);
+          })
+          .catch((reason) => {
+            // reason could be a string or an EffectFailureReason
```
TrevorJoelHarris, line 188: How do you know it's one of these two things? The returned promise is created by the application, they might be throwing a lot of different things out of here. You're going to need some sort of type checking to prevent invalid data potentially being sent up to the host.

Also it would be good if only one type of thing can be thrown: EffectFailureReason. Otherwise, you're passing an app-specified string up to the host, and we don't know what that might contain.

packages/teams-js/test/public/video.spec.ts
```diff
@@ -559,5 +559,123 @@ describe('video', () => {
...
+      framedPlatformMock.sendMessage('video.effectParameterChange', effectId);
+      await invokeHanlderPromise;
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      const messageForRegister = framedPlatformMock.findMessageByFunc('video.videoEffectReadiness');
+      expect(messageForRegister).not.toBeNull();
+      expect(messageForRegister?.args?.length).toBe(3);
+      expect(messageForRegister?.args).toEqual([true, effectId, 'EffectPrepared']);
+    });
+
+    it('FRAMEDLESS - should invoke videoEffectReadiness handler on callback resolved', async () => {
```
TrevorJoelHarris, line 600: should be `FRAMELESS`

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/video.ts
AE-MS, line 117: >VideoEffectCallBack [](http://example.com/codeflow?start=14&length=19)

Everywhere else in here treats callback like one word, which I think is the most common usage. As such, I recommend this be `VideoEffectCallback` (lower-case b) and not `VideoEffectCallBack`

packages/teams-js/src/public/video.ts
```diff
@@ -88,11 +88,33 @@ export namespace video {
...
+
+  enum EffectSuccessReason {
+    EffectPrepared = 'EffectPrepared',
+
+    NoPromiseReturned = 'NoPromiseReturned',
+  }
+
   /**
    * Video effect change call back function definition
+   * Return a Promise which will be resolved when the effect is prepared, or reject with a failure reason
```
AE-MS, line 114: >reject with a failure reason [](http://example.com/codeflow?start=77&length=28)

Does this mean specifically that rejected promises should throw an `EffectFailureReason` value? You should explicitly document that and reference that structure.

packages/teams-js/src/public/video.ts
```diff
@@ -88,11 +88,33 @@ export namespace video {
...
 
+  /**
+   * Predefined failure reasons for preparing the selected video effect
+   * @beta
+   */
+  export enum EffectFailureReason {
+    /**
+     * A wrong effect id is provided
+     */
+    InvalidEffectId = 'InvalidEffectId',
```
AE-MS, line 99: >InvalidEffectId = 'InvalidEffectId', [](http://example.com/codeflow?start=4&length=36)

Under what circumstances would this happen?

AE-MS:
>    * Register the video effect callback, host uses this to notify the video extension the new video effect will by applied

be

---
Refers to: packages/teams-js/src/public/video.ts:170 in c9aebcb. [](commit_id = c9aebcbdacdfc0f7d4913a09be50509f629237f2, deletion_comment = False)

AE-MS:
>    * Register the video effect callback, host uses this to notify the video extension the new video effect will by applied

This could be made clearer and more concise, e.g.,

Registers a callback to be notified when a new video effect is applied.

---
Refers to: packages/teams-js/src/public/video.ts:170 in c9aebcb. [](commit_id = c9aebcbdacdfc0f7d4913a09be50509f629237f2, deletion_comment = False)

AE-MS:
>    * @param callback - The VideoEffectCallback to invoke when registerForVideoEffect has completed

This feels like it would be clearer if it used the same language as the function description, e.g.,

Function to be called when new video effect is applied

---
Refers to: packages/teams-js/src/public/video.ts:172 in c9aebcb. [](commit_id = c9aebcbdacdfc0f7d4913a09be50509f629237f2, deletion_comment = False)

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/video.ts
WengHe: updated, please take a look

packages/teams-js/src/public/video.ts
WengHe: updated

packages/teams-js/test/public/video.spec.ts
WengHe: updated

packages/teams-js/src/public/video.ts
WengHe: updated, add type checking

packages/teams-js/src/public/video.ts
WengHe: Updated

packages/teams-js/src/public/video.ts
WengHe: Updated

packages/teams-js/src/public/video.ts
AE-MS: It may be worth documenting a few of these here, otherwise developers may wonder under what circumstances it would be used.

packages/teams-js/src/public/video.ts
WengHe: updated

WengHe:
> > * Register the video effect callback, host uses this to notify the video extension the new video effect will by applied
> 
> This could be made clearer and more concise, e.g.,
> 
> Registers a callback to be notified when a new video effect is applied.
> 
> Refers to: packages/teams-js/src/public/video.ts:170 in [c9aebcb](https://github.com/OfficeDev/microsoft-teams-library-js/commit/c9aebcbdacdfc0f7d4913a09be50509f629237f2). [](commit_id = [c9aebcb](https://github.com/OfficeDev/microsoft-teams-library-js/commit/c9aebcbdacdfc0f7d4913a09be50509f629237f2), deletion_comment = False)

packages/teams-js/src/public/video.ts
```diff
@@ -88,11 +88,28 @@ export namespace video {
...
   ) => void;
 
+  /**
+   * Predefined failure reasons for preparing the selected video effect
+   * @beta
+   */
+  export enum EffectFailureReason {
+    /**
+     * A wrong effect id is provide.
+     * Use this reason when the effect id is not found or empty, this may indicate a mismatch between the app and its manifest or a bug of the hub.
```
AE-MS, line 98: >hub [](http://example.com/codeflow?start=143&length=3)

Please use the word 'host.' Thank you!

packages/teams-js/src/public/video.ts
```diff
@@ -145,16 +162,35 @@ export namespace video {
...
     }
-    registerHandler('video.effectParameterChange', callback, false);
+
+    const effectParameterChangeHandler = (effectId: string | undefined): void => {
+      callback(effectId)
+        .then(() => {
+          sendMessageToParent('video.videoEffectReadiness', [true, effectId]);
+        })
+        .catch((reason) => {
+          if (reason in EffectFailureReason) {
```
lubobill1990, line 181: Can be simplify as?
```typescript
const validReason = reason in EffectFailureReason ? reason: EffectFailureReason.InitializationFailure;
sendMessageToParent('video.videoEffectReadiness', [
    false,
    effectId,
    validReason
  ]);

```

WengHe: updated as suggested

AE-MS (approved):
:shipit:

TrevorJoelHarris (approved):

PR #1461, opened 2022-11-28
Lakhveer/dialog frame contexts

lakhveerk:

## Description

Deleted sidePanel and meetingRoom frame contexts from ensureInitialized in dialog.submit API. The reason is `submit` supposed to be called only from inside of a dialog (frame context task). We are keeping frameContext content because of mobile team.

### Main changes in the PR:

1. Deleted sidePanel and meetingRoom frame contexts from ensureInitialized in dialog.submit API.

### Validation performed: N/A

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No> Fixed UTs in tasks and dialog file

### End-to-end tests added:

<Yes/No> Not needed

### Change file added:

<Yes/No> Yes

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:


packages/teams-js/src/public/dialog.ts
```diff
@@ -125,7 +125,7 @@ export namespace dialog {
    * @beta
    */
   export function submit(result?: string | object, appIds?: string | string[]): void {
-    ensureInitialized(FrameContexts.content, FrameContexts.sidePanel, FrameContexts.task, FrameContexts.meetingStage);
+    ensureInitialized(FrameContexts.content, FrameContexts.task);
```
AE-MS, line 128: >FrameContexts.content [](http://example.com/codeflow?start=22&length=21)

Is there a bug and/or timeline for when Teams mobile will be fixing this and when we can remove this?

AE-MS, line 128: >FrameContexts.content [](http://example.com/codeflow?start=22&length=21)

You should probably add a comment here that says something like "This should not be here, but it's only here because Teams mobile incorrectly returns FrameContext.content. When that bug is fixed, this will be removed."

AE-MS: This might be overkill but you could probably make it so that `FrameContexts.content` is only allowed when running on Teams Android/iOS (or wherever the bug exists) and that it's not allowed elsewhere.

AE-MS (changes requested):
:clock1:

AE-MS:
Thank you for fixing this! 😁

packages/teams-js/src/public/dialog.ts
lakhveerk: I have asked for the timeline. If I don't get a clear answer, I will pass the bug tracking part to our PMs who are tracking other bugs in iOS as well.

lakhveerk: Adding a comment is a good idea! I can look into restricting context to ios/mobile after UX. If it takes so much time, and feels like an over kill, we can leave it

AE-MS (approved):
:shipit:

shrshindeMSFT (approved):

PR #1462, opened 2022-11-28
Add dataResidency interface in the SDK within getUser API

Ella-ly:

### Main changes in the PR:

1. Added `dataResidency` interface in teams-js SDK within `getUser` API. Exposed a limited set of data residencies information to 1P app developers.

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

Yes.

### End-to-end tests added:

No.

### Change file added:

Yes.

### Related PRs:

### DIFF
--- a/change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
+++ b/change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
@@ -1,7 +1,7 @@
 {
-  "type": "major",
-  "comment": "Added dataResidencySlim interface in teams-js SDK within getUser API. Exposed a limited set of data residencies information to 1P app developers.",
+  "type": "minor",
+  "comment": "Added `dataResidency` property to `UserProfile` interface to expose a limited set of data residency information to 1P app developers.",
   "packageName": "@microsoft/teams-js",
-  "email": "email not defined",
+  "email": "liuella@microsoft.com",
   "dependentChangeType": "patch"
 }
--- a/change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
+++ b/change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Added dataResidencySlim interface in teams-js SDK within getUser API. Exposed a limited set of data residencies information to 1P app developers.",
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -679,7 +679,11 @@ export namespace authentication {
   }
 
   /**
+   * @hidden
    * Limited set of data residencies information exposed to 1P application developers
+   *
+   * @internal
+   * Limited to Microsoft-internal use
    */
   export enum DataResidency {
     /**
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -670,7 +670,7 @@ export namespace authentication {
     ver: string;
     /**
      * @hidden
-     * Stores the residency of the user.
+     * Stores the data residency of the user.
      *
      * @internal
      * Limited to Microsoft-internal use
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -668,6 +668,34 @@ export namespace authentication {
      * Limited to Microsoft-internal use
      */
     ver: string;
+    /**
+     * @hidden
+     * Stores the residency of the user.
+     *
+     * @internal
+     * Limited to Microsoft-internal use
+     */
+    dataResidency: DataResidencySlim;
+  }
+
+  /**
+   * Limited set of data residencies inforamtion exposed to 1P application developers
+   */
+  export enum DataResidencySlim {
+    /**
+     * Public
+     */
+    Public = 'public',
+
+    /**
+     * European Union Data Boundary
+     */
+    EUDB = 'eudb',
+
+    /**
+     * Other, stored to cover fields that will not be exposed
+     */
+    Other = 'other',
   }
 
   /**
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -679,7 +679,7 @@ export namespace authentication {
   }
 
   /**
-   * Limited set of data residencies inforamtion exposed to 1P application developers
+   * Limited set of data residencies information exposed to 1P application developers
    */
   export enum DataResidencySlim {
     /**
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -675,13 +675,13 @@ export namespace authentication {
      * @internal
      * Limited to Microsoft-internal use
      */
-    dataResidency: DataResidencySlim;
+    dataResidency?: DataResidency;
   }
 
   /**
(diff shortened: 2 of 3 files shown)


packages/teams-js/src/public/authentication.ts
```diff
@@ -668,6 +668,34 @@ export namespace authentication {
...
+     * @internal
+     * Limited to Microsoft-internal use
+     */
+    dataResidency: DataResidencySlim;
+  }
+
+  /**
+   * Limited set of data residencies information exposed to 1P application developers
+   */
+  export enum DataResidencySlim {
```
AE-MS, line 684: >DataResidencySlim [](http://example.com/codeflow?start=14&length=17)

Why does this type have "slim" in the name?

packages/teams-js/src/public/authentication.ts
```diff
@@ -668,6 +668,34 @@ export namespace authentication {
...
+
+    /**
+     * European Union Data Boundary
+     */
+    EUDB = 'eudb',
+
+    /**
+     * Other, stored to cover fields that will not be exposed
+     */
+    Other = 'other',
```
AE-MS, line 698: Is this the value that's returned if the host doesn't provide this property when `getUser` is called? Would it be valuable to have a separate enum value to distinguish between "the host gave us a data residency value but it's not public/eudb" and "the host didn't give us a value." ?

packages/teams-js/src/public/authentication.ts
```diff
@@ -668,6 +668,34 @@ export namespace authentication {
      * Limited to Microsoft-internal use
      */
     ver: string;
+    /**
+     * @hidden
+     * Stores the residency of the user.
```
AE-MS, line 673: >residency [](http://example.com/codeflow?start=18&length=9)

Maybe consider calling it "data residency" of the user, since that's the term used elsewhere?

AE-MS:
It seems like unit tests would be necessary to validate this change, but I didn't see any in the change and the PR description just says "no" and doesn't provide a reason why they aren't necessary?

change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
AE-MS, line 2: >major [](http://example.com/codeflow?start=11&length=5)

This doesn't seem like a major change. It doesn't break any existing APIs and there's I don't think there's anything an existing consumer would have to do to integrate with the newer version of this library (e.g. they wouldn't have to change their code)

change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Added dataResidencySlim interface in teams-js SDK within getUser API. Exposed a limited set of data residencies information to 1P app developers.",
```
AE-MS, line 3: >Added dataResidencySlim interface in teams-js SDK within getUser API. Exposed a limited set of data residencies information to 1P app developers. [](http://example.com/codeflow?start=14&length=145)

Thank you for using past tense! Please also surround type names with backticks. Consider something like this:

Added `dataResidency` property to `UserProfile` interface to exposed a limited set of data residency information to 1P app developers.

AE-MS:
Awesome to see these data residency changes making their way across all parts of the stack! Thanks for doing this!

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/authentication.ts
erinha: We used the name `DataResidency` on the host side for an enum that has more values than are being exported publicly. The internal-only `DataResidency` enum values are mapped to the public-facing ones in `DataResidencySlim`.

change/@microsoft-teams-js-9e642ffe-0bf6-4775-a92f-99a31e897e32.json
Ella-ly: Modified, thank you for pointing that out.

packages/teams-js/src/public/authentication.ts
AE-MS: Is that host implementation detail (that it's the "slim" version) something that is important for consumers of teamsjs be aware of?

MarcinJuraszek: I think this should just be `DataResidency` here in TeamsJs.

packages/teams-js/src/public/authentication.ts
MarcinJuraszek: I think because `dataResidency` is optional on `UserProfile` that's already covered by not having it there at all?

AE-MS: Yes, that would do it.

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -617,6 +637,50 @@ describe('Testing authentication capability', () => {
           });
         });
 
+        it(`authentication.getUser should successfully get user profile with data residency info if dataResidency is provided by hosts`, (done) => {
+          utils.initializeWithContext(context).then(() => {
```
shrshindeMSFT, line 641: Is there a way to use async-await instead of callbacks? Is there any particular reason we are using callbacks?
Same goes for all the test cases below.

packages/teams-js/src/public/authentication.ts
```diff
@@ -668,6 +668,34 @@ export namespace authentication {
...
+     * @internal
+     * Limited to Microsoft-internal use
+     */
+    dataResidency?: DataResidency;
+  }
+
+  /**
+   * Limited set of data residencies information exposed to 1P application developers
+   */
+  export enum DataResidency {
```
lakhveerk, line 684: Since data Residency is marked internal and hidden up in L678, does it make sense to mark the interface hidden/internal as well?

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -617,6 +637,50 @@ describe('Testing authentication capability', () => {
           });
         });
 
+        it(`authentication.getUser should successfully get user profile with data residency info if dataResidency is provided by hosts`, (done) => {
```
shrshindeMSFT, line 640: Any particular reason to use functional string here, I don't see any variables in the test title

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -617,6 +637,50 @@ describe('Testing authentication capability', () => {
...
+          utils.initializeWithContext(context).then(() => {
+            const successCallback = (user: authentication.UserProfile): void => {
+              expect(user).toEqual(mockUserWithDataResidency);
+              done();
+            };
+            const failureCallback = (): void => {
+              done();
```
AE-MS, line 647: > done(); [](http://example.com/codeflow?start=13&length=8)

Should this fail the test if the failureCallback is called?

shrshindeMSFT (changes requested):

packages/teams-js/test/public/authentication.spec.ts
anmaur, line 640: Could we also mention expect assertions at the top of each test? as it would help in failing appropriately incase all assertions are not called

packages/teams-js/test/public/authentication.spec.ts
Ella-ly: I am following the current `getUser` unit tests format which are using callbacks.

anmaur (approved):
LGTM!! Thanks Ella

AE-MS (approved):
:shipit:

erinha (approved):
🚀

shrshindeMSFT (approved):
LGTM

ndangudubiyyam (approved):

PR #1463, opened 2022-11-29
noahdarveau/ssr patch

noahdarveau-MSFT:

## Description

### Main changes in the PR:

1. Added teams-js patch to allow SSR app support.

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

No tests were added as the changes do not alter any code flow. The only changes are checks that allow for compilation in an SSR environment.

### End-to-end tests added:

No

### Change file added:

Yes

### Next/remaining steps:

- [ ] Incorporate a Server Side Rendered test app to ensure SSR capabilities are not broken by future changes.
- [ ] Incorporate E2E tests to be run against the SSR Test app in the release pipeline.


lerna-debug.log
```diff
@@ -0,0 +1,53 @@
+0 silly argv {
```
AE-MS, line 1: Should this file be checked in?

AE-MS:
I can't tell from the PR description: are the changes in this PR all that's needed to support SSR apps? Or is there more coming in the future? Is there validation to ensure SSR continues to work?

lerna-debug.log
noahdarveau-MSFT: Nope, that was a mistake. I removed the file.

noahdarveau-MSFT:
> I can't tell from the PR description: are the changes in this PR all that's needed to support SSR apps? Or is there more coming in the future? Is there validation to ensure SSR continues to work?

The changes in this PR are, to the best of my knowledge, all that's needed to support SSR apps. The only thing currently not working with SSR apps is that unchecked references to the `window` object are causing build failures. I have a basic SSR test app coming soon in another PR that will be used in the future to validate SSR capabilities continue to work.

TrevorJoelHarris:
It would be good to fill out the PR template: 
image

change/@microsoft-teams-js-d1f40ab4-e602-4977-9be1-67eb2a22440e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added SSR support patch",
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
```
TrevorJoelHarris, line 5: Make sure that your email address is associated with this checkin

change/@microsoft-teams-js-d1f40ab4-e602-4977-9be1-67eb2a22440e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added SSR support patch",
```
TrevorJoelHarris, line 3: Since this will turn into customer facing patch notes, it would be good to be slightly more explicit. Maybe something like "Fixed bugs preventing the use of this library in SSR applications"

packages/teams-js/src/public/app.ts
```diff
@@ -534,11 +534,16 @@ export namespace app {
...
-      new Error('SDK initialization timed out.'),
-    );
+    if (typeof window !== 'undefined') {
+      return runWithTimeout(
+        () => initializeHelper(validMessageOrigins),
+        initializationTimeoutInMs,
+        new Error('SDK initialization timed out.'),
+      );
+    } else {
+      console.log('window is undefined, continuing without initialization');
```
TrevorJoelHarris, line 544: console.log is not usually how we do logging in this library. If you look a little further down in this file (line 549) you can see `initializeHelperLogger` get initialized and then used. Following a similar pattern will make sure all of our log statements show up in the same way.

packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -36,7 +36,7 @@ export namespace teamsCore {
    * default print handler
    */
   export function print(): void {
-    window.print();
+    typeof window !== undefined ? window.print() : new Error('window does not exist');
```
TrevorJoelHarris, line 39: It looks like if window is not defined you are just creating an error and doing nothing with it. You probably want to throw the error?

packages/teams-js/webpack.config.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
       type: 'umd',
       umdNamedDefine: true,
     },
+    globalObject: 'this',
```
TrevorJoelHarris, line 30: Why was this necessary? I believe you that it is (because we talked about it) but it would be goo to capture in a comment in the code or on the PR so that future developers will know what it is for

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/app.ts
```diff
@@ -534,11 +534,16 @@ export namespace app {
    * @returns Promise that will be fulfilled when initialization has completed, or rejected if the initialization fails or times out
    */
   export function initialize(validMessageOrigins?: string[]): Promise<void> {
-    return runWithTimeout(
-      () => initializeHelper(validMessageOrigins),
-      initializationTimeoutInMs,
-      new Error('SDK initialization timed out.'),
-    );
+    if (typeof window !== 'undefined') {
```
AE-MS, line 537: >if (typeof window !== 'undefined') [](http://example.com/codeflow?start=4&length=34)

Consider extracting this check into a more semantically meaningful function you can call, e.g.:
```
if (!inServerSideRenderingEnvironment()) {
...
} else {
...
}

File: packages/teams-js/webpack.config.js (hunk 6 above)
Comment (noahdarveau-MSFT) [reply packages/teams-js/webpack.config.js]: Originally this was fixing an issue where webpack was compiling our .ts files to `self` which was conflicting with some node.js stuff. I did some more testing, and it seems since I now have nextjs working properly, it's no longer necessary, so I removed it. If I ever get new errors regarding `self` being undefined, I'll re-add it.

File: packages/teams-js/src/public/teamsAPIs.ts (hunk 5 above)
Comment (noahdarveau-MSFT) [reply packages/teams-js/src/public/teamsAPIs.ts]: Yeah, I'm dumb. Should be fixed!

File: packages/teams-js/src/public/app.ts [hunk 8]
```diff
@@ -534,11 +534,16 @@ export namespace app {
...
+      return runWithTimeout(
+        () => initializeHelper(validMessageOrigins),
+        initializationTimeoutInMs,
+        new Error('SDK initialization timed out.'),
+      );
+    } else {
+      console.log('window is undefined, continuing without initialization');
+      return Promise.resolve();
```
AE-MS, line 546: This is going to show my complete naïveté when it comes to SSR usage:
In an SSR scenario, does this code even run? My (likely incorrect) understanding was that in an SSR situation, none of the calls to teamsjs methods actually ever run (because they would all fail since there's no host SDK to answer the calls), they just need to be able to resolve all references/symbols and compile.  Is that true? If that's true, will the log statement here ever be executed?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/app.ts
noahdarveau-MSFT: To the best of my knowledge (and testing), you are correct, the log statement does not get run. I decided to include it to cover any potential scenarios that I may have missed/are unaware of in which it may get run, or if teams-js were to extend functionality in the future that would allow it to run on a server. If you feel like it is unnecessary, I am not opposed to removing it.

AE-MS: I think I would leave it in but make the message much more dire (and use the logging functions mentioned in another comment). Something like:

"This log statement should NEVER actually be written. This code path exists only to enable compilation in server-side rendering environments. If you EVER see this statement in ANY log file, something has gone horribly wrong and a bug needs to be filed. Peace out."

packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -36,7 +36,11 @@ export namespace teamsCore {
    * default print handler
    */
   export function print(): void {
-    window.print();
+    if (typeof window !== 'undefined') {
```
AE-MS, line 39: >typeof window !== 'undefined' [](http://example.com/codeflow?start=8&length=29)

Shouldn't this use your `inServerSideRenderingEnvironment` function?

packages/teams-js/src/public/teamsAPIs.ts
AE-MS: Should this code also never run because it only exists for the SSR case? If so maybe the message of the `Error` should be more explicit. E.g., "This codepath only exists to enable compilation in an SSR environment. If this error has actually been thrown something very wrong has happened and it is a bug."

AE-MS:
>   resolved "https://registry.npmjs.org/@azure/msal-browser/-/msal-browser-2.27.0.tgz"

Did you intend to update all package paths from yarn to npm?

---
Refers to: yarn.lock:7 in 1043fa6. [](commit_id = 1043fa6c25c81f04dfa86311338a1500440af33f, deletion_comment = False)

packages/teams-js/src/public/teamsAPIs.ts
noahdarveau-MSFT: Since the point of the `inServerSideRenderingEnvironment` function was to be more semantically meaningful, I was actually thinking in this instance, since the `print` function prints the window, it would be more meaningful and obvious to just directly reference window in the check. I can also go change it to the `inServerSideRenderingEnvironment` function if you strongly disagree with my reasoning or feel the consistency would be better.

AE-MS: Hmm, well, I agree that since this function is about printing the `window` then directly checking that the `window` object is available makes that connection clear. I think the readability concern I would have for other developers would be if they came to look at this and thought "That's weird, why would the window object ever be undefined?" I guess I could go either way. 🤷‍♂️

noahdarveau-MSFT: I decided to leave the direct check of `window` and supplemented it with a much more descriptive error message that should give other developers a better understanding of why the check is there, and how and answer the "That's weird, why would the window object ever be undefined?" question.

packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -36,7 +36,13 @@ export namespace teamsCore {
...
+      window.print();
+    } else {
+      throw new Error(
+        'This codepath only exists to enable compilation in a server-side redered environment. In standard usage, the window object should never be undefined so this code path should never run. If this error has actually been thrown, something has gone very wrong and it is a bug',
```
AE-MS, line 43: >redered [](http://example.com/codeflow?start=74&length=7)

typo

noahdarveau-MSFT: Fixed!

AE-MS (dismissed):
:shipit:

jadahiya-MSFT (dismissed):

TrevorJoelHarris:
This looks good and I will approve it. My only comment is that I suspect that the bundle size increase you saw originally:

image

Was probably due to the largish string error messages that you added for SSR error cases. It might be worth either making those strings shorter and adding detailed descriptions in comments (which should be removed from the package before increasing size) or making a single shared string we can use for all SSR errors so we only pay the large string cost once.

TrevorJoelHarris (dismissed):

jadahiya-MSFT (approved):

AE-MS (approved):
:shipit:

PR #1466, opened 2022-12-02
Runtime Versioning

jekloudaMSFT:

## Description

This PR implements versioning for the runtime interface to unblock the adaptive card work. With these changes, we can make breaking changes to the runtime interface while still being maintaining compatibility with older versions of the runtime.

### Main changes in the PR:

1. Use the Runtime type alias to reference the latest runtime interface
2. Rename current runtime interface to IRuntimeV1
3. Implement fastForwardRuntime, a function that upgrades outdated runtime objects to use the latest interface
5. Add RUNTIME.md which describes when and how to update the runtime version
6. Add a GitHub Action that will leave a comment on PRs that touch the runtime.ts file to prompt the author to review RUNTIME.md

### Validation performed:

1. Implemented a version 0 runtime interface to test the version upgrade mechanism
2. GitHub Action tested by comments on this PR

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### Next/remaining steps:

This change will unblock the adaptive card implementation currently in PR


jekloudaMSFT:

Added runtime versioning. No updates to current runtime interface so no new version is required

packages/teams-js/src/public/runtime.ts
```diff
@@ -66,14 +98,14 @@ export interface IRuntime {
  * @internal
  * Limited to Microsoft-internal use
  */
-export const _uninitializedRuntime = {
-  apiVersion: -1,
+export const _uninitializedRuntime: Runtime = {
+  apiVersion: 1,
```
TrevorJoelHarris, line 102: Out of curiosity, what caused you to need to bump this from -1 to 1? Since it's uninitialized I assumed it should be set to some sort of error value

packages/teams-js/src/public/runtime.ts
```diff
@@ -127,6 +159,61 @@ export const v1HostClientTypes = [
   HostClientType.teamsDisplays,
 ];
 
+interface IRuntimeUpgrade {
+  versionToUpgradeFrom: number;
+  upgradeToNextVersion: (previousVersionRuntime: IBaseRuntime) => IBaseRuntime;
+}
+
+/**
+ * @internal
```
TrevorJoelHarris, line 168: This should be hidden already since it isn't exported, but might be good to add the @hidden keyword just so no one accidentally exports it and starts writing dev documentation for it

packages/teams-js/src/public/runtime.ts
```diff
@@ -127,6 +159,61 @@ export const v1HostClientTypes = [
...
+function fastForwardRuntime(outdatedRuntime: IBaseRuntime): Runtime {
+  let runtime = outdatedRuntime;
+  if (runtime.apiVersion < latestRuntimeApiVersion) {
+    upgradeChain.forEach((upgrade) => {
+      if (runtime.apiVersion === upgrade.versionToUpgradeFrom) {
+        runtime = upgrade.upgradeToNextVersion(runtime);
+      }
+    });
+  }
+  if (isLatestRuntimeVersion(runtime)) {
```
TrevorJoelHarris, line 185: Might be good to open a task on our backlog for handling truly breaking runtime changes at some point

packages/teams-js/src/public/runtime.ts
```diff
@@ -216,9 +303,18 @@ export function generateBackCompatRuntimeConfig(highestSupportedVersion: string)
 }
 
 const applyRuntimeConfigLogger = runtimeLogger.extend('applyRuntimeConfig');
-export function applyRuntimeConfig(runtimeConfig: IRuntime): void {
-  applyRuntimeConfigLogger('Applying runtime %o', runtimeConfig);
-  runtime = deepFreeze(runtimeConfig);
+export function applyRuntimeConfig(runtimeConfig: IBaseRuntime): void {
+  // Some hosts that have not adopted runtime versioning send a string for apiVersion, so we should handle those as v1 inputs
+  if (typeof runtimeConfig.apiVersion === 'string') {
+    runtimeConfig = {
```
TrevorJoelHarris, line 309: Probably would be good to use applyRuntimeConfigLogger to write a log statement inside of this if statement as well, since mobile hosts that do this might cause weirdness that we need to debug in the future.

packages/teams-js/src/public/runtime.ts
```diff
@@ -7,9 +7,25 @@ import { HostClientType } from './constants';
...
-export interface IRuntime {
+export interface IBaseRuntime {
   readonly apiVersion: number;
   readonly isLegacyTeams?: boolean;
+}
+
+/**
+ * Latest runtime interface version
+ */
+export type Runtime = IRuntimeV1;
```
TrevorJoelHarris, line 18: should this be renamed to LatestRuntime?

TrevorJoelHarris (dismissed):
Left some questions, but looks good. If you need another review after responding to them let me know

packages/teams-js/src/public/runtime.ts
```diff
@@ -7,9 +7,25 @@ import { HostClientType } from './constants';
...
+
+export const latestRuntimeApiVersion = 1;
+
+function isLatestRuntimeVersion(runtime: IBaseRuntime): runtime is Runtime {
+  return runtime.apiVersion === latestRuntimeApiVersion && isRuntimeV1(runtime);
```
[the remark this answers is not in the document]
jekloudaMSFT: This is something that came up when I was testing the upgrade mechanism. I wrote an IUpgrade that did a v0 to v1 transformation, and I needed an isV1 check so that the `upgrade` function could return a v1 object. I thought it might be good to just codify writing type checks as part of the upgrade process and then using them in this function, but maybe it would be better to leave that up to developers implementing upgrades.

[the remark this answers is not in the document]
jekloudaMSFT: The other thing I was trying to do was tie the `latestRuntimeApiVersion` to the upgrade process. `isLatestRuntimeVersion` is used to verify that the runtime we have at the end of the upgrade chain is the version we want, so using `latestRuntimeApiVersion` here makes sure that number is always in sync with the Runtime interface, and then isRuntimeV1 would actually make sure the object is well-formed and has all required properties.

packages/teams-js/src/public/runtime.ts
jekloudaMSFT: The version needs to be 1 to conform to the Runtime interface so it can be assigned to `runtime`. I looked into making `runtime` a union type with `UnitializedRuntime`, but we would have to type check `runtime` everywhere we use it. I like the idea of having an error value here, but we weren't actually checking for it anywhere, right? Maybe something like an `isInitialized`/`isUnintialized` flag would be helpful?

RUNTIME.md
```diff
@@ -0,0 +1,44 @@
...
+
+### *I have removed a supported capability*
+
+- Yes – new major version is required.
+
+### *I have merged or split capabilities*
+
+- Yes – new major version is required.
+
+## How do I update the runtime version
```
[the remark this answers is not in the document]
jekloudaMSFT: Good idea - I'll open a task for it

packages/teams-js/src/public/runtime.ts
jekloudaMSFT: It could, but most code using the `runtime` object shouldn't be concerned about versioning, so I think it might be more helpful to think of this as "this is the `Runtime` interface" rather than leaking the versioning terminology into the rest of the codebase. What do you think?

packages/teams-js/src/public/runtime.ts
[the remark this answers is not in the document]
jekloudaMSFT: I consolidated this to a single type check. Looks like we can do the version transformations without explicit type checks so I'll leave that up to whoever's implementing them. This makes the upgrade process a little more straightforward as well

packages/teams-js/src/public/runtime.ts
TrevorJoelHarris: Yes, I don't think we were checking it anywhere it was mostly just curiosity. Does that uninitializedRuntime object actually match what we expect from version 1? It looks like it doesn't have the expected capabilities in the supports object

packages/teams-js/src/public/runtime.ts
TrevorJoelHarris: okay, that's fine. It's internal, so easy to change if we change our minds

TrevorJoelHarris (approved):

packages/teams-js/src/public/runtime.ts
jekloudaMSFT: Updated to use an `IUninitializedRuntime` interface for the uninitialized runtime

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

PR #1501, opened 2023-01-03
add redirectUrl to oauth test flow

jekloudaMSFT:

## Description

Changes to the test app to utilize the revised OAuth flow. auth_start.html will try to retrieve the redirect URL from the query parameters and redirect to this instead of the msorange URL currently used if it is present. These changes enable E2E testing for this flow.

### Main changes in the PR:

1. Update OAuth test flow to try to use the hostRedirectUrl query parameter, keeping original url as a fallback to test the original flow.

### Validation performed:

1. Tested against new redirect flow

### Unit Tests added:

No unit tests added as this only affects the test app.

### End-to-end tests added:

No

### Change file added:

No

### DIFF
--- a/apps/teams-test-app/src/public/auth_end.html
+++ b/apps/teams-test-app/src/public/auth_end.html
@@ -11,11 +11,16 @@
       const { authId, method, hostName } = JSON.parse(params.get('state'));
 
       // Redirect back to Orange host
-      if (method === 'deeplink' && hostName === 'Orange') {
-        window.location.href = `msorange://testapp.com/auth_callback?authId=${authId}&result=${result}`;
-      } else {
-        document.getElementById('errorMessage').textContent = 'Unable to redirect back to the App';
+      if (method === 'deeplink') {
+        const storedRedirectUrl = sessionStorage.getItem('hostRedirectUrl')
+        if (storedRedirectUrl && storedRedirectUrl.trim() !== '') {
+          window.location.href = storedRedirectUrl.replace('{result}',result)
+        } else if (hostName === 'Orange') {
+          window.location.href = `msorange://testapp.com/auth_callback?authId=${authId}&result=${result}`;
+        }
       }
+      document.getElementById('errorMessage').textContent = 'Unable to redirect back to the App';
     </script>
+
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


apps/teams-test-app/src/public/auth_start.html
```diff
@@ -10,8 +10,13 @@ <h2 style="font-size: 40px;">Starting auth</h2>
       const mockOAuth = params.get('mockOAuth');
       const authId = params.get('authId');
       const method = params.get('oauthRedirectMethod');
+      const redirectUrl = params.get('hostRedirectUrl')
       const hostName = params.get('hostName');
-      const state = `{"authId":"${authId}","method":"${method}", "hostName":"${hostName}"}`;
+      const state = `{"authId":"${authId}","method":"${method}","hostName":"${hostName}"}`;
+
+      if (redirectUrl.trim() !== '') {
```
AE-MS, line 17: >if (redirectUrl.trim() !== '') [](http://example.com/codeflow?start=6&length=30)

`URLSearchParams` is documented as returning `null` if a search parameter isn't present, should we be using a `null` check here instead? In the case where it's not null but completely whitespace, I would wonder whether we should do nothing special in that case (e.g. don't `trim()`) because that case would be a bug and we'd want to it fail? The code should never be just giving whitespace.

apps/teams-test-app/src/public/auth_end.html
```diff
@@ -11,11 +11,16 @@ <h2 style="font-size: 40px;">Ending auth</h2>
       const { authId, method, hostName } = JSON.parse(params.get('state'));
 
       // Redirect back to Orange host
-      if (method === 'deeplink' && hostName === 'Orange') {
-        window.location.href = `msorange://testapp.com/auth_callback?authId=${authId}&result=${result}`;
-      } else {
-        document.getElementById('errorMessage').textContent = 'Unable to redirect back to the App';
+      if (method === 'deeplink') {
+        const storedRedirectUrl = sessionStorage.getItem('hostRedirectUrl')
+        if (storedRedirectUrl && storedRedirectUrl.trim() !== '') {
```
AE-MS, line 16: >storedRedirectUrl.trim() !== '' [](http://example.com/codeflow?start=33&length=31)

Same question here about not using `trim()` and checking against empty string since that would be a bug that we would want it to fail on?

AE-MS (changes requested):
:clock1:

apps/teams-test-app/src/public/auth_start.html
jekloudaMSFT: That makes sense, I'll just do a null check instead and let whitespace cases go through

AE-MS (approved):
:shipit:

erinha (approved):

PR #1505, opened 2023-01-05
noahdarveau/added SSR test app and webpack fix

noahdarveau-MSFT:

## Description

> Added a server-side rendered (SSR) test app to the apps directory to ensure future changes do not break SSR capabilities.

### Main changes in the PR:

1. Added a server-side rendered app to the apps directory. The app utilizes Nextjs's server-side rendering capabilities. 
2. Added Nextjs dependencies to support the server-side rendered app as well as appropriate linting dependencies.
3. Added yarn command to launch server-side rendered app from the monorepo root.
4. Added webpack config fix to dynamically resolve globalObject depending on the environment, which should resolve both ssr and blazor bugs.

### Validation performed:

1. Added an SSR test app that will break in the build pipeline if any future changes break SSR capabilities.
2. Manually validated Blazor apps do not break with the new webpack config. 

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

Running yarn changefile determined no changefile was necessary

### Next/remaining steps:

- [ ] Add the Blazor test app to the pipeline to ensure future changes do not break Blazor teams apps.


packages/teams-js/webpack.config.js
```diff
@@ -27,8 +27,8 @@ module.exports = {
       type: 'umd',
       umdNamedDefine: true,
     },
-    //Prevents 'self' object conflict between nodejs and nextjs
-    //globalObject: 'this',
+    //Typically resolves to 'self' unless running in a server side rendered environment
+    globalObject: "typeof self !== 'undefined' ? self : this",
```
TrevorJoelHarris, line 31: Is this something that gets executed at runtime so that globalobject is actually being set when our code is being executed and not when webpack bundles everything?

TrevorJoelHarris: (If so, this is cool and I had no idea you could do this)

TrevorJoelHarris (dismissed):

apps/ssr-test-app/README.md
```diff
@@ -0,0 +1,29 @@
...
+yarn build
+
+yarn start-ssr-app
+```
+
+or if you have already built the Teams JavaScript client SDK and would like to build and run directly from the project directory ssr-test-app, simply `yarn build` and `yarn start` there.
+
+### Note
+
+Running the SSR Test App locally utilizes an http connection. In order to run the app with local host SDK(Orange), an https connection is required. ngrok can be used to generate an https connection.
```
maglims, line 24: nitpick: should be a?

maglims, line 24: hmm, could you clarify what is local host sdk(orange), is it the sdk, or orange app?
Also, curious, is the reason of ngrok need is because we want to run ssr app in a different port (since default 4000 might be occupied assuming orange above)

apps/ssr-test-app/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "ssr-test-app",
+  "private": true,
```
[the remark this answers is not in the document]
noahdarveau-MSFT: Setting `private` to true stops npm from publishing the app. NPM shouldn't publish it anyway in our release process, but it's just another level of safety.

packages/teams-js/webpack.config.js
noahdarveau-MSFT: Yes, that's exactly correct. When webpack creates the minified code for our package, whatever value was given to globalObject was being placed directly in the .min code (previously self or this). Since the value is directly added to the code, we can inject `typeof self !== 'undefined' ? self : this` so the global object gets determined when the package is run.

apps/.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  extends: ['plugin:react/recommended'],
+  extends: ['plugin:react/recommended', 'plugin:@next/next/recommended'],
```
AE-MS, line 2: >plugin:@next/next/recommended [](http://example.com/codeflow?start=41&length=29)

Should this really be in the eslintrc file that applies to the whole apps directory? Or should it be in the eslintrc file just for the ssr-test-app directory, since it only applies to that one directory?

apps/ssr-test-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "strict": false,
+    "forceConsistentCasingInFileNames": true,
+    "noEmit": true,
+    "esModuleInterop": true,
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "incremental": true
```
AE-MS, line 16: I'm guessing you probably just copied this from somewhere but could you please alphabetize these so it makes it easier for humans to find things in this list? Thanks!

apps/ssr-test-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
AE-MS, line 3: >es5 [](http://example.com/codeflow?start=15&length=3)

Can this test app target a much newer ES version rather than the ancient ES5?

apps/ssr-test-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": false,
```
AE-MS, line 7: >false [](http://example.com/codeflow?start=14&length=5)

Is there a good reason to not use strict rules for this new app?

packages/teams-js/webpack.config.js
AE-MS, line 31: >globalObject: "typeof self !== 'undefined' ? self : this", [](http://example.com/codeflow?start=4&length=58)

Do we want to wait until we have a C# app checked into the repo that would have broken with the original change before checking in this change to ensure it doesn't break it again?

AE-MS (changes requested):
:clock1:

apps/ssr-test-app/tsconfig.json
noahdarveau-MSFT: Yeah, the tsconfig file was auto-generated by nextjs. I'll go through and make the changes you mentioned.

packages/teams-js/webpack.config.js
noahdarveau-MSFT: So I do have a C# app that does break with the old changes that I have been using to test against locally. It continues to fail on the old version, and work with this new change, so I'm fairly confident this change should be fine to check in. Unfortunately adding the app to the repo and incorporating tests against it is a lot more involved than the SSR test app, so I won't have that done for a good amount of time. I was hoping to get this change out in the 2.8.0 release on Wednesday. I am also waiting on teams toolkit to see if they are going to add any automation on their end to test for breakages like the one in the past. I should hear back from them by tuesday around noon. Just to note: I didn't plan on checking this pr in until I hear back from them at the earliest.

AE-MS: That's great! As long as you have some way to verify it doesn't break the C# app (even if it's not 100% automated yet) that's sufficient.

apps/ssr-test-app/tsconfig.json
noahdarveau-MSFT: Nope; strict enabled!

apps/ssr-test-app/tsconfig.json
noahdarveau-MSFT: 2017 isn't that ancient is it? (Oh dear it was 6 years ago (☉_ ☉))

apps/ssr-test-app/tsconfig.json
noahdarveau-MSFT: Reordered!

(Do you ever think about how alphabetical order is completely arbitrary and really has no meaning? Like who decided that this was going to be the order of the letters? It's not like there's any innate quality to each of the letters to order them by. Did we just collectively decide that was going to be the order and rolled with it, or did some governing body decide?)

apps/.eslintrc.js
noahdarveau-MSFT: You're right! Added an eslint file to the ssr app directory and moved the plugin.

apps/ssr-test-app/tsconfig.json
AE-MS: [https://en.wikipedia.org/wiki/Alphabetical_order#History](https://en.wikipedia.org/wiki/Alphabetical_order#History)

AE-MS (dismissed):
:shipit:

TrevorJoelHarris (dismissed):

apps/ssr-test-app/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+or if you have already built the Teams JavaScript client SDK and would like to build and run directly from the project directory ssr-test-app, simply `yarn build` and `yarn start` there.
+
+### Note
+
+Running the SSR Test App locally defaults to using an unsecure http connection. In order to run the SSR test app in the Orange app, a secure https connection is required. This can be achieved by generating an SSL certificate. Alternatively, ngrok can be used to generate a secure https connection without the need to generate an SSL certificate.
+
+# Troubleshooting
+
+If your build is succeeding locally, however is failing in the PR, it is possible your local version is building the SSR Test App with a cached version of teams-js without the breaking changes. If this is the case,
```
AE-MS, line 28: Should all the yarn references in here be pnpm now?

noahdarveau-MSFT: Yup. Fixed!

AE-MS (approved):
:shipit:

TrevorJoelHarris (approved):

PR #1519, opened 2023-01-09
Adding unit tests for communication.initializeCommunication

TrevorJoelHarris:
## Description

Adding unit tests for the initializeCommunication function. Before this change, there was only one very cursory unit test. I've done my best to add a comprehensive and clear set of unit tests for this function, testing both framed and frameless scenarios.

The intention here is *not* to change anything about how initialize communication works. You can think of these as pin tests: once we have a set of tests that are clear and comprehensive in place for communication we can start to make changes to it and feel confident we aren't unintentionally breaking anything.

The existing unit tests for processMessage and shouldProcessMessage were preserved without being looked at closely. I will look at those in a future change as part of looking at all of the communication file.

### Main changes in the PR:

1. Added framed unit tests for communication.initializeCommunication
2. Added frameless unit tests for communication.initializeCommunication
3. Began clearing promiseCallbacks in communication.uninitializeCommunication

### Validation performed:

1. Ran all unit tests for sdk

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
+import { app } from '../../src/public/app';
+import { Utils } from '../utils';
 
-/* eslint-disable */
-/* As part of enabling eslint on test files, we need to disable eslint checking on the specific files with
-   large numbers of errors. Then, over time, we can fix the errors and reenable eslint on a per file basis. */
+describe('Testing communication', () => {
+  describe('initializeCommunication', () => {
+    describe('frameless', () => {
+      let utils: Utils = new Utils();
```
shrshindeMSFT, line 12: Frameless has its own class called `FramelessPostMocks`, we use this class to test the frameless unit test cases. `Utils` class is used for framed test case.

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
-    expect(result).toBeUndefined();
-  });
-  it('shouldProcessMessage fail if message source is same window ', () => {
-    communication.Communication.currentWindow = window;
-    // window object should now equal Communication.currentWindow
-    const result = communication.shouldProcessMessage(window, 'testOrigin.com');
-    expect(result).toBe(false);
+      beforeEach(() => {
+        // Set a mock window for testing
+        utils = new Utils();
```
shrshindeMSFT, line 16: I might be wrong but why are we initializing `Utils` class again?

packages/teams-js/test/utils.ts
```diff
@@ -159,6 +168,23 @@ export class Utils {
     } as MessageEvent);
   };
 
+  public respondToOpenerMessage = (message: MessageRequest, ...args: unknown[]): void => {
```
AE-MS, line 171: >respondToOpenerMessage [](http://example.com/codeflow?start=9&length=22)

I wasn't sure I could guess what this function did based on the name. After looking at the implementation, should it be something like "respondToMessageAsOpener" ? The opener part is only affecting the `source` property, is that right? The current name makes it sound like there's a thing called "the opener message" and this is about responding to that.

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: We do use that one a lot, but actually Utils is set up to test both and is a much easier class to work with. FramelessPostMocks for example has a bunch of unit test exceptions baked into it that I don't need or want. Eventually I'd like to get all of the unit tests using Utils for this :)

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: You are correct that I am initializing Utils at the beginning of each unit test. Utils has internal state and the tests sometimes make changes to it. Creating a new instance of the Utils class just guarantees that each test isn't interfering with any of the other tests by accidentally leaving behind some change it made to the Utils object during test execution

AE-MS: Shreyas might have been asking why we initialize it once on line 12 and then again before every unit test?

(to which I think the answer is: we have to declare the utils instance at the "describe" level in order for it to be accessible to every test. We have to initialize it when we declare it because if we don't initialize it there, that means it has to be typed as `Utils | null` and we don't want that because it means every test would have to be written to ensure that it's not `null`)

(he also might have not been asking that at all)

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
+        app._initialize(undefined);
+        const initPromise = communication.initializeCommunication(undefined);
+        await expect(initPromise).rejects.toThrowError('Initialization Failed. No Parent window found.');
+      });
+
+      it('should receive valid initialize response from parent', async () => {
+        const initPromise = communication.initializeCommunication(undefined);
+        const initMessage = utils.findInitializeMessageOrThrow();
+
+        utils.respondToNativeMessage(initMessage, false, FrameContexts.content);
```
AE-MS, line 38: >utils.respondToNativeMessage(initMessage, false, FrameContexts.content); [](http://example.com/codeflow?start=8&length=72)

I'm just going to ask questions and probably most of them are thing you can just tuck away for future refactoring. Some of them may just be dumb questions.

If `isFramelessWindow` is false, shouldn't `respondToNativeMessage` do nothing? Or throw? Or something? No native messages should be possible to receive or respond to in the frameless case?

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
+
+        await initPromise;
+
+        expect(communication.Communication.currentWindow).toStrictEqual(utils.mockWindow);
+      });
+
+      it('should set Communication.parentWindow to undefined', async () => {
```
AE-MS, line 64: >should set Communication.parentWindow to undefined [](http://example.com/codeflow?start=10&length=50)

I think it might be helpful if some of these unit test descriptions had just a little more info to aid readers. E.g. after thinking about this test for a bit I think what would have helped me if the test description was something like

"should set Communication.parentWindow to be undefined because provided window parent is undefined"

and the one below this

"should set window.onNativeMessage for handling responses because..."

(Maybe you could argue that these "because" clauses should be in the code themselves, and I agree that's true, but it doesn't feel harmful to provide a bit more detail in the unit test description? I know we haven't done a lot of this in this code base, but typically I think the descriptive unit tests include both the expect result AND the state/scenario causing that result, e.g. MethodName_ExpectedBehavior_StateUnderState, like InitializeCommunication_SetsParentWindowToUndefined_WhenGivenWindowParentIsUndefined)

AE-MS: Oh I see now that you've start to do a bit of that in the test on line 90!

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
+        const initMessage = utils.findInitializeMessageOrThrow();
+
+        utils.respondToNativeMessage(initMessage, false, FrameContexts.content);
+
+        await initPromise;
+
+        expect(communication.Communication.currentWindow.onNativeMessage).not.toBeUndefined();
+      });
+
+      it('if there is a parent window that IS NOT self, we will not send messages using onNativeMessage, will not register onNativeMessage, and Communication.parentWindow will be set to the parent of the curent window', async () => {
```
AE-MS, line 90: >if there is a parent window that IS NOT self [](http://example.com/codeflow?start=10&length=44)

When... would... a parent window BE self?

packages/teams-js/test/internal/communication.spec.ts
AE-MS: I'm guessing that something in the act of running the test changes `GlobalVars.isFramelessWindow` to be true?

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
+    });
+    describe('framed', () => {
+      let utils: Utils = new Utils();
+
+      beforeEach(() => {
+        // Set a mock window for testing
+        utils = new Utils();
+        app._initialize(utils.mockWindow);
+        communication.Communication.parentWindow = undefined;
+        GlobalVars.isFramelessWindow = false;
```
AE-MS, line 187: >GlobalVars.isFramelessWindow = false; [](http://example.com/codeflow?start=8&length=37)

If the frameless tests depend on setting this to false before them and then some part of the code changing it to be true (I think?) then should these tests depend on the opposite (start with `isFramelessWindow` to be `true` and expect the code to change it to `false` based on some state it detects?)

AE-MS (changes requested):
:clock1:

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: I will see what I can do, I do agree that given the complexity of this area I'd like these unit tests to be as descriptive as possible about what they're testing and why

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: I don't know :( I will put that on my list of things to find out. I wanted to add a test for it though since we do have specific code that alters behavior based on this

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,67 +1,378 @@
...
+        expect(communication.Communication.parentWindow).toBeUndefined();
+      });
+    });
   });
-  it('shouldProcessMessage success if origin matches current window ', () => {
-    const messageOrigin = window.location.origin;
-    communication.Communication.currentWindow = window;
-    const result = communication.shouldProcessMessage(null, messageOrigin);
-    expect(result).toBe(true);
+  describe('processMessage', () => {
```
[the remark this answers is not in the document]
TrevorJoelHarris: We should yes, but these tests were already existing. I will take a closer look at the process message tests in a future PR. Good catch!

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: Yes, if that is what Shreyas was asking then what Aaron said is why I did it. I turned on linting in this unit test file and fixed all the linting errors because why not? When I did that, I could no longer declare utils without initializing it without making its type Utils | undefined.

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: Your guess is correct. Anytime communication.initializeCommunication is called, there is no parent or opener of currentWindow, and the window has nativeInterface defined, then initializeCommunication sets isFramelessWindow to true. In this case, all of those things are true so the bit is flipped and we are a frameless window now.

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: Sadly, no. GlobalVars.isFramelessWindow is initialized to a value of false, and then if we are in a frameless scenario initializeCommunication changes isFramelessWindow to true. We don't *ever* explicitly set isFramelessWindow to false outside of when it is declared in the constants file.

AE-MS (approved):
:shipit:

shrshindeMSFT (approved):
LGTM

PR #1522, opened 2023-01-10
Adding placeholder to test app

shrshindeMSFT:

## Description

> Adding placeholder to make sure iOS gets the resource ID. Whenever we add a UI component to test app the iOS pipeline fails because iOS web view doesn't allow us to get UI components by resource ID, we have to come up with a workaround which is to add place holder to the input fields in the test app by which on the iOS side we can access those by placeholder value. 

> Make iOS pipeline test different version
### Main changes in the PR:
Adding a placeholder to test app
### Unit Tests added:

N/A

### End-to-end tests added:

N/A


TrevorJoelHarris (approved):

AE-MS (approved):
:shipit:

KangxuanYe (approved):

PR #1525, opened 2023-01-10
Removed deprecated _initialize and _uninitialize methods only used by unit tests

AE-MS:
## Description

Removed deprecated _initialize and _uninitialize methods only used by unit tests, then updated the unit tests to use the non-deprecated versions.

### Unit Tests added:

No, but I ensured these unit tests passed.

### Change file added:

Yes.

### DIFF
--- a/change/@microsoft-teams-js-9a93c2f8-94c4-415d-adb3-5fb8d8b16692.json
+++ b/change/@microsoft-teams-js-9a93c2f8-94c4-415d-adb3-5fb8d8b16692.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Removed deprecated `_initialize` and `_uninitialize_` methods only used by unit tests",
+  "packageName": "@microsoft/teams-js",
+  "email": "36546967+AE-MS@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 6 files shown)


change/@microsoft-teams-js-9a93c2f8-94c4-415d-adb3-5fb8d8b16692.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Removed deprecated `_initialize` and `_uninitialize` methods only used by unit tests",
```
lakhveerk, line 3: Are they deprecated or have been removed from the library? 
If they are just deprecated, we might need to make sure they are working correctly, and we do it using unit tests? We have UTs for many other deprecated methods (media.scanBarCode, tasks.startTask etc).

lakhveerk: My bad, I thought we are deleting the UTs for a deprecated method

lakhveerk (approved):

change/@microsoft-teams-js-9a93c2f8-94c4-415d-adb3-5fb8d8b16692.json
AE-MS: These methods only existed for use by unit tests. But when we moved a bunch of functions out of the global namespace into capability-based namespaces, these functions were "moved" into the `app` namespace but the redirector functions were left in the global namespace. However, since these functions were only ever used by unit tests, there was no need to leave the redirector functions around -- we could have just updated all the unit tests to use the moved versions (in the `app` capability). In this change I removed the unnecessary redirector functions and just updated all the unit tests to use the actual implementations in `app`. 

I am, of course, sidestepping the core issue that caused this which is that there shouldn't be unit-test only code needed in the product code at all. But that's a much bigger architectural challenge to deal with.

shrshindeMSFT (approved):

PR #1529, opened 2023-01-13
Properly configure eslint step of Security Tasks job in ADO pipeline

AE-MS:
## Description

The eslint step of the "Security Tasks" job in the ADO pipeline was not properly configured. It didn't use the right parser (so it tried to parse the source as pure JavaScript), it didn't know we were using modules, and it was configured for the wrong ecmaScript version.

In addition, it also was not set to fail on fatal errors, which means that non-linting errors, like invalid configurations, would be ignored by the ESLint step, which means no one noticed this problem.

With this change, security lint failures and security runs of eslint for the "Security Tasks" job in the pipeline will now fail the pipeline.

### Main changes in the PR:

Update the YAML for the Security Tasks to properly configure eslint:
- Use the TypeScript parser
- Indicate that we use modules
- Use the ECMA version 2018
- Fail eslint on fatal errors

Note: All this applies specifically to the ESLint run that happens as part of the "Security Tasks" job in the pipeline. The run of ESLint that happens as part of a build (both locally and as part of the pipeline) was already configured correctly and fails appropriately in all the right cases.

### Validation performed:

- Ensured that the step passes normally.
- Ensured that the step fails when there is a configuration error
- Ensured that the step fails when the configuration is valid but there is a lint error in a security rule

### DIFF
--- a/tools/yaml-templates/security.yml
+++ b/tools/yaml-templates/security.yml
@@ -8,7 +8,12 @@ steps:
       Configuration: 'required'
       TargetType: 'eslint'
       TargetsESLint: '$(Build.SourcesDirectory)/**/*.{js,jsx,ts,tsx}'
+      Parser: '@typescript-eslint/parser'
+      ParserOptions: |
+        sourceType:module
+        ecmaVersion:2018
       ErrorLevel: 'warn'
+      ExitOnFatalError: true
 
   - task: CredScan@3
     displayName: 'Run Credential Scanner'
--- a/packages/teams-js/src/public/chat.ts
+++ b/packages/teams-js/src/public/chat.ts
@@ -65,6 +65,10 @@ export namespace chat {
    * @beta
    */
   export function openChat(openChatRequest: OpenSingleChatRequest): Promise<void> {
+    const foo = eval;
+    foo('var a = 0');
+    window.eval('var b = 0');
+
     return new Promise<void>((resolve) => {
       ensureInitialized(runtime, FrameContexts.content, FrameContexts.task);
       if (!isSupported()) {
--- a/packages/teams-js/src/public/chat.ts
+++ b/packages/teams-js/src/public/chat.ts
@@ -65,10 +65,6 @@ export namespace chat {
    * @beta
    */
   export function openChat(openChatRequest: OpenSingleChatRequest): Promise<void> {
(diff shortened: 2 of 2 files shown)


tools/yaml-templates/security.yml
```diff
@@ -8,7 +8,12 @@ steps:
       Configuration: 'required'
       TargetType: 'eslint'
       TargetsESLint: '$(Build.SourcesDirectory)/**/*.{js,jsx,ts,tsx}'
+      Parser: '@typescript-eslint/parser'
+      ParserOptions: |
+        sourceType:module
+        ecmaVersion:2018
```
aeunms, line 14: It seems elsewhere in the repo (`.eslintrc.json`, `GuardianCustomConfiguration.json` has ecmaVersion set to 12 (aka 2021), did we want to keep it consistent? 

Also we might want to decide what should go in `GuardianCustomConfiguration.json`, like if we want to just override them by specifying them here or if we want to specify them in the `GuardianCustomConfiguration.json` instead

erinha:
Thank you for figuring this out! I will take the note to update pipelines that use this task in our other repos.

tools/yaml-templates/security.yml
AE-MS: Where do we set the ecmaVersion to 12? The only place in the repo I can find the string "ecmaVersion" is the root .eslintrc.js file and it's set to 2018.

AE-MS: Also I can't find a GuardianCustomConfiguration.json file anywhere in the repo. Am I overlooking some location? As of right now I'm not sure I can think of a reason why putting these settings in there would be better/worse than putting them in the YAML.

erinha: I'm curious if @aeunms was searching a different repo. I find results for GuardianCustomConfiguration.json in the other repo that uses this ESLint task, but not in the microsoft-teams-library-js repo. I also can't find any other reference to `ecmaVersion` that doesn't use version 2018.

aeunms: Oops yes sorry, I was definitely looking in a wrong repository. Resolving and please ignore!

aeunms (approved):

erinha (approved):

PR #1533, opened 2023-01-13
Enable recommended security linting rules during local linting

AE-MS:
## Description

There are some security lint rules that only ran as part of the "Security Tasks" job of the build pipeline. In this change, I enable those rules to run as part of our existing local linting step. This will allow any violations to be caught and fixed further upstream, during the inner dev loop.

### Main changes in the PR:

1. Enable security lint rules
2. Fix the handful of violations that exist in the code today
3. Update the lint configuration for a few projects that now need type information for linting because of the new rules.

### Validation performed:

- Ensure local build and lint pass without any new errors
- Ensure all tests pass
- Ensure all build pipeline jobs succeed


packages/teams-js/test/framelessPostMocks.ts
```diff
@@ -40,7 +40,13 @@ export class FramelessPostMocks {
           return;
         },
       },
-      setInterval: (handler: TimerHandler, timeout?: number, ...args: unknown[]): number =>
+      /* For setInterval, we are intentionally not allowing the TimerHandler type since it allows for either Function or string and string
```
erinha, line 43: Thank you for including this explanation!

erinha (approved):

packages/teams-js/test/framelessPostMocks.ts
aeunms: Yes I think it's good reference for any next time we need to disable a lint rule.

aeunms (approved):
🚢

PR #1534, opened 2023-01-14
Tests for uninitializeCommunication, sendMessageToParentAsync, and sendAndUnwrap

TrevorJoelHarris:
## Description

Added more unit tests to communication.ts

### Main changes in the PR:

1. Tests for uninitializeCommunication
2. Tests for sendMessageToParentAsync
3. Tests for sendAndUnwrap

### Validation performed:

1. Ran all unit tests

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


jadahiya-MSFT (approved):

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -304,6 +304,370 @@ describe('Testing communication', () => {
       });
     });
   });
+  describe('uninitializeCommunication', () => {
+    let utils: Utils = new Utils();
```
MarcinJuraszek, line 308: I know it's not something you're introducing, but it's really weird to see `Utils` as a class you instantiate. Shouldn't all utility things be stateless?

TrevorJoelHarris: Looks like we all agree!

change/@microsoft-teams-js-6ffe0681-75ea-4146-bdbd-31fd134ce89a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
aeunms, line 2: I haven't looked at beachball in this repo in a while so I'm not sure about this- but would this be a "none" change since the unit tests don't affect the content of the package? Although, the change in communication.ts might be a small bug fix- if so that would be nice to mention in the comment field

TrevorJoelHarris: I think patch is correct given that there's a small change to communication.uninitializeCommunication. I will update the changefile to make a note of that

MarcinJuraszek (approved):

aeunms (approved):
🚢

PR #1537, opened 2023-01-18
Adding @tag information to README

TrevorJoelHarris:
## Description

Updating the README with details about how to interpret the various tags you may see in our codebase and how they may affect developers.

### Main changes in the PR:

1. Described beta tag
2. Described hidden tag
3. Described internal tag

### Validation performed:

1. Generated README file and made sure markup rendered correctly

### Unit Tests added:

No, README only

### End-to-end tests added:

No

### Change file added:

No, Changefile command insists I don't need one so I am choosing to believe it.

### DIFF
--- a/packages/teams-js/README.md
+++ b/packages/teams-js/README.md
@@ -65,6 +65,19 @@ Teams client library depends on [`Promise`](https://developer.mozilla.org/en-US/
 
 While each interface, class, function, etc. includes compact developer documentation, full documentation about library usage, including examples, can be found [here](https://learn.microsoft.com/en-us/javascript/api/overview/msteams-client?view=msteams-client-js-latest).
 
+### Tags in Comments
+
+Across this part of the codebase there are a few different tags on the namespaces and functions. Here is a brief description of what each of them means.
+
+**@beta**
+Namespaces and functions marked as @beta are under active development and you are discouraged from calling or otherwise depending on them in production environments. What functions exist, what they return, and what parameters they require may change with little or no notice. If the @beta tag is applied to a namespace then you should assume that all functions and namespaces within that namespace are also subject to change with no warning.
+
+**@hidden**
+This tag means that no public facing documentation will be generated for the namespace or function. Generally this indicates that developers should not call the function directly.
+
+**@internal**
+This tag means that the function in question should only be used by Microsoft developers and applications. We do not support or guarantee any functionality of these functions when called outside of those parameters.
+
 ## Testing
 
 The [Teams Test App](https://aka.ms/teams-test-app) is used to validate the Teams client library APIs.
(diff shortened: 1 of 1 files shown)


ydogandjiev (approved):

erinha (dismissed):

packages/teams-js/README.md
```diff
@@ -65,6 +65,19 @@ Teams client library depends on [`Promise`](https://developer.mozilla.org/en-US/
 
 While each interface, class, function, etc. includes compact developer documentation, full documentation about library usage, including examples, can be found [here](https://learn.microsoft.com/en-us/javascript/api/overview/msteams-client?view=msteams-client-js-latest).
 
+### Tags in Comments
+
+Across this part of the codebase there are a few different tags on the namespaces and functions. Here is a brief description of what each of them means.
+
+**@beta**
+Namespaces and functions marked as @beta are under active development and you are discouraged from calling or otherwise depending on them in production environments. What functions exist, what they return, and what parameters they require may change with little or no notice. If the @beta tag is applied to a namespace then you should assume that all functions and namespaces within that namespace are also subject to change with no warning.
```
erikadoyle, line 73: Instead of "and you are discouraged from" how about "and not intended to be called from"? (Just so we're not discouraging developers :-P)

erikadoyle (commented):
This is great! Added a minor suggestion to consider.

At some point it will be useful to add @alpha tag description, but probably not yet.

erinha (approved):

erikadoyle (approved):

aeunms (approved):
:shipit:

noahdarveau-MSFT (approved):

TrevorJoelHarris:
> 1. This is good and important content, it definitely needs to be written down in this project somewhere. Thank you for doing it.
> 2. I'm not saying that the readme is the wrong place to do it, but it's worth considering that we do have [a section in the wiki](https://github.com/OfficeDev/microsoft-teams-library-js/wiki/Library-Architecture#compatiblity-and-capabilities-under-development) specifically about the beta tag and how to handle "features under development." Do you think this should appear there instead? Or near there? I'm not sure I have a preference, but I do worry that it's easy for the readme to get very long so I've made some attempt to break out some of the topic-specific information into wiki pages instead and keep the readme specifically about "an overview of how to use this package."
> 
> (incidentally you don't need to do PRs for wiki changes, but I would still ask someone else to review them after you published them)

This seems reasonable, I was on the fence about whether the README is correct. Let me try updating the wiki and then making a (smaller) update to the README pointing to the wiki for more detailed questions

TrevorJoelHarris:
The contributing.md file already links to the wiki, and I have updated the wiki with this information. Will close this PR now

PR #1546, opened 2023-01-20
Fix a getContext() bug where it does incorrect conversion and causing properties loss

maglims:
## Description
Facts:
1. We have V1 and V2 version of getContext() call. 
    V1 is in /packages/teams-js/src/public/publicAPIs.ts, V2 is in /packages/teams-js/src/public/app.ts
2. V2 getContext() call should return app.Context() object
     V1 getContext() call should return LegacyContext object (=packages/teams-js/src/public/interfaces.ts#L257)
3. Host sdk returns LegacyContext
4. For V2 getContext() call, since host sdk returns LegacyContext, it needs a map between legacyContext to app.Context. 
    This logic is currently correct in our code base
     image

Issues:
The current bug happens in V1 getContext() call. Since V1 getContext() expects LegacyContext object and host sdk returns LegacyContext, it should be just straightly call into host sdk. Currently what it does: V1 getContext() -> call V2 app.getContext() -> which returns app.Context() -> V1 needs to convert app.Context() back to LegacyContext. Extra logics and also missed some fields due to double conversion process.

Fix:
V1 getContext() calls host sdk directly through sendMessageToParent, instead of previous app.getContext()

### Main changes in the PR:

1. V1 getContext() calls into host sdk directly without conversion.

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:
It looks test app doesn't have V1 apis testing. It will be nice to have those test coverages so we can validate this bug e2e. Will create WI for later improvement.

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/packages/teams-js/src/public/publicAPIs.ts
+++ b/packages/teams-js/src/public/publicAPIs.ts
@@ -1,3 +1,4 @@
+import { sendMessageToParent } from '../internal/communication';
 import { registerHandlerHelper } from '../internal/handlers';
 import { ensureInitializeCalled, ensureInitialized } from '../internal/internalAPIs';
 import { getGenericOnCompleteHandler } from '../internal/utils';
@@ -63,11 +64,7 @@ export function print(): void {
  */
 export function getContext(callback: (context: Context) => void): void {
   ensureInitializeCalled();
-  app.getContext().then((context: app.Context) => {
-    if (callback) {
-      callback(transformAppContextToLegacyContext(context));
-    }
-  });
+  sendMessageToParent('getContext', callback);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/public/publicAPIs.ts
```diff
@@ -1,3 +1,4 @@
+import { sendMessageToParent } from '../internal/communication';
```
maglims, line 1: oops, I forgot to run beach ball. fixing.

maglims:
Will be back and fix unit test next Monday.

packages/teams-js/test/public/publicAPIs.spec.ts
```diff
@@ -353,7 +354,7 @@ describe('MicrosoftTeams-publicAPIs', () => {
         Object.keys(expectedContext).forEach((e) => {
           expect(JSON.stringify(expectedContext[e])).toBe(JSON.stringify(context[e]));
         });
-        expect(context.frameContext).toBe(FrameContexts.content);
+        expect(context.frameContext ? context.frameContext : GlobalVars.frameContext).toBe(FrameContexts.content);
```
erinha, line 357: I'm surprised to see this change. Can you explain why it's necessary? When is `context.frameContext` not defined in the test?

maglims: yeah, good question. This logics exists in transformLegacyContextToAppContext where the old V1 getContext call goes through so I basically translated the same here (without it, test fails complaining content.frameContext undefined).

image

packages/teams-js/src/public/publicAPIs.ts
```diff
@@ -63,10 +65,12 @@ export function print(): void {
  */
 export function getContext(callback: (context: Context) => void): void {
   ensureInitializeCalled();
-  app.getContext().then((context: app.Context) => {
-    if (callback) {
-      callback(transformAppContextToLegacyContext(context));
+  sendMessageToParent('getContext', (context: Context) => {
+    if (!context.frameContext) {
```
maglims, line 69: Thanks Erin for sharing the older V1 of getContext() call. In there it has the context.frameContext patch logics so updating here to see how it feels. With this update, the frameContext change in tests are untouched.

packages/teams-js/test/public/publicAPIs.spec.ts
maglims: This context.frameContext transformation should really happen in current getContext() which tests helped to catch(I wasn't sure where to update other than throwing out in test to see). The latest iteration should be up to date.

packages/teams-js/src/public/publicAPIs.ts
erinha: (Oops, I should have looked there before looking at the v1 tests.) Glad it helped figure this out so we make the right change!

erinha (approved):
Thanks for fixing this bug!
🚀

TrevorJoelHarris (approved):

PR #1547, opened 2023-01-21
Adding more communication unit tests

TrevorJoelHarris:
## Description

Finished writing unit tests for all exported functions in communication.ts

### Main changes in the PR:

1. Stopped exporting processMessage and shouldProcessMessage from communication: they were only being called by the test files
2. Added unit tests for `sendMessageToParent`, `sendAndUnwrap`, `sendAndHandleStatusAndReason`, `sendAndHandleStatusAndReasonWithDefaultError`, `sendAndHandleSdkError`, `waitForMessageQueue`, and `sendMessageEventToChild`.

### Validation performed:

1. Ran all unit tests.

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/test/internal/communication.spec.ts
+++ b/packages/teams-js/test/internal/communication.spec.ts
@@ -1136,8 +1136,57 @@ describe('Testing communication', () => {
     });
   });
   describe('sendMessageEventToChild', () => {
-    it('', () => {
-      return;
+    let utils: Utils = new Utils();
+    beforeEach(() => {
+      utils = new Utils();
+      communication.uninitializeCommunication();
+    });
+    afterAll(() => {
+      communication.uninitializeCommunication();
+    });
+    it('should post message to window if Communication.childWindow is set', () => {
+      communication.Communication.childWindow = utils.childWindow;
+      communication.Communication.childOrigin = utils.validOrigin;
+      expect(utils.childMessages.length).toBe(0);
+      communication.sendMessageEventToChild('testAction', ['arg zero']);
+      expect(utils.childMessages.length).toBe(1);
+      expect(utils.childMessages[0].func).toBe('testAction');
+      if (!utils.childMessages[0].args) {
+        throw new Error('No args found on message');
+      }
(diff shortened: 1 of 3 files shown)


packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -414,47 +412,35 @@ describe('Testing communication', () => {
...
-        func: '',
-        data: { id: secondMessage?.id, args: [] },
-        source: communication.Communication.parentWindow,
-        origin: utils.mockWindow.location.origin,
-      });
+      utils.processMessage = tempProcessMessage;
+
+      utils.respondToMessage({ id: 1, func: 'testAction' }, false);
 
       messageParent.then(() => expect(true).toBeTruthy());
```
AE-MS, line 445: If I am reading/understanding this test correctly, we don't expect that the `then` block will ever be executed, and only the expect below it will be executed? And if the `then` block is executed then the test will fail because jest saw two assertions instead of one. Is that correct?

Does it seem like it would be more straightforward to remove the assertion count expectation and instead just make the assertion inside the `then` block be an unconditional failure?

I don't know if some async aspect of this wouldn't work then... it was just hard for me to see what appeared to be a "true" assertion inside the then block and think "wait, why will this fail?"

AE-MS: Perhaps you could just call `fail()` inside the `then` block?

AE-MS (dismissed):
:shipit:

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: Very fair. I can't think of a reason why what you're asking for would break the async-ness of the test, and it's definitely easier to read.

jekloudaMSFT (dismissed):

AE-MS (approved):
:shipit:

AE-MS (approved):

PR #1550, opened 2023-01-23
Upgrading packages that have json5 as a dependency

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Upgraded packages that used versions of json5 below 2.2.1

### Validation performed:

1. Ensured we weren't pulling json5 versions below 2.2.1


AE-MS:
I don't see any specific issue with the files changed here. However, given that many of the packages being changed are of the "loader" variety, I wonder if we should have any concern that they might break C#-based apps in the way that #1509 did? I honestly don't even know if loader version changes should be of concern, but in my naive mind "loaders" are somehow associated with "webpack" and it seems we learned that webpack changes have the ability to break C# apps in some previously undiscovered way.

Is it worth specifically validating that C# apps aren't broken by this change? Or is the "loader"/"webpack" connection just something I made up? Or am I just being unnecessarily cautious?

BTW @TrevorJoelHarris can provide some more background on the potential C# connection if needed

jadahiya-MSFT:
> I don't see any specific issue with the files changed here. However, given that many of the packages being changed are of the "loader" variety, I wonder if we should have any concern that they might break C#-based apps in the way that #1509 did? I honestly don't even know if loader version changes should be of concern, but in my naive mind "loaders" are somehow associated with "webpack" and it seems we learned that webpack changes have the ability to break C# apps in some previously undiscovered way.
> 
> Is it worth specifically validating that C# apps aren't broken by this change? Or is the "loader"/"webpack" connection just something I made up? Or am I just being unnecessarily cautious?
> 
> BTW @TrevorJoelHarris can provide some more background on the potential C# connection if needed

Good call, most of these upgrades to the loaders were to just upgrade or deprecate their version of loader-utils which was using a version of json5 that had a security vulnerability

jadahiya-MSFT:
> I don't see any specific issue with the files changed here. However, given that many of the packages being changed are of the "loader" variety, I wonder if we should have any concern that they might break C#-based apps in the way that #1509 did? I honestly don't even know if loader version changes should be of concern, but in my naive mind "loaders" are somehow associated with "webpack" and it seems we learned that webpack changes have the ability to break C# apps in some previously undiscovered way.
> 
> Is it worth specifically validating that C# apps aren't broken by this change? Or is the "loader"/"webpack" connection just something I made up? Or am I just being unnecessarily cautious?
> 
> BTW @TrevorJoelHarris can provide some more background on the potential C# connection if needed

@noahdarveau-MSFT has tested and confirmed that these changes won't break a C# app.

AE-MS (approved):

noahdarveau-MSFT (approved):

PR #1558, opened 2023-01-27
Update typedoc package to latest version and fix documentation issues raised by it

AE-MS:
## Description

This change updates the repo to the latest version of the typedoc package and fixes documentation issues it flagged that the previous versions did not.

I was also able to put back the `--emit none` parameter on the `docs:validate` step in `package.json`. Recall that in a previous PR (#1555) I removed this parameter because it was causing all errors and warnings to be silenced for some reason. It seems like it was a bug because with this newer version of typedoc I can replace that parameter and still see the errors/warnings (I verified it).

### Validation performed:

Ensured builds pass, ensured tests pass, ensured documentation successfully validates.

### DIFF
--- a/change/@microsoft-teams-js-fb7d03a6-a374-4fc3-a9cd-6f72c9b0952a.json
+++ b/change/@microsoft-teams-js-fb7d03a6-a374-4fc3-a9cd-6f72c9b0952a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated typedoc version and fixed doc issues raised by it",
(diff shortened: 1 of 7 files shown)


shrshindeMSFT (approved):

AE-MS:
I'm putting this back in draft for now because apparently it requires a newer node version.

AE-MS:
> I'm putting this back in draft for now because apparently it requires a newer node version.

Fixed by removing the soon-to-be-retired node 12 from our pipelines with PR #1559

TrevorJoelHarris (approved):

PR #1564, opened 2023-01-30
Updated documentation links to avoid using locale in URLs and use markdown format for external links

AE-MS:
## Description

External links to other documentation pages (i.e., documentation that is not within this teams-js library) should avoid using the locale in their URLs. The Microsoft documentation site will automatically redirect to the correct locale for the user.

Also, per [the typedoc documentation](https://typedoc.org/tags/link/#:~:text=%7B%40link%7D%201%20TSDoc%20Compatibility%20TypeDoc%20implements%20the%20beta,other%20links.%203%20See%20Also%20The%20%40template%20tag), external links should use markdown and not the `@link` tag. I fixed the one instance of this.

### Validation performed:

1. Built the documentation locally and tested each of the three updated URLs to ensure they worked and navigated to the correct place.
2. Ensured builds and tests passed

### DIFF
--- a/change/@microsoft-teams-js-99eca3fb-24d8-4c34-95d3-d83b96469721.json
+++ b/change/@microsoft-teams-js-99eca3fb-24d8-4c34-95d3-d83b96469721.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Update documentation links to not use locale",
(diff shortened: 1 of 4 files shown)


erikadoyle (approved):
Looks good--double checked links resolve as intended

TrevorJoelHarris (approved):

erinha (approved):

PR #1565, opened 2023-01-30
[OLD] Change onLoad/onBeforeUnload register handlers to beta

alexneyman-MSFT:
## Description

These handlers are being updated from an internal and hidden state to beta. They can be used for app caching in Teams meetings. 

### Validation performed:

1. Ensured the newly generated docs were accurate and linked to the correct locations
2. Checked that all builds and tests passed


packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -46,12 +46,14 @@ export namespace teamsCore {
   }
 
   /**
-   * @hidden
    * Registers a handler to be called when the page has been requested to load.
    *
+   * @see Check out [App Caching in Teams](https://learn.microsoft.com/en-us/microsoftteams/platform/apps-in-teams-meetings/build-tabs-for-meeting?tabs=desktop%2Cmeeting-chat-view-desktop%2Cmeeting-stage-view-desktop%2Cchannel-meeting-desktop#app-caching)
```
AE-MS, line 51: >/en-us [](http://example.com/codeflow?start=70&length=6)

Please remove the locale portion of this URL. The docs site will automatically redirect the user to the locale page most appropriate for them. Thank you! :)

packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -86,13 +88,15 @@ export namespace teamsCore {
   }
 
   /**
-   * @hidden
    * Registers a handler to be called before the page is unloaded.
    *
+   * @see Check out {@link https://learn.microsoft.com/en-us/microsoftteams/platform/apps-in-teams-meetings/build-tabs-for-meeting?tabs=desktop%2Cmeeting-chat-view-desktop%2Cmeeting-stage-view-desktop%2Cchannel-meeting-desktop#app-caching | App Caching in Teams}
```
AE-MS, line 93: >{@link  [](http://example.com/codeflow?start=20&length=7)

Please use markdown formatting for external links (like you did on line 51)

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/teamsAPIs.ts
erikadoyle, line 51: I'm not sure if the Typedoc `@see` tag is supported by MS Docs build system--it might be stripped out upon publish. But I guess we will [at]see . . . :-p

erinha: @erikadoyle  is correct. You can look at the `OpenSingleChatRequest` interface in teams-js/src/public/chat.ts:

https://learn.microsoft.com/en-us/javascript/api/@microsoft/teams-js/opensinglechatrequest?view=msteams-client-js-latest

The link gets stripped out

erinha: Is there a different way to link to the doc we can use? Maybe @remarks?

alexneyman-MSFT: I think I can keep the link in and remove @see and it'll work. Not sure how to test it because both work fine when looking at the generated tsdocs in the repo

jadahiya-MSFT (approved):

packages/teams-js/src/public/teamsAPIs.ts
alexneyman-MSFT: image

This is how it looks on my end

erinha: FWIW there is a different system (MS Docs) that consumes typedoc tags to render the reference documentation, so sometimes there are typedoc tags that aren't supported by their system (even though it looks correct to you). @erikadoyle please correct me if I'm wrong!

AE-MS (approved):
:shipit:

erinha:
Ah shoot, you need a changefile:

Change files are required before merging. Please run 'yarn changefile' from the monorepo root. This would be a patch change. Please type a comment in past tense and surround API function names with backtick quotes, e.g.,

"Added @beta tags to `registerBeforeUnloadHandler` and `registerOnLoadHandler` APIs."

PR #1578, opened 2023-02-08
noahdarveau/blazor test app

noahdarveau-MSFT:

## Description

### Main changes in the PR:

1. Added the blazor test app to the monorepo. It currently is not hooked up to any tests, it just exists and builds alongside the teams-js package. BY adding it to the monorepo, people can manually validate their changes don't break c# apps before I finish getting automated testing set up.
2. changed some yarn commands in package.json so that the blazer app is built when running the general build command

### Validation performed:

1. None, as the test app isn't doing anything at the moment other than existing in the monorepo.

### Change file added:

Yes

### Next/remaining steps:

- [ ] Incorporating the Blazor Test App into end-to-end testing so that it runs in the pipeline.


apps/blazor-test-app/Properties/launchSettings.json
```diff
@@ -0,0 +1,25 @@
...
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      },
+      "dotnetRunMessages": "true",
+      "applicationUrl": "https://localhost:44302;http://localhost:2544",
+      "hotReloadProfile": "aspnetcore"
+    },
+    "WSL": {
+      "commandName": "WSL2",
+      "launchBrowser": true,
+      "launchUrl": "https://teams.microsoft.com/l/app/0794bd56-95ae-4835-9588-d81b26ec3d8d?installAppPackage=true&webjoin=true&appTenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&login_hint=noahdarveau@microsoft.com",
```
TrevorJoelHarris, line 17: should your alias be in the login hint?

TrevorJoelHarris: (in both launchUrls)

package.json
```diff
@@ -8,8 +8,9 @@
     "tools/**/*"
   ],
   "scripts": {
-    "build": "lerna run build --stream",
+    "build": "lerna run build --stream && yarn build-blazor-app",
```
TrevorJoelHarris, line 11: How much time does the blazor app build add to our build command?

TrevorJoelHarris (changes requested):

apps/blazor-test-app/FakesAssemblies/Microsoft.AspNetCore.Components.Server.6.0.0.0.Fakes.messages
```diff
@@ -0,0 +1,139 @@
+C:\Users\noahdarveau\Code\Blazor-Test-App\Blazor-Test-App\Fakes\Microsoft.AspNetCore.Components.Server.fakes : warning : Cannot generate stub for MessagePack.MessagePackCode: Type is not visible to exported or assembly(Microsoft.AspNetCore.Components.Server.6.0.0.0.Fakes).
```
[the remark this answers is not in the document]
noahdarveau-MSFT: Yup, missed this one while I was trimming away unnecessary files.

Deleted!

apps/blazor-test-app/FakesAssemblies/Microsoft.AspNetCore.Components.Server.6.0.0.0.Fakes.xml
```diff
@@ -0,0 +1,668 @@
+<?xml version="1.0"?>
```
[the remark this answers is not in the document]
noahdarveau-MSFT: I assumed I needed it, but looking into it a little more, I think it's for documentation, so I added this to the .gitignore

apps/blazor-test-app/Properties/launchSettings.json
noahdarveau-MSFT: Who put that there?! I sure didn't. Anyways, fixed!

package.json
noahdarveau-MSFT: About 6-10 seconds.

packages/teams-js/src/public/video.ts
```diff
@@ -178,7 +178,7 @@ export namespace video {
           sendMessageToParent('video.videoEffectReadiness', [true, effectId]);
         })
         .catch((reason) => {
-          const validReason = reason in EffectFailureReason ? reason: EffectFailureReason.InitializationFailure;
+          const validReason = reason in EffectFailureReason ? reason : EffectFailureReason.InitializationFailure;
```
noahdarveau-MSFT, line 181: This is just a prettier thing that happened automatically when I built.

TrevorJoelHarris (changes requested):

apps/blazor-test-app/wwwroot/js/MicrosoftTeams.min.js
```diff
@@ -0,0 +1,2 @@
...inedTeams||(e.joinedTeams={})),e.getConfigSetting=function(e){return new Promise((function(t){if(me(Z),!n())throw k;t(je("getConfigSetting",e))}))},e.isSupported=n}(e.fullTrust||(e.fullTrust={}))}($t||($t={})),function(e){var n;function t(){return me(Z),st.isSupported()}function o(e){Qe("video.videoFrameProcessed",[e])}function i(e,t){void 0===t&&(t=n.Warn),Qe("video.notifyError",[e,t])}!function(e){e.Fatal="fatal",e.Warn="warn"}(n=e.ErrorLevel||(e.ErrorLevel={})),e.registerForVideoFrame=function(e,n){if(me(Z,S.sidePanel),!t())throw k;ke("video.newVideoFrame",(function(n){if(n){var t=n.timestamp;e(n,(function(){o(t)}),i)}}),!1),Qe("video.registerForVideoFrame",[n])},e.notifySelectedVideoEffectChanged=function(e,n,o){if(me(Z,S.sidePanel),!t())throw k;Qe("video.videoEffectChanged",[e,n,o])},e.registerForVideoEffect=function(e){if(me(Z,S.sidePanel),!t())throw k;ke("video.effectParameterChange",e,!1),Qe("video.registerForVideoEffect")},e.updatePersonalizedEffects=function(e){if(me(Z,S.sidePanel),!st.isSupported())throw k;Qe("video.personalizedEffectsChanged",[e])},e.isSupported=t,e.notifyFatalError=function(e){if(me(Z),!st.isSupported())throw k;i(e,n.Fatal)}}(Yt||(Yt={}))})(),o})()}));
```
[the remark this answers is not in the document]
noahdarveau-MSFT: We don't. For some reason my brain thought it needed to be to run in the pipeline, but that's not correct.

TrevorJoelHarris (dismissed):

apps/blazor-test-app/Properties/launchSettings.json
```diff
@@ -0,0 +1,25 @@
...
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      },
+      "dotnetRunMessages": "true",
+      "applicationUrl": "https://localhost:44302;http://localhost:2544",
+      "hotReloadProfile": "aspnetcore"
+    },
+    "WSL": {
+      "commandName": "WSL2",
+      "launchBrowser": true,
+      "launchUrl": "https://teams.microsoft.com/l/app/0794bd56-95ae-4835-9588-d81b26ec3d8d?installAppPackage=true&webjoin=true&appTenantId=72f988bf-86f1-41af-91ab-2d7cd011db47",
```
shrshindeMSFT, line 17: Why do we need static `launchURL`?

apps/blazor-test-app/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "blazor-test-app",
+  "private": true,
+  "author": "Microsoft Teams",
+  "description": "Blazor test app to ensure teams-js changes don't break Blazor app functionality",
+  "version": "1.0.0",
+  "scripts": {
+    "compile": "dotnet build",
```
shrshindeMSFT, line 8: I would suggest `build` instead of `compile` to be consistent across test apps.

apps/blazor-test-app/appsettings.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "AllowedHosts": "*",
+  "TeamsFx": {
+    "Authentication": {
+      "ClientId": "$clientId$",
+      "ClientSecret": "$client-secret$",
+      "OAuthAuthority": "$oauthAuthority$"
+    }
+  }
+}
```
shrshindeMSFT, line 17: Some files are missing new lines at EOF.

apps/blazor-test-app/Pages/_Host.cshtml
```diff
@@ -0,0 +1,36 @@
...
+        </environment>
+        <environment include="Development">
+            An unhandled exception has occurred. See browser dev tools for details.
+        </environment>
+        <a href="" class="reload">Reload</a>
+        <a class="dismiss">🗙</a>
+    </div>
+
+    <script src="_framework/blazor.server.js"></script>
+    <script type="module" src="https://unpkg.com/@@fluentui/web-components"></script>
```
shrshindeMSFT, line 34: What is `unpkg.com` ? Is it stable registry? Why not npm registry?

apps/blazor-test-app/Properties/launchSettings.json
```diff
@@ -0,0 +1,25 @@
...
+  "profiles": {
+    "Microsoft Teams (browser)": {
+      "commandName": "Project",
+      "launchBrowser": true,
+      "launchUrl": "https://teams.microsoft.com/l/app/0794bd56-95ae-4835-9588-d81b26ec3d8d?installAppPackage=true&webjoin=true&appTenantId=72f988bf-86f1-41af-91ab-2d7cd011db47",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      },
+      "dotnetRunMessages": "true",
+      "applicationUrl": "https://localhost:44302;http://localhost:2544",
```
shrshindeMSFT, line 11: Why the need of `http`?

package.json
shrshindeMSFT, line 11: Why add separately? `Lerna` looks through all the apps and builds the repository.

packages/teams-js/package.json
```diff
@@ -11,7 +11,7 @@
   "main": "./dist/MicrosoftTeams.min.js",
   "typings": "./dist/MicrosoftTeams.d.ts",
   "scripts": {
-    "build": "yarn lint && webpack && yarn docs:validate",
+    "build": "yarn lint && webpack && yarn docs:validate && node updateBlazorApp.js",
```
shrshindeMSFT, line 14: Blazor App is a test app why the need of `updateBlazorApp` here?

packages/teams-js/updateBlazorApp.js
```diff
@@ -0,0 +1,12 @@
+const fs = require('fs');
```
shrshindeMSFT, line 1: No need to copy new build of `teams-js` file to `blazorApp` directory. Is there any other way to work around this?

shrshindeMSFT (changes requested):
Quick question: Other apps directly use `teams-js` package and not need to copy from one directory to another. AFAIK it's not a best practice to consume built package.

apps/blazor-test-app/Properties/launchSettings.json
noahdarveau-MSFT: The static launchURL is used when debugging the Blazor app in Visual Studio. It allows for debugging the Blazor app inside of an instance of Teams web, rather than just the app on its own. So, it is not entirely necessary, as we are using the app on its own inside Orange for testing, but I feel it is nice to have the option to debug inside of Teams web, should the need arise.

apps/blazor-test-app/Pages/_Host.cshtml
noahdarveau-MSFT: *Note: this file was auto-generated by the TeamsFX toolkit* 
unpkg.com is a third party, open-source CDN for npm packages. The TeamsFX team seems to prefer it as their CDN of choice for (at least) their Blazor applications. I know it is very fast and reliable, but mostly I wanted to keep the CDN consistent with other Blazor apps customers would have, so I didn't want to change it to npm.

apps/blazor-test-app/appsettings.json
noahdarveau-MSFT: Added new lines

apps/blazor-test-app/Properties/launchSettings.json
noahdarveau-MSFT: It was added by default, we're not using it, so I removed it.

noahdarveau-MSFT:
@shrshindeMSFT 
So, a bunch of your questions/comments are actually all tied together, so I'm going to answer them all in one giant comment.
To start, you asked why we need to copy over the teams-js package to the Blazor app, while all of the other test apps can just reference it directly. Unfortunately, we can't do this with the Blazor app since all files referenced in the code need to be contained within the `blazor-test-app` package. In a typical c# project, you do have the ability to link in external files without making a copy of them, but unfortunately that does not work in our case since we are referencing a .js file not a .cs. Part of how Blazor is able to interact the c# with the js is by running a webserver at runtime that handles all of the javascript functionality. The files included in the webserver is everything contained in the `wwwroot` folder in the project. So a copy of our teams-js library needs to be included in the `wwwroot` folder at compile time. 
Which brings us to the package scripts. Since we need a copy of the Teams-js library inside the `wwwroot` folder at compile time, I wrote a simple js script `updateBlazorApp.js` that moves a copy of the built teams-js library to the `wwwroot` folder and updated the teams-js build command to execute the script once it is completed building. Since we need teams-js to finish its build command before the blazor app can start compiling, I had to add the blazor build command separately from the Lerna build command. Lerna runs all of the build commands in parallel and AFAIK there isn't a way to configure Lerna such that it ensure teams-js finishes building before the blazor app starts. As such I also named the build command `compile` in the blazor test app rather than `build` as it would be included in the Lerna build if it was named `build`

packages/teams-js/updateBlazorApp.js
noahdarveau-MSFT: see mega-comment

packages/teams-js/package.json
noahdarveau-MSFT: see mega-comment

package.json
noahdarveau-MSFT: see mega-comment

apps/blazor-test-app/package.json
noahdarveau-MSFT: see mega-comment

apps/blazor-test-app/Properties/launchSettings.json
shrshindeMSFT: What if the same issue arises in Outlook or office? Also, I'm little bit worried about public display of `appTenantID` in the URL.

apps/blazor-test-app/Pages/_Host.cshtml
shrshindeMSFT: I would suggest it to have `npm` packages, we cannot rely on 3rd party CDNs.

shrshindeMSFT:
> @shrshindeMSFT So, a bunch of your questions/comments are actually all tied together, so I'm going to answer them all in one giant comment. To start, you asked why we need to copy over the teams-js package to the Blazor app, while all of the other test apps can just reference it directly. Unfortunately, we can't do this with the Blazor app since all files referenced in the code need to be contained within the `blazor-test-app` package. In a typical c# project, you do have the ability to link in external files without making a copy of them, but unfortunately that does not work in our case since we are referencing a .js file not a .cs. Part of how Blazor is able to interact the c# with the js is by running a webserver at runtime that handles all of the javascript functionality. The files included in the webserver is everything contained in the `wwwroot` folder in the project. So a copy of our teams-js library needs to be included in the `wwwroot` folder at compile time. Which brings us to the package scripts. Since we need a copy of the Teams-js library inside the `wwwroot` folder at compile time, I wrote a simple js script `updateBlazorApp.js` that moves a copy of the built teams-js library to the `wwwroot` folder and updated the teams-js build command to execute the script once it is completed building. Since we need teams-js to finish its build command before the blazor app can start compiling, I had to add the blazor build command separately from the Lerna build command. Lerna runs all of the build commands in parallel and AFAIK there isn't a way to configure Lerna such that it ensure teams-js finishes building before the blazor app starts. As such I also named the build command `compile` in the blazor test app rather than `build` as it would be included in the Lerna build if it was named `build`

Instead of copying the `teams-js` try to use `webpack.config.js` see if we can leverage it. Also, no to need modify the `package.json` file as `webpack.config.js` can output the file to the blazor directory.

apps/blazor-test-app/Properties/launchSettings.json
noahdarveau-MSFT: Very well, I will remove it.

apps/blazor-test-app/Pages/_Host.cshtml
noahdarveau-MSFT: So Blazor also just outright does not support using npm. I found a hacky way of possibly getting it to work, but it would likely be far less reliable than just using unpkg. The npm documentation for the fluentui/web-components package actually uses unpkg. I also saw it recommended by other microsoft employees in the github threads I was looking at to try and find a way to get it working with npm, so leaving it using unpkg will be fine.

apps/blazor-test-app/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+.DS_Store
+.env.teamsfx.local
+subscriptionInfo.json
+build
+.fx/configs/config.local.json
+.fx/states/state.local.json
+appsettings.Development.json
+FakesAssemblies/Microsoft.AspNetCore.Components.Server.6.0.0.0.Fakes.xml
+FakesAssemblies/Microsoft.AspNetCore.Components.Server.6.0.0.0.Fakes.messages
+wwwroot/js/MicrosoftTeams.min.js
```
AE-MS, line 14: I know that I am a broken record about such things but alphabetizing this list would make it easier for people to look through it to see if something is present.

apps/blazor-test-app/README.md
```diff
@@ -0,0 +1,25 @@
...
+// Ensuring you have installed and built the Teams JavaScript client SDK
+yarn install
+yarn build
+
+yarn start-blazor-app
+```
+
+or if you have already built the Teams JavaScript client SDK and would like to build and run directly from the project directory blazor-test-app, simply `yarn compile` and `yarn start` there.
+
+Once starting the app, it will run on https://localhost:44302. Upon visiting the page, the text `Congratulations` on the page indicates the c# app is working properly, otherwise and error will be thrown in the console and the webpage will not render.
```
AE-MS, line 23: >and [](http://example.com/codeflow?start=178&length=3)

an

noahdarveau-MSFT:
>Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory

I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.

apps/blazor-test-app/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+_NOTE: The Blazor Test App Needs to have the latest compiled version of MicrosoftTeams.min.js located inside the blazor-test-app/wwwroot/js directory. This can be copied over manually, however this will be done automatically when building the entire teams-js package from the monorepo root, so that course of action is recommended_
```
AE-MS, line 25: >_NOTE [](http://example.com/codeflow?start=0&length=5)

Did you intend for their to be a trailing underscore here as well for markdown formatting?

apps/blazor-test-app/.gitignore
AE-MS: (unless this is an auto generated file)

AE-MS:
> export function initializeAsync() {

Out of curiosity are the contents of this file provided by Teams Toolkit or did you author it? It seems like all this boilerplate interop code should be provided by the toolkit to avoid making every C# developer write it themselves?

---
Refers to: apps/blazor-test-app/wwwroot/js/TeamsJsBlazorInterop.js:1 in 1fb07f4. [](commit_id = 1fb07f4bf855e6e4d8a5404a94f8a2b5bd3931fa, deletion_comment = False)

apps/blazor-test-app/Pages/_Host.cshtml
AE-MS: Given that the goal of this app is to accurately represent what a customer would do to avoid breaking them, using the Teams Toolkit defaults is the right thing to do here.

AE-MS:
> /* eslint-disable @typescript-eslint/no-var-requires */

Generally speaking I would check in this change separate from adding the C# app. By making these changes independent, it may be easier/clearer if something breaks which to revert. Moreover, if this change breaks something and we have to revert, you won't also get other code unnecessarily rolled back.

This is more just general guidance, since you've already made the change this way now.

---
Refers to: packages/teams-js/webpack.config.js:1 in 1fb07f4. [](commit_id = 1fb07f4bf855e6e4d8a5404a94f8a2b5bd3931fa, deletion_comment = False)

packages/teams-js/src/public/video.ts
AE-MS: Happened to me too in a recent PR. We'll race to see who checks it in first!

AE-MS (changes requested):
:clock1:

noahdarveau-MSFT:
> > export function initializeAsync() {
> 
> Out of curiosity are the contents of this file provided by Teams Toolkit or did you author it? It seems like all this boilerplate interop code should be provided by the toolkit to avoid making every C# developer write it themselves?
> 
> Refers to: apps/blazor-test-app/wwwroot/js/TeamsJsBlazorInterop.js:1 in [1fb07f4](https://github.com/OfficeDev/microsoft-teams-library-js/commit/1fb07f4bf855e6e4d8a5404a94f8a2b5bd3931fa). [](commit_id = [1fb07f4](https://github.com/OfficeDev/microsoft-teams-library-js/commit/1fb07f4bf855e6e4d8a5404a94f8a2b5bd3931fa), deletion_comment = False)

Most of it was autogenerated, then I added some functionality to it.

apps/blazor-test-app/.gitignore
noahdarveau-MSFT: Yes, this was auto generated. Yes, I also fixed it for you.

apps/blazor-test-app/wwwroot/js/TeamsJsBlazorInterop.js
```diff
@@ -0,0 +1,61 @@
...
+  microsoftTeams.pages.config.registerOnSaveHandler((saveEvent) => {
+    microsoftTeams.pages.config.setConfig(settings);
+    saveEvent.notifySuccess();
+  });
+
+  microsoftTeams.pages.config.setValidityState(true);
+}
+
+// Come from here: https://github.com/wictorwilen/msteams-react-base-component/blob/master/src/useTeams.ts
+export function inOrange() {
```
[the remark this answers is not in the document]
noahdarveau-MSFT: >Is this function trying to determine if the app is being run as a Teams app vs a plain web app?

Yes that's correct
>If so, the official way to do this is to call app.initialize() and see if it succeeds or fails.

So the `app.initialize()` method actually doesn't work. For some reason (that I haven't completely worked out),  calling `app.initialize()` from the js method will always result in a failure, even when running inside a host (in this case orange). The js side of blazor works by hosting a web server so when app.initialize() is called, its being called in a server environment, so it always fails. However, I'm not sure why the same function, when called from the .razor file, works perfectly fine.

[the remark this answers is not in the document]
noahdarveau-MSFT: For context, I adapted this method from the autogenerated isInTeams function, which was doing similar checks, except they didn't work with Orange, so I had to slightly change it. (I don't remember exactly what it was checking for the existence of).

shrshindeMSFT:
> >Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory
> 
> 
> 
> I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.



Did you try passing arguments via command line interface ? 

`webpack --args copy-blazer-app`

noahdarveau-MSFT:
> > > Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory
> > 
> > 
> > I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.
> 
> Did you try passing arguments via command line interface ?
> 
> `webpack --args copy-blazer-app`

Yeah, I looked into webpack's cli and there isn't anything available for copying the output to multiple directories. The underlying issue is that webpack does not have any copying functionality. It also doesn't have any capabilities to output to more locations than there are entry points. (i.e. output to ./dist and ./wwwroot for entry: ./teams-js), so I had to use the plugin to do the copying.

shrshindeMSFT:
> > > > Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory
> > > 
> > > 
> > > I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.
> > 
> > 
> > Did you try passing arguments via command line interface ?
> > `webpack --args copy-blazer-app`
> 
> Yeah, I looked into webpack's cli and there isn't anything available for copying the output to multiple directories. The underlying issue is that webpack does not have any copying functionality. It also doesn't have any capabilities to output to more locations than there are entry points. (i.e. output to ./dist and ./wwwroot for entry: ./teams-js), so I had to use the plugin to do the copying.

File Manager Plugin looks good to me. But if we are using plugin then we don't need to explicitly state to build blazer app in package.json. `yarn build-blazor-app` it should be handled by lerna and webpack

noahdarveau-MSFT:
> > > > > Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory
> > > > 
> > > > 
> > > > I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.
> > > 
> > > 
> > > Did you try passing arguments via command line interface ?
> > > `webpack --args copy-blazer-app`
> > 
> > 
> > Yeah, I looked into webpack's cli and there isn't anything available for copying the output to multiple directories. The underlying issue is that webpack does not have any copying functionality. It also doesn't have any capabilities to output to more locations than there are entry points. (i.e. output to ./dist and ./wwwroot for entry: ./teams-js), so I had to use the plugin to do the copying.
> 
> File Manager Plugin looks good to me. But if we are using plugin then we don't need to explicitly state to build blazer app in package.json. `yarn build-blazor-app` it should be handled by lerna and webpack

Well we still need to ensure that teams-js finishes building before starting to build the blazor app. Lerna handles all of the builds in parallel, so we can't guarantee teams-js finishes before the blazor app (its actually likely the blazor app would finish first most times). Having the distinct build-blazor-app command lets us ensure the order of operations there.

shrshindeMSFT:
> > > > > > Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory
> > > > > 
> > > > > 
> > > > > I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.
> > > > 
> > > > 
> > > > Did you try passing arguments via command line interface ?
> > > > `webpack --args copy-blazer-app`
> > > 
> > > 
> > > Yeah, I looked into webpack's cli and there isn't anything available for copying the output to multiple directories. The underlying issue is that webpack does not have any copying functionality. It also doesn't have any capabilities to output to more locations than there are entry points. (i.e. output to ./dist and ./wwwroot for entry: ./teams-js), so I had to use the plugin to do the copying.
> > 
> > 
> > File Manager Plugin looks good to me. But if we are using plugin then we don't need to explicitly state to build blazer app in package.json. `yarn build-blazor-app` it should be handled by lerna and webpack
> 
> Well we still need to ensure that teams-js finishes building before starting to build the blazor app. Lerna handles all of the builds in parallel, so we can't guarantee teams-js finishes before the blazor app (its actually likely the blazor app would finish first most times). Having the distinct ` yarn build-blazor-app` command lets us ensure the order of operations.

Lerna builds the packages in parallel but we can specify in `devDependencies` that we need to build `teams-js` first then `blazor-razor-app`. More details can be found [here](https://stackoverflow.com/questions/50769518/lerna-specify-run-order).

noahdarveau-MSFT:
> > > > > > > Instead of copying the teams-js try to use webpack.config.js see if we can leverage it. Also, no to need modify the package.json file as webpack.config.js can output the file to the blazor directory
> > > > > > 
> > > > > > 
> > > > > > I wasn't able to find any way to have webpack output teams-js to multiple directories for the same entry, but I was able to use a webpack plugin to do the copying of the file.
> > > > > 
> > > > > 
> > > > > Did you try passing arguments via command line interface ?
> > > > > `webpack --args copy-blazer-app`
> > > > 
> > > > 
> > > > Yeah, I looked into webpack's cli and there isn't anything available for copying the output to multiple directories. The underlying issue is that webpack does not have any copying functionality. It also doesn't have any capabilities to output to more locations than there are entry points. (i.e. output to ./dist and ./wwwroot for entry: ./teams-js), so I had to use the plugin to do the copying.
> > > 
> > > 
> > > File Manager Plugin looks good to me. But if we are using plugin then we don't need to explicitly state to build blazer app in package.json. `yarn build-blazor-app` it should be handled by lerna and webpack
> > 
> > 
> > Well we still need to ensure that teams-js finishes building before starting to build the blazor app. Lerna handles all of the builds in parallel, so we can't guarantee teams-js finishes before the blazor app (its actually likely the blazor app would finish first most times). Having the distinct ` yarn build-blazor-app` command lets us ensure the order of operations.
> 
> Lerna builds the packages in parallel but we can specify in `devDependencies` that we need to build `teams-js` first then `blazor-razor-app`. More details can be found [here](https://stackoverflow.com/questions/50769518/lerna-specify-run-order).

Ahh I didn't know that, thanks for sharing the link! I'll go ahead and make the changes.

apps/blazor-test-app/.gitignore
AE-MS: ezgif com-gif-maker

AE-MS (dismissed):
:shipit:

apps/blazor-test-app/README.md
```diff
@@ -0,0 +1,25 @@
...
+cd {monorepo root}
+
+// Ensuring you have installed and built the Teams JavaScript client SDK
+yarn install
+yarn build
+
+yarn start-blazor-app
+```
+
+or if you have already built the Teams JavaScript client SDK and would like to build and run directly from the project directory blazor-test-app, simply `yarn compile` and `yarn start` there.
```
shrshindeMSFT, line 21: `yarn compile` -> `yarn build`

shrshindeMSFT (changes requested):
Looks good so far only small, tiny `readme` change.

apps/blazor-test-app/README.md
noahdarveau-MSFT: Good catch. Thanks!

shrshindeMSFT (approved):
LGTM

TrevorJoelHarris (approved):

PR #1598, opened 2023-02-21
Make app.initialize return an object containing only valid capabilities and functions that an app developer can use

TrevorJoelHarris:
## Description

This was my FHL project: change app.initialize to return an object that contains all of the capabilities and functions that an app developer can call successfully, and none of the ones that they can't. Filtering out "unusable" functions is done by looking at a couple of different things:

1. Whether or not the host supports each capability
2. The current frame context being used
3. Whether or not the app developer has requested the microsoft-only capabilities from the private folder

The bulk of this work is done in the new supportedCapabilities.ts file. In essence, we have script go through the runtime object, check whether each capability is supported, and if it is we put it and all of its functions onto a new object. If it isn't supported, we don't put that capability or any of its subcapabilities on the object. Here is the object that is returned: 

image

Each capability file now also includes a class that is an implementation of the new CapabilityMetadata abstract class in capability.ts. This lets us associate metadata with that capability (like what framecontexts are supported by each function in the capability).

This new object is currently being returned from the app.initialize call.

You can look at app.spec.ts for unit tests showing how this flow would work from an app developer's POV. At a high level...

**User calls app.initialize**
```
const supportedCapabilities = app.initialize(...);
```

**If a capability is supported**
```
supportedCapabilities.barCode.isSupported(); // returns true
supportedCapabilities.barCode.requestPermission(); // works as expected
```

**If a capability is *not* supported**
```
supportedCapabilities.barCode.isSupported(); // returns false
supportedCapabilities.barCode.requestPermission // is undefined
supportedCapabilities.barCode.requestPermission(); // throws an error because the function is undefined
```

**If a specific function in a capability doesn't work due to frame context**
```
// Frame context is set to FrameContexts.sidePanel
supportedCapabilities.dialog.isSupported(); // returns false
supportedCapabilities.dialog.open(); // works correctly
supportedCapabilities.barCode.submit // undefined because submit is not supported in sidePanel
supportedCapabilities.barCode.submit(); // throws an error
```

**If a user tries to call microsoft only functions without explicitly saying that they want them in app.initialize**
```
// throws an error because microsoftOnly is undefined, which is what contains all capabilities in the private folder
supportedCapabilities.microsoftOnly?.teams.getChannels();
```

We need to decide if we think this is an improvement for app developers: It certainly makes it harder to call functions that can't possibly be useful to them which would hopefully cut down on how many open issues happen due to calling functions with the wrong frame context for example.

Assuming we decide to carry through with this, remaining work off the top of my head:

1. Make sure we are still generating reasonable documentation for everything
2. Figure out how to get a second app.initialize function with an overloaded signature so we can still have the old one that returns void. That way the new signature can be marked with the beta tag.
3. Clean up supportedCapabilities.ts (and the other files). There is a good amount of readability and refactoring required.
4. Add unit tests for making sure the right frame contexts are being checked for each function. Could possibly just update existing unit tests?
5. Does CapabilityMetadata need to be an abstract class? Should each function that checks frameContext be checking metadata instead of duplicating the framecontext list?

### Main changes in the PR:

1. supportedCapabilities.ts generated object returned by app.initialize
2. capabilityMetadata in capability.ts. Each capability file includes an implementation of this.
3. app.initialize changed to return the new object, and unit tests in app.spec.ts

### Validation performed:

1. Unit tests run

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/change/@microsoft-teams-js-5e0d3f05-925c-4b22-9301-ffe29414b433.json
+++ b/change/@microsoft-teams-js-5e0d3f05-925c-4b22-9301-ffe29414b433.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "`app.initialize` now returns an object that only contains functions that can be called legally by the app developer.",
+  "packageName": "@microsoft/teams-js",
+  "email": "trharris@microsoft.com",
+  "dependentChangeType": "patch"
+}


TrevorJoelHarris:
Closing: this isn't intended for check in at this point

PR #1600, opened 2023-02-23
Noahdarveau/comp gov fix

noahdarveau-MSFT:
Removing the Microsoft.QualityTools.Testing.Fakes dependency as it uses the System.Drawing.Common v4.7.0 dependency, which contains an identified security vulnerability. Since Microsoft.QualityTools.Testing.Fakes is not necessary for the app, simply removing it is the best solution.


apps/blazor-test-app/Blazor-Test-App.csproj
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="Microsoft.Fast.Components.FluentUI" Version="1.5.3" />
     <PackageReference Include="Microsoft.QualityTools.Testing.Fakes" Version="16.11.230815" />
     <PackageReference Include="Microsoft.TeamsFx" Version="1.2.*" />
+    <PackageReference Include="System.Drawing.Common" Version="4.7.2" />
```
maglims, line 28: Looking at here, it says System.drawing.Common only supported on Windows?
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-And starting with .net6, we might start to build warning when running on non-windows machine.

so it feels we could evaluate a few options: is that possible that we find a replacement for System.Drawing.Common to be used in test app? and if not, maybe we could find a way to run this blazor test app on window agent from mobile side?

maglims (approved):

TrevorJoelHarris (changes requested):
This change itself looks good, but can you please update the PR description to say why you need to remove this reference? It took me a few minutes to figure out why deleting this was the right thing to do, and I have the context for what's going on. Just a brief description of why this package was causing a problem would be great

TrevorJoelHarris (approved):

PR #1603, opened 2023-02-27
Add more documentation to dialog and tasks capabilities

AE-MS:
## Description

Some of the documentation for the `tasks` and `dialog` namespaces/functions were terse and didn't contain links to relevant details. In this change, I have expanded the documentation to clarify which areas should be used in which cases and provided links to relevant data structures and documentation.

### Change file added:

Yes


packages/teams-js/src/public/dialog.ts
```diff
@@ -21,7 +21,15 @@ import {
 import { runtime } from './runtime';
 
 /**
- * Namespace to interact with the dialog module-specific part of the SDK.
+ * This group of capabilities enables apps to show dialogs. There are two primary types of dialogs: URL-based dialogs and [Adaptive Card](https://learn.microsoft.com/adaptive-cards/) dialogs.
+ * Both types of dialogs are shown in a modal dialog on top of your app.
```
TrevorJoelHarris, line 25: Is this also true for mobile dialogs?

AE-MS: Hmm... I think we're getting deep into UX terminology territory. The behavior on mobile is "modal" in the sense that the user cannot interact with the underlying app while the "dialog" is showing. However, the tricky bit really is that the UX isn't a dialog in the classical sense of the word (it's just another full screen page on top of the full screen app page).

Maybe I can sidestep whether "modal" is the right word here by instead changing it to describe the behavior. How does the updated version look?

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/dialog.ts
```diff
@@ -21,7 +21,15 @@ import {
 import { runtime } from './runtime';
 
 /**
- * Namespace to interact with the dialog module-specific part of the SDK.
+ * This group of capabilities enables apps to show dialogs. There are two primary types of dialogs: URL-based dialogs and [Adaptive Card](https://learn.microsoft.com/adaptive-cards/) dialogs.
```
erikadoyle, line 24: Maybe introduce the modal dialog concept in the 1st sentence:

This group of capabilities enables apps to show **modal** dialogs.

packages/teams-js/src/public/dialog.ts
```diff
@@ -21,7 +21,15 @@ import {
...
+ * Both types of dialogs are shown on top of your app, preventing interaction with your app while they are displayed.
+ * - URL-based dialogs allow you to specify a URL for which the contents will be shown inside the dialog.
+ *   - For URL dialogs, you should use the functions and interfaces in the {@link dialog.url} namespace.
```
erikadoyle, line 27: For URL dialogs, **use the** functions and interfaces . . . 

(Drop "you should" to be more concise, per [MS Docs styleguide](https://review.learn.microsoft.com/en-us/help/contribute/writing-principles?branch=main#3-write-concisely))

packages/teams-js/src/public/dialog.ts
```diff
@@ -21,7 +21,15 @@ import {
...
+ * - Adaptive Card-based dialogs allow you to provide JSON describing an Adaptive Card that will be shown inside the dialog.
+ *   - For Adaptive Card dialogs, you should use the functions and interfaces in the {@link dialog.adaptiveCard} namespace.
```
erikadoyle, line 29: "use the functions"

packages/teams-js/src/public/dialog.ts
```diff
@@ -21,7 +21,15 @@ import {
...
+ *
+ * @remarks Note that dialogs were previously called "task modules". While they have been renamed for clarity, the functionality has been maintained.
+ * For more details about dialogs, see [Task Modules](https://learn.microsoft.com/microsoftteams/platform/task-modules-and-cards/what-are-task-modules)
```
erikadoyle, line 32: We're [in the process](https://github.com/MicrosoftDocs/msteams-docs/pull/7853) of updating Teams docs from Task Module-> Dialog terminology. How about:

For more details, see [Dialogs](https://learn.microsoft.com/microsoftteams/platform/task-modules-and-cards/what-are-task-modules).

erikadoyle (commented):
A few minor suggestions, otherwise looks great!

packages/teams-js/src/public/dialog.ts
AE-MS: The experience of "dialogs" on mobile hosts is different than on web/Windows hosts. Specifically, the UX implementation isn't really a "dialog" per se -- e.g., it doesn't look like a pop-up box on top of the other content (which is what I typically think of as a dialog). On mobile, the implementation is basically another full screen element that looks like the app that launched it (e.g., the chrome/container is the same and not differentiated). The only thing that makes it remotely like a dialog is that it has an X in the upper left corner that, when tapped, will bring the user back to the underlying app. We can't change the "dialog" capability name, but I'm wondering if we should still use words like modal and dialog in the description?
It is still "modal" in the sense that user cannot do anything else in the host while it's showing but I didn't know if using the modal terminology was still appropriate here when there's no obvious "dialog" (even though I know that technically "modal" doesn't have to apply to a dialog)

erikadoyle (approved):

TrevorJoelHarris (approved):

Issue #1621, opened 2023-03-03
Cannot find name 'VideoFrame'


vvolodin:
After updating to 2.9.0 teams-js breaks my build again with:

```
main.ts depends on '@microsoft/teams-js'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies



Error: node_modules/@microsoft/teams-js/dist/MicrosoftTeams.d.ts:6815:37 - error TS2304: Cannot find name 'VideoFrame'.

6815                         videoFrame: VideoFrame;
                                         ~~~~~~~~~~


Error: node_modules/@microsoft/teams-js/dist/MicrosoftTeams.d.ts:6823:97 - error TS2304: Cannot find name 'VideoFrame'.

6823                 type VideoFrameCallback = (receivedVideoFrame: MediaStreamFrameData) => Promise<VideoFrame>;
                                                                                                     ~~~~~~~~~~
```

nwojod-MSFT:
@vvolodin thanks for reporting this. We are currently investigating this issue. We will update you on a fix.

TrevorJoelHarris:
@vvolodin Could you tell me a little about how your build process is put together and what commands you ran to upgrade to 2.9.0 and then build? You also caught a similar issue a few months ago: https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

Seems pretty clear to me that we have a hole in our automation somewhere and I'd like to figure out what that is so we can avoid making more releases with this problem in the future.

TrevorJoelHarris:
We've released 2.9.1 to fix, please try that out and let us know if you have any more problems.

lubobill1990:
> We've released 2.9.1 to fix, please try that out and let us know if you have any more problems.

Hi @TrevorJoelHarris @nwojod-MSFT,

It seems we need to add a pipeline, like a test app to consume and build with the built SDK package.

vvolodin:
2.9.1 fixed my issue. I don't have any special update procedure, I check for new versions of npm packages using `npm-check` utility and it installs updates via `npm i`.

AE-MS:
> > We've released 2.9.1 to fix, please try that out and let us know if you have any more problems.
> 
> Hi @TrevorJoelHarris @nwojod-MSFT,
> 
> It seems we need to add a pipeline, like a test app to consume and build with the built SDK package.

[...]

PR #1586, opened 2023-02-13
new API: video.mediaStream.registerForVideoFrame

WengHe:
## Description

Added a new sub-namespace: `video.mediaStream` and provided a new API under this namespace: `video.mediaStream.registerForVideoFrame`. The video app should use this new API to get video frames when the capability is supported by the host.

### Main changes in the PR:

1. Added a new API `video.mediaStream.registerForVideoFrame`.
2. Renamed video.VideoFrame to video.VideoFrameData to avoid name collision with the JS native VideoFrame interface.

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

PR #1628, opened 2023-03-07
Wenghe/media stream namespace

WengHe:
## Description
(This PR recovers the PR #1586 , which was reverted because of a project dependencies issue #1621 .)
Added a new sub-namespace: `video.mediaStream` and provided a new API under this namespace: `video.mediaStream.registerForVideoFrame`. The video app should use this new API to get video frames when the capability is supported by the host.

### Main changes in the PR:

1. Added a new API `video.mediaStream.registerForVideoFrame`.
2. Renamed video.VideoFrame to video.VideoFrameData to avoid name collision with the JS native VideoFrame interface.

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/package.json
```diff
@@ -20,6 +20,8 @@
     "test": "jest"
   },
   "dependencies": {
+    "@types/dom-mediacapture-transform": "0.1.2",
```
TrevorJoelHarris, line 23: I assume this is the main difference between this change and the previous version that broke a customer. Before we approve this, you will need to be able to explain to us why this change will fix that customer issue. Also, you will need to be able to reproduce the customer reported issue: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1628

Once you can reproduce their issue, please verify that this fix will no longer break them.

TrevorJoelHarris:
Also, it looks like there are merge conflicts you will need to resolve by hand before this can be checked in.

TrevorJoelHarris (changes requested):
Please verify that this change will fix the reported customer issue and resolve the merge conflicts.

WengHe:
Before fix:
image

After fix, with the locally built package:
image

packages/teams-js/package.json
WengHe: Reproduced the issue and confirmed the fix won't bring the same issue again, please see the comment I posted below.

TrevorJoelHarris:
I talked with @WengHe on the side: the issue here was that we were exporting types that were only included as dev dependencies. This works if the test app is inside of the same monorepo (as ours is), but will cause problems once someone externally tries to import the package. We'll have someone look into building new automation to catch this scenario going forward.

TrevorJoelHarris (approved):

packages/teams-js/src/public/video.ts
```diff
@@ -216,4 +222,159 @@ export namespace video {
...
+    type MediaStreamInfo = {
+      streamId: string;
+    };
+
+    /**
+     * @beta
+     * Register to read the video frames from the media stream provided by the host.
+     * @param frameCallback - The callback to invoke when recieve a video frame from the media stream.
+     * @example
+     * ```typescript
```
aeunms, line 279: Nit: Might be good to include comment on what this is commented out/saved for

aeunms (approved):
🚢

packages/teams-js/src/public/video.ts
TrevorJoelHarris: This is how you put a code example in the documentation that gets formatted nicely. See here for an example that's already live: https://learn.microsoft.com/en-us/javascript/api/@microsoft/teams-js/search?view=msteams-client-js-latest

PR #1632, opened 2023-03-08
Fixing incorrect version name being chosen in prepBetaRelease script

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Converting pkgParts and betaParts values into numbers before comparisons. String comparison checker was comparing '10' to '9' and that caused '10' < '9' due to checking each character at a time rather than the full number.

### DIFF
--- a/change/@microsoft-teams-js-856c0e77-b263-4d92-ba73-eeb35a3c66b1.json
+++ b/change/@microsoft-teams-js-856c0e77-b263-4d92-ba73-eeb35a3c66b1.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Modified prepBetaRelease script",
+  "packageName": "@microsoft/teams-js",
+  "email": "jdahiya8719@outlook.com",
+  "dependentChangeType": "none"
+}
--- a/packages/teams-js/prepBetaRelease.js
+++ b/packages/teams-js/prepBetaRelease.js
@@ -75,10 +75,12 @@ function getNewerPrefix(currBetaVer, currPkgJsonVer) {
   const pkgParts = currPkgPrefix.split('.');
 
   for (let i = 0; i < betaParts.length; i++) {
-    if (betaParts[i] > pkgParts[i]) {
+    const betaPart = Number(betaParts[i]);
+    const pkgPart = Number(pkgParts[i]);
+    if (betaPart > pkgPart) {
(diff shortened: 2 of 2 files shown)


change/@microsoft-teams-js-856c0e77-b263-4d92-ba73-eeb35a3c66b1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Modified prepBetaRelease script",
```
AE-MS, line 3: >Modified prepBetaRelease script [](http://example.com/codeflow?start=14&length=31)

I know that this doesn't matter as much for external people reading the changelog but you might consider elaborating just a bit, e.g. "Modified prepBetaRelease script to use proper number comparison instead of string comparison for numbers."

packages/teams-js/prepBetaRelease.js
```diff
@@ -75,10 +75,12 @@ function getNewerPrefix(currBetaVer, currPkgJsonVer) {
...
+    const betaPart = Number(betaParts[i]);
+    const pkgPart = Number(pkgParts[i]);
+    if (betaPart > pkgPart) {
       return currBetaPrefix;
     }
-    if (pkgParts[i] > betaParts[i]) {
+    if (pkgPart > betaPart) {
       return currPkgPrefix;
     }
   }
```
AE-MS, line 86: I know that this isn't something you're changing as part of this but would it be prudent to put code after this for loop that throws/crashes/whatever to indicate that we got to a place in the function that we never expected to get to?

AE-MS (approved):
:shipit:

TrevorJoelHarris (approved):

change/@microsoft-teams-js-856c0e77-b263-4d92-ba73-eeb35a3c66b1.json
jadahiya-MSFT: This won't actually be added to the changelog due to the "none" status, hence why I elected to keep this short and instead elaborate more on the PR description itself.

packages/teams-js/prepBetaRelease.js
jadahiya-MSFT: Good idea, since this was merged I will put out a separate PR with those changes.

PR #1636, opened 2023-03-09
Bug fix on sharing.isSupported() returning true while not supported on mobile

maglims:

## Description

During my last DRI session, there was a bug reported on sharing.isSupported() returning true while we can repro that it is not  supported on mobile. The root cause is: if host doesn't send empty object to app sdk during runtime, app sdk will provide a default empty object, meaning that capability is supported. Sharing is in the list of this default setting regardless platform. In this PR, we are removing the default support and instead we start to specify in which version that it gets supported and also provide platform info(currently, sharing is only supported on desktop and web)

### Main changes in the PR:

1. Removing sharing from supported default list
2. Providing a default version on when sharing starts to be supported on desktop and web

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:
Yes, updated existing

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/change/@microsoft-teams-js-0921e3b3-3094-48eb-9bc3-26740e08336f.json
+++ b/change/@microsoft-teams-js-0921e3b3-3094-48eb-9bc3-26740e08336f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix sharing.isSupported bug which returns true while not supported on mobile",
+  "packageName": "@microsoft/teams-js",
+  "email": "magli@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-0921e3b3-3094-48eb-9bc3-26740e08336f.json
+++ b/change/@microsoft-teams-js-0921e3b3-3094-48eb-9bc3-26740e08336f.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Added a bug fix for sharing.isSupported which returns true while not supported on mobile",
+  "comment": "`sharing.isSupported` now returns the correct value on mobile platforms",
   "packageName": "@microsoft/teams-js",
   "email": "magli@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-0921e3b3-3094-48eb-9bc3-26740e08336f.json
+++ b/change/@microsoft-teams-js-0921e3b3-3094-48eb-9bc3-26740e08336f.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "fix sharing.isSupported bug which returns true while not supported on mobile",
+  "comment": "Added a bug fix for sharing.isSupported which returns true while not supported on mobile",
   "packageName": "@microsoft/teams-js",
   "email": "magli@microsoft.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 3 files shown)


TrevorJoelHarris (dismissed):

AE-MS:
>   '2.0.1': [

Does any version of Teams desktop (Angular or React) pass a "clientSupportedSDKVersion" in response to the `initialize` call? If so, what version do they pass?

---
Refers to: packages/teams-js/src/public/runtime.ts:311 in 46c2e22. [](commit_id = 46c2e226494cca25369f66192f0e3b8e073eae49, deletion_comment = False)

AE-MS (changes requested):
:clock1:

AE-MS:
>   '2.0.1': [

I looked into it. It seems like Angular Teams (pre-host SDK) doesn't pass a version at all, so it will be 2.0.1 in that case (since that's the default version this library assumes when no version is returned from the initialize call).

However, React Teams, which has integrated with the host SDK, looks like it will return 2.0.0 since that's what the host SDK returns (see the link I sent you). That makes me think that this addition needs to go in to the 2.0.0 section otherwise it won't work in React Teams and only Angular Teams.

---
In reply to: [1462540334](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1636#issuecomment-1462540334) [](http://example.com/codeflow?ancestors=1462540334)

---
Refers to: packages/teams-js/src/public/runtime.ts:311 in 46c2e22. [](commit_id = 46c2e226494cca25369f66192f0e3b8e073eae49, deletion_comment = False)

AE-MS:
> describe('sharing_v1', () => {

Should there be tests which validate that when the client type is returned from the init call is mobile that sharing is not supported? And that it is supported for the versions we expect when not on mobile?

---
Refers to: packages/teams-js/test/public/sharing.spec.ts:13 in 46c2e22. [](commit_id = 46c2e226494cca25369f66192f0e3b8e073eae49, deletion_comment = False)

maglims:
> > describe('sharing_v1', () => {
> 
> Should there be tests which validate that when the client type is returned from the init call is mobile that sharing is not supported? And that it is supported for the versions we expect when not on mobile?
> 
> Refers to: packages/teams-js/test/public/sharing.spec.ts:13 in [46c2e22](https://github.com/OfficeDev/microsoft-teams-library-js/commit/46c2e226494cca25369f66192f0e3b8e073eae49). [](commit_id = [46c2e22](https://github.com/OfficeDev/microsoft-teams-library-js/commit/46c2e226494cca25369f66192f0e3b8e073eae49), deletion_comment = False)

maglims:
> > > describe('sharing_v1', () => {
> > 
> > 
> > Should there be tests which validate that when the client type is returned from the init call is mobile that sharing is not supported? And that it is supported for the versions we expect when not on mobile?
> > Refers to: packages/teams-js/test/public/sharing.spec.ts:13 in [46c2e22](https://github.com/OfficeDev/microsoft-teams-library-js/commit/46c2e226494cca25369f66192f0e3b8e073eae49). [](commit_id = [46c2e22](https://github.com/OfficeDev/microsoft-teams-library-js/commit/46c2e226494cca25369f66192f0e3b8e073eae49), deletion_comment = False)

Great suggestions! thanks

TrevorJoelHarris (approved):

AE-MS (approved):
:shipit:

PR #1637, opened 2023-03-10
Use a different event name when calling video.mediaStream.registerForVideoFrame

WengHe:
## Description
When incidentally calling both video.registerForVideoFrame and video.mediaStream.registerForVideoFrame, two events with the same name are sent to the host and the host can't identify what triggered the event.

### Main changes in the PR:
Changed the event name to "video.mediaStream.registerForVideoFrame" in the API `video.mediaStream.registerForVideoFrame`.

### Unit Tests added:
Yes

### End-to-end tests added:
No

### Change file added:
Yes

### DIFF
--- a/change/@microsoft-teams-js-ad46a5c7-5cc0-4176-8539-0b126be3e1f5.json
+++ b/change/@microsoft-teams-js-ad46a5c7-5cc0-4176-8539-0b126be3e1f5.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix bug that when calling both video.registerForVideoFrame and video.mediaStream.registerForVideoStream an error happens: \"stream id is already registered\"",
+  "comment": "Updated message sent to host by `video.registerForVideoFrame` to avoid duplication with message sent by `video.mediaStream.registerForVideoStream` (bug would manifest in \"stream id is already registered\" error message)",
   "packageName": "@microsoft/teams-js",
   "email": "wenghe@microsoft.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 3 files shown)


lubobill1990 (dismissed):

packages/teams-js/test/public/video.spec.ts
```diff
@@ -721,6 +721,20 @@ describe('video', () => {
       expect(registeredStreamId).toEqual(streamId);
     });
 
+    it('should send event to parent to inform the registeration', async () => {
```
AE-MS, line 724: >registeration [](http://example.com/codeflow?start=50&length=13)

"registration"

change/@microsoft-teams-js-ad46a5c7-5cc0-4176-8539-0b126be3e1f5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix bug that when calling both video.registerForVideoFrame and video.mediaStream.registerForVideoStream an error happens: \"stream id is already registered\"",
```
AE-MS, line 3: >Fix bug that when calling both video.registerForVideoFrame and video.mediaStream.registerForVideoStream an error happens: \"stream id is already registered\" [](http://example.com/codeflow?start=14&length=157)

Since this goes into the public changelog, we have a particular format (past tense and fixed with symbol names).

I recommend something like this:

Updated message sent to host by `video.registerForVideoFrame` to avoid duplication with message sent by `video.mediaStream.registerForVideoStream` (bug would manifest in \"stream id is already registered\" error message)

AE-MS (changes requested):
:clock1:

change/@microsoft-teams-js-ad46a5c7-5cc0-4176-8539-0b126be3e1f5.json
WengHe: updated as suggested

packages/teams-js/test/public/video.spec.ts
WengHe: Fixed

AE-MS (approved):

erinha (approved):

PR #1643, opened 2023-03-15
Change buttons' name to unblock iOS UI tests

MengyiGong:
## Description
Renamed the `Has Permission` and `Request Permission` buttons for GeoLocation API and BarCode API.

What changes: Renaming button labels (which is the `title` field in the codebase) of `Has Permission` and `Request Permission` the same as the button name.
Why made this change: iOS UI test lib `XCTest` can get the item by label only. If more than one button have the same label name, we could use index for accessing different buttons with same label name but then the UI layers change may break index order, which cause E2E test failure. This is a scenario we hope could be avoided, thus, in iOS side, we would like to edit button label name to make them to be unique and function as what box selector id does.

### Main changes in the PR:

1. Renaming `Has Permission` and `Request Permission` buttons under GeoLocationAPI to `Has GeoLocation Permission` and `Request GeoLocation Permission`
2. Renaming `Has Permission` and `Request Permission` buttons under BarCodeAPI to `Has BarCode Permission` and `Request BarCode Permission`

### Validation performed:

1. Build and run iOS orange app
2. Open test-app in the iOS orange and run the GeoLocation UI test locally: from the local log, iOS UI testing lib find the button by label.
3. Manually tested all the GeoLocation APIs made sure they all worked as expected.

### Unit Tests added:

No button naming covered in the unit tests.

<No>

### End-to-end tests added:
Will add UI tests in the iOS codebase
<No>

## Additional Requirements
No
### Change file added:
Checking for changes against "origin/main"
fetching latest from remotes "origin/main"
Validating no private package among package dependencies
 OK!

No change files are needed

<Yes>

### Next/remaining steps:
Start implementing GeoLocation E2E tests in iOS-hub-sdk

### Screenshots:

| Before     | After      |
| ---------- | ---------- |
|  geolocation-before| geolocation_after|

### DIFF
--- a/apps/teams-test-app/src/components/BarCodeAPIs.tsx
+++ b/apps/teams-test-app/src/components/BarCodeAPIs.tsx
@@ -31,7 +31,7 @@ const ScanBarCode = (): React.ReactElement =>
 const HasBarCodePermission = (): React.ReactElement =>
   ApiWithoutInput({
     name: 'hasBarCodePermission',
-    title: 'Has Permission',
+    title: 'Has BarCode Permission',
     onClick: async () => {
       const result = await barCode.hasPermission();
       return JSON.stringify(result);
@@ -41,7 +41,7 @@ const HasBarCodePermission = (): React.ReactElement =>
 const RequestBarCodePermission = (): React.ReactElement =>
   ApiWithoutInput({
     name: 'requestBarCodePermission',
-    title: 'Request Permission',
+    title: 'Request BarCode Permission',
     onClick: async () => {
       const result = await barCode.requestPermission();
       return JSON.stringify(result);
--- a/apps/teams-test-app/src/components/GeoLocationAPIs.tsx
+++ b/apps/teams-test-app/src/components/GeoLocationAPIs.tsx
@@ -58,7 +58,7 @@ const ShowLocation = (): React.ReactElement =>
 const HasGeoLocationPermission = (): React.ReactElement =>
   ApiWithoutInput({
     name: 'hasGeoLocationPermission',
-    title: 'Has Permission',
+    title: 'Has GeoLocation Permission',
     onClick: async () => {
       const result = await geoLocation.hasPermission();
       return JSON.stringify(result);
@@ -68,7 +68,7 @@ const HasGeoLocationPermission = (): React.ReactElement =>
 const RequestGeoLocationPermission = (): React.ReactElement =>
   ApiWithoutInput({
     name: 'requestGeoLocationPermission',
-    title: 'Request Permission',
+    title: 'Request GeoLocation Permission',
     onClick: async () => {
       const result = await geoLocation.requestPermission();
       return JSON.stringify(result);


KangxuanYe:
`iOS UI test lib XCTest can get the item by label only. If more than one button have the same label name, it will lead iOS UI test always fail.`

For this line, it is actually not 100% true. 

We could use index for accessing different buttons with same label name **but then the UI layers change may break index order**, which cause E2E test failure. This is a scenario we hope could be avoided, thus, in iOS side, we would like to edit button label name to make them to be unique and function as what box selector id does.

KangxuanYe (approved):

apps/teams-test-app/src/components/BarCodeAPIs.tsx
maglims, line 1: We normally follow this instruction to make changes to TeamsJSSDK repo, although I am not 100% sure that the rule applies if the change only is in test app. @AE-MS might know.
https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md

MengyiGong: Thanks for your time, Maggie! I followed the steps in the above link, and I got `No change files are needed` after I run yarn changefile. Just in case I miss anything, I will wait @AE-MS 's approval.

AE-MS:
I'm signing off on this but given the failures in the pipeline runs for these, I'm guessing you have to make changes in the E2E tests to support the new names first.

AE-MS (approved):

jadahiya-MSFT:
As Aaron mentioned, you will need to update the E2E tests and put out a PR before this change will pass on this PR. For my own curiosity, why does the iOS UI Tests look for the display label rather than the object name?

MengyiGong:
> 

Hi @jadahiya-MSFT Thanks for the guiding! 
To answer your question of `why does the iOS UI Tests look for the display label rather than the object name` : The Swift UI test library is constrained by its inability to detect visual elements beyond the displayed text on the screen. @KangxuanYe  Please feel free to correct me if it is inaccurate.

AE-MS:
> > 
> 
> Hi @jadahiya-MSFT Thanks for the guiding! To answer your question of `why does the iOS UI Tests look for the display label rather than the object name` : The Swift UI test library is constrained by its inability to detect visual elements beyond the displayed text on the screen. @KangxuanYe Please feel free to correct me if it is inaccurate.

@jadahiya-MSFT it's reasonable question but I promise you that @KangxuanYe can talk your ear off about why it has to be that way. Maybe schedule some coffee time with him.  😜

MengyiGong:
I tested the E2E tests in the web-hub-sdk, and all API tests passed locally. I would not worries about the E2E tests for Android and iOS hub-sdk since both of the mobile platform have not implemented the E2E tests for the geoLocation and barCode E2E tests yet. I also checked the web-hub-sdk codebase, looks like web-hub-sdk only used the name of the button while the changes in this PR are all about titles. I will try to reach out to someone who familiar with the pipeline and get back to this PR change.

MengyiGong:
> Details

Thanks a lot @AE-MS for taking time review the changes!

PR #1656, opened 2023-03-22
New sub namespace for registerForVideoFrame

WengHe:
## Description
Move video.registerForVideoFrame to a new namespace: video.sharedFrame.registerForVideoFrame

### Unit Tests added:

No, only update a part of the UTs.

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/src/private/videoEx.ts
TrevorJoelHarris, line 1: You'll need to make more changes to videoEx than this in this PR. It's okay with me if you aren't adding the mediaStream function overloads here yet, but you'll need to reflect the changes you're making to add the new sharedFrame namespace. videoEx's isSupported function currently looks like this:

image

The current function in there is mirroring the one in video.sharedFrame. You'll probably want to go in and also add a sharedFrame namespace to videoEx and move videoEx.registerForVideoFrame into it.

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/video.ts
```diff
@@ -226,7 +134,7 @@ export namespace video {
   /**
    * @beta
    * Namespace to get video frames from a media stream.
-   * When the host supports this capability, developer should call {@link mediaStream.registerForVideoFrame} to get the video frames instead of {@link registerForVideoFrame} to get the video frames, callback of {@link registerForVideoFrame} will be ignored when the host supports this capability.
+   * When the host supports this capability, developer should call {@link mediaStream.registerForVideoFrame} to get the video frames instead of {@link sharedFrame.registerForVideoFrame}, callback of {@link sharedFrame.registerForVideoFrame} will be ignored when the host supports this capability.
```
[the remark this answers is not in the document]
WengHe: Revise the comments a bit

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/video.ts
```diff
@@ -134,7 +134,7 @@ export namespace video {
   /**
    * @beta
    * Namespace to get video frames from a media stream.
-   * When the host supports this capability, developer should call {@link mediaStream.registerForVideoFrame} to get the video frames instead of {@link sharedFrame.registerForVideoFrame}, callback of {@link sharedFrame.registerForVideoFrame} will be ignored when the host supports this capability.
+   * When the host supports this capability, developer should call {@link mediaStream.registerForVideoFrame} to get the video frames.
```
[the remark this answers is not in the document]
WengHe: > I know that your plan (currently) is that mediaStream and sharedFrame would never both be supported at the same time. 

Yes

> In case that changes in the future, it may be worth adding a comment to mediaStream saying that if both mediaStream and sharedFrame are supported the developer should use mediaStream.

I can't predict what change will be in the future, so I prefer not to add such comments at the moment.

apps/teams-test-app/src/components/VideoAPIs.tsx
TrevorJoelHarris, line 1: These test app changes look correctly written but incomplete. All functions in the capability should have entry points here if they are callable by app developers. Functions in the top level `video` capability are missing (I think just isSupported after a quick scan), and the new child capabilities (sharedFrame and mediaStream) should have their own `APIs` files

apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
TrevorJoelHarris, line 1: These test app changes look correctly written but incomplete. All functions in the capability should have entry points here if they are callable by app developers. Functions in the top level `videoEx` capability are missing (I think just registerForVideoEffect, notifyFatalError, and isSupported after a quick scan), and the new child capability (sharedFrame) should have their own `APIs` files

TrevorJoelHarris (approved):

shilcare (dismissed):

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

erinha (approved):

PR #1664, opened 2023-03-23
added pnpm commands to force blazor build in pipeline

noahdarveau-MSFT:
Added pnpm commands so Lerna does not build the blazor-test-app when simply running `pnpm build`. Instead, I added a `build-force-blazor` option that will build everything, including the blazor-test-app. I then changed all of the build commands that get run in the pipeline to use the `build-force-blazor` command instead.


TrevorJoelHarris (approved):

shrshindeMSFT (approved):
LGTM! 🚢 it.

PR #1672, opened 2023-03-27
Remove unnecessary documentation on enums properties

AE-MS:
The documentation for some `enum` properties was wrong (out of date) and unnecessary since it was just literally listing the enum values. I have removed said documentation.

### DIFF
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -178,12 +178,12 @@ export namespace app {
    */
   export interface AppHostInfo {
     /**
-     * The name of the host client. Possible values are: Office, Orange, Outlook, Teams
+     * Identifies which host is running your app
      */
     name: HostName;
 
     /**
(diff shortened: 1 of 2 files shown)


packages/teams-js/src/public/app.ts
```diff
@@ -178,12 +178,12 @@ export namespace app {
    */
   export interface AppHostInfo {
     /**
-     * The name of the host client. Possible values are: Office, Orange, Outlook, Teams
+     * Identifies which host is running your app
```
erinha, line 181: Do we use second person pronouns elsewhere in the documentation? Also, I think we've used "application" rather than "app".

erinha (changes requested):
Suggested a small edit.

packages/teams-js/src/public/app.ts
erinha: Discussed offline. This is recommended language. Approving.

erinha (approved):

PR #1673, opened 2023-03-27
Magli/fix ipad os auth

maglims:
## Description

Currently, ipados is not listed in the host client type check for auth.auth flow. With this, the auth flow will open a browser, which has no knowledge of its parent window and flows broke. In this PR, we are adding ipados client type to the host client list, and it will be able to leverage auth.auth flow existing in ios hub sdk and work properly.

### Main changes in the PR:

1. added ipados to auth flow check
2. added ipados to isHostClientMobile check

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -91,13 +91,17 @@ export function isCurrentSDKVersionAtLeast(requiredVersion: string = defaultSDKV
...
  * @internal
  * Limited to Microsoft-internal use
  */
 export function isHostClientMobile(): boolean {
-  return GlobalVars.hostClientType == HostClientType.android || GlobalVars.hostClientType == HostClientType.ios;
+  return (
+    GlobalVars.hostClientType == HostClientType.android ||
+    GlobalVars.hostClientType == HostClientType.ios ||
+    GlobalVars.hostClientType == HostClientType.ipados
+  );
```
maglims, line 104: The bug https://office.visualstudio.com/MetaOS/_workitems/edit/7715609 mentions teamsphones.
I have checked tat in teams repo but not able to find any entries. I'd image it maybe one of depreciated stuff but will double check with Yuri and update the PR accordingly if needed.

TrevorJoelHarris: This change makes sense, but wanted to check: After this change some functions in media.ts will work differently on ipad. After a quick scan, I think it is mainly media.selectMedia. I *think* that it would have thrown an exception here on ipad before your change:

image

Does this function now work on ipad? Is it supposed to? I *assume* it's fine, but wanted to check and see what you thought

TrevorJoelHarris (approved):
Had one question about how this change would affect media.selectMedia

packages/teams-js/src/internal/internalAPIs.ts
maglims: well, it looks media is really a complex api(if not messy).It feels it should be fine mainly because of the line from 606-610 I think.

With adding ipados in isHostClient, it will go to else section in this screenshot
image
If it goes to else, it should be caught earlier from line 606 since they are doing the same thing.

AE-MS:
>         GlobalVars.hostClientType === HostClientType.teamsDisplays

For this bug, doing this fix is the right thing to do. But I have a larger question about whether this list should exist in teamsjs at all. Should the question about whether or not teamsjs is responsible for opening the auth window vs funneling it to the host be something that the host tells teamjs to do, rather than teamsjs decides to do?  E.g., should part of the initialization call, in addition to the runtime, have other "instructions" like "should you try to open the auth window or should you ask me to do it?"

I think in general I am very worried about teamsjs using OS information for anything, except maybe passing it to the app in context. It doesn't feel sustainable.

@TrevorJoelHarris may also have thoughts.

---
Refers to: packages/teams-js/src/public/authentication.ts:144 in 925ec35. [](commit_id = 925ec35f568b028154e87463a8500963e299a42f, deletion_comment = False)

AE-MS (approved):
:shipit:

PR #1700, opened 2023-04-06
Fixed eslint error for ssr test app

noahdarveau-MSFT:
Eslint requires that parserOptions.project property be set in the eslint file.  This wasn't set when I initially checked in the ssr test app, so I'm adding it now to remove the error displayed at build time. 

>I'm not sure why there are so many commits since they've already been checked in.


TrevorJoelHarris (approved):

AE-MS (approved):
:shipit:

PR #1706, opened 2023-04-10
Update yml files to enable CodeQL scanning

TrevorJoelHarris:
## Description

Enabled CodeQL static analysis in this repository, for javascript/typescript and C#. Most of the code is Javascript, but the SSR test app is C#. The "interesting" part of the configuration is here:

image

This says that we will run CodeQL in the following cases:

- whenever we push to a branch that matches one of those 4 regular expressions
- whenever a pull request is prepared or changed for a change going into the main branch
- once a week, saturday at 2am UTC

### Main changes in the PR:

1. Enabled CodeQL workflow for javascript/typescript and C#

### Validation performed:

Ran pipelines and made sure codeQL was running and that issues were reported in github

### Unit Tests added:

No, no code changes that are testable included in this change

### End-to-end tests added:

No

### Change file added:

None needed, change not in code

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,75 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: 'CodeQL'
+
+on:
+  push:
+    branches: ['main', archive/*, main*, release-*, release/*]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: ['main']
+  schedule:
+    - cron: '0 2 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ['csharp', 'javascript']
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Use only 'java' to analyze code written in Java, Kotlin or both
+        # Use only 'javascript' to analyze code written in JavaScript, TypeScript or both
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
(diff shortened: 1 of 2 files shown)


jekloudaMSFT:
Where can we see the results after the scan runs?

jekloudaMSFT (approved):

TrevorJoelHarris:
> Where can we see the results after the scan runs? Asking specifically about the result of the "Uploading results" step:
> 
> ```
> Uploading results
> Processing sarif files: ["/home/runner/work/microsoft-teams-library-js/results/csharp.sarif"]
> Uploading results
> Successfully uploaded results
> ```

If you go into the "Security/Code Scanning" section of our repo and make your query this:

`is:open pr:1706` You can see the 5 alerts that we have after this change. Once I check in that should start showing up in that section by default.

jadahiya-MSFT (approved):

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,75 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: 'CodeQL'
+
+on:
+  push:
+    branches: ['main', archive/*, main*, release-*, release/*]
```
jadahiya-MSFT, line 16: We might not want to run this on release-* and instead only run it on release-1.13.* and up instead, because those have automation in place that we control.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,75 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: ['main']
+  schedule:
+    - cron: '0 2 * * 6'
```
aeunms, line 21: For anyone wondering how the heck to interpret cron other than what Trevor said, I ended up doing some googling to find out: [crontab](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07)

aeunms (approved):
🚢

.github/workflows/codeql.yml
TrevorJoelHarris: That's fair, I've pushed an update that should match 1.13.* and higher

jadahiya-MSFT (approved):

jekloudaMSFT (dismissed):

.github/workflows/codeql.yml
```diff
@@ -13,7 +13,7 @@ name: 'CodeQL'
 
 on:
   push:
-    branches: ['main', archive/*, main*, release-*, release/*]
+    branches: ['main', archive/*, main*, release-1\.(1[3-9]|[2-9]\d|\d{3,})\.+.*, release/*]
```
aeunms, line 16: What do you think about adding a comment here to explain the regex? Personally I have to do some extra work to see what it means, and even if I figured it out, I don't think I would understand why we do 1.13.* and higher until I dug my way to the PR

TrevorJoelHarris: No problem

jekloudaMSFT (approved):

PR #1711, opened 2023-04-17
upgrade image and improve yaml file

KangxuanYe:
## Description

Currently we are facing an issue on internal-macos-11 VM image in Azure pipeline that XCTest runner could not be launched. This issue could be intimidated by upgrading to internal-macos12 VM image.

Meanwhile, this PR remove the bash script dependencies in iOS host SDK repo so that Teams-JS library side will has less chance to be broken by the change in iOS side.

### Main changes in the PR:

1. Upgrade VM image from internal-macos-11 to internal-macos12
2. Bash scripts dependencies are removed.

## Validation

CI pipeline checkin in iOS should not be timeout. It should pass after this PR.

### Validation performed:

1. All CI pipelines check are past.

### DIFF
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -36,10 +36,6 @@ steps:
     displayName: 'Install XCTestHtmlReport for publishing result'
     workingDirectory: '$(System.DefaultWorkingDirectory)'
 
-  - bash: 'sh build_ios_host_sdk.sh'
-    displayName: 'Build iOS Host SDK'
-    workingDirectory: '$(Agent.BuildDirectory)/iOSHost/jobs/ciDevTools'
-
   - task: NodeTool@0
     inputs:
       versionSpec: '14.x'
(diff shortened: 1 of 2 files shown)


tools/yaml-templates/ios-test.yml
```diff
@@ -77,17 +71,43 @@ steps:
...
     workingDirectory: '$(ClientSdkProjectDirectory)'
 
-  - bash: 'sh disable_slow_animation.sh'
+  - bash: cd ~/Library/Preferences
+      defaults write com.apple.iphonesimulator SlowMotionAnimation -bool NO
     displayName: 'Disable Slow Animation'
-    workingDirectory: '$(Agent.BuildDirectory)/iOSHost/jobs/ciDevTools'
+    workingDirectory: '$(System.DefaultWorkingDirectory)'
 
-  - bash: 'sh parallel_ui_testing_for_publishing.sh'
```
maglims, line 84: Are these two sh files being used by ios hub sdk? If they are used by two places, one potential issue is for any fixes/improvement done in the script, teamsjssdk won't be able to pick it up and we have to remember to update here manually

maglims (approved):

tools/yaml-templates/ios-test.yml
KangxuanYe: I separated them from Teams-JS library and iOS host last time. Also, to reduce dependency, I am trying to get rid of these scripts so that we won't accidentally edit it and then break the pipelines in Teams-JS library side.

KangxuanYe:
PR closed as it is merged in Shreyas PR for fixing

PR #1728, opened 2023-05-03
Fixing CodeQL issues

TrevorJoelHarris:
## Description

CodeQL flagged a few minor issues in our codebase. This change fixes them.

### Main changes in the PR:

1. Updated gitCommands.ts to use execFileSync instead of execFile based on CodeQL's guidance. Added unit tests for the first time to the tools folder to test those functions. Slightly modified them to throw errors if obviously incorrect input is passed in.
2. Updated prerelease.js to use escaped . characters (\.) instead of just . characters because they were being used in regular expressions which made them more permissive than the author intended

### Validation performed:

1. Wrote new unit tests, made sure all unit tests passed
2. Did some quick validation that calling execFileSync as shown in this change will still execute the same command as before

### Unit Tests added:

Yes!

### End-to-end tests added:

No!

### Change file added:

No, all changes in tooling, test, and setup files


tools/cli/preRelease.js
```diff
@@ -60,11 +60,11 @@ const updateVersionAndIntegrity = async (absolutePath, version, integrityHash) =
   const result = readme
     .replace(/integrity=\".*?\"/, `integrity="${integrityHash}"`)
     .replace(
-      /res.cdn.office.net\/teams-js\/.*\/js\/MicrosoftTeams.min.js/g,
+      /res\.cdn\.office.net\/teams-js\/.*\/js\/MicrosoftTeams\.min\.js/g,
```
jadahiya-MSFT, line 63: I will need to test this out to make sure it doesn't alter artifact naming.

tools/bundle-size-tools/src/utilities/gitCommands.ts
```diff
@@ -3,19 +3,25 @@
  * Licensed under the MIT License.
  */
 
-import { execSync } from 'child_process';
+import { execFileSync, execSync } from 'child_process';
```
jadahiya-MSFT, line 6: I think we can remove `execSync`, since it looks like it is no longer being used in this file

jadahiya-MSFT (approved):

tools/cli/preRelease.js
jadahiya-MSFT: This works as intended.

tools/bundle-size-tools/src/utilities/gitCommands.ts
AE-MS: I don't want to make you take this too far right now but in addition to agreeing with jadahiya, if we had linting on this file I think this would have been caught automatically? Or maybe prettier?

Your call whether that's really worth it.

TrevorJoelHarris: Alright, I figured out why the linter wasn't working (the eslintrc file was so old that it didn't do anything). I basically copied our existing rules from packages and then fixed all the easy looking problems and ignored lines with ones I couldn't quickly address. Doing this caught the unused import

AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

PR #1743, opened 2023-05-08
Adding secondaryBrowser capability to allow M365 Apps enable browsing experience

anmaur:

## Description

Enabling M365 Apps with browsing experience using a cross-platform API

### Main changes in the PR:

1. Adding `secondaryBrowser` capability and its `open` API
2. Test App changes to enable testing of this capability and new `open` API functionality
3. Unit tests

### Validation performed:

1. Unit tests added to validate the expected behavior from teamsJs side
2. Validated with the sample host application

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:


jekloudaMSFT (approved):

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+   * Pls note that any error that happens after starting navigation is delegated on external browsers to handle in their generic way
+   * @beta
+   */
+  export function open(url: URL): Promise<boolean> {
+    ensureInitialized(runtime, FrameContexts.content);
+    if (!isSupported() || !isHostClientMobile()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    if (!url || !isValidHttpsURL(url)) {
```
maglims, line 31: nit: maybe this check can be inside isValidHttpsURL?

maglims: just to be more clear, I meant to say !url can be inside the function isValidHttpsURL :)

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
+import { sendAndHandleSdkError as sendAndHandleError } from '../internal/communication';
+import { ensureInitialized, isHostClientMobile } from '../internal/internalAPIs';
+import { isValidHttpsURL } from '../internal/utils';
+import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
+import { ErrorCode } from './interfaces';
+import { runtime } from './runtime';
+
+/**
+ * Namespace to power up the in-app browser experiences in the Host App.
+ * For eg. opening a URL in the Host App inside a browser
```
AE-MS, line 10:
```suggestion
 * For e.g., opening a URL in the Host App inside a browser
```

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @beta
+ */
+export namespace secondaryBrowser {
+  /**
+   * Open a URL in the secondary browser aka in-app browser
+   *
+   * @param url Url to open in the browser
+   * @returns Promise that resolve to true if the URL successfully opens in the secondaryBrowser
```
AE-MS, line 19:
```suggestion
   * @returns Promise that resolves to true if the URL successfully opens in the secondaryBrowser
```

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+   * or throws an error {@link SdkError} incase of failure before starting navigation
+   *
+   * Pls note that any error that happens after starting navigation is delegated on external browsers to handle in their generic way
```
AE-MS, line 22:
```suggestion
   * @remarks Any error that happens after navigation begins is handled by the platform browser component and not returned from this function.
```

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @beta
+   */
+  export function open(url: URL): Promise<boolean> {
+    ensureInitialized(runtime, FrameContexts.content);
+    if (!isSupported() || !isHostClientMobile()) {
```
AE-MS, line 27: Is checking `isHostClientMobile` needed here? Shouldn't the presence or absence of the capability in the runtime be sufficient to just use `isSupported`?

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw errorNotSupportedOnPlatform;
+    }
+
+    if (!url || !isValidHttpsURL(url)) {
+      throw { errorCode: ErrorCode.INVALID_ARGUMENTS, message: 'Invalid Url: Only https URL is allowed' };
+    }
+    return new Promise<boolean>((resolve) => {
```
AE-MS, line 34: Is it necessary to wrap and resolve the promise returned from `sendAndHandleError`? Can you just do this:
```
return sendAndHandleError('secondaryBrowser.open', url.toString())
```

AE-MS (changes requested):
A few suggestions; I've only looked at a bit so far; will look more later!

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+   * Open a URL in the secondary browser aka in-app browser
+   *
+   * @param url Url to open in the browser
+   * @returns Promise that resolve to true if the URL successfully opens in the secondaryBrowser
+   * or throws an error {@link SdkError} incase of failure before starting navigation
+   *
+   * Pls note that any error that happens after starting navigation is delegated on external browsers to handle in their generic way
+   * @beta
+   */
+  export function open(url: URL): Promise<boolean> {
```
AE-MS, line 25: Is there a reason this needs to be a `Promise<boolean>` rather than just `Promise<void>`? It seems like success or failure is communicated by whether the promise is resolved or rejected; is additionally having a `boolean` value duplicative?

change/@microsoft-teams-js-623bd4b5-7341-4b40-8392-62413766d29a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added secondaryBrowser capability and its open API to enable browsing experience for M365 Apps",
```
AE-MS, line 3: >secondaryBrowser [](http://example.com/codeflow?start=20&length=16)

Please use backticks around this

AE-MS, line 3: >M365 Apps [](http://example.com/codeflow?start=99&length=9)

Just use "apps" for this, since "M365 apps" isn't a brand we use.

packages/teams-js/src/public/secondaryBrowser.ts
anmaur: In my opinion for good Developer's experience its always helpful to provide an acknowledgement of the request if its completed successfully or failed. Promise<void> would not allow to convey the success side and only the failure side of the story. Let me know if that answers your question.

AE-MS: I agree that it's necessary to indicate whether the request succeeded or failed, but I'm not sure I understand the distinction you are making. Isn't the point of either resolving or rejecting a promise to indicate success or failure? If you also include a boolean then you're indicating creating a matrix of return possibilities that seems unclear to developers. For example, what does it mean if the promise is resolved but the boolean is false? Is that a success or a failure? Or both?

maglims (approved):

packages/teams-js/src/public/secondaryBrowser.ts
anmaur: Discussed offline on this and sent an update.

AE-MS (approved):

PR #1759, opened 2023-05-22
E2E Testing Json Data Migration

KangxuanYe:
## Description

This PR is to add testing json data under Teams-JS library repo so that data could be shared to each host SDK team. Teams could use the data to drive/build their own E2E test and the data in Teams-JS library side have access for the single control and change on E2E test for each host. If anyone want to do change on any case, the data will now make sure the change takes effect on other hosts at the same time, which reduce the cost of time and labor and make across-platform developing easier. 

backward-compatible E2E testing data is extended by adding two fields, `platforms` for capability-wide control and `platformsExcluded` for case-wide control. For example, `platforms: '*'` means testing cases under this capability should be supported by all platforms (web, iOS and Android in the future). And `platforms: 'Web'` means testing cases under this capability are only be supported in Web Host SDK side. `platformsExcluded` field provides the ability to exclude a case for a specific platform, i.e. `platformsExcluded: 'iOS'` will exclude corresponding test case for iOS platform (case will not be built in iOS side).  

### Main changes in the PR:

1. Testing json data files
2. Json data schema

### Validation performed:

1. You need to have permission and access to each host repo to check the migrating PR and see if it works.

### DIFF
--- a/apps/teams-test-app/e2e-test-data/authentication.json
+++ b/apps/teams-test-app/e2e-test-data/authentication.json
@@ -1,0 +1,22 @@
+{
+  "name": "Authentication",
+  "platforms": "*",
+  "testCases": [
+    {
+      "title": "getUser API Call - Success",
+      "version": "1.x || >2.0.0-beta.2",
+      "type": "callResponse",
+      "platformsExcluded": ["iOS"],
+      "boxSelector": "#box_getUser",
+      "expectedTestAppValue": "{\"oid\":\"mockoid\",\"tid\":\"mocktid\",\"upn\":\"mockupn\",\"loginHint\":\"mockLoginHint\",\"displayName\":\"mockName\",\"dataResidency\":\"public\"}"
+    },
+    {
+      "title": "getUser API Call - Success",
+      "version": "2.0.0-beta.2",
+      "type": "callResponse",
+      "platformsExcluded": ["iOS"],
+      "boxSelector": "#box_getUser",
+      "expectedTestAppValue": "Success: {\"oid\":\"mockoid\",\"tid\":\"mocktid\",\"upn\":\"mockupn\",\"loginHint\":\"mockLoginHint\",\"displayName\":\"mockName\",\"dataResidency\":\"public\"}"
+    }
+  ]
+}
--- a/apps/teams-test-app/e2e-test-data/chat.json
+++ b/apps/teams-test-app/e2e-test-data/chat.json
@@ -1,0 +1,58 @@
+{
+  "name": "Chat",
+  "version": ">2.0.0-beta.0",
+  "platforms": "*",
+  "checkIsSupported": {
+    "version": ">2.0.0-beta.3"
+  },
+  "testCases": [
+    {
+      "title": "openChat API Call - Success",
+      "type": "callResponse",
+      "version": ">2.0.0-beta.3",
+      "boxSelector": "#box_openChat",
+      "inputValue": {
+        "user": ["testUpn"],
+        "message": "testMessage"
+      },
+      "expectedAlertValue": "openChat called with {\"members\":[\"testUpn\"],\"message\":\"testMessage\"}",
+      "expectedTestAppValue": "chat.openChat() was called, but there was no response from the Host SDK."
+    },
+    {
+      "title": "openGroupChat API Call - Success",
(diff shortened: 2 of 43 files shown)


maglims (dismissed):

apps/teams-test-app/e2e-test-data/dialog.json
shrshindeMSFT, line 1: `dialog.json` has another file called `dialog.ts` which has test cases as well. Don't we need to add that file as well?

KangxuanYe: The backward compatible E2E testing system in web side will run those data-driven test first and then come to the tests specifically written in that file. Thus, we don't need to add it as they are viewed as platform-specific test cases which cannot be driven by data.

jadahiya-MSFT (approved):

apps/teams-test-app/e2e-test-data/chat.json
```diff
@@ -0,0 +1,58 @@
...
+      "boxSelector": "#box_openGroupChat2",
+      "inputValue": {
+        "users": ["testUpn1", "testUpn2"],
+        "message": "testMessage"
+      },
+      "expectedAlertValue": "openChat called with {\"members\":[\"testUpn1\",\"testUpn2\"],\"message\":\"testMessage\"}",
+      "expectedTestAppValue": "chat.openGroupChat() was called"
+    }
+  ]
+}
```
erinha, line 58: The source version of the file that has been copied here has 135 lines in the file, whereas this only has 58. I think maybe it needs to be updated?

erinha: Ah I see, they're in conversation.json now.

erinha (approved):
Looks good! Thanks for making this change!

apps/teams-test-app/e2e-test-data/conversation.json
```diff
@@ -0,0 +1,124 @@
+{
+  "name": "Conversation",
+  "platforms": "*",
+  "testCases": [
+    {
```
erinha, line 5: Should there be `isSupported` test cases in conversations.json?

erinha (changes requested):
Request to add `isSupported` tests to conversations.json (and to the test app to match the UI)

apps/teams-test-app/e2e-test-data/conversation.json
erinha: Discussed offline. We'll track that separately.

erinha (approved):

erinha (approved):
Looks good to me, thank you!

apps/teams-test-app/e2e-test-data/authentication.json
```diff
@@ -0,0 +1,22 @@
...
+      "type": "callResponse",
+      "platformsExcluded": ["iOS"],
+      "boxSelector": "#box_getUser",
+      "expectedTestAppValue": "{\"oid\":\"mockoid\",\"tid\":\"mocktid\",\"upn\":\"mockupn\",\"loginHint\":\"mockLoginHint\",\"displayName\":\"mockName\",\"dataResidency\":\"public\"}"
+    },
+    {
+      "title": "getUser API Call - Success",
+      "version": "2.0.0-beta.2",
+      "type": "callResponse",
+      "platformsExcluded": ["iOS"],
```
shrshindeMSFT, line 17: Why there is a need to include platformExcluded ? Can't platforms field handle it ? Just curious is auth is not supported on iOS?

jadahiya-MSFT (approved):

apps/teams-test-app/e2e-test-data/authentication.json
KangxuanYe: there is a one for capability-wide control and what you labelled here is for case-wide control. Saying maybe for a capability there are 20 cases and each of them is driven by corresponding data. But there might be one or two cannot be used for iOS/Android platform. Thus, we need another field in each case to precisely exclude it for a specific platform.

Auth is supported in iOS. Same as implemented in web sdk, most of auth capability E2E test cases are implemented directly on the platform. And the data of getUser case here is excluded because the user info provided from web is different from the one in iOS and web side might need to change to get aligned with that in iOS so I excluded it temporarily.

PR #1761, opened 2023-05-22
Fixing video import/export naming

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Removing `export type = other type` convention from `videoEx.ts`. This is because exporting a type that was imported as {insert name here} can not properly generate the type when making a .d.ts file. Replacing it with proper convention to export type as {GENERIC_TYPE_NAME} and to import as {LOCAL_TYPE_NAME} without exporting the local version.


TrevorJoelHarris (dismissed):

erinha:
Thank you for fixing this so quickly! It would be helpful to update the description with a synopsis of the problem that is occurring and the reason this change fixes it, especially since I think this problem (or a version of it) has occurred before. 

Also I think you'll need a change file.

erinha:
> Thank you for fixing this so quickly! It would be helpful to update the description with a synopsis of the problem that is occurring and the reason this change fixes it, especially since I think this problem (or a version of it) has occurred before.

Lol thanks for updating so quickly!

erinha (approved):
🚀

TrevorJoelHarris (approved):

TrevorJoelHarris (approved):

jadahiya-MSFT:
@nandin-borjigin @WengHe as an FYI on this fix

erinha (approved):

PR #1767, opened 2023-05-23
U/kennethba/208803 add new function to api

KennethBarbozaC:

## Description

> Add a new function to the search service API to close and clear the search box from a hub

### Main changes in the PR:

1. Add closeSearch function to the search.ts file, alternative name for the funtion is clearSearch

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

> Simple addition of function Unit Test not required.

<Yes/No>

### End-to-end tests added:

No

### Change file added:

Yes


change/@microsoft-teams-js-fdf3465b-f6f4-4631-b88c-9bcf27caf0ec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added closeSearch function",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Added `closeSearch` to `search` capability",
```

change/@microsoft-teams-js-fdf3465b-f6f4-4631-b88c-9bcf27caf0ec.json
```diff
@@ -0,0 +1,7 @@
...
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
```
TrevorJoelHarris, line 5: your github email address should show up here

packages/teams-js/src/public/search.ts
```diff
@@ -127,4 +128,14 @@ export namespace search {
...
     return ensureInitialized(runtime) && runtime.supports.search ? true : false;
   }
+
+  /**
+   * Clear the host M365 application's search box
+   *
+   * @beta
+   */
+  export function closeSearch(): void {
+    ensureInitialized(runtime);
```
TrevorJoelHarris, line 138: The other functions in this capability (registerHandlers, unregisterHandlers) both say that they have to be called in the FrameContexts.content context (see those function's ensureInitialized calls). That means that those functions will not work if called outside the context of a full screen app. Why is closeSearch able to be called from any FrameContext, shouldn't it be the same as the other functions?

TrevorJoelHarris:
We didn't do a good job of enforcing this originally, but please add unit tests for the new closeSearch function. To do so, you'll need to add a search.spec.ts file and add unit tests to fully test what the function is doing here in the teams-js layer.

TrevorJoelHarris:
This change should also include an update to the teams-test-app to add the ability to test search.closeSearch (the test app will be used by the host sdk code to run end-to-end tests against your new function). See apps\teams-test-app\src\components\SearchAPI.tsx for the file you'll need to update. There are instructions on how to run the teams test app [here](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/apps/teams-test-app/README.md)

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/search.ts
```diff
@@ -127,4 +128,14 @@ export namespace search {
   export function isSupported(): boolean {
     return ensureInitialized(runtime) && runtime.supports.search ? true : false;
   }
+
+  /**
+   * Clear the host M365 application's search box
+   *
+   * @beta
+   */
+  export function closeSearch(): void {
```
SidGulatiMsft, line 137: Have we finalized on the name? Is it going to be closeSearch or clearSearch?

KennethBarbozaC: I do have to agree with Trevor's comment on the matter I think we first need to define exactly what we want with the method if it is just to clear the text that appears on the search box or to exit completely from the experience

packages/teams-js/src/public/search.ts
```diff
@@ -1,8 +1,20 @@
-import { sendMessageToParent } from '../internal/communication';
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
TrevorJoelHarris, line 1: You shouldn't need to disable linting rules for this change. Why was this necessary?

packages/teams-js/src/public/search.ts
```diff
@@ -1,8 +1,20 @@
...
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import {
+  Communication,
+  sendAndHandleStatusAndReason as sendAndHandleError,
```
TrevorJoelHarris, line 5: I know we do this import _bla_ as sendAndHandlerError in other places in the code, but it would be good to just import sendAndHandleStatusAndReason directly instead of aliasing it. Aliasing it confuses other developers and helps disguise bugs.

packages/teams-js/src/public/search.ts
```diff
@@ -17,6 +29,12 @@ export namespace search {
   const onClosedHandlerName = 'search.queryClose';
   const onExecutedHandlerName = 'search.queryExecute';
 
+  /** Search Function Messages */
```
TrevorJoelHarris, line 32: This doesn't appear to be used anywhere, and exporting something like this makes it callable by external developers and creates a documentation page for it, which probably isn't what you want. Generally, we don't create a const holder like this, but I'm not *necessarily* opposed. I would suggest that you delete this, but if you want to keep it please do the following:

1) Use the values where you are passing messages
2) Add a @hidden comment above this const so no documentation is generated
3) Add *all* of the search messages that go up to the host to this const instead of just one.

TrevorJoelHarris: (Like I said though, if I were you I would just move the const string up to line 31 to be with the rest of the consts)

packages/teams-js/src/public/search.ts
```diff
@@ -127,4 +145,74 @@ export namespace search {
...
+   * @beta
+   */
+  export function closeSearch(): Promise<void> {
+    return new Promise<void>((resolve) => {
+      ensureInitialized(runtime, FrameContexts.content);
+      if (!isSupported()) {
+        throw new Error('Not supported');
+      }
+
+      resolve(sendAndHandleError('search.closeSearch', version));
```
TrevorJoelHarris, line 161: Why do you need to pass the package version up to the host in order to close search?

packages/teams-js/src/public/search.ts
```diff
@@ -127,4 +145,74 @@ export namespace search {
...
+  }
+
+  /**
+   * @hidden
+   * Undocumented function used to set a mock window for unit tests
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export function _initialize(hostWindow: any): void {
```
TrevorJoelHarris, line 172: This doesn't seem like a function that should live in the search capability. I assume you're only calling it from your unit tests, in which case you should probably move it there. There's nothing search specific about this and putting it here will add to the package size for external developers.

packages/teams-js/src/public/search.ts
```diff
@@ -127,4 +145,74 @@ export namespace search {
...
+  }
+
+  /**
+   * @hidden
+   * Undocumented function used to clear state between unit tests
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export function _uninitialize(): void {
```
TrevorJoelHarris, line 183: Same with this: unit test only functions should live in unit test files and not in the public exported capabilities

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/search.ts
KennethBarbozaC: I see what happened it was my mistake I used the search initialize instead of the app one, will correct it now.

packages/teams-js/src/public/search.ts
KennethBarbozaC: Sound good will change it

packages/teams-js/src/public/search.ts
KennethBarbozaC: That was for a previous version for test it is unused now will delete it

packages/teams-js/src/public/search.ts
KennethBarbozaC: That was actually one of the questions we had for you if it was necessary, if you also see it as something not needed, I will delete it too

packages/teams-js/src/public/search.ts
KennethBarbozaC: Yeah it was also for an old version of tests will delete it

packages/teams-js/src/public/search.ts
KennethBarbozaC: This one also

packages/teams-js/test/public/search.spec.ts
```diff
@@ -0,0 +1,264 @@
+import { errorLibraryNotInitialized } from '../../src/internal/constants';
+import { GlobalVars } from '../../src/internal/globalVars';
+import { app } from '../../src/public/app';
+import { FrameContexts } from '../../src/public/constants';
+import { M365ContentAction } from '../../src/public/interfaces';
+import { _minRuntimeConfigToUninitialize } from '../../src/public/runtime';
+import { search } from '../../src/public/search';
+import { Utils } from '../utils';
+
+/* eslint-disable */
```
TrevorJoelHarris, line 10: Please enable linting in new unit test files

packages/teams-js/test/public/search.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+/* As part of enabling eslint on test files, we need to disable eslint checking on the specific files with
+   large numbers of errors. Then, over time, we can fix the errors and reenable eslint on a per file basis. */
+
+const dataError = 'Something went wrong...';
+
+/**
+ * Type guard to determine if an action item is of M365Content Type
+ */
+function isM365ContentType(actionItem: unknown): actionItem is M365ContentAction {
```
TrevorJoelHarris, line 19: This appears unused, why is it here?

packages/teams-js/test/public/search.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+  // eslint-disable-next-line no-prototype-builtins
+  return actionItem && Object.prototype.hasOwnProperty.call(actionItem, 'secondaryId');
+}
+
+describe('Testing search capability', () => {
+  const mockErrorMessage = 'Something went wrong...';
```
TrevorJoelHarris, line 25: This const is the same value as the dataError const above. Probably you only need one?

TrevorJoelHarris (changes requested):

SidGulatiMsft (approved):

PR #1793, opened 2023-06-12
Support mediaStream with metadata in videoEx module

shilcare:
## Description

support mediaStream with metadata in videoEx module.

### Main changes in the PR:

1. add optional flag metadataInTexture in mediaStreamInfo to indicate the texture contains metadata or not, metadata contains only audioInferenceResult for now, but can be extended to support new types of metadata in the future
2. parse texture to extract metadata and real video frame when metadata is in texture.
3. register message handler for an internal message for whether we should pass the audioInferenceResult to video app.
4. add effectParameter in the videoEffectReadiness message

### Validation performed:

Used the new API in test app and verified it works fine on both T1/T2.1

### Unit Tests added:

Yes

### End-to-end tests added:

Yes

### Change file added:

Yes

### Related PRs:

#1750

### DIFF
--- a/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
@@ -80,26 +80,44 @@ const CheckIsSupported = (): React.ReactElement =>
     },
   });
 
-const RegisterForVideoFrame = (): React.ReactElement =>
+const MediaStreamRegisterForVideoFrame = (): React.ReactElement =>
   ApiWithoutInput({
-    name: 'videoExSharedFrameRegisterForVideoFrame',
-    title: 'registerForVideoFrame',
+    name: 'videoExMediaStreamRegisterForVideoFrame',
+    title: 'medisStream - registerForVideoFrame',
     onClick: async (setResult) => {
-      const onFrameCallback: videoEx.VideoBufferHandler = async () => {
-        setResult('video frame received');
-      };
       try {
-        const audioInferenceModel = new ArrayBuffer(8);
-        const view = new Uint8Array(audioInferenceModel);
-        for (let i = 0; i < view.length; i++) {
-          view[i] = i;
-        }
         videoEx.registerForVideoFrame({
-          videoBufferHandler: onFrameCallback,
+          videoFrameHandler: async (frame) => {
+            setResult('video frame received');
+            return frame.videoFrame;
+          },
+          videoBufferHandler: (buffer) => buffer,
           config: {
             format: video.VideoFrameFormat.NV12,
-            requireCameraStream: false,
-            audioInferenceModel,
+          },
+        });
+      } catch (error) {
+        return `Faild to register for video frame: ${JSON.stringify(error)}`;
+      }
+      return generateRegistrationMsg('it is invoked on video frame received');
+    },
+  });
+
+const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'videoExSharedFrameRegisterForVideoFrame',
+    title: 'sharedFrame - registerForVideoFrame',
+    onClick: async (setResult) => {
+      try {
+        videoEx.registerForVideoFrame({
+          videoFrameHandler: async (frame) => {
+            return frame.videoFrame;
+          },
+          videoBufferHandler: () => {
+            setResult('video frame received');
+          },
+          config: {
+            format: video.VideoFrameFormat.NV12,
           },
         });
       } catch (error) {
@@ -114,7 +132,8 @@ const VideoExAPIs = (): React.ReactElement => (
     <UpdatePersonalizedEffects />
     <NotifySelectedVideoEffectChanged />
     <RegisterForVideoEffect />
-    <RegisterForVideoFrame />
+    <MediaStreamRegisterForVideoFrame />
+    <SharedFrameRegisterForVideoFrame />
     <NotifyFatalError />
     <CheckIsSupported />
   </ModuleWrapper>
--- a/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
@@ -110,13 +110,14 @@ const MediaStreamRegisterForVideoFrame = (): React.ReactElement =>
           },
         });
       } catch (error) {
-        return `Faild to register for video frame: ${JSON.stringify(error)}`;
+        return `Failed to register for video frame: ${JSON.stringify(error)}`;
       }
       return generateRegistrationMsg('it is invoked on video frame received');
     },
   });
 
-const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
+// To be removed, use the one below with typo fix
+const SharedFrameRegisterForVideoFrameToBeRemoved = (): React.ReactElement =>
   ApiWithoutInput({
     name: 'videoExSharedFrameRegisterForVideoFrame',
     title: 'sharedFrame - registerForVideoFrame',
@@ -147,12 +148,44 @@ const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
     },
   });
 
+const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'videoExSharedFrameRegisterForVideoFrame-1',
+    title: 'sharedFrame - registerForVideoFrame',
+    onClick: async (setResult) => {
+      try {
+        const audioInferenceModel = new ArrayBuffer(8);
+        const view = new Uint8Array(audioInferenceModel);
+        for (let i = 0; i < view.length; i++) {
+          view[i] = i;
+        }
+        videoEx.registerForVideoFrame({
+          videoFrameHandler: async (frame) => {
+            return frame.videoFrame;
+          },
+          videoBufferHandler: () => {
+            setResult('video frame received');
+          },
+          config: {
+            format: video.VideoFrameFormat.NV12,
+            requireCameraStream: false,
+            audioInferenceModel,
+          },
+        });
+      } catch (error) {
+        return `Failed to register for video frame: ${JSON.stringify(error)}`;
+      }
+      return generateRegistrationMsg('it is invoked on video frame received');
+    },
+  });
+
 const VideoExAPIs = (): React.ReactElement => (
   <ModuleWrapper title="VideoEx">
     <UpdatePersonalizedEffects />
     <NotifySelectedVideoEffectChanged />
     <RegisterForVideoEffect />
     <MediaStreamRegisterForVideoFrame />
+    <SharedFrameRegisterForVideoFrameToBeRemoved />
     <SharedFrameRegisterForVideoFrame />
     <NotifyFatalError />
     <CheckIsSupported />
--- a/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
@@ -45,7 +45,7 @@ const RegisterForVideoEffect = (): React.ReactElement =>
     name: 'videoExRegisterForVideoEffect',
     title: 'videoEx - registerForVideoEffect',
     onClick: async (setResult) => {
-      const onVideoEffectChanged = (effectId: string | undefined, effectParam?: string): Promise<void> => {
+      const onVideoEffectChanged = async (effectId: string | undefined, effectParam?: string): Promise<void> => {
         if (effectId === 'anInvalidEffectId') {
           setResult(`failed to change effect to ${JSON.stringify(effectId)}, param: ${JSON.stringify(effectParam)}`);
           throw video.EffectFailureReason.InvalidEffectId;
@@ -54,7 +54,6 @@ const RegisterForVideoEffect = (): React.ReactElement =>
             `video effect changed to ${JSON.stringify(effectId)}, effect param: ${JSON.stringify(effectParam)}`,
           );
         }
-        return Promise.resolve();
       };
       videoEx.registerForVideoEffect(onVideoEffectChanged);
       return generateRegistrationMsg('it is invoked on video effect changed');
--- a/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
@@ -86,6 +86,11 @@ const MediaStreamRegisterForVideoFrame = (): React.ReactElement =>
     title: 'medisStream - registerForVideoFrame',
     onClick: async (setResult) => {
       try {
+        const audioInferenceModel = new ArrayBuffer(8);
+        const view = new Uint8Array(audioInferenceModel);
+        for (let i = 0; i < view.length; i++) {
+          view[i] = i;
+        }
         videoEx.registerForVideoFrame({
           videoFrameHandler: async (frame) => {
             setResult('video frame received');
@@ -94,6 +99,8 @@ const MediaStreamRegisterForVideoFrame = (): React.ReactElement =>
           videoBufferHandler: (buffer) => buffer,
           config: {
             format: video.VideoFrameFormat.NV12,
+            requireCameraStream: false,
+            audioInferenceModel,
           },
         });
       } catch (error) {
@@ -109,6 +116,11 @@ const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
     title: 'sharedFrame - registerForVideoFrame',
     onClick: async (setResult) => {
       try {
+        const audioInferenceModel = new ArrayBuffer(8);
+        const view = new Uint8Array(audioInferenceModel);
+        for (let i = 0; i < view.length; i++) {
+          view[i] = i;
+        }
         videoEx.registerForVideoFrame({
           videoFrameHandler: async (frame) => {
             return frame.videoFrame;
@@ -118,6 +130,8 @@ const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
           },
           config: {
             format: video.VideoFrameFormat.NV12,
+            requireCameraStream: false,
+            audioInferenceModel,
           },
         });
       } catch (error) {
--- a/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
@@ -150,7 +150,7 @@ const SharedFrameRegisterForVideoFrameToBeRemoved = (): React.ReactElement =>
 
 const SharedFrameRegisterForVideoFrame = (): React.ReactElement =>
   ApiWithoutInput({
-    name: 'videoExSharedFrameRegisterForVideoFrame-1',
+    name: 'videoExSharedFrameRegisterForVideoFrame1',
     title: 'sharedFrame - registerForVideoFrame',
     onClick: async (setResult) => {
       try {
--- a/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
@@ -46,7 +46,14 @@ const RegisterForVideoEffect = (): React.ReactElement =>
     title: 'videoEx - registerForVideoEffect',
     onClick: async (setResult) => {
       const onVideoEffectChanged = (effectId: string | undefined, effectParam?: string): Promise<void> => {
-        setResult(`video effect changed to ${JSON.stringify(effectId)}, effect param: ${JSON.stringify(effectParam)}`);
+        if (effectId === 'anInvalidEffectId') {
+          setResult(`failed to change effect to ${JSON.stringify(effectId)}, param: ${JSON.stringify(effectParam)}`);
+          throw video.EffectFailureReason.InvalidEffectId;
+        } else {
+          setResult(
+            `video effect changed to ${JSON.stringify(effectId)}, effect param: ${JSON.stringify(effectParam)}`,
+          );
+        }
         return Promise.resolve();
       };
       videoEx.registerForVideoEffect(onVideoEffectChanged);
--- a/packages/teams-js/src/internal/videoUtils.ts
+++ b/packages/teams-js/src/internal/videoUtils.ts
@@ -3,10 +3,22 @@ import { videoEx } from '../private/videoEx';
 import { errorNotSupportedOnPlatform } from '../public/constants';
 import { video } from '../public/video';
 import { sendMessageToParent } from './communication';
-import { AllowSharedBufferSource, VideoFrameBufferInit, VideoFrameInit } from './VideoFrameTypes';
+import {
+  AllowSharedBufferSource,
+  PlaneLayout,
+  VideoFrameBufferInit,
+  VideoFrameCopyToOptions,
+  VideoFrameInit,
+  VideoPixelFormat,
+} from './VideoFrameTypes';
 
 interface VideoFrame {
+  readonly codedWidth: number;
+  readonly codedHeight: number;
+  readonly format: VideoPixelFormat | null;
+  readonly timestamp: number;
   close(): void;
+  copyTo(destination: AllowSharedBufferSource, options?: VideoFrameCopyToOptions): Promise<PlaneLayout[]>;
 }
 
 /**
@@ -153,7 +165,7 @@ function createProcessedStreamGeneratorWithMetadata(
           const timestamp = originalFrame.timestamp;
           if (timestamp !== null) {
             try {
-              const { videoFrame, metadata: { audioInferenceResult } = {} } = extractVideoFrameAndMetadata(
+              const { videoFrame, metadata: { audioInferenceResult } = {} } = await extractVideoFrameAndMetadata(
                 originalFrame,
                 notifyError,
               );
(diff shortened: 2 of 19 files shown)


packages/teams-js/webpack.config.js
shilcare, line 1: For dev and debugging. will remove this change when ready to merge.

packages/teams-js/tsconfig.json
shilcare, line 1: Update for debugging. will remove changes to this file when ready to merge.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -108,6 +141,142 @@ function createProcessedStreamGenerator(
   return generator;
 }
 
+/**
+ * The function to create a processed video track from the original video track.
+ * It reads frames from the video track and pipes them to the video frame callback to process the frames.
+ * The processed frames are then enqueued to the generator.
+ * The generator can be registered back to the media stream so that the host can get the processed frames.
+ */
+function createProcessedStreamGeneratorWithMetadata(
```
WengHe, line 150: There are some repeated code in this function and `createProcessedStreamGenerator`, try to make them reusable.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -125,12 +294,12 @@ export function createEffectParameterChangeCallback(callback: VideoEffectCallBac
   return (effectId: string | undefined, effectParam?: string): void => {
     callback(effectId, effectParam)
       .then(() => {
-        sendMessageToParent('video.videoEffectReadiness', [true, effectId]);
+        sendMessageToParent('video.videoEffectReadiness', [true, effectId, effectParam ?? null, null]);
```
WengHe, line 297: Please move `effectParam` to the end of the list so that the normal video app processing logic won't need to update.

shilcare: done

packages/teams-js/src/internal/videoUtils.ts
shilcare: refactored

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +108,142 @@ function createProcessedStreamGenerator(
...
+        originalFrame.close();
+        this.notifyError(error);
+      }
+    } else {
+      this.notifyError('timestamp of the original video frame is null');
+    }
+  };
+}
+
+class TransformerWithMetadata {
```
WengHe, line 141: Suggestion: to reduce more repeated code, you may make TransformerWithMetadata and DefaultTransformer both extend from an abstract parent class that has the basic implementation of the `transform` function and an abstract function named `getFrameProcessedByAppFromVideoFrameHandler`. Then TransformerWithMetadata and DefaultTransformer will only need to provide the implementation of `getFrameProcessedByAppFromVideoFrameHandler` without the need to implement the `transform` function

packages/teams-js/src/private/videoEx.ts
```diff
@@ -117,38 +173,63 @@ export namespace videoEx {
...
-    if (!parameters.videoBufferHandler) {
-      throw new Error('parameters.videoBufferHandler must be provided');
+
+    if (ensureInitialized(runtime, FrameContexts.sidePanel)) {
+      if (runtime.supports.video?.mediaStream) {
+        registerHandler(
+          'video.startVideoExtensibilityVideoStream',
+          async (mediaStreamInfo: { streamId: string; metadataInTexture?: boolean }) => {
+            const { streamId, metadataInTexture } = mediaStreamInfo;
+            const generator = metadataInTexture
```
WengHe, line 194: question: when the `metadataInTexture` will by falsy?

packages/teams-js/src/internal/videoUtils.ts
shilcare: seems it won't save more lines, and will be more prune to errors like "videoFrame not closed".

packages/teams-js/src/private/videoEx.ts
shilcare: the audioInferenceModel are optional, so if the app don't need audioInference, there will be no metadata in the texture

packages/teams-js/src/private/videoEx.ts
```diff
@@ -117,38 +173,63 @@ export namespace videoEx {
...
+   * @param parameters - Callbacks and configuration to process the video frames. A host may support either {@link VideoFrameHandler} or {@link VideoBufferHandler}, but not both.
+   * To ensure the video effect works on all supported hosts, the video app must provide both {@link VideoFrameHandler} and {@link VideoBufferHandler}.
+   * The host will choose the appropriate callback based on the host's capability.
    *
    * @internal
    * Limited to Microsoft-internal use
    */
   export function registerForVideoFrame(parameters: RegisterForVideoFrameParameters): void {
-    ensureInitialized(runtime, FrameContexts.sidePanel);
-    if (!isSupported()) {
```
ydogandjiev, line 128: Why are we removing the isSupported check?

packages/teams-js/src/private/videoEx.ts
```diff
@@ -117,38 +173,63 @@ export namespace videoEx {
...
+                () => {
+                  notifyVideoFrameProcessed(timestamp);
+                },
+                notifyError,
+              );
+            }
+          },
+          false,
+        );
+        sendMessageToParent('video.registerForVideoFrame', [parameters.config]);
```
ydogandjiev, line 221: The registerHandler call is already sending a message to Teams. Why do we need a separate one here?

packages/teams-js/src/private/videoEx.ts
```diff
@@ -117,38 +173,63 @@ export namespace videoEx {
...
+            const { streamId, metadataInTexture } = mediaStreamInfo;
+            const generator = metadataInTexture
+              ? await processMediaStreamWithMetadata(streamId, parameters.videoFrameHandler, notifyError)
+              : await processMediaStream(streamId, parameters.videoFrameHandler, notifyError);
+            // register the video track with processed frames back to the stream
+            !inServerSideRenderingEnvironment() &&
+              window['chrome']?.webview?.registerTextureStream(streamId, generator);
+          },
+          false,
+        );
```
ydogandjiev, line 202: The registerHandler call above is already sending a message to Teams. Why do we need a separate one here?

ydogandjiev (changes requested):

packages/teams-js/src/private/videoEx.ts
```diff
@@ -80,9 +83,24 @@ export namespace videoEx {
...
+     */
+    data: Uint8ClampedArray;
+  };
+
   /**
    * @hidden
-   * Video frame call back function
+   * Video frame call back function definition
+   * The callback will be called on every frame when running on the supported host.
+   * We require the frame rate of the video to be at least 22fps for 720p, thus the callback should process a frame timely.
```
TrevorJoelHarris, line 101: This line of documentation doesn't seem very helpful for developers calling the function, please consider removing it.

packages/teams-js/src/private/videoEx.ts
TrevorJoelHarris: I will be more explicit: you can't remove the isSupported check.

packages/teams-js/src/private/videoEx.ts
TrevorJoelHarris: Same for other copies of this line on other functions.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +108,142 @@ function createProcessedStreamGenerator(
...
+      x: 0,
+      y: headerDataView[6], // multiStreamHeaderRowOffset
+      width: texture.codedWidth,
+      height: texture.codedHeight - headerDataView[6], // multiStreamHeaderRowOffset
+    };
+    const metadataBuffer = new ArrayBuffer((metadataRect.width * metadataRect.height * 3) / 2);
+    await texture.copyTo(metadataBuffer, { rect: metadataRect });
+    const metadata = new Uint32Array(metadataBuffer);
+    let audioInferenceResult: Uint8Array | undefined;
+    for (let i = 0, index = 0; i < headerDataView[7] /* multiStreamCount */; i++) {
```
TrevorJoelHarris, line 220: 7? How do you know how many entries there are in this array? You should have protection against index out of bounds

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +108,142 @@ function createProcessedStreamGenerator(
...
+    // it's in NV12 format, but the real data is in the y plane only.
+    const headerBuffer = new ArrayBuffer((headerRect.width * headerRect.height * 3) / 2);
+    await texture.copyTo(headerBuffer, { rect: headerRect });
+    const headerDataView = new Uint32Array(headerBuffer);
+    // const [ oneTextureId, version, frameRowOffset, frameFormat, frameWidth, frameHeight, multiStreamHeaderRowOffset, multiStreamCount ] = headerDataView;
+    const metadataRect = {
+      x: 0,
+      y: headerDataView[6], // multiStreamHeaderRowOffset
+      width: texture.codedWidth,
+      height: texture.codedHeight - headerDataView[6], // multiStreamHeaderRowOffset
```
TrevorJoelHarris, line 214: 6? How do you know how many entries there are in this array? You should have protection against index out of bounds

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +108,142 @@ function createProcessedStreamGenerator(
...
+    };
+    const metadataBuffer = new ArrayBuffer((metadataRect.width * metadataRect.height * 3) / 2);
```
TrevorJoelHarris, line 216: Why are you multiplying by 3 and dividing by 2? There are a lot of [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) in this code, please either remove them, or store them as consts and clearly document why they need to be these specific values.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +108,142 @@ function createProcessedStreamGenerator(
...
+        audioInferenceResult = streamData;
+      }
+    }
+
+    return {
+      videoFrame: new VideoFrame(texture as unknown as CanvasImageSource, {
+        timestamp: texture.timestamp,
+        visibleRect: {
+          x: 0,
+          y: headerDataView[2], // frameRowOffset
```
TrevorJoelHarris, line 235: 3 more magic numbers

apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
```diff
@@ -80,22 +86,54 @@ const CheckIsSupported = (): React.ReactElement =>
...
+          },
+          videoBufferHandler: (buffer) => buffer,
+          config: {
+            format: video.VideoFrameFormat.NV12,
+            requireCameraStream: false,
+            audioInferenceModel,
+          },
+        });
+      } catch (error) {
+        return `Faild to register for video frame: ${JSON.stringify(error)}`;
```
TrevorJoelHarris, line 113:
```suggestion
        return `Failed to register for video frame: ${JSON.stringify(error)}`;
```

packages/teams-js/test/public/video.spec.ts
```diff
@@ -314,7 +314,7 @@ describe('video', () => {
       });
 
       it('should invoke videoEffectReadiness handler on callback resolved', async () => {
-        expect.assertions(3);
+        expect.assertions(4);
```
TrevorJoelHarris, line 317: You don't have any changes to the video file in this PR, why do you need to update the unit tests?

packages/teams-js/test/private/videoEx.spec.ts
```diff
@@ -6,6 +6,8 @@ import { errorNotSupportedOnPlatform, FrameContexts } from '../../src/public/con
 import { video } from '../../src/public/video';
 import { Utils } from '../utils';
 
+mockMediaStreamAPI();
```
TrevorJoelHarris, line 9: This isn't the right way to make a mock: it looks like after a quick scan of this function you're replacing defining window functions, but I don't see anywhere that you are undoing this at the end of these tests. Assuming I'm right, this means this unit test file will now be potentially causing problems in other unit test files. Please make sure that you have cleaned up all state including mocks at the end of your tests. Jest should have hooks you can use to run code at the beginning and end of unit test suites.

packages/teams-js/test/private/videoEx.spec.ts
```diff
@@ -249,6 +243,71 @@ describe('videoEx', () => {
...
+          utils.respondToFramelessMessage({
+            data: {
+              func: 'video.startVideoExtensibilityVideoStream',
+              args: [{ streamId: 'stream id' }],
+            },
+          } as DOMMessageEvent);
+          await utils.flushPromises();
+
+          // Assert
+          expect(webview.getTextureStream).toHaveBeenCalledTimes(1);
```
TrevorJoelHarris, line 277: I don't see any unit tests that are verifying that the `video.mediaStream.audioInferenceDiscardStatusChange` handler is getting registered for successfully

packages/teams-js/src/private/videoEx.ts
```diff
@@ -117,38 +173,63 @@ export namespace videoEx {
...
+
+    if (ensureInitialized(runtime, FrameContexts.sidePanel)) {
+      if (runtime.supports.video?.mediaStream) {
+        registerHandler(
+          'video.startVideoExtensibilityVideoStream',
+          async (mediaStreamInfo: { streamId: string; metadataInTexture?: boolean }) => {
+            const { streamId, metadataInTexture } = mediaStreamInfo;
+            const generator = metadataInTexture
+              ? await processMediaStreamWithMetadata(streamId, parameters.videoFrameHandler, notifyError)
+              : await processMediaStream(streamId, parameters.videoFrameHandler, notifyError);
```
TrevorJoelHarris, line 196: This call seems like it can throw errors that you aren't testing for in the unit tests. All error scenarios need unit tests.

TrevorJoelHarris (changes requested):

packages/teams-js/src/private/videoEx.ts
shilcare: `isSupport` check is not actually removed, the if-else branches below this line already covered all the supported/unsupported cases. 
I'll add it back if you prefer explicit checks.

packages/teams-js/src/private/videoEx.ts
shilcare: This is for backward compatible reasons, the initial implementation somehow used a separate message.

packages/teams-js/src/private/videoEx.ts
shilcare: For backward compatible reasons.

packages/teams-js/test/public/video.spec.ts
shilcare: The video module use createEffectParameterChangeCallback in internal/videoUtils.

apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
shilcare: hmm, fixing this typo will cause e2e failure, because hub-sdk also checks the typo string. Maybe I should create a new input with the typo fix and update the hub sdk and then remove this typo one

packages/teams-js/src/private/videoEx.ts
shilcare: removed

packages/teams-js/src/private/videoEx.ts
```diff
@@ -94,6 +112,40 @@ export namespace videoEx {
...
+     *
+     * @internal
+     * Limited to Microsoft-internal use
+     */
+    audioInferenceResult?: Uint8Array;
+  };
+
+  /**
+   * @hidden
+   * Video frame call back function definition.
```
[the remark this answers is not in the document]
shilcare: removed

packages/teams-js/src/private/videoEx.ts
```diff
@@ -94,6 +112,40 @@ export namespace videoEx {
...
+   * The callback will be called on every frame when running on the supported host.
+   * We require the frame rate of the video to be at least 22fps for 720p, thus the callback should process a frame timely.
+   * The video app should resolve the promise to notify a successfully processed video frame.
+   * The video app should reject the promise to notify a failure. When the failures accumulate to a certain number, the host will see the app is "frozen" and ask the user to close it or not.
```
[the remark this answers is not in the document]
shilcare: Added (determined by the host)

packages/teams-js/src/internal/videoUtils.ts
shilcare: added protection at line 259

packages/teams-js/src/internal/videoUtils.ts
shilcare: added comments above to explain the memory layout

apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
shilcare: fixed

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +108,142 @@ function createProcessedStreamGenerator(
...
+
+    if (texture.format !== 'NV12') {
+      this.notifyError('Unsupported video frame format');
+      throw new Error('Unsupported video frame format');
+    }
+
+    /**
+     * stream id for audio inference data
+     */
+    const AUDIO_INFERENCE_RESULT_STREAM_ID = 0x31646961;
```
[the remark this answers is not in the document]
shilcare: Added comments to explain the code

packages/teams-js/src/internal/videoUtils.ts
shilcare: Added comments to explain the size of NV12 format

packages/teams-js/test/private/videoEx.spec.ts
shilcare: added

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -1,11 +1,25 @@
...
+  AllowSharedBufferSource,
+  PlaneLayout,
+  VideoFrameBufferInit,
+  VideoFrameCopyToOptions,
+  VideoFrameInit,
+  VideoPixelFormat,
+} from './VideoFrameTypes';
 
 interface VideoFrame {
+  readonly codedWidth: number;
```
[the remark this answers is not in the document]
shilcare: done

packages/teams-js/test/private/videoEx.spec.ts
shilcare: added reset operation after test end

packages/teams-js/src/private/videoEx.ts
shilcare: Added case when videoFrame handler rejects.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +135,186 @@ function createProcessedStreamGenerator(
...
+   */
+  private extractVideoFrameAndMetadata = async (
+    texture: VideoFrame,
+  ): Promise<{ videoFrame: VideoFrame; metadata: { audioInferenceResult?: Uint8Array } }> => {
+    if (inServerSideRenderingEnvironment()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    if (texture.format !== 'NV12') {
+      this.notifyError('Unsupported video frame format');
```
TrevorJoelHarris, line 239: Please only define this string once, using either a const string or a const Error

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +135,186 @@ function createProcessedStreamGenerator(
...
+    const metadataBuffer = new ArrayBuffer((metadataRect.width * metadataRect.height * 3) / 2);
+    await texture.copyTo(metadataBuffer, { rect: metadataRect });
+    const metadata = new Uint32Array(metadataBuffer);
+    let audioInferenceResult: Uint8Array | undefined;
+    for (let i = 0, index = 0; i < headerDataView[7] /* multiStreamCount */; i++) {
+      const streamId = metadata[index++];
+      const streamDataOffset = metadata[index++];
+      const streamDataSize = metadata[index++];
+      const streamData = new Uint8Array(metadataBuffer, streamDataOffset, streamDataSize);
+      if (streamId === AUDIO_INFERENCE_RESULT_STREAM_ID) {
```
TrevorJoelHarris, line 296: when would streamId be some other value? This seems like the only streamId value you're ever setting

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +135,186 @@ function createProcessedStreamGenerator(
...
+    // 4: frameWidth
+    // 5: frameHeight
+    // 6: multiStreamHeaderRowOffset
+    // 7: multiStreamCount
+    if (headerDataView.length < 8) {
+      this.notifyError('Invalid video frame header');
+      throw new Error('Invalid video frame header');
+    }
+    // ensure the texture layout is supported
+    if (headerDataView[0] !== ONE_TEXTURE_LAYOUT_ID) {
```
TrevorJoelHarris, line 273: I've asked you in other places below to please not use magic numbers, but it doesn't look like you've fixed those. Rather than creating a Uint32Array and then needing to read one of the 8 indices (remembering what each of them means), please make a new class that takes in headerBuffer and exposes named getters to retrieve these values (oneTextureLayoutId, version etc.) It's okay if internal to that class you're using an array, but I don't want to mix that complicated logic in with the rest of this function. That will give us a class that has all the logic of how to work with this array and nothing else, and then this code will just use that class to pull out the named pieces of data it needs.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +135,186 @@ function createProcessedStreamGenerator(
...
+    // 0: oneTextureLayoutId
+    // 1: version
+    // 2: frameRowOffset
+    // 3: frameFormat
+    // 4: frameWidth
+    // 5: frameHeight
+    // 6: multiStreamHeaderRowOffset
+    // 7: multiStreamCount
+    if (headerDataView.length < 8) {
+      this.notifyError('Invalid video frame header');
```
TrevorJoelHarris, line 269: everywhere you're throwing errors with duplicated strings, please only define the strings once.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -78,36 +135,186 @@ function createProcessedStreamGenerator(
...
+
+    // The rectangle of pixels to copy from the texture. Metadata are at the bottom.
+    const metadataRect = {
+      x: 0,
+      y: headerDataView[6], // multiStreamHeaderRowOffset
+      width: texture.codedWidth,
+      height: texture.codedHeight - headerDataView[6], // multiStreamHeaderRowOffset
+    };
+    // allocate buffer for the metadata, buffer size = (the size of the Y plane + the size of the UV plane)
+    // (metadataRect.width * metadataRect.height) + (metadataRect.width * metadataRect.height) / 2
```
TrevorJoelHarris, line 286: the math in this comment does not appears to match the math on the line below it. I assume that's a mistake

packages/teams-js/src/internal/videoUtils.ts
TrevorJoelHarris: I appreciate the comments, but you still have magic numbers that are not consts, and there is no explanation of why you are using 2 or 3. Someone else looking at this code later would not be able to look at this code and understand what is doing.

packages/teams-js/src/private/videoEx.ts
TrevorJoelHarris: Does that mean that you're sending a message up from here to the host with the message name `video.registerForVideoFrame`, but then you're getting a response with the name `video.newVideoFrame`?

TrevorJoelHarris (changes requested):

packages/teams-js/src/private/videoEx.ts
shilcare: yes

packages/teams-js/src/internal/videoUtils.ts
shilcare: This is the only supported metadata type for now.

packages/teams-js/src/internal/videoUtils.ts
shilcare: refactored

packages/teams-js/src/internal/videoUtils.ts
shilcare: updated

apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
```diff
@@ -45,9 +45,15 @@ const RegisterForVideoEffect = (): React.ReactElement =>
     name: 'videoExRegisterForVideoEffect',
     title: 'videoEx - registerForVideoEffect',
     onClick: async (setResult) => {
-      const onVideoEffectChanged = (effectId: string | undefined, effectParam?: string): Promise<void> => {
-        setResult(`video effect changed to ${JSON.stringify(effectId)}, effect param: ${JSON.stringify(effectParam)}`);
-        return Promise.resolve();
+      const onVideoEffectChanged = async (effectId: string | undefined, effectParam?: string): Promise<void> => {
+        if (effectId === 'anInvalidEffectId') {
```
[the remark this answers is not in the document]
shilcare: We use the `anInvalidEffectId` as a special effectId in the tests to mock the case when video app reject the promise of effect change change callback. And check the message payload of `video.videoEffectReadiness` to be [false, 'anInvalidEffectId', InvalidEffectId, undefined] when video app rejects the promise.

WengHe (dismissed):

packages/teams-js/src/internal/videoUtils.ts
TrevorJoelHarris: Still have magic numbers here. The comment above doesn't really help me. You have a line of math that shows two things being equal ((metadataRect.width * metadataRect.height) + (metadataRect.width * metadataRect.height) / 2 and (metadataRect.width * metadataRect.height * 3) / 2) but those two things aren't equal?

apps/teams-test-app/src/components/privateApis/VideoExAPIs.tsx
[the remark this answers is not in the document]
TrevorJoelHarris: I see, the piece i was missing was that you are wrapping this handler with videoUtils.createEffectParameterChangeCallback. That code is messaging back to the host.

TrevorJoelHarris (changes requested):
Please address all feedback, including the open comment in videoUtils from the last PR request.

packages/teams-js/src/internal/videoUtils.ts
shilcare: I added a link to NV12 format document at line 345. The calculation is straightforward when looking at the memory layout of the NV12 format.
> // it's in NV12 format (https://learn.microsoft.com/en-us/windows/win32/medfound/recommended-8-bit-yuv-formats-for-video-rendering#nv12), so the
    // number bytes of the header is (the size of the Y plane + the size of the UV plane)

TrevorJoelHarris: I appreciate that to you this math is trivial. Comments and clearly named constants outlining why those values are being used are still necessary. "Go read this very long piece of external documentation" isn't sufficient. You could link to a specific part of that page which outlines how the math works, or you could explain it here. Asking other developers who have to understand this code to go to that page and search for "3" is not maintainable.

shilcare: Added the explanation in the comment

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

ydogandjiev (approved):

PR #1802, opened 2023-06-15
Add marketplace capability

wills2133:

## Description

To enable Teams users a more flexible purchasing experience, we introduce the marketplace capability to make checkout flow can be powered by a Teams app. A Teams app can leverage the marketplace capability to get and update a cart owned by the Teams client.

### Main changes in the PR:

1. add `marketplace` capability.
2. add `getCart`, `addOrUpdateCartItems`, `removeCartItems`, and `updateCartStatus` to interface with `cart` in the host.

### Validation performed:

1. Verify  `getCart`, `addOrUpdateCartItems`, `removeCartItems`, and `updateCartStatus` can modify the `cart` object in the MetaOS host.

### Unit Tests added:

Yes

### End-to-end tests added:

Yes

### Change file added:

Yes

### Related PRs:

adding marketplace capability in hub sdk - PR#2001855


packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * Represents the basic cart item information.
+ *
+ * @beta
+ */
+interface Item {
+  /** The id of the cart item. */
+  readonly id: string;
+  /** The display name of the cart item. */
```
TrevorJoelHarris, line 65: these 3 (or at least name and price) seem like they should be readonly: I assume those values won't change after construction of the Item. `readonly` prevents changing the value of the property after construction, but you (or devs) can modify by making copies of the Item and replacing certain values at construction time.

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @beta
+ */
+export enum CartStatus {
+  /** Cart is created but not checked out yet. */
+  Open = 'Open',
+  /** Cart is checked out but not completed yet. */
+  Processing = 'Processing',
+  /**
+   * Indicate checking out is completed and the host should
```
TrevorJoelHarris, line 108: I don't think I know what this sentence means. Is the end of the sentence missing?

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+ */
+export interface UpdateCartStatusParams {
+  /** Status of the cart. */
+  cartStatus: CartStatus;
+  /** Extra info to the status. */
+  statusInfo?: string;
+}
+
+/**
+ * Namespace for a vendor app to support a checkout flow by interacting with the marketplace cart in the host.
```
TrevorJoelHarris, line 131: What is a "vendor app"? How would app developers know whether or not that applies to them?

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+    });
+  }
+  /**
+   * Add or update cart items in the cart owned by the host.
+   *
+   * @param cartItems - A list of cart items, if item id exists, overwrite the item, otherwise add new items to cart.
+   *
+   * @beta
+   */
+  export function addOrUpdateCartItems(cartItems: CartItem[]): Promise<void> {
```
TrevorJoelHarris, line 155: It seems like you would want to pass up the cart Id at least here in case you need that in the host. May also be worth considering whether this should return Promise<Cart> instead of Promise<void>, in case developers want to render the updated card state after calling this method

TrevorJoelHarris, line 155: this seems like it should be Item instead of CartItem

TrevorJoelHarris: Is the array of items in here intended to just be the changed or added items? What happens if a value is passed in that is the same as the current value in the cart?

TrevorJoelHarris, line 155: This function should have validation of the cart items that have been passed in to attempt to catch and prevent illegal values (no negative number of items, no negative prices, no invalid price numbers like 25.24242, is imageUrl an actual URL that is valid? etc.

TrevorJoelHarris: (with corresponding unit tests)

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+    });
+  }
+  /**
+   * Remove cart items from the cart owned by the host.
+   *
+   * @param cartItemIds - A list of cart id, delete the cart item accordingly.
+   *
+   * @beta
+   */
+  export function removeCartItems(cartItemIds: string[]): Promise<void> {
```
TrevorJoelHarris, line 171: Is calling this function the same as calling addOrUpdate and setting the quantity to 0? If so, you could consider having these functions actually send the same message and data to the host under the covers. That would simplify the host code you have to write.

TrevorJoelHarris, line 171: Again, validation + unit tests needed. Make sure id is a valid format and value, array is non-empty, etc.

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+   *
+   * @param updateCartStatusParams
+   *
+   * updateCartStatusParams.cartStatus - cart status.
+   *
+   * updateCartStatusParams.message - extra info to the status.
+   *
+   * @beta
+   */
+  export function updateCartStatus(updateCartStatusParams: UpdateCartStatusParams): Promise<void> {
```
TrevorJoelHarris, line 191: consider adding cart id to this. If this is ever service backed you will need to know things like cart id, and it will save you plumbing in the future.

packages/teams-js/src/public/marketplace.ts
TrevorJoelHarris, line 171: consider adding cart id to this

packages/teams-js/test/public/marketplace.spec.ts
TrevorJoelHarris, line 1: I will look more closely at your tests once the other comments have been addressed

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+  /** The country market where the products are selling. */
+  readonly market: Market;
+  /** The identifier to tell the cart is checked out by admin or information worker */
+  readonly intent: Intent;
+  /** Locale for the user */
+  readonly locale: Locale;
+  /** The status of the cart. */
+  status: CartStatus;
+  /** Currency code for the cart item price. */
+  readonly currency: Currency;
```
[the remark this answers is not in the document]
wills2133: yes, we expect the items inside a cart should always have one currency.

packages/teams-js/src/public/marketplace.ts
wills2133: we allow the app to update the accessories of an item, for example, the app getCart from the host that the host intends to check out but the app cannot provide accessory A for one of an item due to lack of inventory, so it needs to update the latest cart items data. At this moment, even if the host cancels the checkout by closing the app, it won't end up with a stale cart.

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Represents the cart object for the app checkout flow.
+ *
+ * @beta
+ */
+export interface Cart {
+  /** Version of the cart. */
+  readonly version: CartVersion;
+  /** The id of the cart. */
+  readonly id: string;
```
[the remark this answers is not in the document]
wills2133: in some use cases, for example, the partner app of Teams Admin Center device store uses an integer item id as the identifier, we hope the id field could be compatible for this case.

packages/teams-js/src/public/marketplace.ts
wills2133: will make name readonly, but we allow the app to change price and quantity because the store in the host only provides the catalog service, and users are actually purchasing from the app. That means the price and quantity should respect what the app offers when the price is stale or the inventory is low. In this case, we will need the app to update the price/quantity of the host cart, through the addOrUpdateCartItems method.

packages/teams-js/src/public/marketplace.ts
wills2133: Will through error when quantity is 0

packages/teams-js/src/public/marketplace.ts
wills2133: will add price and quantity validation,  but imageUrl is readonly and from the host, I assume we don't need to validate it?

TrevorJoelHarris: Please still validate it: we see a lot of bugs where the hosts incorrectly implements functionality.

TrevorJoelHarris:
Closing this in favor of the newer PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1810/files

PR #1803, opened 2023-06-15
Wrapped live share code in liveShare namespace

huntj88:
## Description
This is primarily a cleanup PR to align the liveShare code with the rest of the code. Other features are wrapped in their own namespace, which was never done for Live Share.

LiveShareHost is still outside the namespace. Moving that inside the namespace would be a breaking change for LiveShare users.

removed @beta comments, updated doc comments.

### Validation performed:

1. Tested LiveShare apps with teams-js changes

### Unit Tests added: no

No logic changes were introduced, just aligning code patterns.

### DIFF
--- a/packages/teams-js/src/public/liveShareHost.ts
+++ b/packages/teams-js/src/public/liveShareHost.ts
@@ -3,9 +3,16 @@ import { ensureInitialized } from '../internal/internalAPIs';
 import { FrameContexts } from './constants';
 import { runtime } from './runtime';
 
+/**
+ * APIs involving Live Share, a framework for building real-time collaborative apps.
+ * For more information, visit https://aka.ms/teamsliveshare
+ *
+ * @see LiveShareHost
+ */
 export namespace liveShare {
   /**
-   * Meeting Roles.
+   * The meeting roles of a user.
+   * Used in Live Share for its role verification feature.
    */
   export enum UserMeetingRole {
     /**
@@ -27,7 +34,8 @@ export namespace liveShare {
   }
 
   /**
-   * State of the current Live Share sessions backing fluid container.
+   * State of the current Live Share session's Fluid container.
+   * This is used internally by the `LiveShareClient` when joining a Live Share session.
    */
   export enum ContainerState {
     /**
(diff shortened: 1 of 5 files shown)


packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -4,149 +4,184 @@ import { FrameContexts } from './constants';
...
- * State of the current Live Share sessions backing fluid container.
- *
- * @beta
- */
-export enum ContainerState {
+export namespace liveShare {
   /**
-   * The call to `LiveShareHost.setContainerId()` successfully created the container mapping
-   * for the current Live Share session.
+   * @hidden
```
TrevorJoelHarris, line 14: Why is this hidden? Adding this tag means dev docs won't be created for this enum, which would be pretty unusual for a non-beta enum that is used in exported and documented functions.

packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -4,149 +4,184 @@ import { FrameContexts } from './constants';
...
+   * The meeting roles of a user.
+   * Used in Live Share for its role verification feature.
```
TrevorJoelHarris, line 16: Is there documentation for the role verification feature? If so, it would be useful to link that here.

packages/teams-js/src/public/liveShareHost.ts
TrevorJoelHarris: I think I just have that question for all @hidden documentation in this file. Why don't you want public documentation for them?

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/liveShareHost.ts
ryanbliss: Developers don’t use those APIs directly, they are used in a very specific workflow within the Live Share SDK and have little utility outside of it. We thought about making them public but it’s not something anybody would know how to do, so we figured it’d be more confusing than not to document it.

packages/teams-js/src/public/liveShareHost.ts
ryanbliss: Sure, @huntj88 please add the following link to the doc: https://learn.microsoft.com/microsoftteams/platform/apps-in-teams-meetings/teams-live-share-capabilities?tabs=javascript#role-verification-for-live-data-structures

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

shrshindeMSFT (approved):
🚢

ryanbliss (approved):

PR #1810, opened 2023-06-21
Xjwu/add marketplace capability

wills2133:
## Description

To enable Teams users a more flexible purchasing experience, we introduce the marketplace capability to make checkout flow can be powered by a Teams app. A Teams app can leverage the marketplace capability to get and update a cart owned by the Teams client.

### Main changes in the PR:

1. add `marketplace` capability.
2. add `getCart`, `addOrUpdateCartItems`, `removeCartItems`, and `updateCartStatus` to interface with `cart` in the host.

### Validation performed:

1. Verify  `getCart`, `addOrUpdateCartItems`, `removeCartItems`, and `updateCartStatus` can modify the `cart` object in the MetaOS host.

### Unit Tests added:

Yes

### End-to-end tests added:

Yes

### Change file added:

Yes

### Related PRs:

adding marketplace capability in hub sdk - PR#2001855
[adding marketplace capability in hub sdk - PR#1802](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1802)

### DIFF
--- a/packages/teams-js/src/internal/marketplaceUtils.ts
+++ b/packages/teams-js/src/internal/marketplaceUtils.ts
@@ -1,59 +1,145 @@
+import { validate } from 'uuid';
+
 import { marketplace } from '../public';
 
 /**
  * @hidden
- * Validates the cart item properties are valid
+ * Validate the cart item properties are valid
  * @param cartItems The cart items
- * @returns [true, undefined] if all properties of all cart items are valid, [false, error message] otherwise
  *
  * @internal
  * Limited to Microsoft-internal use
  */
-export function validateCartItems(cartItems: marketplace.CartItem[]): [boolean, string | undefined] {
-  for (const cartItem of cartItems) {
-    const priceValidationResult = validatePrice(cartItem.price);
-    if (!priceValidationResult[0]) {
-      return priceValidationResult;
-    }
-    const quantityValidationResult = validateQuantity(cartItem.quantity);
-    if (!quantityValidationResult[0]) {
-      return quantityValidationResult;
-    }
+export function validateCartItems(cartItems: marketplace.CartItem[]): void {
+  if (!Array.isArray(cartItems) || cartItems.length === 0) {
+    throw new Error('cartItems must be a non-empty array');
   }
-  return [true, undefined];
+  for (const cartItem of cartItems) {
+    validateBasicCartItem(cartItem);
+    validateAccessoryItems(cartItem.accessories);
+  }
 }
 
 /**
  * @hidden
- * Validates the cart item properties are valid
- * @param price The price to be validated
- * @returns [true, undefined] if price is valid, [false, error message] otherwise
+ * Validate accessories
+ * @param accessoryItems The accessories to be validated
  *
  * @internal
  * Limited to Microsoft-internal use
  */
-export function validatePrice(price: number): [boolean, string | undefined] {
+export function validateAccessoryItems(accessoryItems: marketplace.Item[] | undefined | null): void {
+  if (accessoryItems === null || accessoryItems === undefined) {
+    return;
+  }
+  if (!Array.isArray(accessoryItems) || accessoryItems.length === 0) {
+    throw new Error('CartItem.accessories must be a non-empty array');
+  }
+  for (const accessoryItem of accessoryItems) {
+    if (accessoryItem['accessories']) {
+      throw new Error('Item in CartItem.accessories cannot have accessories');
+    }
+    validateBasicCartItem(accessoryItem);
+  }
+}
+
+/**
+ * @hidden
+ * Validate the basic cart item properties are valid
+ * @param basicCartItem The basic cart item
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validateBasicCartItem(basicCartItem: marketplace.Item): void {
+  if (!basicCartItem.name) {
+    throw new Error('cartItem.name must not be empty');
+  }
+  validatePrice(basicCartItem.price);
+  validateQuantity(basicCartItem.quantity);
+  validateUrl(basicCartItem.imageURL);
+}
+
+/**
+ * @hidden
+ * Validate the id is valid
+ * @param id A uuid string
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validateUuid(id: string | undefined | null): void {
+  if (id === undefined || id === null) {
+    return;
+  }
+  if (!id) {
+    throw new Error('id must not be empty');
+  }
+  if (validate(id) === false) {
+    throw new Error('id must be a valid UUID');
+  }
+}
+
+/**
+ * @hidden
+ * Validate the cart item properties are valid
+ * @param price The price to be validated
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validatePrice(price: number): void {
   if (typeof price !== 'number' || price < 0) {
-    return [false, `price ${price} must be a number not less than 0`];
+    throw new Error(`price ${price} must be a number not less than 0`);
   }
   if (parseFloat(price.toFixed(3)) !== price) {
-    return [false, `price ${price} must have at most 3 decimal places`];
+    throw new Error(`price ${price} must have at most 3 decimal places`);
   }
-  return [true, undefined];
 }
 
 /**
  * @hidden
- * Validates quantity
+ * Validate quantity
  * @param quantity The quantity to be validated
- * @returns [true, undefined] if quantity is valid, [false, error message] otherwise
  *
  * @internal
  * Limited to Microsoft-internal use
  */
-export function validateQuantity(quantity: number): [boolean, string | undefined] {
+export function validateQuantity(quantity: number): void {
   if (typeof quantity !== 'number' || quantity <= 0 || parseInt(quantity.toString()) !== quantity) {
-    return [false, `quantity ${quantity} must be an integer greater than 0`];
+    throw new Error(`quantity ${quantity} must be an integer greater than 0`);
   }
-  return [true, undefined];
+}
+
+/**
+ * @hidden
+ * Validate url
+ * @param url The url to be validated
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validateUrl(url: string | undefined | null): void {
+  if (url === undefined || url === null) {
+    return;
+  }
+  try {
+    new URL(url);
+  } catch (e) {
+    throw new Error(`url ${url} is not valid`);
+  }
+}
+
+/**
+ * @hidden
+ * Validate cart status
+ * @param cartStatus The cartStatus to be validated
+ *
+ * @internal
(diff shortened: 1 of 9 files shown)


packages/teams-js/src/internal/marketplaceUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+    const priceValidationResult = validatePrice(cartItem.price);
+    if (!priceValidationResult[0]) {
+      return priceValidationResult;
+    }
+    const quantityValidationResult = validateQuantity(cartItem.quantity);
+    if (!quantityValidationResult[0]) {
+      return quantityValidationResult;
+    }
+  }
+  return [true, undefined];
```
TrevorJoelHarris, line 23: Rather than have this return [boolean, string | undefined], it would be simpler to just have it return only a boolean, and throw if it ever detects an invalid state. That way, if other functions use this in the future they will have the same error handling behavior

packages/teams-js/src/internal/marketplaceUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @hidden
+ * Validates the cart item properties are valid
+ * @param cartItems The cart items
+ * @returns [true, undefined] if all properties of all cart items are valid, [false, error message] otherwise
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validateCartItems(cartItems: marketplace.CartItem[]): [boolean, string | undefined] {
+  for (const cartItem of cartItems) {
```
TrevorJoelHarris, line 13: Other validation I'd like to see in here (or explanations as to why it's not needed):

- id is not null and not empty
- name is not null and not empty
- imageURL, if present, is a valid URL (this is very easy to do in typescript: 
image)
- accessories, if present, is in a valid state:
     - Array is either absent or contains items
     - Items in the array do not themselves contain nested accessories arrays (accessories can't have more accessories)

packages/teams-js/src/internal/marketplaceUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Limited to Microsoft-internal use
+ */
+export function validatePrice(price: number): [boolean, string | undefined] {
+  if (typeof price !== 'number' || price < 0) {
+    return [false, `price ${price} must be a number not less than 0`];
+  }
+  if (parseFloat(price.toFixed(3)) !== price) {
+    return [false, `price ${price} must have at most 3 decimal places`];
+  }
+  return [true, undefined];
```
TrevorJoelHarris, line 42: Similar comment about throwing as above

packages/teams-js/src/internal/marketplaceUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @returns [true, undefined] if quantity is valid, [false, error message] otherwise
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validateQuantity(quantity: number): [boolean, string | undefined] {
+  if (typeof quantity !== 'number' || quantity <= 0 || parseInt(quantity.toString()) !== quantity) {
+    return [false, `quantity ${quantity} must be an integer greater than 0`];
+  }
+  return [true, undefined];
```
TrevorJoelHarris, line 58: Similar comment about throwing as above

packages/teams-js/test/public/marketplace.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+          expect(marketplace.isSupported()).toBeTruthy();
+        });
+
+        it('should throw if called before initialization', () => {
+          utils.uninitializeRuntimeConfig();
+          expect(() => marketplace.isSupported()).toThrowError(new Error(errorLibraryNotInitialized));
+        });
+      });
+
+      describe('Testing marketplace.getCart function', () => {
```
TrevorJoelHarris, line 84: Should be a unit test validating that the marketplace.getCart message is sent to the host

packages/teams-js/test/public/marketplace.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+                  FrameContexts.content,
+                  FrameContexts.task,
+                ])}. Current context: "${context}".`,
+              );
+            });
+          }
+        });
+      });
+
+      describe('Testing marketplace.addOrUpdateCartItems function', () => {
```
TrevorJoelHarris, line 110: this function should also be testing that the correct errors are thrown when invalid cartItems are passed in

TrevorJoelHarris: I realize you have tests for the utility functions above, but we should fully test all exported functions to help document how they work

packages/teams-js/test/public/marketplace.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+        it('marketplace.removeCartItems should throw error with empty array input', async () => {
+          await utils.initializeWithContext(FrameContexts.content);
+          utils.setRuntimeConfig({ apiVersion: 2, supports: { marketplace: {} } });
+          expect(marketplace.removeCartItems([])).rejects.toThrowError(
+            new Error('cartItemIds must be a non-empty array'),
+          );
+        });
+
+        Object.values(FrameContexts).forEach((context) => {
+          if (context === FrameContexts.content || context === FrameContexts.task) {
```
TrevorJoelHarris, line 176: should have unit tests that test the "not an array" and "empty array" cases

TrevorJoelHarris: (The test case you have above only tests empty array, and it should b e moved into this forEach so we test it on all valid frameContexts)

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,323 @@
...
+   * @param updateCartStatusParams
+   * updateCartStatusParams.cartStatus - cart status.
+   * updateCartStatusParams.statusInfo - extra info to the status.
+   *
+   * @beta
+   */
+  export function updateCartStatus(updateCartStatusParams: UpdateCartStatusParams): Promise<void> {
+    return new Promise<void>((resolve) => {
+      ensureInitialized(runtime, FrameContexts.content, FrameContexts.task);
+      if (!isSupported()) {
```
TrevorJoelHarris, line 304: There should be a check to verify that updateCartStatusParams is not null

TrevorJoelHarris (changes requested):

TrevorJoelHarris (changes requested):
Almost there!

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,361 @@
...
+   * @hidden
+   * Represents the cart object for the app checkout flow.
+   * @beta
+   */
+  export interface Cart {
+    /**
+     * @hidden
+     * Version of the cart.
+     */
+    readonly version: CartVersion;
```
AE-MS, line 23: Were I to be a developer looking at this property and documentation, I don't think I would understand what this property is for. I know what a version is but... what should I be using it for?

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,361 @@
...
+    /**
+     * @hidden
+     * The display name of the cart item.
+     */
+    readonly name: string;
+    /**
+     * @hidden
+     * The quantity of the cart item.
+     */
+    readonly quantity: number;
```
AE-MS, line 131: >number [](http://example.com/codeflow?start=23&length=6)

Not sure how deep you want to go to ensuring misuse is avoided but you could consider a custom type to prevent this value from ever being negative (assuming it's invalid to have a negative quantity of items)

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,361 @@
...
+    /**
+     * @hidden
+     * The price of the single cart item.
+     */
+    readonly price: number;
+    /**
+     * @hidden
+     * The thumbnail imageURL of the cart item.
+     */
+    readonly imageURL?: string;
```
AE-MS, line 141: >string [](http://example.com/codeflow?start=24&length=6)

Can this be a `URL` type?

packages/teams-js/src/public/marketplace.ts
```diff
@@ -0,0 +1,361 @@
...
+   * @beta
+   */
+  export function getCart(): Promise<Cart> {
+    return new Promise<Cart>((resolve) => {
+      ensureInitialized(runtime, FrameContexts.content, FrameContexts.task);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+      resolve(sendAndHandleSdkError('marketplace.getCart'));
+    });
```
AE-MS, line 284: Is this "return new Promise" pattern necessary? I see it's used multiple places in this file. It seems like it adds an extra unnecessary layer of promise wrapping/etc. Could this just be written more straightforwardly as:
```
      if (!isSupported()) {
      }
      return sendAndHandleSdkError('marketplace.getCart');
    });
```
?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/marketplace.ts
AE-MS: Makes sense to me! I would suggest you update the docs just to make it clear to developers that they can just ignore this field (e.g., just add something like "this is for internal use only")

packages/teams-js/src/public/marketplace.ts
AE-MS: My general principle is this: we should use the most specific, structured types when we can. When we can't (e.g., serialization), then the transformation and validation of that data should happen as early as possible. For example, if we get raw data "over the wire" (e.g., from a post message), the first thing the code should do is transform it into a non-raw better type, and only that type should be used to access the data from then on. In the other direction, it should be in the exact type until the very last second (e.g., immediately before serialization).

This approach that as much code as possible can access the data as safely as possible without repeated validation. For example, even if you validate that something is a URL after receiving the data, but keep it stored as a string, then no one else who uses that data later will know that the data has been validated and doesn't need to be validated again (because the type makes no such guarantees).

In practice, the way this _could_ be implemented here is that as soon as you receive the message from the hub SDK, it should validate it as a URL and only ever expose it to users/developers as a `URL` type. The fact that it had to be a string for serialization purposes is an implementation detail that doesn't need to leak out to the rest of the code.

In practice, this means that you may need to define two interfaces -- the one used for serialization and the one used "for the rest of the code to safely interact with."

Of course, there may be some reason why this doesn't work in this case, but I think it's something worth reflecting on to see if it can be applied here: The serialization format is an implementation detail, not a type definition to be used by the rest of the code.

Thoughts?

wills2133: add URL type, serializer and deserializer

AE-MS (dismissed):

TrevorJoelHarris (approved):

TrevorJoelHarris (approved):

AE-MS (approved):

PR #1819, opened 2023-06-27
update webpack to enable local self-signed certs to be used

alexneyman-MSFT:
## Description

The changes in this PR enable someone to use self-signed certificates when running the teams test app. The code looks for local certificates and if they exist, uses them when starting the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ dts/
 # Test App gitignore areas
 yarn-error.log
 *.log
+*.pem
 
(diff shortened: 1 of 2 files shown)


jadahiya-MSFT (dismissed):

TrevorJoelHarris (approved):

TrevorJoelHarris (approved):

alexneyman-MSFT:
As of right now, we don't need this so I'm closing it

PR #1827, opened 2023-07-05
Add testing Json data for marketplace capability

KangxuanYe:

## Description

A recent check-in PR in Web Host SDK side added marketplace capability and thus new testing json data is added in web Host SDK side. e2e-test-data under teams-test-app needs to keep updating for migration purpose as well.

### Main changes in the PR:

1. marketplace.json (testing json data) is added and specifically label it to run on web platform only.

### DIFF
--- a/apps/teams-test-app/e2e-test-data/marketplace.json
+++ b/apps/teams-test-app/e2e-test-data/marketplace.json
@@ -1,0 +1,136 @@
+{
+  "name": "Marketplace",
+  "platforms": "Web",
+  "version": ">2.12.0",
+  "checkIsSupported": {
+    "capabilityName": "Marketplace",
+    "toggleId": "marketplaceToggle",
+    "expectedOutput": "marketplace module is not supported"
+  },
+  "testCases": [
+    {
+      "title": "GetCart API Call - Success",
+      "type": "callResponse",
+      "boxSelector": "#box_getCart",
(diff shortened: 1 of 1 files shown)


lakhveerk (approved):

apps/teams-test-app/e2e-test-data/marketplace.json
```diff
@@ -0,0 +1,136 @@
...
+          {
+            "id": "2",
+            "name": "Item 1",
+            "price": 10,
+            "quantity": 1,
+            "accessories": [{ "id": "11", "name": "Item 11", "price": 5, "quantity": 1 }]
+          }
+        ]
+      },
+      "expectedTestAppValue": "Error: Error: id must be a valid UUID"
```
jadahiya-MSFT, line 112: Not saying this has to be fixed here, but the Error: Error: seems very strange to me. Might be worth creating a work item to go around and fix this.

jadahiya-MSFT (approved):

apps/teams-test-app/e2e-test-data/marketplace.json
KangxuanYe: I think I have seen some places for the same prompts when we do negative E2E testing. I am not sure if it is a bug or the way that teams test app handling errors. I'll also tag @erinha for awareness because if we change the way for producing prompts in teams test app, there might be more places need to be updated.

jadahiya-MSFT: Yea you would have to make the changes on the Teams Test App and the E2E Tests as well as this file. It will be a bigger change than the one you are trying to do here, which is why I suggested creating a work item to track this for later. You can assign that to me if you'd like.

KangxuanYe: Thank you. A task has been created. Please check. https://office.visualstudio.com/MetaOS/_workitems/edit/8084720

PR #1836, opened 2023-07-10
appNotifications API added to the teamsJS

bernice-msft:
A new appNotification capability to the teamsJS codebase, this allow MOS app developer utilize this capability on their apps.This is the first step in the implementation- including a .ts file all the necessary functions required for this capability to fully work

Changes Made
1)Included the appNotification.ts file which consists of all the relevant functions implemented.
2)With my changes incorporated,I was able to successfully build without a fail,however I’m still working on completing the Unit tests for all the functions to ensure it passes/fails at the necessary test cases.
3)I also included the appNotification capability on the teamsRuntimeConfig object(just the apiVersion2) in the runtime.ts file
4)I exported the appNotification capability in the index.ts as well.

There are further changes that would coming along listed down below:
1)A fully implemented unit test file to properly test all the functions for the capability
2)A fully implemented end to end test file
3)A sample test app created


packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    return urlPattern.test(Url);
+  }
+
+  /**
+   * function definition for appNotification title string length validation
+   *
+   * @param notificationdisplayparam
+   * @returns a boolean to represent whether the length of the Title and Content strings are valid
+   */
+  export function isValidSTitleLength(Title: string): boolean {
```
lakhveerk, line 44: nit: 'S' in `isValidSTitleLength` seems extra

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+  export function isValidUrl(Url: string): boolean {
+    // Regular expression pattern for URL validation
+    const urlPattern = /^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$/i;
+    return urlPattern.test(Url);
+  }
+
+  /**
+   * function definition for appNotification title string length validation
+   *
+   * @param notificationdisplayparam
```
lakhveerk, line 41: parameter name is title

packages/teams-js/src/public/appNotification.ts
lakhveerk: Also 'T' for Title should be lower case

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    return Title.length <= maxTitleLength ? true : false;
+  }
+
+  /**
+   * function definition for appNotification content string length validation
+   *
+   * @param notificationdisplayparam
+   * @returns a boolean to represent whether the length of the Title and Content strings are valid
+   */
+  export function isValidContentLength(Content: string): boolean {
```
lakhveerk, line 55: you can have one general function that checks the length instead of two separate ones for content and title.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    const maxContentLength = 45;
+    return Content.length <= maxContentLength ? true : false;
+  }
+
+  /**
+   * This function displays notification from a METAOS application
```
lakhveerk, line 61: I think we advertise it as Teams app.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  export function isValidSTitleLength(Title: string): boolean {
+    const maxTitleLength = 20;
+    return Title.length <= maxTitleLength ? true : false;
+  }
+
+  /**
+   * function definition for appNotification content string length validation
+   *
+   * @param notificationdisplayparam
```
lakhveerk, line 52: parameter is content here not notificationDisplayParam

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    const maxContentLength = 45;
+    return Content.length <= maxContentLength ? true : false;
+  }
+
+  /**
+   * This function displays notification from a METAOS application
+   * @param notificationdisplayparam
+   * @returns a promise resolution upon conclusion
+   */
+  export function notiificationDisplay(notificationdisplayparam: NotificationDisplayParam): Promise<void> {
```
lakhveerk, line 65: nit: notiificationDisplay should have single i

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+   * @returns a boolean to represent whether the length of the Title and Content strings are valid
+   */
+  export function isValidContentLength(Content: string): boolean {
+    const maxContentLength = 45;
+    return Content.length <= maxContentLength ? true : false;
+  }
+
+  /**
+   * This function displays notification from a METAOS application
+   * @param notificationdisplayparam
```
lakhveerk, line 62: describe in one line what `notificationdisplayparam` is

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    if (notificationdisplayparam.title && !isValidSTitleLength(notificationdisplayparam.title)) {
+      throw 'Invalid notification title length';
+    }
+
+    if (notificationdisplayparam.Content && !isValidContentLength(notificationdisplayparam.Content)) {
+      throw 'Invald notification content length';
```
lakhveerk, line 82: nit: Invalid spelling

packages/teams-js/src/public/appNotification.ts
lakhveerk: nit: change notificationdisplayparam to `notificationDisplayParam`

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    if (notificationdisplayparam.url && !isValidUrl(notificationdisplayparam.url)) {
+      throw 'INVALID URL';
+    }
+    return sendAndHandleSdkError('appNotification.displayNotification', notificationdisplayparam);
+  }
+
+  /**
+   * Checks if appNotification is supported by the host on which the MOS app is currently launched
+   * @returns boolean to represent whether the appNotification capability is supported
```
lakhveerk, line 93: You can add @throws Error if {@linkcode app.initialize} has not successfully completed

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Namespace to interact with the appNotifications specific part of the SDK
+ */
+export namespace appNotification {
+  /**
+   * Data structure to represent appNotification information
+   */
+  export interface NotificationDisplayParam {
+    /**The heading of the notiification to be delivered,it would have a maximum string length of 20*/
```
lakhveerk, line 14: We usually use this comment pattern, instead of one liner
```
/**
 * some comment
 */
  ```

**Comment by lakhveerk** [comment]
The pipelines are failing because of no beachball. Whenever we make a new change, we want to keep track of that, so we can list them when we publish a new version. We do it using beachball. Learn more about this here.  https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball

To fix the error, you can run `pnpm changefile` this will prompt you to select what kind of changes you are trying to make. Select that and add a comment describe what this change is about. It will automatically commit your change, but you would need to push them

File: packages/teams-js/src/public/appNotification.ts (hunk 10 above)
Comment (AE-MS) [reply packages/teams-js/src/public/appNotification.ts:14]: JFYI it's stronger than "usually." It's required that all our public interfaces/properties/functions be documented using the typedoc syntax so it can be published to the Microsoft docs site. I suspect that if you don't do this, the build will fail (it should fail) -- it didn't fail in the pipeline for this run because it failed and aborted when it couldn't find the beachball changefile.

FYI you can and should run the build locally (using `pnpm build`) to help validate this before publishing the PR. I don't expect that building locally will catch the beachball problem, but it should catch the incorrect documentation problem.

Comment (AE-MS) [inline packages/teams-js/src/public/appNotification.ts:14]: >notiification [](http://example.com/codeflow?start=26&length=13)

Spelling

Comment (AE-MS) [inline packages/teams-js/src/public/appNotification.ts:14]: >, [](http://example.com/codeflow?start=55&length=1)

spaces after commas, please.

**Comment by AE-MS** [comment]
First PR! w00t!
image

File: packages/teams-js/src/public/appNotification.ts [hunk 11]
```diff
@@ -0,0 +1,98 @@
...
+    title: string;
```
AE-MS, line 15: We should be consistent on the naming. If we are calling this the header, then the property should be called header. If we are calling it the title, then we should use that word in the documentation.

packages/teams-js/src/public/appNotification.ts
AE-MS, line 14: >The heading of the notiification to be delivered,it would have a maximum string length of 20 [](http://example.com/codeflow?start=7&length=92)

To be consistent with how other documentation in the library is phrased, avoid using words like "the" or "this would." For example, for this property consider:

``Heading of the notification (maximum length: 30 characters)"

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    /**the body of information which would be displayed in the notification box,it would have a maximum string length of 40*/
+    Content: string;
+    /**this would represent an icon that can be displayed on the notification*/
```
AE-MS, line 18: >/**this would represent an icon that can be displayed on the notification*/ [](http://example.com/codeflow?start=4&length=75)

Since it looks like the icon is optional, can you document the behavior when no icon is specificed?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    icon?: HTMLImageElement;
```
AE-MS, line 19: >HTMLImageElement [](http://example.com/codeflow?start=11&length=16)

Can you say more about how you chose this type for this property? Since this icon is really just an image, it seems like it would be extra and unnecessary work for a developer to create an HTML DOM `img` element to hold the icon. Moreover, I'm not sure how passing a DOM element like this across the iframe boundary could work. The IMG element will be created inside the app's IFRAME, and it couldn't really be access by the host SDK or the hosting app. Have you considered just using either a URL to the image or the bytes of the image? Are there any restrictions on its size?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    /**This would specify how long a notification would be displayed on the screen for*/
+    itemId: number;
```
AE-MS, line 21: >itemId [](http://example.com/codeflow?start=4&length=6)

Can you specify the units for this? E.g., is it seconds, milliseconds, minutes, etc.?  Putting that in the name and the comments will make the usage of this property clearer, e.g., displayDurationInSeconds

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    /**a url link to the page in which the notification would direct the user to.*/
+    url?: string;
```
AE-MS, line 23: >string [](http://example.com/codeflow?start=10&length=6)

Can you use the `URL` type for this?

AE-MS, line 23: >url [](http://example.com/codeflow?start=4&length=3)

Can you use a more specific name for this that describes its usage, e.g. `notificationActionUrl`

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * //function definition to check validity of the URL
+   *
+   * @param url
+   * @returns a boolean to represent whether the URL passed is a valid one
+   */
+  export function isValidUrl(Url: string): boolean {
```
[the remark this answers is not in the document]
AE-MS: How did you decide which types of URLs are valid? E.g., why are ftp sites valid?

Per Lakhveer's comment (and mine on the interface above), using the `URL` type will prevent from having to manually write validation code, enabling you to just check the `protocol` property.

AE-MS, line 32: >export [](http://example.com/codeflow?start=2&length=6)

Does this function need to be exported? I didn't see it being used outside of this file.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    // Regular expression pattern for URL validation
+    const urlPattern = /^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$/i;
+    return urlPattern.test(Url);
+  }
+
+  /**
+   * function definition for appNotification title string length validation
```
AE-MS, line 39: >function definitio [](http://example.com/codeflow?start=5&length=18)

You don't need to start these with "function definition for" -- it's obvious that it is a function. :) If you look at some of the other documentation in the library, you can follow the documentation pattern there,
e.g.,
"Validates that notification title is of property length"

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * @param notificationdisplayparam
+   * @returns a boolean to represent whether the length of the Title and Content strings are valid
```
AE-MS, line 42: >a boolean [](http://example.com/codeflow?start=14&length=9)

You don't need to document the type it returns, that is already clear from the function definition. You can just say ```
@returns true if length of title is shorter than the limit

AE-MS, line 42: >whether the length of the Title and Content strings are valid [](http://example.com/codeflow?start=37&length=61)

The comment indicates that it's checking the length of both the title and content strings but it looks like it is only checking the title lenth?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  export function isValidSTitleLength(Title: string): boolean {
+    const maxTitleLength = 20;
+    return Title.length <= maxTitleLength ? true : false;
```
AE-MS, line 46: >return Title.length <= maxTitleLength ? true : false; [](http://example.com/codeflow?start=4&length=53)

You can just shorten this to
```
return title.length <= maxTitleLength
```

packages/teams-js/src/public/appNotification.ts
AE-MS: You can't use the term "metaos" publicly.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+      FrameContexts.stage,
+      FrameContexts.sidePanel,
+      FrameContexts.meetingStage,
+    );
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    if (notificationdisplayparam.title && !isValidSTitleLength(notificationdisplayparam.title)) {
+      throw 'Invalid notification title length';
```
AE-MS, line 78: >throw [](http://example.com/codeflow?start=6&length=5)

All of your `throw` statements should throw `Error` objects and not strings.

AE-MS, line 78: This error would be most helpful if it also let the developer know what the length limit was and how long the title was that was given.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    if (notificationdisplayparam.Content && !isValidContentLength(notificationdisplayparam.Content)) {
+      throw 'Invald notification content length';
+    }
+
+    if (notificationdisplayparam.url && !isValidUrl(notificationdisplayparam.url)) {
+      throw 'INVALID URL';
```
AE-MS, line 86: >throw 'INVALID URL'; [](http://example.com/codeflow?start=6&length=20)

Why is this one in all caps?

packages/teams-js/src/public/appNotification.ts
AE-MS, line 61: Given how much this function will do, it feels like it deserves a more completely description to help set developers expectations on what it does and how to use it. If you were a developer reading this, would you know what this function did?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+      FrameContexts.content,
+      FrameContexts.stage,
+      FrameContexts.sidePanel,
+      FrameContexts.meetingStage,
+    );
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    if (notificationdisplayparam.title && !isValidSTitleLength(notificationdisplayparam.title)) {
```
AE-MS, line 77: >notificationdisplayparam.title && !isValidSTitleLength(notificationdisplayparam.title) [](http://example.com/codeflow?start=8&length=86)

This statement seems to say "if we were given a title, and it's NOT valid, then throw an error."

Do you want to handle the case that no title was given? Is that an error? Or are you assuming that you will always be given a title?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
+import { sendAndHandleSdkError } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
+import { runtime } from './runtime';
+
+/**
```
AE-MS, line 6: So awesome to see these changes become a reality! Nice work!

AE-MS (changes requested):
👏

packages/teams-js/src/public/appNotification.ts
bernice-msft: Woops I missed that edgecase! I've included a check for that!

packages/teams-js/src/public/appNotification.ts
bernice-msft: I don't think so..I've updated that

jadahiya-MSFT, line 32: nit: `Url` -> `url` lowercase instead.

packages/teams-js/src/public/appNotification.ts
bernice-msft: Updated this

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  /**
+   * Displays appNotification after making a validiity check on all of the required parameters, by calling the validateNotificationDisplayParams helper function
+   * An interface object containing all the required parameters to be displayed on the notification would be passed in here
+   * @param notificationdisplayparam - Interface object with all the parameters required to display an appNotificiation
+   * @returns a promise resolution upon conclusion
+   */
+  export function displayInAppNotification(notificationDisplayparam: NotificationDisplayParam): Promise<void> {
```
lakhveerk, line 121: notificationDisplayparam should be `notificationDisplayParam`

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,145 @@
...
+      FrameContexts.stage,
+      FrameContexts.sidePanel,
+      FrameContexts.meetingStage,
+    );
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+    validateNotificationDisplayParams(notificationDisplayparam);
+    return sendAndHandleSdkError('appNotification.displayNotification', notificationDisplayparam);
```
lakhveerk, line 134: In our meeting with Trevor, we decided we will be sending a string type of url in notificationDisplayparam to hubSDK.

.vscode/settings.json
```diff
@@ -11,5 +11,8 @@
   "typescript.preferences.quoteStyle": "single",
   "prettier.singleQuote": true,
   "cSpell.words": ["adal", "uninitialize"],
-  "jest.jestCommandLine": "pnpm test -- --"
+  "jest.jestCommandLine": "pnpm test -- --",
+  "[typescript]": {
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
lakhveerk, line 16: What is this for?

change/@microsoft-teams-js-cc1d35f9-c5f8-4d2b-9161-2d3c53d6d593.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Included appNotification.ts file alongside  changes",
```
lakhveerk, line 3: `alongside changes` have extra space

packages/teams-js/src/public/appNotification.ts
lakhveerk: We can talk more about this in our 1:1 today

AE-MS:
What are all the files in your change under the ".history" directory? I don't think those should be committed...

change/@microsoft-teams-js-cc1d35f9-c5f8-4d2b-9161-2d3c53d6d593.json
AE-MS, line 3: >Included appNotification.ts file alongside  changes [](http://example.com/codeflow?start=14&length=51)

This comment should be a summary of the whole change you are committing and understandable by people reading the changelog. Based on that, I would suggest something like "Added new `appNotification` capability for apps to send notifications to the user through the host"

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+   * Checks the valididty of the URL passed
+   *
+   * @param url
+   * @returns True if a valid url was passed
+   */
+  function isValidUrl(url: URL): boolean {
+    const validProtocols = ['http:', 'https:', 'ftp:'];
```
AE-MS, line 90: >'http:', 'https:', 'ftp:' [](http://example.com/codeflow?start=28&length=25)

Are ftp sites really valid for this? Should we restrict to https sites for privacy/security reasons?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,98 @@
...
+  export function isValidContentLength(Content: string): boolean {
+    const maxContentLength = 45;
+    return Content.length <= maxContentLength ? true : false;
+  }
+
+  /**
+   * This function displays notification from a METAOS application
+   * @param notificationdisplayparam
+   * @returns a promise resolution upon conclusion
+   */
```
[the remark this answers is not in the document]
AE-MS: It looks like the errors thrown are still undocumented. E.g. you can look at other capabilities and where they use `@throws` for examples.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Notification content (maximum length: 40 characters)
+     */
+    content: string;
+    /**
+     * This would represent an optional icon that can be displayed on the notification. It should have a max size of 49 pixels by 49 pixels
+     * If no icon is provided, the notification card would be displayed without an icon
+     * The url link to where the icon is stored should be provided as the input string
+     */
+    icon?: string;
```
AE-MS, line 27: >string [](http://example.com/codeflow?start=11&length=6)

Since you are able to use the `URL` type in this interface (i.e. you use it for the notification action URL), should this also be a `URL` then?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    displayDurationInSeconds: number;
+    /**
+     * A url link to the page in which the notification would direct the user to.
+     */
+    notificationActionUrl: URL;
+  }
+
+  /**
+   * Data structure with simplified parameters that would be passed across the iframe
```
AE-MS, line 40: > with simplified parameters  [](http://example.com/codeflow?start=19&length=28)

What does it mean that it has "simplified parameters?" Would a developer reading this know what that meant? Is there a more complex version of this interface they could optionally choose to use?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+    displayDurationInSeconds: number;
+    /**
+     * A url string type to the page in which the notification would direct the user to.
+     */
+    notificationActionUrlAsString: string;
+  }
+
+  /**
+   * This converts the notifcationActionUrl from a URL type to a string type for proper flow across the iframe
+   * @param ndParam
```
AE-MS, line 70: > @param ndParam [](http://example.com/codeflow?start=4&length=15)

Should this be `notificationDisplayParam`?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+   * @returns a NotificationDisplay object with a string type parameter for the notificationAtionUrl
```
AE-MS, line 71: > @returns a NotificationDisplay object with a string type parameter for the notificationAtionUrl [](http://example.com/codeflow?start=4&length=96)

I'm not sure I see how this describes what is going on here?

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+      throw new Error(
+        `Invalid notification content length: Maximum content length ${maxContentLength}, content length supplied ${notificationDisplayParam.content.length}`,
+      );
+    }
+
+    if (!notificationDisplayParam.notificationActionUrl) {
+      throw new Error('Must supply notification url');
+    }
+    if (!isValidUrl(notificationDisplayParam.notificationActionUrl)) {
+      throw new Error('Invalid url');
```
AE-MS, line 139: >Invalid url' [](http://example.com/codeflow?start=23&length=12)

Can you clarify in this error message which URL is invalid?

packages/teams-js/test/public/appNotification.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+  });
+
+  afterEach(() => {
+    app._uninitialize();
+    GlobalVars.isFramelessWindow = false;
+  });
+
+  const displayNotificationParam: appNotification.NotificationDisplayParam = {
+    title: 'New Missed Call',
+    content: 'You just received a missed call from Paul Sayer',
```
AE-MS, line 38: >Paul Sayer [](http://example.com/codeflow?start=51&length=10)

You can't use names in our source code unless they are on the official approved list of fictitious names. Is that where you got the name from?

Even if you weren't intending to refer to a real person, there can be legal issues from people who actually have that name.

You should ensure you are using a name from the approved list -- or, probably more appropriate here, just don't use a name, e.g., "You just received a missed call."

packages/teams-js/test/public/appNotification.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+          utils.setRuntimeConfig({ apiVersion: 2, supports: { appNotification: {} } });
+
+          const displayNotificationParam: appNotification.NotificationDisplayParam = {
+            title: 'Update: Maintenance extended to include server upgrades.',
+            content:
+            `Dear valued users,
+            We would like to inform you that scheduled maintenance will take place on our Microsoft platform tomorrow, August 1st, 2023, starting at 10:00 AM UTC. During this time, Microsoft services will be temporarily unavailable as we perform necessary upgrades to enhance performance and security.
+            Expected Downtime: Approximately 4 hours.
+            We apologize for any inconvenience this may cause and assure you that our team at Microsoft is working diligently to complete the maintenance as quickly as possible.`,   
+            displayDurationInSeconds: -1,
```
AE-MS, line 130: >-1 [](http://example.com/codeflow?start=38&length=2)

The title of this test suggests that you meant for this to be zero?

packages/teams-js/test/public/appNotification.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+                We would like to inform you that scheduled maintenance will take place on our Microsoft platform tomorrow, August 1st, 2023, starting at 10:00 AM UTC. During this time, Microsoft services will be temporarily unavailable as we perform necessary upgrades to enhance performance and security
+                Expected Downtime: Approximately 4 hours.
+                We apologize for any inconvenience this may cause and assure you that our team at Microsoft is working diligently to complete the maintenance as quickly as possible.`,
+            displayDurationInSeconds: 30,
+            notificationActionUrl: new URL('http://www.example.com'),
+          };
+
+          expect(() => appNotification.displayInAppNotification(displayNotificationParam)).toThrowError(
+            new Error(
+              `Invalid notification title length: Maximum title length ${maxTitleLength}, title length supplied ${displayNotificationParam.title.length}`,
```
AE-MS, line 114: >Invalid notification title length [](http://example.com/codeflow?start=15&length=33)

In this test, it looks like you have both a title string that is too long and a content string that is too long. Your test is validating that the error thrown is about the title length.

Unless you are trying to guarantee that whenever there are multiple parameter violations the title is ALWAYS validated first (which I don't think you are trying to guarantee), then you probably don't want to specify both an invalid title AND invalid content. Otherwise someone could change the order in which the title and content lengths are validated in the code, and this test would start failing, even though the behavior was still correct.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Data structure with simplified parameters that would be passed across the iframe
+   */
+  export interface NotificationDisplayParamForAppHost {
+    /**
+     * Notification title(maximum length: 20 characters)
+     */
+    title: string;
+    /**
+     * Notification content (maximum length: 40 characters)
```
AE-MS, line 48: >(maximum length: 40 characters) [](http://example.com/codeflow?start=28&length=31)

`maxContentLength` is set to 1500... but this says 40 is the maximum?

packages/teams-js/test/public/appNotification.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+          const callbackId = message.id;
+          utils.respondToFramelessMessage({
+            data: {
+              id: callbackId,
+              args: [{ errorCode: ErrorCode.INTERNAL_ERROR }],
+            },
+          } as DOMMessageEvent);
+          await expect(promise).rejects.toEqual({ errorCode: ErrorCode.INTERNAL_ERROR });
+        });
+      } else {
```
AE-MS, line 203: >else [](http://example.com/codeflow?start=8&length=4)

What does this `else` statement do here?

packages/teams-js/src/public/appNotification.ts
bernice-msft: Yes it should be! Updated that

packages/teams-js/test/public/appNotification.spec.ts
bernice-msft: No, I didn't get it from an approved list of fictitious names. Alright I didn't know that could come up as an issue, thanks for the enlightenment.

Alright, I'll take off the name.

packages/teams-js/test/public/appNotification.spec.ts
bernice-msft: It's the corresponding else statement for if the current frame context is included in the list of allowedContexts 
    line 65)  if (allowedContexts.some((allowedContext) => allowedContext === context)) { ....

change/@microsoft-teams-js-cc1d35f9-c5f8-4d2b-9161-2d3c53d6d593.json
bernice-msft: Alright,I'll take note of this for further commits.

packages/teams-js/src/public/appNotification.ts
bernice-msft: Okay, Are you also suggesting taking off the http as well?

packages/teams-js/src/public/appNotification.ts
bernice-msft: Yes it should be.
I updated this after the meeting, but I made this commit prior to the meeting.

packages/teams-js/src/public/appNotification.ts
bernice-msft: Alright, I'll reword this for better clarity.

bernice-msft: Does this sound better?

\**
* Data structure to represent appNotification information that would be sent across the iframe to the hubsdk
   * The notificationActionUrl in this data structure has been been serialized from a URL type to a string type for efficient transfer of data across the iframe
   */

packages/teams-js/src/public/appNotification.ts
bernice-msft: Yes, that sounds like a better idea!
I've implemented that change.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  /**
+   * Displays appNotification after making a validiity check on all of the required parameters, by calling the validateNotificationDisplayParams helper function
+   * An interface object containing all the required parameters to be displayed on the notification would be passed in here
+   * The notificationDisplayParam would be serialized before passing across to the message handler to ensure all objects passed contain simple parameters that would properly pass across the Iframe
+   * @param notificationdisplayparam - Interface object with all the parameters required to display an appNotificiation
+   * @returns a promise resolution upon conclusion
+   * @throws
```
AE-MS, line 161: >@throws [](http://example.com/codeflow?start=5&length=7)

You'll also need to actually document _what_ this throws, not just that it throws. You can look at other functions in this library as an example, e.g.,

```
     * @throws Error if {@linkcode app.initialize} has not successfully completed
```

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,187 @@
...
+    displayDurationInSeconds: number;
+    /**
+     * A url string type to the page in which the notification would direct the user to.
+     */
+    notificationActionUrlAsString: string;
+  }
+
+  /**
+   * This converts the notifcationActionUrl from a URL type to a string type for proper flow across the iframe
+   * @param notificationDisplayParam
```
AE-MS, line 70: >notificationDisplayParam [](http://example.com/codeflow?start=12&length=24)

You'll have to document the parameter is for, not just that it exists.

AE-MS (approved):

AE-MS (approved):

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,188 @@
...
+     */
+    displayDurationInSeconds: number;
+    /**
+     * A url link to the page in which the notification would direct the user to.
+     */
+    notificationActionUrl: URL;
+  }
+
+  /**
+   * Data structure to represent appNotification information that would be sent to the host SDK
```
lakhveerk, line 40: we can mark this internal & hidden. App developers don't need to know this exists

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,188 @@
...
+     */
+    notificationActionUrlAsString: string;
+  }
+
+  /**
+   * This converts the notifcationActionUrl from a URL type to a string type for proper flow across the iframe
+   * @param notificationDisplayParam - appNotification display parameter with the notificationActionUrl as a URL type
+   * @returns a serialized object that can be sent to the host SDK
+   */
+  function serializeParam(notificationDisplayParam: NotificationDisplayParam): NotificationDisplayParamForAppHost {
```
lakhveerk, line 73: same for this. Mark it @internal and @hidden

lakhveerk: We don't need to generate documentation on the functions/interfaces that app developers don't need. 
Tagging @TrevorJoelHarris to double check this.

packages/teams-js/src/public/appNotification.ts
```diff
@@ -0,0 +1,188 @@
...
+      throw new Error('Must supply notification display duration in seconds');
+    }
+
+    if (notificationDisplayParam.displayDurationInSeconds < 0) {
+      throw new Error('Notification display time must be greater than zero');
+    }
+  }
+
+  /**
+   * Displays appNotification after making a validiity check on all of the required parameters, by calling the validateNotificationDisplayParams helper function
```
lakhveerk, line 156: This is great information, but do we need this as part of function summary? Tagging @erikadoyle and @TrevorJoelHarris (I am sorry for multiple tags. Just want to make sure we are not doing documentation correctly :)

.vscode/settings.json
jadahiya-MSFT: I think this is just the personal `.vscode` config that got accidentally committed and should probably be removed

.vscode/settings.json
```diff
@@ -11,5 +11,5 @@
   "typescript.preferences.quoteStyle": "single",
   "prettier.singleQuote": true,
   "cSpell.words": ["adal", "uninitialize"],
-  "jest.jestCommandLine": "pnpm test -- --"
+  "jest.jestCommandLine": "pnpm test -- --",
```
lakhveerk, line 14: do you need this comma?

lakhveerk (approved):

AE-MS (approved):

PR #1839, opened 2023-07-12
[Reference Improvement] [secondary Browser + compatibility matrix]

mobajemu-msft:
add verbiage to secondary browser reference doc and update combability matrix in conceptual doc. linked to https://github.com/MicrosoftDocs/msteams-docs/pull/9015

### DIFF
--- a/packages/teams-js/src/public/secondaryBrowser.ts
+++ b/packages/teams-js/src/public/secondaryBrowser.ts
@@ -13,8 +13,8 @@ import { runtime } from './runtime';
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser, aka in-app browser. 
-   * The user will be redirected to an external browser on web and desktop
+   * Open a URL in the secondary browser (also known as in-app browser). 
+   * On web and desktop, the user will be redirected to an external browser.
    *
    * @param url Url to open in the browser
    * @returns Promise that successfully resolves if the URL  opens in the secondaryBrowser
--- a/packages/teams-js/src/public/secondaryBrowser.ts
+++ b/packages/teams-js/src/public/secondaryBrowser.ts
@@ -6,15 +6,16 @@ import { ErrorCode } from './interfaces';
 import { runtime } from './runtime';
 
 /**
- * Namespace to power up the in-app browser experiences in the Host App.
- * For e.g., opening a URL in the Host App inside a browser
+ * Namespace to power up the in-app browser experiences in the host app.
+ * For e.g., opening a URL in the host app inside a browser
  *
  * @beta
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser (also known as in-app browser). 
-   * On web and desktop, the user will be redirected to an external browser.
+   * Open a URL in the secondary browser 
+   * On mobile, this is the in-app browser
+   * On web and desktop, the user will be redirected to an external browser
    *
    * @param url Url to open in the browser
    * @returns Promise that successfully resolves if the URL  opens in the secondaryBrowser
--- a/packages/teams-js/src/public/secondaryBrowser.ts
+++ b/packages/teams-js/src/public/secondaryBrowser.ts
@@ -13,7 +13,8 @@ import { runtime } from './runtime';
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser aka in-app browser
+   * Open a URL in the secondary browser, aka in-app browser. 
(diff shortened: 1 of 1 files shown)


packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -13,7 +13,8 @@ import { runtime } from './runtime';
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser aka in-app browser
+   * Open a URL in the secondary browser, aka in-app browser.
```
erikadoyle, line 16: Let's avoid abbreviations like "aka". How about:

Opens a URL in the secondary browser (also known as in-app browser).

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -13,7 +13,8 @@ import { runtime } from './runtime';
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser aka in-app browser
+   * Open a URL in the secondary browser, aka in-app browser. 
+   * The user will be redirected to an external browser on web and desktop
```
erikadoyle, line 17: This sentence only applies to web/desktop:

On web and desktop, the user will be redirected to an external browser.

erikadoyle (changes requested):
Thanks @mobajemu-msft! Left a couple suggestions.

erikadoyle:
@mobajemu-msft, while you're in this file, would you mind sprucing up the namespace description too? Decapitalize "Host App" and use "For example," instead of "For e.g.,". @alicialu-MSFT is this description still correct:

```
* Namespace supporting in-app browser experiences of the host app.
* For example, open a URL inside a simplified browser within the host app.
```

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -13,7 +13,8 @@ import { runtime } from './runtime';
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser aka in-app browser
+   * Open a URL in the secondary browser (also known as in-app browser). 
+   * On web and desktop, the user will be redirected to an external browser.
```
AE-MS, line 17: Should this be written so that it uses a parallel structure for each case, rather than using a parenthetical to describe the mobile behavior and a separate sentence to describe web/desktop? e.g.,

   * Open a URL in a secondary browser
   * On mobile platforms, this is an in-app browser 
   * On web and desktop, the user will be redirected to an external browser.

(I don't have any idea if there is a preferred style here, just a thought I had while reading it).

AE-MS: While I'm at it, do we have specific terminology we use to refer to these platforms that users already understand?

I imagine most users would know what "mobile" refers to

I'm not sure if people know what "desktop" means? Would we ever say "large screen" instead?

What would users think iPad counts as?

Should we just use words like iOS, Android, Windows, Mac, and web instead of web, desktop, mobile, etc.?

I have no idea about any of this, just musing and asking questions. 😅

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/secondaryBrowser.ts
erikadoyle: Great idea (first comment) on the parallel sentence structure! Makes the differences more clear.

Desktop/Web/Mobile are pretty standard terms for tech industry content. Depending on the context, we've been enumerating individual platforms where it's helpful in Teams docs. Here in TJS reference, it's probably most pragmatic to use broad categories and rely on the [compatibility matrix](https://learn.microsoft.com/en-us/microsoftteams/platform/m365-apps/teamsjs-support-m365) to be the single source of truth on hub support details.

AE-MS (approved):
:shipit:

erikadoyle (approved):

TrevorJoelHarris:
Hi @mobajemu-msft ! It looks like this build isn't going to finish because you're trying to contribute from a fork of our repo. That *used* to be allowed, but for some boring reasons we don't allow builds in the pipeline anymore from forked repos.

In order to contribute this change, you'll need to clone the repo, make a branch in your local copy of our repo, and then push a new PR from there. There are more instructions here: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1853

Let me know if you have any problems!

mobajemu-msft:
new PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1866

PR #1852, opened 2023-07-19
Check for duplicate comment when running the runtime.ts check

jekloudaMSFT:

## Description

Adding a step to the runtime versioning check so that the GitHub bot won't keep adding the same comment to your PR every time it's updated.

### Main changes in the PR:

1. Check comments on PR for bot for the runtime change message before posting it.

### Validation performed:

1. Ran changes against this PR with multiple pushes. Only one comment was left by the github bot.

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

No

### DIFF
--- a/.github/workflows/runtimeCheck.yml
+++ b/.github/workflows/runtimeCheck.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            commentBody = 'This pull request contains changes to the runtime.ts file. If you, as the author of this PR, have made changes to the Runtime interface please review RUNTIME.md to determine if a new runtime version is required. Please reply to this comment stating what changes, if any, were made to the runtime object and whether a new runtime version was required.'
+            commentBody = 'This pull request contains changes to the runtime.ts file. If you, as the author of this PR, have made changes to the Runtime interface please review RUNTIME.md to determine if a new runtime version is required. Please reply to this comment stating what changes, if any, were made to the runtime object and whether a new runtime version was required.';
             comments = await github.rest.issues.listComments({
               owner: context.repo.owner,
               repo: context.repo.repo,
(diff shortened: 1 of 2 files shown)


jadahiya-MSFT (dismissed):

jekloudaMSFT:

Just added a newline to test the pipeline. I can remove it now if it shouldn't be there

.github/workflows/runtimeCheck.yml
```diff
@@ -20,6 +20,14 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
+            comments = await github.rest.issues.listComments({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number
+            });
+            if (comments.data.some(comment => comment.body.includes('This pull request contains changes to the runtime.ts file.'))) {
```
AE-MS, line 28: >'This pull request contains changes to the runtime.ts file.' [](http://example.com/codeflow?start=68&length=60)

Can you extract this statement into a single constant so it can be used both here and on line 35? Does YAML allow that?

KangxuanYe (approved):

AE-MS (changes requested):
:clock1:

.github/workflows/runtimeCheck.yml
KangxuanYe: Good question... That's something I haven't seen but could take a look

KangxuanYe: I think it could be defined as a variable(?) at the top of file and then be used later

jekloudaMSFT: Good suggestion, I'll give it a try

jekloudaMSFT: Ok, I just used a variable in the script, but they're using the same string now

AE-MS (approved):

TrevorJoelHarris (approved):

PR #1857, opened 2023-07-23
Update Location data for device permission and supporting different platforms

KangxuanYe:
## Description

Data is updated for supporting back-compatible E2E testing in iOS and Web Host SDK

### Main changes in the PR:

Separate testing json data under location capability for iOS and web platform.

The reason is that design of consent dialog between iOS and web are different. In web platform, there are only two options for consent dialog which are either true or false. So in the snippet below, by design,      
```
      "requestPermissionBeforeThisCall": {
        "boxSelector": "#box_requestGeoLocationPermission",
        "consentPermission": true
        }
```
`consentPermission` is set as a boolean type.

However, in mobile platform, there are three options for consent dialog, which are "Allow", "Not Allow" and "Remind Me Later". This cannot be designed in back-compat support as Web platform did by setting consentPermission as a boolean type value. Thus a new design comes out as below for mobile.
```
      "requestPermissionBeforeThisCall": {
        "boxSelector": "#box_requestGeoLocationPermission",
        "consentPermission": "Allow",
        "expectedOutput": "true"
      },
```

To be more specific, you could not apply the design for iOS to Web or for Web to iOS because it will not drive/build E2E testing correctly.


apps/teams-test-app/e2e-test-data/location.json
```diff
@@ -19,6 +20,7 @@
     {
       "title": "showLocation API Call - Success",
       "type": "callResponse",
+      "platformsExcluded": ["iOS"],
```
MengyiGong, line 23: Just curious, why we will exclude ios platform on this one? As I see we exclude the web platform on "getLocation API Call - Success"

KangxuanYe: Updated in description and please feel free to let me know if it makes sense.

MengyiGong (approved):

apps/teams-test-app/e2e-test-data/location.json
MengyiGong: LGTM and approved from my side. Would be great to include more approver from whoever familiar with teamsjs codebase.

lakhveerk: Do we need to exclude android as well?

KangxuanYe: Ideally yes, but for now, Android doesn't support back-compat so we don't care. There will be another person update here in the future for Android.

lakhveerk (approved):

PR #1866, opened 2023-08-01
Update secondaryBrowser.ts

mobajemu-msft:


AE-MS (approved):

AE-MS:
You'll have to create a changefile by running `pnpm changefile`

packages/teams-js/src/public/secondaryBrowser.ts
```diff
@@ -6,14 +6,16 @@ import { ErrorCode } from './interfaces';
...
- * For e.g., opening a URL in the Host App inside a browser
+ * Namespace to power up the in-app browser experiences in the host app.
+ * For e.g., opening a URL in the host app inside a browser
  *
  * @beta
  */
 export namespace secondaryBrowser {
   /**
-   * Open a URL in the secondary browser aka in-app browser
+   * Open a URL in the secondary browser
```
erikadoyle, line 16: Line breaks in source don't necessarily render in the published doc, so these three lines will probably publish as a run-on sentence. Could you add punctuation?

erikadoyle (changes requested):
Minor change I just noticed, otherwise looks good

change/@microsoft-teams-js-4885f2ff-afee-4756-9192-4731c184d1ba.json
TrevorJoelHarris, line 1: You probably shouldn't be deleting an existing changefile

TrevorJoelHarris (changes requested):
It looks like you're deleting someone else's change, which is probably not what you intended to do

change/@microsoft-teams-js-4885f2ff-afee-4756-9192-4731c184d1ba.json
mobajemu-msft: yea looks like something went wrong. thanks for the callout. will fix that

mobajemu-msft:
Third time is the charm #1889

PR #1878, opened 2023-08-08
Adding lifecycle handlers

lakhveerk:

## Description
This PR is adding `lifecycle` handlers to `app.ts` file. These handlers are duplicates of `teamsCoreAPI` handlers (`beforeUnload` and `onLoad`) with some changes. 
1. The callback in new handler `beforeResumeOrTerminate` does not need to return a Boolean value unlike beforeUnload
2. app-developers are required to call the handler readyToSuspndOrTerminate in beforeResumeOrTerminate unlike beforeUnload that gives allows app developers an option to not call the handler. 
3. On beforeResumeOrTerminate call, it does not send 'readyToSuspndOrTerminate' if the handler is not registered.

### Main changes in the PR:

1. Added an app.lifecycle subcapability and  app.lifecycle.caching subcapability to let users participate in full caching or delayed-termination.
4. Added the APIs to test-app
5. deprecate the old handlers named `registerOnLoadHandler` and `registerBeforeUnload` handler

### Validation performed:

1. <Step 1>
6. <Step 2>

### Unit Tests added: yes

<Yes/No>

### End-to-end tests added: No
The host sdk does not have neccessary changes yet to perform e2e tests

### Change file added: yes

<Yes/No>

### Related PRs:

> Remove this section if n/a


packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+
+    /**
+     * Checks if app.lifecycle is supported by the host.
+     * @returns boolean to represent whether the lifecycle capability is supported
+     * @throws Error if {@linkcode app.initialize} has not successfully completed
+     *
+     * @beta
+     */
+    export function isSupported(): boolean {
+      return ensureInitialized(runtime) && runtime.supports.app && runtime.supports.app.lifecycle ? true : false;
```
AE-MS, line 872: > ? true : false; [](http://example.com/codeflow?start=97&length=16)

I think you can probably get rid of the ? and everything after it?

lakhveerk: lifecycle has a subcapability `caching` in it. Without true/false, I would have to write something like `runtime.supports.app.lifecycle != undefined` which does not shorten the code. I am going to leave this as it is :)

packages/teams-js/src/internal/handlers.ts
```diff
@@ -18,8 +19,16 @@ class HandlersPrivate {
...
   public static themeChangeHandler: (theme: string) => void;
+  /**
+   * @deprecated
+   */
   public static loadHandler: (context: LoadContext) => void;
+  /**
+   * @deprecated
+   */
   public static beforeUnloadHandler: (readyToUnload: () => void) => boolean;
+  public static beforeSuspedOrTerminateHandler: (readyToSuspendOrTerminate: () => void) => void;
```
AE-MS, line 30: >beforeSuspedOrTerminateHandler [](http://example.com/codeflow?start=16&length=30)

I believe the letter "n" is missing in suspend

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+    }
+
+    /**
+     * Namespace to opt into full caching of an app
+     *
+     * @beta
+     */
+    export namespace caching {
+      /**
+       * Registers a handler to be called when the page has been requested to load.
```
AE-MS, line 882: >load [](http://example.com/codeflow?start=78&length=4)

Is it worth (and would it be correct) to expand on this to say "requested to resume from being suspended?"

Clearly there is a longer narrative needed to explain the app lifecycle, and I'm sure that we'll work with Erika to document and publish that, and then link to it from the code. So I'm not suggesting that the full flow needs to be documented here. But maybe a few details to improve the summary docs might help?

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+       *
+       * @param handler - The handler to invoke when the page is loaded.
+       *
+       * @beta
+       */
+      export function registerOnResumeHandler(handler: registerOnResumeHandlerFunctionType): void {
```
AE-MS, line 888: >registerOnResumeHandler [](http://example.com/codeflow?start=22&length=23)

Maybe we talked about this as part of the dev plan and I just forgot. From the perspective of the developer API, is there value in separating out this registration function into its own subcapability? Yes, there will be hosts which don't fully support the suspend/resume lifecycle, but since the only function here is the registration, and registering a handler in a hub which doesn't support it will just result it in the handler never getting called, can this just go directly into the `lifecycle` capability?

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+    export function isSupported(): boolean {
+      return ensureInitialized(runtime) && runtime.supports.app && runtime.supports.app.lifecycle ? true : false;
+    }
+
+    /**
+     * Namespace to opt into full caching of an app
+     *
+     * @beta
+     */
+    export namespace caching {
```
AE-MS, line 880: >caching [](http://example.com/codeflow?start=21&length=7)

Depending on how you feel about my comment below about the existence of this namespace all together, I wonder if this name would be better expressed as "resuming" or "suspension" or something, since you are modeling this as part of the app lifecycle (using appropriate terms like suspend and resume), and the developer-facing terminology moved away from "caching"

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+     * Registers a handler to be called before the page is suspended or terminatd.
+     *
+     * @param handler - The handler to invoke before the page is suspended or terminated. The page should
+     * invoke the readyToSuspendOrTerminate function provided to it once it's ready.
+     *
+     */
+    export function registerBeforeSuspendOrTerminateHandler(
+      handler: registerBeforeSuspendOrTerminateHandlerFunctionType,
+    ): void {
+      handler && ensureInitialized(runtime);
```
AE-MS, line 857: >handler && ensureInitialized(runtime); [](http://example.com/codeflow?start=6&length=38)

I think I might know the answer to this but why do we have to do this here? Doesn't `ensureInitialized` get called from within `isSupported` ?

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+      if (handler && !isSupported()) {
```
AE-MS, line 858: >handler [](http://example.com/codeflow?start=10&length=7)

Similarly to the earlier comment, why do we have to check for `handler` to be defined and non-null here?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/app.ts
lakhveerk: We divided caching into two: delayed termination and full caching. Hosts can choose to support only delayed termination (In web, they don't have full caching option) or full caching.
To trace that behavior to teams-js, we have 'lifecycle' that indicates delayed termination and subcapability 'caching' that is full caching. 
In teams-js, if we have both handlers in one capability and a host is supporting only delayed termination, what should isSupported function return: true or false? 
If we return true, app developer might get confused, because app is not being cached, while capability shows supported. 

However, you can also argue that app might not get cached for other reasons like memory limitation etc even though  hub supports full caching, and is-supported for subcapability 'caching' returns true. 

Last time in our caching sync, the votes for subcapability were higher. If you think nested subacapility is causing more harm than benefits, we are open to change it to one capability :)

packages/teams-js/src/public/app.ts
lakhveerk: We don't. Thank you for pointing. I changed the code to explicitly throw error when handler is not provided.

packages/teams-js/src/public/app.ts
lakhveerk: I am not sure the reason, but we have been calling the ensureInitialized and isSuuported for all (afaik) APIs. @TrevorJoelHarris  can you help us please?

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,100 @@ export namespace app {
...
+
+    /**
+     * Registers a handler to be called before the page is suspended or terminatd.
+     *
+     * @param handler - The handler to invoke before the page is suspended or terminated. The page should
+     * invoke the readyToSuspendOrTerminate function provided to it once it's ready.
+     *
+     */
+    export function registerBeforeSuspendOrTerminateHandler(
+      handler: registerBeforeSuspendOrTerminateHandlerFunctionType,
```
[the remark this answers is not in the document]
lakhveerk: We can abstractify the function and not expose the `readyToSuspendOrTerminate` at all. 
Once the suspendOrTerminate event is dispatched from host sdk, teams-js would trigger the `registerBeforeSuspendOrTerminateHandler` handler. Right next, teams-js will send `readyToSuspendOrTerminate` back to host. Since the `registerBeforeSuspendOrTerminateHandler` handler is synchronous,  `readyToSuspendOrTerminate` won't be sent until handler execution is done. This way, we can keep the registerHandler as of empty function type, so app developers won't need to deal with all the complexity.
`type registerBeforeSuspendOrTerminateHandlerFunctionType = () => void;`
 @erinha @alexneyman-MSFT

packages/teams-js/src/public/app.ts
lakhveerk: Yep, it makes sense to change the name to resuming. I am going to wait on our decision on whether or not we need this sub capability before making the change. Thank you for the suggestion! 😊

packages/teams-js/src/public/app.ts
lakhveerk: Yep talking to Erica about it is on my list!

packages/teams-js/src/public/app.ts
lakhveerk: I talked to Trevor, and the reason is,  

Hosts don't alter what they say are supported in the runtime object based on the current framecontext. Because of this, we have some capabilities where different combinations of functions work depending on what the frameContext is. 

For example, dialog.url.open, hosts always say they support dialog.url, no matter what the framecontext is, but if the app is in the tasks context, we don't want that function to work because the app shouldn't be able to make a dialog that opens another dialog.

packages/teams-js/src/public/app.ts
AE-MS: I swear I responded to this comment and it feels like it somehow got lost in GitHub but I don't see how that's possible.

Anyway I wrote something like this:
I believe you can write something like this:
```
return ensureInitialized(runtime) && !!(runtime.supports.app?.lifecycle);
```
...but I don't know if the `!!` notation is common enough that all readers would know what it was doing.

packages/teams-js/src/public/app.ts
AE-MS: I can see how that would apply if the parameters being passed to `ensureInitialized` were different (e.g., `dialog.bot.open` passes three frame contexts and `dialog.bot.isSupported` passes none). But at least here, the same frame contexts are being passed in `registerBeforeSuspendOrTerminateHandler` and in `isSupported`?

This isn't a big deal if you want to leave it this way.

packages/teams-js/src/public/app.ts
```diff
@@ -806,6 +816,98 @@ export namespace app {
...
+  export namespace lifecycle {
+    /**
+     * Register on resume handler function type
+     *
+     * @param context - Data structure to be used to pass the context to the app.
+     */
+    type registerOnResumeHandlerFunctionType = (context: ResumeContext) => void;
+
+    /**
+     * Register before suspendOrTerminate handler function type
```
AE-MS, line 838: >Register before suspendOrTerminate handler function type [](http://example.com/codeflow?start=7&length=56)

Should the comment here describe the expectation on this function? E.g., "When this function is called, your app should do any cleanup needed before suspension or termination. Upon returning, the app will be suspended or terminated." ?

packages/teams-js/src/public/app.ts
AE-MS: Maybe I'm misunderstanding what you are saying -- I don't see any code that throws when`handler` is null/undefined?

packages/teams-js/src/public/app.ts
AE-MS: First, I don't want to be in the business of overruling democratically elected capability names.

I will state my position and y'all, but do treat it as just one input.

Because caching is an optimization, no developer should ever be relying on it. Even in hosts where caching exists, as you say, it might not ever happen for a variety of reasons. As such, the `isSupported` method on the `caching` namespace (as you currently have it) is actually not helpful since it doesn't guarantee anything (other than `registerOnResumeHandler` won't throw an error). Since the act of registering a handler is a supported operation on all hosts, and there's no host where you can guarantee that it will be called, I didn't see the reason to increase developer and code complexity by putting it into its own subcapability, at least from the teamsjs perspective, since they aren't getting any value from that.

lakhveerk: Thank you for more feedback! We decided to combine both under one subCapability. The isSupported function will return true if hubs provide lifeCycleService.

packages/teams-js/src/public/app.ts
lakhveerk: We are deleting this extra sub-sub capability and combining both handlers under `lifecycle`. Closing this comment for that reason

packages/teams-js/src/public/app.ts
lakhveerk: I am sorry, I had to change it back because unit tests fail if we don't check for handler before calling isSupported. I think the reason is we are assigning null to the handlers in uninitialize call. I have opened another PR to fix it: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1915

packages/teams-js/src/public/app.ts
lakhveerk: Added the comment :)

packages/teams-js/src/public/app.ts
lakhveerk: This is fixed now.

packages/teams-js/src/public/app.ts
lakhveerk: deleted the ensure initialized call :)

PR #1891, opened 2023-08-22
API Telemetry Change - First Batch

KangxuanYe:
## Description

This PR is to make the change to help on differentiating old/deprecated and new APIs. Additional argument, which is called apiVersionTag, is set when Teams-Js library sends SDK requests to Hub SDKs. The old/deprecated API will be labelled as v1_[API_NAME], i.e. v1_getContext, and new API will then be labelled as v2_[API_NAME], i.e. v2_App.getContext. The v0_testing value is to indicate that the project for labelling API is ongoing and it is meaningless to indicate API is either new or old for now.

Some of core functions under communication.ts file are copied and edited to make sure the working flow to send out SDK requests with apiVersion will not affect the original flow, which makes the progress clean and easy to track.

Here are reasons for version upgrade in tag, which is to be shared internally:

1. New API introduced to perform same task in a new namespace, i.e. getContext -> app.getContext
2. New API introduced to break out tasks from a former namespace as sub-capabilities in a new namespace, i.e. task.startTask with `taskInfo` param and an optional `submitHandler` -> dialog.url.open
3. API changed the callback to a promise return type, i.e. authentication.getUser(userRequest: UserRequest): void - the deprecated function accepts callbacks as param and returns void, while the new function returns promise -> authentication.getUser(): Promise\<UserProfile\>

Most of API versions are controlled on file levels and there are some exceptions, which we mix V1 and V2 APIs in the same capability file. Please take a look on comment at the beginning part of each capability file and a rule for defining/updating ApiVersionNumber and ApiName will be raised in a PR soon.

### Main changes in the PR:

1. Updated core functions to send out SDK requests in communication.ts file
2. Labeled all of APIs under App, Task, Dialog, Location, GeoLocation, Navigation, Page capabilities.
3. Updated unit tests for navigation capabilities as helper function involved.

### Validation performed:

1. Please check changelog or internal resource (excel from PM) to determine which api is old/deprecated and which is new
2. For deprecated API, it should be labeled as v1_[API_NAME].
3. For new API, it should be labeled as v2_[API_NAME].

### Unit Tests added: Yes

Add some of unit tests for checking pattern of tags.


change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
TrevorJoelHarris, line 2: This shouldn't be a major change, major indicates that this is a breaking change for developers that will require them to make a code change on their side. I would this this is a patch since it is not changing developer facing functionality and won't require them to "do" anything.

change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "draft solution with optional apiVersion",
```
TrevorJoelHarris, line 3: This comment should be in the past tense and be written to communicate to external developers what has changed. Maybe something like "Added additional telemetry to `app`, `dialog`, <etc. etc.> capabilities"

change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
```diff
@@ -0,0 +1,7 @@
...
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
```
TrevorJoelHarris, line 5: this should be the email address associated with your github account.

packages/teams-js/src/internal/communication.ts
```diff
@@ -90,7 +90,7 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
     // Send the initialized message to any origin, because at this point we most likely don't know the origin
     // of the parent window, and this message contains no data that could pose a security risk.
     Communication.parentOrigin = '*';
-    return sendMessageToParentAsync<[FrameContexts, string, string, string]>('initialize', [
+    return sendMessageToParentAsyncWithVersion<[FrameContexts, string, string, string]>('v2', 'initialize', [
```
TrevorJoelHarris, line 93: Why is this v2? Is there a v1 veresion of this being passed from somewhere? Was surprised to see this as v2

packages/teams-js/src/internal/communication.ts
```diff
@@ -124,6 +124,14 @@ export function uninitializeCommunication(): void {
   CommunicationPrivate.promiseCallbacks = {};
 }
 
+/**
+ * @internal
```
TrevorJoelHarris, line 128: I know this file doesn't do good at documenting functions, but brief descriptions of what these functions are for would be really great!

packages/teams-js/src/internal/communication.ts
```diff
@@ -132,6 +140,20 @@ export function sendAndUnwrap<T>(actionName: string, ...args: any[]): Promise<T>
   return sendMessageToParentAsync(actionName, args).then(([result]: [T]) => result);
```
TrevorJoelHarris, line 140: Curious: should developers ever call this function after your change? Do we want all calls up to the host to be passing some form of version number? Maybe this is an intermediate step and eventually functions like this will be removed or altered to pass version numbers?

packages/teams-js/src/internal/communication.ts
```diff
@@ -169,6 +227,22 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
   });
 }
 
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function sendMessageToParentAsyncWithVersion<T>(
+  apiVersion: string,
```
TrevorJoelHarris, line 235: It would be good to do some validation here that the version is being passed in a consistent format: not empty or null/undefined, starts with a v, lower case, etc. Want to prevent someone who isn't you from adding a change in the future that passes in an apiVersion that is differently formatted which could cause problems downstream.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -157,7 +162,7 @@ export function registerHandlerHelper(
  */
 export function registerOnThemeChangeHandler(handler: (theme: string) => void): void {
   HandlersPrivate.themeChangeHandler = handler;
-  handler && sendMessageToParent('registerHandler', ['themeChange']);
+  handler && sendMessageToParentWithVersion('v2', 'registerHandler', ['themeChange']);
```
TrevorJoelHarris, line 165: Similar question to one I left in communication: why are these v2? Shouldn't they be v1 because they're the first version we're tracking telemetry of?

packages/teams-js/src/public/app.ts
```diff
@@ -714,7 +714,7 @@ export namespace app {
   export function getContext(): Promise<app.Context> {
     return new Promise<LegacyContext>((resolve) => {
       ensureInitializeCalled();
-      resolve(sendAndUnwrap('getContext'));
+      resolve(sendAndUnwrapWithVersion('v2', 'getContext'));
```
TrevorJoelHarris, line 717: same question: why not v1?

packages/teams-js/src/public/dialog.ts
```diff
@@ -101,7 +101,89 @@ export namespace dialog {
     }
   }
 
+  function updateResizeHelper(apiVersion = 'v1', dimensions: DialogSize): void {
```
TrevorJoelHarris, line 104: Can you explain what the benefit of moving these implementations out into helper funtions is? It doesn't seem to be adding much value that we wouldn't get by just adding apiVersion into the original implementations, and it makes the code harder to follow because the function signatures and implementations are in separate places.

packages/teams-js/src/public/tasks.ts
```diff
@@ -42,10 +42,10 @@ export namespace tasks {
       : undefined;
     if (taskInfo.card === undefined && taskInfo.url === undefined) {
       ensureInitialized(runtime, FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
-      sendMessageToParent('tasks.startTask', [taskInfo as DialogInfo], submitHandler);
+      sendMessageToParentWithVersion('v1', 'tasks.startTask', [taskInfo as DialogInfo], submitHandler);
```
TrevorJoelHarris, line 45: You're using v1 both here and in dialogs. Should dialogs be v2 so we can tell those calls apart from these ones?

packages/teams-js/src/public/pages.ts
```diff
@@ -198,9 +183,9 @@ export namespace pages {
         throw errorNotSupportedOnPlatform;
       }
       if (runtime.isLegacyTeams) {
-        resolve(sendAndHandleStatusAndReason('executeDeepLink', createTeamsAppLink(params)));
+        resolve(sendAndHandleStatusAndReasonWithVersion('v2', 'executeDeepLink', createTeamsAppLink(params)));
```
TrevorJoelHarris, line 186: Why do we need different version numbers here? I would assume the same version numbers would be fine since there are different function names being used.

packages/teams-js/test/public/navigation.spec.ts
```diff
@@ -38,9 +39,13 @@ describe('MicrosoftTeams-Navigation', () => {
     Object.values(FrameContexts).forEach((context) => {
       it(`navigation.returnFocus should successfully call pages.returnFocus when initialized with ${context} context`, async () => {
         await utils.initializeWithContext(context);
-        const pagesReturnFocus = jest.spyOn(pages, 'returnFocus');
+
+        const pagesUtilReturnFocusHelper = jest.spyOn(pagesUtil, 'returnFocusHelper');
         returnFocus(true);
-        expect(pagesReturnFocus).toHaveBeenCalled();
+        expect(pagesUtilReturnFocusHelper).toHaveBeenCalled();
+        // const pagesReturnFocus = jest.spyOn(pages, 'returnFocus');
```
TrevorJoelHarris, line 46: remove or uncomment this

TrevorJoelHarris: (Pretend I typed this for every instance in this file)

TrevorJoelHarris (changes requested):

change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
KangxuanYe: Thank you for letting me know. I will change it in next push. Do we have any documentation or somewhere defined what each term means (major, minor, patch)? I am just curious about what kind of change is defined as minor.

packages/teams-js/src/internal/communication.ts
KangxuanYe: @TrevorJoelHarris , the goal of this PR is to differentiate the versions of API by giving them 'apiVersion' labels. And remember very early ago we said that
```
No matter what API  it is, if it is used in Teams-Js library V1.X.Y, or in the other word, under Teams-Js library V2.0.0, it should be labeled as V1
```

'Initialize' is a funcName argument used in SDK request to help communicating between SDKs and Teams-Js library.

Here is my thought and please correct me because I am probably wrong. 'Initialize' represents API 'initialize' (public API) in Teams-Js library V1.X.Y or API 'App.Initialize' in Teams-Js library ^V2.0.0(V2.X.Y). So, we will assign 'V1' to 'Initialize' API as a public API, and thus code here, which is to send out request for 'App.Initialize', is assigned as 'V2'.

packages/teams-js/src/internal/communication.ts
KangxuanYe: Once this PR gets checked in, I would recommend developers who is developing new capability to call those functions to pass apiVersion number to Host SDKs. If it is possible, I would like to make it be "should" rather than "suggest to" (but it might request me to take a look on each PR in Teams-Js library to make sure developers use those functions who can send out version numbers)

For your last question, yes, this is an intermediate step. Even if developers who checked in capability without using functions in communication to send out api version numbers to Host SDKs, I will edit functions in that capability later, i.e. @MengyiGong 's VisualMedia Capability. At the end of this project, those functions written as 'send..........WithVersion...' will all get removed (or let's say renamed as original function name and remove original function to make sure all of APIs send out request with apiVersion number using the same name API)

packages/teams-js/src/internal/handlers.ts
KangxuanYe: @TrevorJoelHarris please take a look on comment that I left for 'initialize' function.

This place keeps the same logic. register* functions are public APIs in Teams-Js library V1.X.Y. Thus, those in Teams-Js library V1.X.Y will be labeled as V1. Then, in Teams-Js library V2.X.Y, those register related APIs are moved to each capability, i.e. registerFullScreenHandler API is moved to Pages capability, so, for now, it needs to be labeled as V2.

packages/teams-js/src/public/app.ts
KangxuanYe: 'GetContext' as public API will be labeled as V1 in Teams-Js library V1.X.Y
'GetContext' is then moved to App capability so it is V2 here starting from V2.0.0

packages/teams-js/src/public/dialog.ts
KangxuanYe: The code is moved out as a helper to resolve the difficulty of sending out different api version number.

In previous design, sendMessageToParent(WithVersion) API is embedded into updateResize function. And this function is used inside of Task capability.

KangxuanYe: @TrevorJoelHarris, those helper functions are not in the proper files. I have moved them from public/dialog file to internal/dialogUtil file. I think it will make more sense to people now because helpers are used in both Task capability and Dialog capability.

packages/teams-js/src/public/tasks.ts
KangxuanYe: I have made an update and now v1 is used here for tasks.startTask - with `taskInfo` param and an optional `submitHandler` - and for dialog.url.open (its replacement), it should be 'v2' now.

packages/teams-js/src/public/pages.ts
KangxuanYe: Ideally, navigateToApp is a 'v1' API but it calls executeDeepLink internally and executeDeepLink in Teams-Js library is a 'v2' API. Let's hold this discussion and I need to ask PM about this problem. Good catch! Thank you.

change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
KangxuanYe: Updated. Please take another look.

change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
KangxuanYe: Added. But it seems like in Mac I need to manually add this. It cannot relate to my email automatically, which is odd.

packages/teams-js/src/internal/communication.ts
KangxuanYe: I tried my best understand on those functions and write some small comments on APIs I added. Please take a look. And also do let me know if you want me to add for the others.

packages/teams-js/src/internal/communication.ts
KangxuanYe: I added a check function in run-time but I don't think it is in best practice. Let's start from this point and discuss how to improve that.

packages/teams-js/src/public/tasks.ts
```diff
@@ -47,14 +48,14 @@ export namespace tasks {
...
     } else if (taskInfo.card) {
       ensureInitialized(runtime, FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage);
-      sendMessageToParent('tasks.startTask', [taskInfo as DialogInfo], submitHandler);
+      sendMessageToParentWithVersion('v1', 'tasks.startTask', [taskInfo as DialogInfo], submitHandler);
     } else if (taskInfo.completionBotId !== undefined) {
-      dialog.url.bot.open(getBotUrlDialogInfoFromTaskInfo(taskInfo), dialogSubmitHandler);
+      botUrlOpenHelper(getBotUrlDialogInfoFromTaskInfo(taskInfo), dialogSubmitHandler);
```
maglims, line 56: This is purely from a person's point of view who doesn't interact with teamsjs regularly so this comment is mostly a personal opinions as headup :)
How do we feel to make apiVersion as required in botUrlOpenHelper (and also other helpers), meaning here we will also pass V1 instead of letting it pick up the default V1 when not specifying? The way I am correlating is: Tasks is an old capability, so anything coming out of it should be V1. If we pass v1 specially here, I don't have to go into botUrlOpenHelper to check what version being sent to it. It applies to other capabilities being specific about versioning.

on a related note, I found it is really helpful to be specific passing v1 in  Dialog.ts

packages/teams-js/src/public/dialog.ts
```diff
@@ -179,7 +159,7 @@ export namespace dialog {
         throw errorNotSupportedOnPlatform;
       }
 
-      sendMessageToParent('messageForParent', [message]);
+      sendMessageToParentWithVersion('v1', 'messageForParent', [message]);
```
maglims, line 162: is this v1?

packages/teams-js/src/internal/communication.ts
```diff
@@ -194,6 +311,58 @@ function waitForResponse<T>(requestId: number): Promise<T> {
...
+    args = argsOrCallback;
+  }
+
+  if (!isFollowApiVersionLabelFormat(apiVersion)) {
+    throw Error(
+      `apiVersion: ${apiVersion} passed in doesn't follow the pattern starting with 'v' followed by digits, please check.`,
+    );
+  }
+
+  // APIs with v0 represents beta changes haven't been implemented on them
```
maglims, line 357: This comment is really helpful.. and it also makes me to think:
how do you think of the idea that having an enum for versioning? (in there for v0 we can also clarify it as beta like what you are doing here)
That may reduce of chancing that people accidentally passing an bad format of apiversion.

packages/teams-js/src/internal/communication.ts
```diff
@@ -178,7 +293,9 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
  */
 export function sendMessageToParentAsync<T>(actionName: string, args: any[] | undefined = undefined): Promise<T> {
   return new Promise((resolve) => {
-    const request = sendMessageToParentHelper(actionName, args);
+    // APIs with v0 represents beta changes haven't been implemented on them
+    // Otherwise, minimum version number will be v1
+    const request = sendMessageToParentHelper('v0', actionName, args);
```
maglims, line 298: why this is v0? sorry I know you added comments on what v0 is, but I couldn't tell why it is v0 here.

packages/teams-js/src/public/dialog.ts
KangxuanYe: Yes, it doesn't have any other function (which we call it deprecated) performs the same task

packages/teams-js/src/public/pages.ts
KangxuanYe: @TrevorJoelHarris, could you please explain a little bit more what happened if runtime is legacy Teams? Are we sending out different requests based on runtime? I feel like we might need a clear definition here to figure out whether it is v1, v2, or something else.

packages/teams-js/src/public/tasks.ts
KangxuanYe: I think it makes more sense. Version number is much more important so we could tolerate a little bit inconvenience on writing code. Will make the change and force developers to specifically label the version for APIs.

packages/teams-js/src/internal/communication.ts
KangxuanYe: Yeah that is something we can do. I think I had an idea before but then probably forget about it. Thanks for reminding and will add it.

Also we have another layer of protection that if you pass in a value doesn't following the pattern 'v(digits)', you will fail on build because there is a check when sdk request is to be sent out.

packages/teams-js/src/internal/communication.ts
KangxuanYe: I think you got the answer in other places by reading my comment?

packages/teams-js/src/internal/communication.ts
KangxuanYe: Updated with an ApiVersion enum to set version number.

packages/teams-js/src/public/app.ts
```diff
@@ -761,15 +635,15 @@ export namespace app {
...
   export function notifyAppLoaded(): void {
     ensureInitializeCalled();
-    sendMessageToParent(Messages.AppLoaded, [version]);
+    sendMessageToParentWithVersion(ApiVersion.V_2, Messages.AppLoaded, [version]);
   }
 
   /**
    * Notifies the frame that app initialization is successful and is ready for user interaction.
    */
   export function notifySuccess(): void {
```
ndangudubiyyam, line 644: Since you are doing the mirror API's (old vs new) shouldn't this PR include the V1 version of these handlers in appInitializationts file?

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -149,3 +149,16 @@ export function processAdditionalValidOrigins(validMessageOrigins: string[]): vo
...
+
+/**
+ * @hidden
+ * Check if apiVersion developer sends follows the pattern starting with a lowercase 'v' and it is then
+ * followed by one or more digits only. If yes, return true. Otherwise, return false.
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function isFollowApiVersionLabelFormat(apiVersion: string): boolean {
```
ndangudubiyyam, line 161: It would be nice if you can add unit tests for all the Util files you introduced for code coverage purposes.

packages/teams-js/src/public/pages.ts
```diff
@@ -199,9 +184,9 @@ export namespace pages {
         throw errorNotSupportedOnPlatform;
       }
       if (runtime.isLegacyTeams) {
-        resolve(sendAndHandleStatusAndReason('executeDeepLink', createTeamsAppLink(params)));
+        resolve(sendAndHandleStatusAndReasonWithVersion(ApiVersion.V_3, 'executeDeepLink', createTeamsAppLink(params)));
```
ndangudubiyyam, line 187: nit: How about V_21 instead of V3, it is a call from the version 2 but running in legacy teams (which is old in some sense). I think V3 should be reserved for future revisions rather than address peculiar flow.

KangxuanYe: Not a nit. It is a unique 1-1 matching and 21 will mess the order of numbering.

packages/teams-js/src/public/app.ts
KangxuanYe: Yes, APIs will be gone through file by file. But for this batch, let's just focus APIs under app file and some of them under publicAPIs file (because some changes in app will affect them)

packages/teams-js/src/internal/internalAPIs.ts
KangxuanYe: Test case added.

I don't think rest of util files need to add cases because they are just moved there from part of original APIs. In the other word, the original test cases have already covered them.

ndangudubiyyam (approved):

packages/teams-js/src/internal/communication.ts
```diff
@@ -587,3 +771,15 @@ function createMessageEvent(func: string, args?: any[]): MessageRequest {
     args: args || [],
   };
 }
+
+/**
```
maglims, line 775: how do you think to put this in a util class?

packages/teams-js/src/internal/constants.ts
```diff
@@ -190,3 +190,98 @@ export const errorRuntimeNotSupported = 'The runtime version is not supported';
 
 /** @hidden */
 export const errorCallNotStarted = 'The call was not properly started';
+
+/**
```
maglims, line 194: How about call this file something like: TelemetryConstants.ts?

packages/teams-js/src/internal/communication.ts
KangxuanYe: @TrevorJoelHarris  same idea here, would you want me to move it to another file which might be called telemetryUtil.ts?

packages/teams-js/src/internal/constants.ts
KangxuanYe: I personally agree with that. Do you have any thoughts? @TrevorJoelHarris

packages/teams-js/src/internal/appUtil.ts
```diff
@@ -0,0 +1,162 @@
+import * as Handlers from '../internal/handlers'; // Conflict with some names
```
TrevorJoelHarris, line 1: The functions in these new util files I think would be better to leave into the original files (app.ts, dialog.ts, pages.ts), and we should just do whatever we need to to not export them for external customers. I totally get why you moved them to new utils files (they're shared between multiple files and putting them here makes it easy to not export them outside of the library), but files like tasks are eventually going to get deleted, at which point we wouldn't really want functionality to live in a utils file that was only called by one file (dialog.ts)

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: Talked with Kangxuan, my understanding was correct

packages/teams-js/src/internal/constants.ts
TrevorJoelHarris: These should actually also go in `telemetry.ts`

packages/teams-js/src/internal/handlers.ts
```diff
@@ -94,14 +101,38 @@ export function callHandler(name: string, args?: unknown[]): [true, unknown] | [
   }
 }
 
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function registerHandlerWithVersion(
```
jekloudaMSFT, line 108: Is it safe to assume that all of the "_____WithVersion" functions in this PR will replace the original implementations in a future iteration?

packages/teams-js/src/internal/constants.ts
```diff
@@ -190,3 +190,98 @@ export const errorRuntimeNotSupported = 'The runtime version is not supported';
...
+ * Limited to Microsoft-internal use
+ */
+export enum ApiVersionNumber {
+  V_0 = 'v0',
+  V_1 = 'v1',
+  V_2 = 'v2',
+  V_3 = 'v3',
+}
+
+export enum ApiName {
```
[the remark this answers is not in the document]
jekloudaMSFT: This might be the same question, but are these meant to correspond to the teams-js name of the function or the function name that gets sent up to the host?

change/@microsoft-teams-js-24933635-de42-4b04-b53f-935c1799929a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added additional apiVersionTag argument on App, Dialog, GeoLocation, Location, Navigation, Pages, Tasks capabilities for telemetry",
```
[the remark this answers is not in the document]
KangxuanYe: Sounds good, updated! Thanks for your suggestion.

packages/teams-js/src/internal/appUtil.ts
KangxuanYe: functions under [app/pages/dialog]Utils are moved back to original files but abstract as helper functions still.

packages/teams-js/src/internal/communication.ts
KangxuanYe: This function is moved to telemetry.ts file as support function for telemetry.

packages/teams-js/src/internal/constants.ts
KangxuanYe: They are moved to telemetry.ts

packages/teams-js/src/internal/constants.ts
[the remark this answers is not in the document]
KangxuanYe: It's to be corresponded to the teams-js name.

For example. we have dialog.url.open API in teams-js library but function name sent to hub sdks is tasks.startTask.

There might be easy to remember by one sentence: function names sent to Hub SDKs are aligned with API names in Teams-Js library V1 so for Teams-Js library V2, they are not necessarily required to be the same, and under most scenarios, they are different.

packages/teams-js/src/internal/handlers.ts
KangxuanYe: No it will not replace the original implementation but "imagining all of APIs gonna use these. xxxxxWithVersion APIs, and then I will revert names back to the one without 'WithVersion'., which means those xxxxxWithVersion functions will no longer exist and original functions have no differences except there is an additional apiVersionTag parameter added to be passed in.

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -149,3 +149,17 @@ export function processAdditionalValidOrigins(validMessageOrigins: string[]): vo
...
+/**
+ * @hidden
+ * Check if apiVersionTag developer sends follows the pattern starting with a lowercase 'v', then
+ * followed by one or more digits, then concated with underscore and some characters to indicate api name.
+ * For example, 'v2_app.getContext'. If yes, return true. Otherwise, return false.
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function isFollowApiVersionTagFormat(apiVersionTag: string): boolean {
```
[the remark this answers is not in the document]
KangxuanYe: Suggestion accepted. Thanks.

packages/teams-js/src/internal/constants.ts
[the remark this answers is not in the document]
jekloudaMSFT: So, if I may try restating what you wrote: the values in the enum ApiName correspond to the function names in teams-js, so, like Trevor said, if we update a function name in teams-js, we should update the ApiName as well, right? So if I have an api `myModule.doSomethingCool()`, then the ApiName should be "myModule.doSomethingCool". Then, if I rename it `myModule.doSomethingCooler()`, then the ApiName should be "myModule.doSomethingCooler", right? Then would the resulting version tags be "v1_myModule_doSomethingCool" and "v1_myModule_doSomethingCooler"?

packages/teams-js/src/internal/handlers.ts
```diff
@@ -154,9 +213,9 @@ export function registerHandlerHelper(
  * @internal
  * Limited to Microsoft-internal use
  */
-export function registerOnThemeChangeHandler(handler: (theme: string) => void): void {
+export function registerOnThemeChangeHandler(apiVersionTag: string, handler: (theme: string) => void): void {
   HandlersPrivate.themeChangeHandler = handler;
-  !isNullOrUndefined(handler) && sendMessageToParent('registerHandler', ['themeChange']);
+  !isNullOrUndefined(handler) && sendMessageToParentWithVersion(apiVersionTag, 'registerHandler', ['themeChange']);
```
TrevorJoelHarris, line 218: Why doesn't this hardcode V_2 the way that the other handlers right below this do? In the existing code this handler is only called from one place

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -149,3 +149,17 @@ export function processAdditionalValidOrigins(validMessageOrigins: string[]): vo
   });
   GlobalVars.additionalValidOrigins = combinedOriginUrls;
 }
+
+/**
+ * @hidden
+ * Check if apiVersionTag developer sends follows the pattern starting with a lowercase 'v', then
+ * followed by one or more digits, then concated with underscore and some characters to indicate api name.
```
TrevorJoelHarris, line 156:
```suggestion
 * followed by one or more digits, then concatenated with underscore and some characters to indicate api name.
```

packages/teams-js/src/internal/internalAPIs.ts
```diff
@@ -149,3 +149,17 @@ export function processAdditionalValidOrigins(validMessageOrigins: string[]): vo
...
+ * For example, 'v2_app.getContext'. If yes, return true. Otherwise, return false.
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function isFollowingApiVersionTagFormat(apiVersionTag: string): boolean {
```
TrevorJoelHarris, line 162: Would it make sense to also put this into the `telemetry.ts` file?

packages/teams-js/src/public/app.ts
```diff
@@ -2,18 +2,23 @@
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+/**
+ * v2 APIs telemetry file: All of APIs in this capability file should send out API version v2 ONLY
```
TrevorJoelHarris, line 6: It would help enforce this if you made a const for the version number to use in this file and then references that const everyone you say V_2 right now.

packages/teams-js/src/public/app.ts
```diff
@@ -33,6 +38,146 @@ import {
...
 import { version } from './version';
 
+/**
+ * Number of milliseconds we'll give the initialization call to return before timing it out
+ */
+const initializationTimeoutInMs = 5000;
+
+const appLogger = getLogger('app');
+
+export function appInitializeHelper(apiVersionTag: string, validMessageOrigins?: string[]): Promise<void> {
```
TrevorJoelHarris, line 48: What other files are calling this function instead of the main `app.initialize` function?

TrevorJoelHarris: (I ask because we only need to move this helper out of the namespace and export it if another file internally needs to call it with a different api version)

packages/teams-js/src/public/app.ts
TrevorJoelHarris: Pretend I typed this for every file with a comment like this

packages/teams-js/src/internal/handlers.ts
KangxuanYe: Yes, there is another call comes from publicAPIs if you take a look on line 112 of publicAPIs.ts file.

packages/teams-js/src/public/geoLocation.ts
```diff
@@ -46,7 +50,14 @@ export namespace geoLocation {
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
-    return sendAndHandleSdkError('location.getLocation', { allowChooseLocation: false, showMap: false });
+    return sendAndHandleSdkErrorWithVersion(
+      getApiVersionTag(ApiVersionNumber.V_2, ApiName.GeoLocation_GetCurrentLocation),
```
TrevorJoelHarris, line 54: Should this file also be one of the "every function in this file is a V_2 function" comment files?

packages/teams-js/src/internal/internalAPIs.ts
KangxuanYe: marked as will be updated in next commit.

packages/teams-js/src/internal/internalAPIs.ts
KangxuanYe: Okay. I rethink of it and I think it worth to say we need to put everything related to apiVersionTag also to under telemetry.ts because that is it serves for. Will updated in next commit .

packages/teams-js/src/public/app.ts
TrevorJoelHarris:
```
const appTelemetryVersion: ApiVersionNumber = ApiVersionNumber.V_2;
```
Then you would use `appTelemetryVersion` instead of `ApiVersionNumber.V_2` below. That would make it a lot harder for other people in the future to add the "wrong" version. If they go look at another function in the same file they will see that they are using `appTelemetryVersion` and use that one too.

packages/teams-js/src/internal/handlers.ts
TrevorJoelHarris: Makes sense, v_1 is from publicAPI.ts and v_2 is from app.ts. Thanks!

TrevorJoelHarris (approved):

ndangudubiyyam (approved):

PR #1892, opened 2023-08-23
Trharris/communicationlogs

TrevorJoelHarris:
## Description

Right now, if incoming postMessages to teamsjs are ignored, the library are ignored we do very little to help engineers investigate why that is. This PR is an attempt to add some helpful logging that will point them in the right direction

### Main changes in the PR:

1. Added logging to communication.ts, specifically in the parts that are used for processing incoming post messages.
2. Added logging to utils.ts, specifically in the function that is used to validate origins.

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -276,7 +276,7 @@ const processMessageLogger = communicationLogger.extend('processMessage');
 function processMessage(evt: DOMMessageEvent): void {
   // Process only if we received a valid message
   if (!evt || !evt.data || typeof evt.data !== 'object') {
-    processMessageLogger('Unrecognized message format received by app, message being ignored');
+    processMessageLogger('Unrecognized message format received by app, message being ignored. Message: %o', evt);
     return;
   }
 
@@ -287,7 +287,7 @@ function processMessage(evt: DOMMessageEvent): void {
   const messageOrigin = evt.origin || (evt.originalEvent && evt.originalEvent.origin);
   if (!shouldProcessMessage(messageSource, messageOrigin)) {
     processMessageLogger(
-      'Message being ignored by app because it is not coming from a different window with a valid origin',
+      'Message being ignored by app because it is either coming from the current window or a different window with an invalid origin',
     );
     return;
   }
@@ -316,7 +316,7 @@ function shouldProcessMessage(messageSource: Window, messageOrigin: string): boo
   // Process if message source is a different window and if origin is either in
   // Teams' pre-known whitelist or supplied as valid origin by user during initialization
   if (Communication.currentWindow && messageSource === Communication.currentWindow) {
-    shouldProcessMessageLogger('Should not process message because it is coming from the same window');
+    shouldProcessMessageLogger('Should not process message because it is coming from the current window');
     return false;
   } else if (
     Communication.currentWindow &&
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -267,6 +267,8 @@ function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequ
   return request;
 }
 
+const processMessageLogger = communicationLogger.extend('processMessage');
+
 /**
  * @internal
  * Limited to Microsoft-internal use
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -267,21 +267,28 @@ function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequ
...
+const processMessageLogger = communicationLogger.extend('processMessage');
+
 /**
  * @internal
  * Limited to Microsoft-internal use
  */
 function processMessage(evt: DOMMessageEvent): void {
   // Process only if we received a valid message
   if (!evt || !evt.data || typeof evt.data !== 'object') {
+    processMessageLogger('Unrecognized message format received by app, message being ignored');
```
AE-MS, line 279: >'Unrecognized message format received by app, message being ignored' [](http://example.com/codeflow?start=25&length=68)

Could we log the evt object here to help aid in debugging?

packages/teams-js/src/internal/communication.ts
```diff
@@ -267,21 +267,28 @@ function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequ
...
 
   // Process only if the message is coming from a different window and a valid origin
-  // valid origins are either a pre-known
+  // valid origins are either a pre-known origin or one specified by the app developer
+  // in their call to app.initialize
   const messageSource = evt.source || (evt.originalEvent && evt.originalEvent.source);
   const messageOrigin = evt.origin || (evt.originalEvent && evt.originalEvent.origin);
   if (!shouldProcessMessage(messageSource, messageOrigin)) {
+    processMessageLogger(
+      'Message being ignored by app because it is not coming from a different window with a valid origin',
```
AE-MS, line 290: >a valid origin' [](http://example.com/codeflow?start=90&length=15)

Could we log the origin it is coming from here to help aid in debugging?

AE-MS, line 290: >not coming from a different window with a valid origin [](http://example.com/codeflow?start=50&length=54)

For some reason I had a hard time parsing this message. Is this the intent:
"Message being ignore by app because it is either coming from this own window or it's coming from a different window but with an invalid origin" ?

AE-MS: (reading further, maybe this will be logged by some of the other log statements?)

packages/teams-js/src/internal/utils.ts
```diff
@@ -36,13 +37,16 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
...
+const validateOriginLogger = getLogger('validateOrigin');
+
 /**
  * @internal
  * Limited to Microsoft-internal use
  */
 export function validateOrigin(messageOrigin: URL): boolean {
   // Check whether the url is in the pre-known allowlist or supplied by user
   if (!isValidHttpsURL(messageOrigin)) {
+    validateOriginLogger('Origin %s is invalid because it is not using https protocol', messageOrigin);
```
AE-MS, line 49: >not using https protocol [](http://example.com/codeflow?start=61&length=24)

Perhaps this is pointless because "not https" basically means "http" but you could consider logging `messageOrigin.protocol`

packages/teams-js/src/internal/utils.ts
```diff
@@ -58,6 +62,10 @@ export function validateOrigin(messageOrigin: URL): boolean {
     }
   }
 
+  validateOriginLogger(
+    'Origin %s is invalid because it is not an origin approved by this library or the call to app.initialize',
```
AE-MS, line 66: >an origin approved by this library [](http://example.com/codeflow?start=44&length=34)

Not sure if this is the place to do it, but could we/do we dump the valid origins to the log stream at some point to aid in debugging?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: Yes, should be captured elsewhere

TrevorJoelHarris: Yes, will update the language

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: Done

packages/teams-js/src/internal/utils.ts
TrevorJoelHarris: I don't think it serves much of a purpose, but it also doesn't hurt and is easy to add so I did

packages/teams-js/src/internal/utils.ts
TrevorJoelHarris: No problem. Also added the list of origins that were passed to app.initialize

AE-MS (approved):
:shipit:

shrshindeMSFT (approved):
🚢

PR #1895, opened 2023-08-25
registerTextureStream before getTextureStream to avoid frame lost

WengHe:
## Description

Call `registerTextureStream` before getting an input texture stream, with this change we can ensure that when we get an input texture stream, the output stream can be returned to the client immediately, otherwise, the first few frames may be lost if the output stream is not registered yet.

### DIFF
--- a/change/@microsoft-teams-js-b3109727-ff31-4dcb-bafc-b21748d8ce7b.json
+++ b/change/@microsoft-teams-js-b3109727-ff31-4dcb-bafc-b21748d8ce7b.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Changed the other of registering texture stream",
+  "packageName": "@microsoft/teams-js",
+  "email": "wenghe@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-b3109727-ff31-4dcb-bafc-b21748d8ce7b.json
+++ b/change/@microsoft-teams-js-b3109727-ff31-4dcb-bafc-b21748d8ce7b.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Changed the other of registering texture stream",
+  "comment": "Fixed issue that resulted in dropping first few frames when using the video capability",
   "packageName": "@microsoft/teams-js",
   "email": "wenghe@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/teams-js/src/internal/videoUtils.ts
+++ b/packages/teams-js/src/internal/videoUtils.ts
@@ -74,7 +74,7 @@ export async function processMediaStream(
   pipeVideoSourceToGenerator(
     await getInputVideoTrack(streamId, notifyError, videoPerformanceMonitor),
     new DefaultTransformer(notifyError, videoFrameHandler),
-    generator,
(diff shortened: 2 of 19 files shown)


packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -124,28 +130,38 @@ async function getInputVideoTrack(
...
+ * The function to create a MediaStreamTrack generator.
+ * The generator can then get the processed frames as media stream source.
+ * The generator can be registered back to the media stream so that the host can get the processed frames.
+ */
+function createProcessedStreamGeneratorWithoutSource(): MediaStreamTrack {
+  if (inServerSideRenderingEnvironment()) {
+    throw errorNotSupportedOnPlatform;
+  }
+  const MediaStreamTrackGenerator = window['MediaStreamTrackGenerator'];
+  const generator = new MediaStreamTrackGenerator({ kind: 'video' });
```
lubobill1990, line 143: return directly

change/@microsoft-teams-js-b3109727-ff31-4dcb-bafc-b21748d8ce7b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Changed the other of registering texture stream",
```
lubobill1990, line 3: The comment is too vague.

lubobill1990 (approved):

TrevorJoelHarris:
It seems like there should be some tests that verify that this change is working correctly

TrevorJoelHarris (changes requested):

AE-MS (changes requested):
Just waiting for other feedback to be addressed
:clock1:

WengHe:
> It seems like there should be some tests that verify that this change is working correctly

The change is internal, all the UTs are passed, and I did an end to end test by using the new package in a video app and verified it works well.

packages/teams-js/src/internal/videoUtils.ts
```diff
@@ -124,28 +130,38 @@ async function getInputVideoTrack(
...
-  }
+  generator: MediaStreamTrack,
+): void {
   const MediaStreamTrackProcessor = window['MediaStreamTrackProcessor'];
   const processor = new MediaStreamTrackProcessor({ track: videoTrack });
   const source = processor.readable;
-  const MediaStreamTrackGenerator = window['MediaStreamTrackGenerator'];
-  const generator = new MediaStreamTrackGenerator({ kind: 'video' });
-  const sink = generator.writable;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
WengHe: updated, now we don't need it any more

change/@microsoft-teams-js-b3109727-ff31-4dcb-bafc-b21748d8ce7b.json
WengHe: updated

packages/teams-js/src/internal/videoUtils.ts
WengHe: updated

TrevorJoelHarris (approved):

AE-MS (approved):

PR #1898, opened 2023-08-27
Code update for android and macOS

shrshindeMSFT:

## Description

While completing the MacOS and Android feature for clipboard, there were two issues one with `blob` type in Android API 27 as it was not able to read the `blob.text()` function, so this PR resolves it by using `FileReader` instead of blob functions, second one was related to MacOS platform which is the communication should be frameless rather than framed added a check in the `clipboard` api.

### Main changes in the PR:

1. Update Test App to use `FileReader` to read data from `blob` type.
2. Add macOS to clipboard api check to receive frameless messages.

### Unit Tests added:

Yes

### End-to-end tests added:

N/A

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/public/clipboard.ts
+++ b/packages/teams-js/src/public/clipboard.ts
@@ -1,7 +1,8 @@
 import { sendAndHandleSdkError } from '../internal/communication';
+import { GlobalVars } from '../internal/globalVars';
 import { ensureInitialized, isHostClientMobile } from '../internal/internalAPIs';
 import * as utils from '../internal/utils';
-import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
+import { errorNotSupportedOnPlatform, FrameContexts, HostClientType } from './constants';
 import { ClipboardParams, ClipboardSupportedMimeType } from './interfaces';
 import { runtime } from './runtime';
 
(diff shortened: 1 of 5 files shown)


packages/teams-js/src/public/clipboard.ts
```diff
@@ -54,7 +55,7 @@ export namespace clipboard {
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
-    if (isHostClientMobile()) {
+    if (isHostClientMobile() || GlobalVars.hostClientType === HostClientType.macos) {
```
Jessica-Koch, line 58: is macOS being handled in a future PR? (sorry if you've already answered this)

Jessica-Koch (approved):

lakhveerk (approved):
I think, you need to mark every function @beta along with @beta on top of file. Check with Trevor or Erika. :) Otherwise, changes look good to me :)

packages/teams-js/src/public/clipboard.ts
shrshindeMSFT: This change is just to add a check for macOS to the platform. I thought that macOS will be `framed` but actually it uses same as `iOS` which is `frameless` post messages.

PR #1905, opened 2023-09-01
add monitor to videoEx videoFrameHandler

WengHe:
## Description

Add monitor to videoEx frame handler to know the video frame processing performance.

### Unit Tests added:

Yes

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/private/videoEx.ts
+++ b/packages/teams-js/src/private/videoEx.ts
@@ -201,26 +201,12 @@ export namespace videoEx {
           'video.startVideoExtensibilityVideoStream',
           async (mediaStreamInfo: { streamId: string; metadataInTexture?: boolean }) => {
             const { streamId, metadataInTexture } = mediaStreamInfo;
-            const generator = metadataInTexture
-              ? await processMediaStreamWithMetadata(
-                  streamId,
-                  videoPerformanceMonitor
-                    ? createMonitoredVideoFrameHandler(parameters.videoFrameHandler, videoPerformanceMonitor)
-                    : parameters.videoFrameHandler,
-                  notifyError,
-                  videoPerformanceMonitor,
-                )
-              : await processMediaStream(
-                  streamId,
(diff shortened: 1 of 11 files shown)


packages/teams-js/src/private/videoEx.ts
```diff
@@ -244,6 +249,20 @@ export namespace videoEx {
     }
   }
 
+  function createMonitoredVideoFrameHandler(
+    videoFrameHandler: VideoFrameHandler,
+    videoPerformanceMonitor?: VideoPerformanceMonitor,
```
TrevorJoelHarris, line 254: where are you passing videoPerformanceMonitor into this function? It doesn't seem like you have anywhere that this is actually being specified

packages/teams-js/src/private/videoEx.ts
```diff
@@ -244,6 +249,20 @@ export namespace videoEx {
     }
   }
 
+  function createMonitoredVideoFrameHandler(
```
TrevorJoelHarris, line 252: the name of this function seems misleading if the monitor is optional. It seems like it could result in instances like what you have here, where you call a function called createMonitoredVideoFrameHandler but the handler you are making is actually unmonitored because no monitor was passed in.

packages/teams-js/src/private/videoEx.ts
TrevorJoelHarris: I wonder if this function is getting confused and using the one defined as a const up at the top of the file? Either way, something confusing going on.

TrevorJoelHarris (changes requested):

packages/teams-js/src/private/videoEx.ts
WengHe: updated, make it required

TrevorJoelHarris (approved):

shrshindeMSFT (approved):
🚢 LGTM!

PR #1915, opened 2023-09-07
Lakhveer/uninitialize handlers

lakhveerk:

## Description

Uninitalize handlers without calling handler APIs with null. 

### Main changes in the PR:

1. Wrote uninitiliazeHandlers function in handlers file to use in UTs
2. changed _uninitialize function to call uninitiliazeHandlers instead of calling handlers with 'null'
3. Reset all the registered handlers
4. Fixed UT in pages

### DIFF
--- a/packages/teams-js/src/internal/handlers.ts
+++ b/packages/teams-js/src/internal/handlers.ts
@@ -20,26 +20,40 @@ class HandlersPrivate {
   public static themeChangeHandler: (theme: string) => void;
   public static loadHandler: (context: LoadContext) => void;
   public static beforeUnloadHandler: (readyToUnload: () => void) => boolean;
+
+  /**
+   * @internal
+   * Limited to Microsoft-internal use
+   * Initializes the handlers.
+   */
+  public static initializeHandlers(): void {
+    // ::::::::::::::::::::MicrosoftTeams SDK Internal :::::::::::::::::
+    HandlersPrivate.handlers['themeChange'] = handleThemeChange;
+    HandlersPrivate.handlers['load'] = handleLoad;
+    HandlersPrivate.handlers['beforeUnload'] = handleBeforeUnload;
+    pages.backStack._initialize();
+  }
+
+  /**
+   * @internal
+   * Limited to Microsoft-internal use
+   * Uninitializes the handlers.
+   */
+  public static uninitializeHandlers(): void {
+    HandlersPrivate.handlers = {};
+    HandlersPrivate.themeChangeHandler = null;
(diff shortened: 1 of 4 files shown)


packages/teams-js/src/public/app.ts
```diff
@@ -687,26 +685,7 @@ export namespace app {
       return;
     }
 
-    if (GlobalVars.frameContext) {
```
lakhveerk, line 690: I am not sure, why we were un-initializing (setting null) only for some of the handlers. For example, all the handlers in remoteCamera, meetingRoom, videoEx etc are not reset.
In addition, some handlers are reset only during certain frameContexts. e.g. registerOnSaveHandler on L701 below.

AE-MS:
OMG I love this change

packages/teams-js/src/public/app.ts
AE-MS: I'm willing to bet that these were just oversights and not part of some master plan.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -33,6 +33,12 @@ export function initializeHandlers(): void {
   HandlersPrivate.handlers['beforeUnload'] = handleBeforeUnload;
   pages.backStack._initialize();
 }
+export function uninitializeHandlers(): void {
+  HandlersPrivate.handlers = {};
+  HandlersPrivate.themeChangeHandler = null;
+  HandlersPrivate.loadHandler = null;
+  HandlersPrivate.beforeUnloadHandler = null;
```
AE-MS, line 40: Should the implementation of clearing the handler state live as a class method on the `HandlersPrivate` class rather than inside this function? You know, encapsulation/abstraction all that good stuff?

(this function can still be exported and called but it should rely on the class to do the actual work?)

AE-MS:
This is a nice change that helps make the code clearer and easier to maintain. Thanks!

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/handlers.ts
lakhveerk: There are other functions like `initializeHandlers`, `callHandler`, `registerHandler` that are in the same file but outside of the `HandlersPrivate` class. I am wondering if we want to put those inside `HandlersPrivate` class? 
I am moving initializeHandlers and uninitializeHandlers into `HandlersPrivate` because they are very related to each other.

AE-MS (approved):
:shipit:

shrshindeMSFT (approved):
LGTM 🛳️

PR #1922, opened 2023-09-12
Prototype of talking from embedded teamsjs directly to outer host

TrevorJoelHarris:
## Description

This is a prototype of letting teams apps that are embedded in other teams apps message directly to the outer host instead of trying to route through their parent and then up to the host.

### Main changes in the PR:

1. Make message ids uuids instead of numbers. This prevents conflicts in ids when messaging directly to the outer host from multiple instances of teamsjs at the same time.
2. Update the test app so it has an embedded iframe hosting the test app again. Makes testing embedded app scenarios easier.
3. Made embedded apps send a dummy message directly to their parent (instead of the outer host) so that the outer host remembered who their child was. This allows the parent to pass responses from the outer host down to the child successfully.
4. Updated communication so that if it received a response from its parent with an id it has never seen before it will just send that response to their child, assuming it's for the child.
5. Made utils.generateGuid not cause errors when called in an SSR environment

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -270,25 +270,6 @@ function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequ
   return request;
 }
 
-const registerWithParentLogger = communicationLogger.extend('registerWithParent');
-
-export function registerWithParentIfNecessary(): void {
-  const parent = ssrSafeWindow().parent;
-  if (parent && parent === ssrSafeWindow().top) {
-    registerWithParentLogger('Not in an embedded iframe, no parent registration needed');
-    return;
-  }
-
-  registerWithParentLogger('Registering with parent app');
-  // Since we can't check our parent's origin and we haven't received anything from them yet, this first
-  // message needs to go to *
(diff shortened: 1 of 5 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -66,7 +66,8 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
   Communication.currentWindow = Communication.currentWindow || ssrSafeWindow();
   Communication.parentWindow =
     Communication.currentWindow.parent !== Communication.currentWindow.self
-      ? Communication.currentWindow.parent
+      ? Communication.currentWindow
+          .top /* Obviously, don't want to call this Communication.parentWindow anymore. Also, frameless */
```
erinha, line 70: Curious what you mean by "Also, frameless"?

TrevorJoelHarris: @erinha frameless mode doesn't use post messages to talk with the host, so we'd have to make sure what we're doing also works in those scenarios

TrevorJoelHarris:
Talked with Erin, this PR is no longer needed in its current form

PR #1930, opened 2023-09-19
Kx/back compat secondary browser

KangxuanYe:
Adding SecondaryBrowser testing JSON data. Check iOS Host SDK back-compat pipeline to see validity.


apps/teams-test-app/e2e-test-data/secondaryBrowser.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "SecondaryBrowser",
+  "version": ">=2.12.0",
```
MengyiGong, line 3: Question: if the version changes here, do we also need to make sure the version matches in https://office.visualstudio.com/ISS/_git/metaos-hub-sdk-iOS/pullrequest/2273698?

KangxuanYe: We don't have to. Version number is to tell back combat tool when this data should be taken for building E2E test. The update is to make sure that when it is the time to build data, it can process and build files properly leveraging on callAndResponse API as base.

MengyiGong (approved):

PR #1938, opened 2023-09-22
Fix Chrome connection failure issue

KangxuanYe:
## Description

**In short, there is a team upgraded the Chrome Driver from V116 to V117 on the image Ubuntu 20.04, which is used for executing most of our E2E tests in Web Host SDK, and there is an open bug on Chrome/Chromium V117 that will raise ECONNRESET error causing connection failure when using Cypress whose version is lower than (<)12.15.0. In Web Host SDK, we are using Cypress V9.7.0, which falls into this scenario. The upgrade change documentation, open bug ticket page, comparison on CI runs are listed below and** 

Upgrade document change: https://github.com/actions/runner-images/commit/22da280029d7b0ac4328e3d4a4c34c44ad7b93e9

Open bug ticket: https://github.com/cypress-io/cypress/issues/27804

**This issue now fails all of E2E testing pipelines running on both Teams-Js library and Web Host SDK, which heavily impact our customers. So this PR is to find an easy and long-term solution to fix the issue.**

Solution:

There are three alternatives we could take:
1. Change image (Not accepted): Using pool: Azure-Pipelines-EO-Ubuntu20.04-Office, which is used for Build & Unit Test, might work because Chrome browser version on that image is V116. However, there might be other reasons to use Ubuntu20.04 on E2E tests, so I didn't take that route.

2. Upgrade Cypress version to be equal or higher than (>=)version 12.15.0 (Not accepted): I tried and there are a lot of unmet peer dependencies and missing peer dependencies to resolve, which are not easy work.

3. **Downgrade Chrome Browser on CI (ACCEPTED): I wrote some of script to remove original chrome browser and download the version of 116.0.5845.179(release here: https://chromereleases.googleblog.com/search/label/Stable%20updates)**

### Main changes in the PR:

1. add bash script on build-host-app.yml file to downgrade chrome browser to V116

## Validation

CI pipeline should now all pass

### Change file added: No

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,7 +33,7 @@ jobs:
     steps:
       - template: tools/yaml-templates/build-app-host.yml
         parameters:
-          appHostGitPath: git://$(AppHostingSdkGitPath)@$(AppHostingSdkGitRef)
+          appHostGitPath: git://$(AppHostingSdkGitPath)@kx/fix_chrome_connection_issue
 
       - bash: 'node tools/cli/runAppsWithE2ETests.js --appUrl=https://localhost:4002 --reportFileName=e2e-tests-report-perf --envType=perf'
         displayName: 'Run E2E Perf tests'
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -110,11 +110,7 @@ jobs:
     displayName: 'E2E Test - CDN (only runs on release builds)'
     # This test only runs after deployment from a release branch and the new CDN version has been deployed
     # This check will run on the PR to merge the release branch back into main
-    condition: and(
```
KangxuanYe, line 113: This is changed to trigger release process and test if E2E test also work successfully on releasing process. It will be reverted back once everything else is good to be checked in.

erinha (dismissed):

azure-pipelines.yml
KangxuanYe: Reverted.

erinha (approved):

jekloudaMSFT (approved):

PR #1942, opened 2023-09-26
Add auto-retry (up to 3) for iOS E2E testing

KangxuanYe:
## Description

Add auto-retry for failed test case in YAML file to make E2E test more stable.

### DIFF
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -80,7 +80,10 @@ steps:
       -sdk iphonesimulator
       -parallel-testing-enabled YES
       -parallel-testing-worker-count 2
-      -destination 'platform=iOS Simulator,name=iPhone 14,OS=16.2'
+      -destination 'platform=iOS Simulator,name=iPhone 14 Pro,OS=16.2'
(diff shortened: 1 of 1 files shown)


jadahiya-MSFT (approved):

tools/yaml-templates/ios-test.yml
```diff
@@ -80,7 +80,10 @@ steps:
       -sdk iphonesimulator
       -parallel-testing-enabled YES
       -parallel-testing-worker-count 2
-      -destination 'platform=iOS Simulator,name=iPhone 14,OS=16.2'
+      -destination 'platform=iOS Simulator,name=iPhone 14 Pro,OS=16.2'
+      -screenshot-enabled="YES"
```
jadahiya-MSFT, line 84: Are screenshots only taken on failures or are they taken for every test?

jadahiya-MSFT (approved):

tools/yaml-templates/ios-test.yml
KangxuanYe: By default it is only taken when the case fails.

Ella-ly (approved):

PR #1951, opened 2023-09-29
Add a correct manifest for the test app along with instructions

TrevorJoelHarris:
## Description

The manifest file that was checked in alongside the test app was old and out of date. Replaced it with a basically functional manifest.json file that only needs to have the URL the test app is running at added to it. Also added instructions for how to use the manifest file.

This *isn't* being used for automation or anything else, the manifest file is just there as a convenience for people who want to sideload the test app in a host.

### Main changes in the PR:

1. Added `manifest.json` file along with 2 app icon files
2. Added `README.md` containing instructions on how to use the manifest file to sideload the test app

### Validation performed:

1. Manifest file was used by @ryanbliss to sideload the test app.

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

None needed, change only to app used for testing.


apps/teams-test-app/public/manifest.json
```diff
@@ -1,10 +1,52 @@
...
+      "name": "Test",
+      "contentUrl": "https://<REPLACE WITH URL OF THE APP>",
+      "websiteUrl": "https://<REPLACE WITH URL OF THE APP>",
+      "scopes": ["personal"]
+    },
+    { "entityId": "about", "scopes": ["personal"] }
+  ],
+  "validDomains": [],
+  "webApplicationInfo": {
+    "id": "34d8e67c-2758-40b5-836f-eb7defa875d3",
```
ydogandjiev, line 48: Is this actually needed for anything? If you expect this app to also be usable for SSO (i.e. calls to getAuthToken), then the resource below needs to also be completed. Otherwise, I would recommend removing this whole section.

apps/teams-test-app/public/manifest.json
```diff
@@ -1,10 +1,52 @@
...
   ],
-  "start_url": "/",
-  "display": "standalone",
-  "theme_color": "#000000",
-  "background_color": "#ffffff"
+  "staticTabs": [
+    {
+      "entityId": "test",
+      "name": "Test",
+      "contentUrl": "https://<REPLACE WITH URL OF THE APP>",
```
ydogandjiev, line 40: Why not list the actual endpoint that you guys are hosting this app at?

apps/teams-test-app/public/manifest.json
```diff
@@ -1,10 +1,52 @@
 {
-  "short_name": "Teams Test App",
-  "name": "Teams Test App",
-  "icons": [
+  "$schema": "https://developer.microsoft.com/en-us/json-schemas/teams/v1.16/MicrosoftTeams.schema.json",
+  "version": "1.0.0",
+  "manifestVersion": "1.16",
+  "id": "a79a2374-ec2a-486e-a3dc-b913fdb1f365",
+  "packageName": "com.package.name",
+  "name": { "short": "M365 Test App", "full": "" },
```
ydogandjiev, line 7: Maybe add a full name too just in case that becomes a validation check in the future. :D

ydogandjiev (dismissed):

apps/teams-test-app/public/manifest.json
TrevorJoelHarris: Great question! We actually don't currently deploy to that endpoint anymore. It's still there, but hasn't been updated in quite a while. Debating whether to do the work to re-enable it, but until then it's likely that people will want to point this at a locally run version of the app.

ryanbliss (approved):
LGTM

erinha (approved):

PR #1952, opened 2023-10-02
Update runtime unit tests to use objects instead of strings

AE-MS:
## Description

Today, the runtime unit tests validate that the Teams default runtime gets applied by using strings and regex on the objects. This was fine for the need at the time, but it has limitations. In particular, the code does not support subcapabilities, only comparing top level capabilities (foreshadowing).

Before adding subcapabilities to the runtime objects in a subsequent change, I'm updating this code to use the actual objects when doing comparisons (rather than trying to extend the string-based comparisons) so that it can now support subcapabilities (to any level of nesting).

### Main changes in the PR:

Update the runtime tests to use object comparisons rather than strings and regex

### Validation performed:

* Ensure tests continued to pass
* Broke something to ensure the tests failed when they should

### Unit Tests added:

No new unit tests were added since this change only focuses on changing the implementation of the current unit tests.

### End-to-end tests added:

No; this change is limited to unit test implementation.

### Change file added:

No, the tool said no changefile was required.

### Next/remaining steps:

1. My subsequent PR will update how the version tests are handled in this file to ensure they're always using a lower version and not just hardcoding 1.4.0 and assuming it is the lowest version
2. Subsequent to that, I will be creating another PR to update the default Teams config to indicate that Teams Mobile does not support various subcapabilities (e.g., pages.tabs)

### DIFF
--- a/packages/teams-js/test/public/runtime.spec.ts
+++ b/packages/teams-js/test/public/runtime.spec.ts
@@ -132,31 +132,92 @@ describe('runtime', () => {
     });
   });
 
+  // Determines whether the given "subset" runtime object is a subset of the given "superset" runtime object.
+  // This is used to determine whether all capabilities supported in "subset" are also supported in "superset"
+  function isSubset(subset: object, superset: object): boolean {
+    for (const key in subset) {
+      if (typeof subset[key] === 'object' && typeof superset[key] === 'object') {
+        if (!isSubset(subset[key], superset[key])) {
+          return false;
+        }
+      } else if (superset[key] === undefined) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  // Can recursively decompose an object into an array of objects, where each object in the array is a path to a leaf
+  // node in the original object.
+  // For example,
+  // {
+  //     pages: {
+  //         appButton: {},
+  //         tabs: {},
+  //     }
+  // }
+  // would be decomposed into
+  // [
+  //     { pages: { appButton: {} } },
+  //     { pages: { tabs: {} } }
+  // ]
+  // This can be a useful helper when identifying which capability defined in 'obj' is not defined in a runtime (because
+  // you can decompose a runtime object using this function, then compare each capability/subcapability one at a time to find
+  // any that are missing)
+  function decomposeObject(obj: object): object[] {
+    const result: object[] = [];
+
+    function recurse(current: object, path: string[] = []): void {
+      for (const key in current) {
+        const newPath = [...path, key];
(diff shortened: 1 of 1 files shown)


packages/teams-js/test/public/runtime.spec.ts
```diff
@@ -132,31 +132,92 @@ describe('runtime', () => {
     });
   });
 
+  // Determines whether the given "subset" runtime object is a subset of the given "superset" runtime object.
+  // This is used to determine whether all capabilities supported in "subset" are also supported in "superset"
+  function isSubset(subset: object, superset: object): boolean {
+    for (const key in subset) {
+      if (typeof subset[key] === 'object' && typeof superset[key] === 'object') {
```
TrevorJoelHarris, line 139: I am not sure I follow how this test works, and when I try to paste this code into typescript playground it won't execute.

AE-MS: Which test are you asking about? This function is just a helper and not a test in itself. The tests below leverage it.

I'm not sure why it wasn't working when you tried it, I pasted it into TS Playground with some sample objects and it seems to work [when I tried it](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABDAzgZRAIxQUygChS1ygC5E5MArHaAGkSIAccAnE8ym6ASnMzhwANjgCGSAN4BYAFCJEwOK0T4ICFFEQBrHAE9kSItjw9E0ufOTAVUXSzjWjJANo7dAXUQBeH4gDkXLRQfogAZKGItvaOICzseK56nj5e-oHQfqbmlpYw1vgAhKgYxgROCW7uDMxsLpU8WbI5Oax4IKxIwKJCuADcTc0AvgPyg4g4PThWKjXxUIke3r7gACY4wDBgOCuNFi1tHQrdfSOIwxbn8q1Q7UhQrCA4-TLnsmpgGogPYLAAtjgARm8ZgGEAAFqIyGZBnRQd0hOQJDCBkxRABzHAoRGneRQUTYRHIi6wl6yN7qTTfP44ABMwOyiAg8MJJPkqIxWJBe0seIJ0NZZxJg2e5I+lPA1IAzPTTuzMdjuTlREwmAAhEBQKAIFmnImvGTvFDCHAAOiEcDR+AABsViHh8FSYP8aQxHf8AXxEAASCS20oOiVO2muwPuniDK08Z6G41mi3Wv0kAM-IOSkMpsPkH2J+1unBpr6hwHhyPPIA)

TrevorJoelHarris (approved):

shrshindeMSFT (approved):
LGTM!

PR #1953, opened 2023-10-02
Update runtime merging code to support merging recursively and properly handle subcapability merging

AE-MS:
## Description

The current runtime object merging code doesn't properly handle merging when there are subcapabilities (e.g., nested objects) because it only uses the spread operator at the top-level. This means that any subcapabilities to be added will overwrite any from the baseline. Here is a minimal example demonstrating the problem:

```
const obj1 = {
    pages: {
        backstack: {},
    }
}

const obj2 = {
    pages: {
        tabs: {},
    }
}

const obj3 = {
    ...obj1,
    ...obj2,
}

console.log(JSON.stringify(obj3));
``` 
Actual Output:
`"{"pages":{"tabs":{}}}"`
Desired Output:
`"{"pages":{"backstack":{}, "tabs":{}}}"`

This change updates the merging algorithm to recursively merge the objects to arbitrary levels to properly support subcapabilities.

Spoilers: A future change will need to leverage this so that pages.tabs can be removed from the baseline but added back only in  non-Teams-Mobile platforms.

### Main changes in the PR:

- Update the runtime merging code to use the new recursive object merging function.
- Introduce a seam into the merging code to allow more direct unit testing: pass the values it operates on as parameters rather than having the code assume it should use globals

### Validation performed:

- Ensured that all unit tests pass indicating that runtime objects continue to be merged as previously.

### Unit Tests added:

Yes, added new unit tests that leverage the newly introduced seam to specifically test cases with nested capabilities and do not require the actual default Teams data to be of any particular form.

### End-to-end tests added:

No since this is all internal implementation.

### Change file added:

Yessir.

### Related PRs:

#1952 

### Next/remaining steps:

- Creating another PR to update the default Teams config to indicate that Teams Mobile does not support various subcapabilities (e.g., pages.tabs)


AE-MS:

No changes were made to the runtime object and, consequently, no new runtime version was required.

packages/teams-js/src/public/runtime.ts
```diff
@@ -439,6 +439,42 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
...
+function mergeRuntimeCapabilities(baselineRuntime: object, runtimeToMergeIntoBaseline: object): object {
+  const merged: object = { ...baselineRuntime };
+
+  for (const key in runtimeToMergeIntoBaseline) {
+    if (Object.prototype.hasOwnProperty.call(runtimeToMergeIntoBaseline, key)) {
+      if (typeof runtimeToMergeIntoBaseline[key] === 'object' && !Array.isArray(runtimeToMergeIntoBaseline[key])) {
+        merged[key] = mergeRuntimeCapabilities(baselineRuntime[key] || {}, runtimeToMergeIntoBaseline[key]);
+      } else {
+        if (!(key in baselineRuntime)) {
+          merged[key] = runtimeToMergeIntoBaseline[key];
```
aeunms, line 469: The doc said that the order of the parameters does not matter, but I think this will make it so merged[key] is determined by whatever is in baselineRuntime if both baselineRuntime and runtimeToMergeIntoBaseline have the same key.

aeunms (commented):
Still not done looking but leaving some comments first

packages/teams-js/src/public/runtime.ts
```diff
@@ -439,6 +439,42 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
...
+ *
+ * @param baselineRuntime the baseline runtime object
+ * @param runtimeToMergeIntoBaseline the runtime object to merge into the baseline
+ * @returns the merged runtime object
+ */
+function mergeRuntimeCapabilities(baselineRuntime: object, runtimeToMergeIntoBaseline: object): object {
+  const merged: object = { ...baselineRuntime };
+
+  for (const key in runtimeToMergeIntoBaseline) {
+    if (Object.prototype.hasOwnProperty.call(runtimeToMergeIntoBaseline, key)) {
```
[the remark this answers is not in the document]
AE-MS: The `Object.prototype.hasOwnProperty.call` method is used to check if an object has a property as its own property (as opposed to inheriting it from its prototype chain). This method is preferred over calling the `hasOwnProperty` method on the object itself because it is possible for the object to have a property with the name `hasOwnProperty` that shadows the method from `Object.prototype`. This can cause unexpected behavior or even errors if the `hasOwnProperty` property on the object is not a function. By calling the method from `Object.prototype` directly, we can ensure that we are using the correct method to check for the object’s own properties.

packages/teams-js/src/public/runtime.ts
AE-MS: Yes, although this only applies for properties that are not objects (e.g., strings), which shouldn't be in the runtime anyway. I've added a comment to indicate this.

change/@microsoft-teams-js-fd2fe2b1-7f23-493a-8cad-78ced187ec23.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
TrevorJoelHarris, line 2: I spent a couple minutes thinking about whether this should be a "none" change. From the consumer of teamsjs's perspective, this should result in no change to the library correct?

TrevorJoelHarris: (Maybe it's correct to leave this as patch since it could certainly result in a bug or change in the library. I've just been thinking lately about whether there's any utility in notifying external developers of these sorts of changes in the patch notes)

TrevorJoelHarris (approved):

change/@microsoft-teams-js-fd2fe2b1-7f23-493a-8cad-78ced187ec23.json
aeunms: I thought it would be a patch since there was a bug in the way it would merge capabilities before.

packages/teams-js/src/public/runtime.ts
aeunms: Isn't it possible that it is an object since something that would be `typeof runtimeToMergeIntoBaseline[key] === 'object'` but not `!Array.isArray(runtimeToMergeIntoBaseline[key]` would go into the `else`?

AE-MS: The only two types that have `typeof === object` are actual objects and arrays.

This runtime should never have anything but objects (actual objects) in it. Many things throughout the code would break if arrays or strings or anything else were in here.

As far as I can tell, in order for us to get into the else case, it means that either it's not an object (`typeof runtimeToMergeIntoBaseline[key] !== 'object'`) or it is an "object" but it's an Array (`Array.isArray(runtimeToMergeIntoBaseline[key])`)

Does this address your concern?

aeunms (approved):

PR #1954, opened 2023-10-02
Fix unit tests that assume the runtime always supports them

AE-MS:
## Description

Some of the unit tests make assumptions that the capability they are testing will always be supported (i.e., they don't explicitly ensure it is supported as part of their set up). This is fine today, when the default runtime does support them. But it's not fine tomorrow when I update the default runtime to no longer support them (i.e., a future change will remove pages.tabs and from being default supported, since it's not correct for it to be supported on Teams mobile).

This change makes their assumptions explicit by ensuring they are always setting that their capability is supported (in the case of navigation and stageView) and ensuring that the default runtime being tested against is actually specific to the version and platform being specified (in the case of app).

### Main changes in the PR:

1. Update navigation and stageView unit tests to explicitly set their dependent capability as supported
2. Update app unit tests to compare against the appropriate version and platform runtime object, rather than assuming the version and platform agnostic one is correct

### Validation performed:

- Ensured unit tests continued to pass

### Unit Tests added:

No, not changing or adding functionality, just updating the implementation of existing unit tests.

### End-to-end tests added:

No.

### Change file added:

The tool said it wasn't needed.

### Related PRs:

#1952 
#1953 

### Next/remaining steps:

- In addition to merging the above PRs, I believe the only remaining step to complete this journey will be to update the Teams fallback runtime to properly indicate that capabilities like pages.tabs are not supported on Teams Mobile

### DIFF
--- a/packages/teams-js/test/public/app.spec.ts
+++ b/packages/teams-js/test/public/app.spec.ts
@@ -21,6 +21,7 @@ import {
 } from '../../src/public/interfaces';
 import {
   _minRuntimeConfigToUninitialize,
+  generateVersionBasedTeamsRuntimeConfig,
   latestRuntimeApiVersion,
   runtime,
   versionAndPlatformAgnosticTeamsRuntimeConfig,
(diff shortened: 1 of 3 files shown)


aeunms (approved):
Thanks for these great changes! Out of curiosity, how was it that you realized these specific tests needed the update? Were they failing when you were testing the default supported capability changes and you decided to make a different PR for the test updates?

AE-MS:
> Thanks for these great changes! Out of curiosity, how was it that you realized these specific tests needed the update? Were they failing when you were testing the default supported capability changes and you decided to make a different PR for the test updates?

Yes, exactly that. I have a monster PR where I changed the defaults and then fixed all the tests that failed because of it. Then I separated out some of the fixes into individual PRs for ease of understanding and reviewing. This is one of those PRs. 😁

TrevorJoelHarris (approved):

Issue #1964, opened 2023-10-09
Chat service types not aligned with the expected values downstream


dinoilic:
# Issue

I was trying to work with chat openings from an app, and following the [docs](https://learn.microsoft.com/en-us/javascript/api/@microsoft/teams-js/chat?view=msteams-client-js-latest#@microsoft-teams-js-chat-openchat) and the [interface](https://learn.microsoft.com/en-us/javascript/api/@microsoft/teams-js/opensinglechatrequest?view=msteams-client-js-latest) for the open chat props I tried to do something simple as:

```ts
if (chat.isSupported()) {
  await chat.openChat({ user });
}
```

Where `user` is a single string containing a UPN. After this change has been deployed I noticed that there was an error, something along the lines "members.map is not a function". This led me to believe that somehow, somewhere downstream in `teams-js` code it's somehow expected that we pass in an array of people instead of a single UPN string. To test that I did the following:

```ts
if (chat.isSupported()) {
  await chat.openChat({ user: [user] as any });
}
```

Due to satisfying the interface, I had to do an explicit cast as `any`. Surprisingly, this worked well, so I looked a bit into the code.

The [interface](https://github.com/OfficeDev/microsoft-teams-library-js/blob/1a6f8c34275e605c6166b4fe9baa9b93c589008a/packages/teams-js/src/public/chat.ts#L26C3-L26C3) indeed expects a single UPN, however further down in the code we have two if branches:
1. If it's `runtime.isLegacyTeams` - the SDK invokes `createTeamsDeepLinkForChat()` with `user` in an array like `[openChatRequest.user]` as seen [here](https://github.com/OfficeDev/microsoft-teams-library-js/blob/1a6f8c34275e605c6166b4fe9baa9b93c589008a/packages/teams-js/src/public/chat.ts#L77C19-L77C19) however
2. If it's not `isLegacyTeams` then SDK invokes `sendAndHandleStatusAndReason` and sends and object where `members` is set to a single user: `members: openChatRequest.user`

Since the `args` that go into `sendAndHandleStatusAndReason` aren't typed, I can't figure out what's actually expected, however I believe this `members` is the same `members` that breaks on non existent `map` function, which makes sense as I pass in a single UPN.

Same issue would appear in `openGroupChat` as well where we have this block

```ts
if (openChatRequest.users.length === 1) {
  const chatRequest: OpenSingleChatRequest = {
    user: openChatRequest.users[0],
[...]

PR #1966, opened 2023-10-09
Magli/openchatfix

maglims:
This PR is to address the github issue: https://github.com/OfficeDev/microsoft-teams-library-js/issues/1964

## Description
Currently, teamsjs supports both single chat and group chat. For single chat, according to the public document, developers can pass user as OpenSingleChatRequest object which holds user as a string https://learn.microsoft.com/en-us/javascript/api/@microsoft/teams-js/opensinglechatrequest?view=msteams-client-js-latest. However, teams client is expecting app sdk passing in an user array for both single and group chat. That causes the issue that if developers are passing in user as a string as part of opensinglechatrequest object, they will encounter the error: "members.map is not a function".

The fix in this PR is to convert the user string as user array in teams js.

### Main changes in the PR:

1. Convert user string to be user array for single chat 
2. Update the test

## Validation
Validated with local test app. After the updating, developer are able to pass parameters like 
image
 for single chat.
Without this fix, it will repro the error "members.map is not a function".

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/change/@microsoft-teams-js-23b47bb4-1712-4f2d-80e1-07d270a51f0c.json
+++ b/change/@microsoft-teams-js-23b47bb4-1712-4f2d-80e1-07d270a51f0c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed the bug that openchat doesn't allow users to pass string for single user",
+  "packageName": "@microsoft/teams-js",
+  "email": "magli@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/public/chat.ts
+++ b/packages/teams-js/src/public/chat.ts
@@ -79,7 +79,7 @@ export namespace chat {
         );
       } else {
         const sendPromise = sendAndHandleStatusAndReason('chat.openChat', {
-          members: openChatRequest.user,
+          members: [openChatRequest.user],
           message: openChatRequest.message,
         });
         resolve(sendPromise);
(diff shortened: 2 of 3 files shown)


TrevorJoelHarris:
@maglims Do we know that all of the host sdks will keep working when we change this to an array?

maglims:
> @maglims Do we know that all of the host sdks will keep working when we change this to an array?

Great question! Here is the test I have done on different platforms:
image

In summary, 

- openchat doesn't seem work on teams with or wihout the fix. It gives the error "invalid url". It doesn't appear it is supported on iOS yet.
- openchat works fine with string and array **on teams web**
- openchat works fine with array **on teams desktop**, but not with string

At this point, it makes me believe that the bug should come to teams desktop. why desktop has different behavior as web?

as for mobile support on openchat, it seems a question that should come to teams mobile.

thoughts?

maglims (commented):
Discussed with Teams. The fix should come from hub sdk side.
Bug created: https://office.visualstudio.com/MetaOS/_workitems/edit/8432543

PR #1985, opened 2023-10-13
Fix local typedoc generation and export unexported referenced types

AE-MS:
## Description

Running `pnpm run docs` to generate docs locally on a dev box was broken. I discovered this was because the command run as part of the build to prevent it breaking was `pnpm run docs:validate` which runs typedoc in a similar way to `pnpm run docs` except that it uses an extra plugin which suppresses certain warnings.

The warnings that are suppressed are warnings about documented types referencing other types that aren't exported. E.g., if you had
```
export function foo(myBar: Bar): void;

type Bar = string;
```
It would show a warning because you were exporting a function that referenced a type which was not exported. This is a problem because if you are exporting a function, you will want all the types referenced as part of the function's definition to be exported (so others can use them and, for the purposes of typedoc, see their published documentation).

There are two primary changes in this PR:
1. Remove the reference to the typedoc plugin which suppressed these warnings
2. Fix all the non-exported types referenced by exported functions to be exported

Note that #1 also fixes the problem where `pnpm run docs` doesn't do the same thing as `pnpm run docs:validate` and therefore also makes `pnpm run docs` work again on your local dev box (so you can see and validate doc changes you make).

### Validation performed:

Ensured that `pnpm run docs` and `pnpm run docs:validate` both function and that the former successfully generates docs locally.

### DIFF
--- a/packages/teams-js/src/public/videoEffects.ts
+++ b/packages/teams-js/src/public/videoEffects.ts
@@ -17,9 +17,9 @@ export namespace videoEffects {
     : new VideoPerformanceMonitor(sendMessageToParent);
 
   /** Notify video frame processed function type */
-  type notifyVideoFrameProcessedFunctionType = () => void;
+  export type notifyVideoFrameProcessedFunctionType = () => void;
   /** Notify error function type */
-  type notifyErrorFunctionType = (errorMessage: string) => void;
+  export type notifyErrorFunctionType = (errorMessage: string) => void;
 
   /**
    * Represents a video frame
--- a/packages/teams-js/package.json
+++ b/packages/teams-js/package.json
@@ -14,7 +14,7 @@
     "build": "pnpm lint && webpack && pnpm docs:validate",
     "clean": "rimraf ./dist",
     "docs": "pnpm typedoc",
-    "docs:validate": "pnpm typedoc --plugin typedoc-plugin-missing-exports --emit none",
+    "docs:validate": "pnpm typedoc --emit none",
     "lint": "pnpm eslint ./src ./test --max-warnings 0 --fix --ext .ts",
     "prettier": "prettier --write '**/*.{ts,js,css,html}'",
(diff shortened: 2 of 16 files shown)


noahdarveau-MSFT (approved):

packages/teams-js/src/public/videoEffects.ts
```diff
@@ -17,9 +17,9 @@ export namespace videoEffects {
     : new VideoPerformanceMonitor(sendMessageToParent);
 
   /** Notify video frame processed function type */
-  type notifyVideoFrameProcessedFunctionType = () => void;
+  export type notifyVideoFrameProcessedFunctionType = () => void;
```
shrshindeMSFT, line 20: Just curious what value are we adding to the documentation by add this type? Also, similar question to previous exported types which are used only in their capabilities files.

AE-MS: I'm not quite sure if you are specifically asking about why there is even a type defined for `() => void` or why there's value in exporting it and documenting it? If the former, I unfortunately don't know since I didn't write this code. I could opine about whether I think there's value in defining a custom time for such a simple callback type but that's outside the scope of this PR.  😉

The value that is being added is two-fold, one having to do with "correctness" and one having to do with documentation:
1. If a particular function is exported, then it seems to reason that all the types upon which it depends (and which developers are expected to use) should also be exported.
2. From a documentation point of view, if those dependent types are not exported, then they don't show up in the documentation. For example, here is a before and after for `fullScreenChangeFunctionType`:

**Before**
image

Note that the `fullScreenChangeFunctionType` type listed is not clickable and that there is no documentation that describes it.

**After**
image

image


`fullScreenChangeFunctionType` is now clickable and has a doc section of its own to describe it

Does that answer your question?

shrshindeMSFT (approved):
LGTM! 🚢

PR #1989, opened 2023-10-16
Lakhveer/lifecycle handlers update

lakhveerk:

## Description

In [PR1878](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1878), two handlers - beforeSuspendOrTermiante and onResume were added in app.lifecycle capability. They were updated versions of old teamsCore handlers - beforeUnload and onLoadHandler. Before, we were using different strings to communicate to the host sdk that the handlers are registered. This would cause issues in scenarios like: If the user has registered for both pair of handlers, what pair will be called when the user navigate away or navigate to an app? 

In this PR, I have update the handlers in app.lifecycle so that they will use the strings `beforeUnload` and `load` to talk to hostSDK. If the user register for both pair of handlers, only the lifecycle.beforeSuspendOrTerminate and lifecycle.resume will be called. Since the teamsCore handlers are marked beta, this is a minor change. 

### Main changes in the PR:

1. Update the handlers in app.lifecycle so that they will use the strings `beforeUnload` and `load` to talk to hostSDK
2. Deleted deprecated tag from old handlers because new ones are not working yet. 

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/packages/teams-js/src/public/teamsAPIs.ts
+++ b/packages/teams-js/src/public/teamsAPIs.ts
@@ -53,9 +53,6 @@ export namespace teamsCore {
    *
    * @param handler - The handler to invoke when the page is loaded.
    *
-   * @deprecated
-   * As of 2.14.1, please use {@link app.lifecycle.registerOnResumeHandler} instead.
-   *
    * @beta
    */
   export function registerOnLoadHandler(handler: registerOnLoadHandlerFunctionType): void {
@@ -101,7 +98,6 @@ export namespace teamsCore {
    * @param handler - The handler to invoke before the page is unloaded. If this handler returns true the page should
    * invoke the readyToUnload function provided to it once it's ready to be unloaded.
    *
-   * @deprecated
    * @beta
    */
   export function registerBeforeUnloadHandler(handler: registerBeforeUnloadHandlerFunctionType): void {
--- a/change/@microsoft-teams-js-79b01e32-a70f-4798-9d2a-8112e7c5495f.json
+++ b/change/@microsoft-teams-js-79b01e32-a70f-4798-9d2a-8112e7c5495f.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Updated `teamsCore` handlers will be overwritten by `app.lifecycle` handlers",
+  "comment": "Updated app.lifecycle handlers, registerBeforeSuspendOrTerminateHandler and registerOnResumeHandler, so that they will overwrite teamsCore's registered handlers, registerBeforeUnloadHandler and registerOnLoadHandler, respectively.",
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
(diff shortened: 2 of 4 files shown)


packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -53,9 +53,6 @@ export namespace teamsCore {
    *
    * @param handler - The handler to invoke when the page is loaded.
    *
-   * @deprecated
-   * As of 2.14.1, please use {@link app.lifecycle.registerOnResumeHandler} instead.
-   *
```
AE-MS, line 58: Is this no longer deprecated? I understood from the PR description that you want to maintain the same messages across the TJS->Host boundary (for compatibility and confusion reasons), but presumably we would still prefer developers not add new uses of the functions in this file? Or is that not true?

AE-MS:
> /* eslint-disable @typescript-eslint/ban-types */

It seems like the intent, from your PR description, is to ensure that both the old and new handlers use the same message when communicating to the host.

The changes in this file seem more extensive than just updating the message -- can you say more about the motivation for that?

---
Refers to: packages/teams-js/src/internal/handlers.ts:1 in c6b7f14. [](commit_id = c6b7f14c695393eaf68267c0ed76430b109778ba, deletion_comment = False)

AE-MS (changes requested):
:clock1:

lakhveerk:
> > /* eslint-disable @typescript-eslint/ban-types */
> 
> It seems like the intent, from your PR description, is to ensure that both the old and new handlers use the same message when communicating to the host.
> 
> The changes in this file seem more extensive than just updating the message -- can you say more about the motivation for that?
> 
> Refers to: packages/teams-js/src/internal/handlers.ts:1 in [c6b7f14](https://github.com/OfficeDev/microsoft-teams-library-js/commit/c6b7f14c695393eaf68267c0ed76430b109778ba). [](commit_id = [c6b7f14](https://github.com/OfficeDev/microsoft-teams-library-js/commit/c6b7f14c695393eaf68267c0ed76430b109778ba), deletion_comment = False)

This commit opens this code
image
where an internal function is renamed to `readyToUnload` from  `readyToSuspendOrTerminate`. This is not customer facing function. So, name change would not matter much. 'readyToUnload` matches with `beforeUnload`, and that is the reason of this particular change. What do you think is extensive about this change? 

The more extensive change is If the user register for both pair of handlers, only the lifecycle.beforeSuspendOrTerminate and lifecycle.resume will be called. Since the teamsCore handlers are marked beta, I believe this is still a minor change even though it can actually be a breaking one if the developer registers a different handler for beforeSuspendOrTermiante than what they have in beforeUnload.

alexneyman-MSFT (approved):

AE-MS (approved):
:shipit:

PR #1990, opened 2023-10-16
Enabling Dynamic Domain List

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/internal/utils.ts
```diff
@@ -6,6 +6,7 @@ import { GlobalVars } from '../internal/globalVars';
 import { minAdaptiveCardVersion } from '../public/constants';
 import { AdaptiveCardVersion, SdkError } from '../public/interfaces';
 import { pages } from '../public/pages';
+import { Communication, sendAndUnwrap } from './communication';
```
AE-MS, line 9: I am concerned about creating a circular dependency between utils.ts and communication.ts. They each import each other.

In my experience, this is one of those things which sometimes seems to work but then randomly breaks later when someone makes an unrelated change because some internal compiler thing gets tripped.

Is it possible to extract some of this code into a different place to avoid creating the circular dependency?

packages/teams-js/src/internal/utils.ts
```diff
@@ -39,6 +40,13 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
 
 const validateOriginLogger = getLogger('validateOrigin');
 
+function validateOriginHandler(messageOrigin: URL): Promise<boolean> {
+  return new Promise<boolean>((resolve) => {
```
AE-MS, line 44: >new Promise<boolean> [](http://example.com/codeflow?start=9&length=20)

Is double wrapping the `Promise` necessary here? Can this function just return the `Promise` returned from `sendAndUnwrap`?

packages/teams-js/src/internal/utils.ts
```diff
@@ -39,6 +40,13 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
...
+    const messageOriginUrl = JSON.parse(JSON.stringify(messageOrigin));
```
AE-MS, line 45: >JSON.parse(JSON.stringify(messageOrigin)); [](http://example.com/codeflow?start=29&length=42)

What is the goal of this line? It's not obvious to someone reading it why we would stringify an object and then immediately parse it.

packages/teams-js/src/internal/communication.ts
```diff
@@ -23,6 +22,7 @@ export class Communication {
   public static parentWindow: Window | any;
   public static childWindow: Window;
   public static childOrigin: string;
+  public static validatedDomains: Map<string, boolean> = new Map();
```
AE-MS, line 25: >validatedDomains [](http://example.com/codeflow?start=16&length=16)

Based on how it looks like this is being used in utils, I _think_ the name `validatedDomains` is confusing, because it implies that the domains contained within "have been validated." In fact, I think it is the case that the domains in here _have been checked for validation_ but have not necessarily actually _been validated_. Is that right? I wouldn't want anyone using this to infer that if the domain was in this object it meant that the domain was okay to use.

Consider a name like `domainsAlreadyCheckedForValidity` or something more catchy.  😜

packages/teams-js/src/internal/utils.ts
```diff
@@ -54,18 +62,28 @@ export function validateOrigin(messageOrigin: URL): boolean {
     return false;
   }
   const messageOriginHost = messageOrigin.host;
-
-  if (validOrigins.some((pattern) => validateHostAgainstPattern(pattern, messageOriginHost))) {
-    return true;
+  let isValid = Communication.validatedDomains.get(messageOrigin.host);
```
AE-MS, line 65: >let isValid = Communication.validatedDomains.get(messageOrigin.host); [](http://example.com/codeflow?start=2&length=69)

It seems like this code would be clearer if `isValid` was always a `boolean` value and not sometimes undefined (despite the fact that JS is fast and loose with the relationship between `undefined` and `false`).

The usage of `isValid` would probably be clearer if this line was written like this:
```
let isValid = Communication.validatedDomains.get(messageOrigin.host) ?? false;
```

packages/teams-js/src/internal/utils.ts
```diff
@@ -39,6 +40,13 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
 
 const validateOriginLogger = getLogger('validateOrigin');
 
+function validateOriginHandler(messageOrigin: URL): Promise<boolean> {
```
AE-MS, line 43: >validateOriginHandler [](http://example.com/codeflow?start=9&length=21)

Given what it looks like this function is doing, consider a name that more clearly expresses the action it performs, e.g., `askHostIfDomainIsValid` or `checkIfDomainIsValid` or `validateDomainWithHost` or, my personal preference reflecting the `boolean` return value, `isDomainValidAccordingToHost`

packages/teams-js/src/internal/utils.ts
```diff
@@ -54,18 +62,28 @@ export function validateOrigin(messageOrigin: URL): boolean {
     return false;
   }
   const messageOriginHost = messageOrigin.host;
-
-  if (validOrigins.some((pattern) => validateHostAgainstPattern(pattern, messageOriginHost))) {
-    return true;
+  let isValid = Communication.validatedDomains.get(messageOrigin.host);
+  if (isValid === undefined) {
+    validateOriginHandler(messageOrigin).then((value: boolean) => {
```
AE-MS, line 67: >then [](http://example.com/codeflow?start=41&length=4)

Should there be a handler here to handle the situation when the `Promise` from `validateOriginHandler` is rejected? At least to log that fact?

packages/teams-js/src/internal/utils.ts
```diff
@@ -54,18 +62,28 @@ export function validateOrigin(messageOrigin: URL): boolean {
...
-  }
 
+    for (const domainOrPattern of GlobalVars.additionalValidOrigins) {
+      const pattern = domainOrPattern.substring(0, 8) === 'https://' ? domainOrPattern.substring(8) : domainOrPattern;
+      if (validateHostAgainstPattern(pattern, messageOriginHost)) {
+        return true;
+      }
+    }
+  } else {
+    return true;
```
AE-MS, line 85: Having many nested `if`s can sometimes make code difficult to read, particularly the longer it gets. Two potential approaches to consider:

1. Rather than writing this as "if !isValid { do some longish code} else return true, consider writing this as "if isValid return true; now do some longish code"
The idea being that when people read the code, when they start reading the longish code, they can already know "Okay we've already gotten the heck out of this function if we established validity" and you don't make people wait a long time (and read the code for handling the opposite case) to figure out "ooh what happens if `isValid` was true"

packages/teams-js/src/internal/communication.ts
```diff
@@ -325,8 +324,11 @@ function shouldProcessMessage(messageSource: Window, messageOrigin: string): boo
     messageOrigin === Communication.currentWindow.location.origin
   ) {
     return true;
+  } else if (Communication.parentOrigin && messageOrigin && messageOrigin === Communication.parentOrigin) {
```
AE-MS, line 327: >Communication.parentOrigin && messageOrigin && messageOrigin === Communication.parentOrigin [](http://example.com/codeflow?start=13&length=91)

What is this new check doing / why is it doing it now?

The code in this file is overall pretty hard to follow and this function in particular can be tricky to understand. One thing that I would encourage to help make it a little easier to read is to consider extracting this new condition into a short one-line boolean function that is named with a clear name.

I don't know what this is doing so I'll make something up just to illustrate my point. Compare:
```
  } else if (Communication.parentOrigin && messageOrigin && messageOrigin === Communication.parentOrigin) {
    return true;
```
to
```
  } else if (isMessageComingFromParentOrigin()) {
    return true;
```

Again I don't know what this new check is checking but a function name that describes the check could help.

packages/teams-js/src/internal/communication.ts
```diff
@@ -325,8 +324,11 @@ function shouldProcessMessage(messageSource: Window, messageOrigin: string): boo
...
+    return true;
   } else {
-    const isOriginValid = validateOrigin(new URL(messageOrigin));
+    const messageOriginUrl: URL = new URL(messageOrigin);
```
AE-MS, line 330: >URL [](http://example.com/codeflow?start=38&length=3)

Thank you for using appropriate typing here.  😀

AE-MS:
Are there new unit tests to validate the new behavior being added in this change?

packages/teams-js/test/private/privateAPIs.spec.ts
```diff
@@ -206,7 +206,7 @@ describe('AppSDK-privateAPIs', () => {
     await initPromise;
 
     // Now the getContext call should have been dequeued
-    expect(utils.messages.length).toBe(2);
+    expect(utils.messages.length).toBe(3);
```
AE-MS, line 209: >3 [](http://example.com/codeflow?start=39&length=1)

Are all these number incrementing changes required because now a new message is being sent in this flow (to check the origin with the host)?

If so, this model in the unit tests doesn't exactly seem sustainable (e.g., these unit tests are assuming that they know how many messages get sent out before them).

I'm not saying you have to fix this now with this change, but it would be good to file a bug to ensure we update these to make fewer assumptions about how the infrastructure is working.

AE-MS:
It's so cool to see this change to increase the library's agility in updating the domain list. Thanks for working on this!

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/communication.ts
```diff
@@ -305,6 +312,70 @@ function processMessage(evt: DOMMessageEvent): void {
 
 const shouldProcessMessageLogger = communicationLogger.extend('shouldProcessMessage');
 
+function sendValidateDomainRequest(messageOrigin: URL): Promise<boolean> {
+  const messageOriginUrl = sanitizeAndStrigifyURL(messageOrigin);
+  return sendAndUnwrap('validateDomains', messageOriginUrl);
+}
+
+export function validateOrigin(messageOrigin: URL): boolean {
```
AE-MS, line 320: >boolean [](http://example.com/codeflow?start=52&length=7)

This should return a `Promise<boolean>` since one of the function calls in here (`sendValidateDomainRequest`) is returning a `Promise` -- this ensures that any callers know there is async stuff that needs to be waited on and can wait on it.

packages/teams-js/src/internal/utils.ts
AE-MS: Thanks for updating this to have a clearer name (`sanitizeAndStrigifyURL`). Can you say more about how doing this roundtrip sanitizes the URL? I couldn't think of a way it was doing that but I'm not an expert in this type of thing. I asked ChatGPT to explain the code as well and it seemed unclear there, too (though AI can obviously be wrong). Do you have an example of something this prevents/fixes?

image

packages/teams-js/src/internal/communication.ts
```diff
@@ -305,6 +312,70 @@ function processMessage(evt: DOMMessageEvent): void {
...
+  }
+  if (isVerified) {
+    return true;
+  }
+  if (fallbackValidateOrigin(messageOriginHost)) {
+    return true;
+  }
+  if (validateAdditionalOrigins(messageOriginHost)) {
+    return true;
+  }
```
AE-MS, line 348: Is this working the way you expect? I would think that all the code after the else block (e.g., starting with `if (isVerified)` is going to run _before_ the code inside the `then` block. If you need to guarantee that code runs **after** calling an async/`Promise` it needs to be in the `then` block.

Try running this sample and check out the ordering of the console statements that are output:

[https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABDMYCmAnAFAZwBZwgA2AJgEpoBWa0AXIgEZxxFoCGYAlPQAoZwBbGDjQAeAG5wYJAHyIA3gFgAUIkQY0UEBiT5CpCtWiIA-Ij6DhaAHQajULJ0S9+QkbbQ4W4tI4DcKgC+KiqgkLAIiIRQmI4uliKiTCzsYHJKqsioscBsRCKc1lB4aGBYjogAvOmIEAherNZEcADmWABExaWMzRAA1u2cfoiBhRBsUBB45RpsXmD0HACeTtUKdWANNs1t7eOTeD1w-YN+oypqG1tNrR0AInCeiMXCiC2aiDvvJIgAQgCiADEAPJkf5RDCINjAGKQgDkXSQDF6fThplOF3Umm0SAsbhsGgaPiwUAwIDQQyCIWU0ViQyAA](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABDMYCmAnAFAZwBZwgA2AJgEpoBWa0AXIgEZxxFoCGYAlPQAoZwBbGDjQAeAG5wYJAHyIA3gFgAUIkQY0UEBiT5CpCtWiIA-Ij6DhaAHQajULJ0S9+QkbbQ4W4tI4DcKgC+KiqgkLAIiIRQmI4uliKiTCzsYHJKqsioscBsRCKc1lB4aGBYjogAvOmIEAherNZEcADmWABExaWMzRAA1u2cfoiBhRBsUBB45RpsXmD0HACeTtUKdWANNs1t7eOTeD1w-YN+oypqG1tNrR0AInCeiMXCiC2aiDvvJIgAQgCiADEAPJkf5RDCINjAGKQgDkXSQDF6fThplOF3Umm0SAsbhsGgaPiwUAwIDQQyCIWU0ViQyAA)

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1188,4 +1189,105 @@ describe('Testing communication', () => {
...
+      const result = validateOrigin(messageOrigin);
+      expect(result).toBe(true);
+    });
+    it('validateOrigin returns true if origin for subdomains in teams pre-known allowlist', () => {
+      const messageOrigin = new URL('https://subdomain.teams.microsoft.com');
+      const result = validateOrigin(messageOrigin);
+      expect(result).toBe(true);
+    });
+    it('validateOrigin returns false if origin is not in teams pre-known allowlist', () => {
+      const messageOrigin = new URL('https://badorigin.com');
```
AE-MS, line 1210: >https://badorigin.com') [](http://example.com/codeflow?start=37&length=23)

Generally speaking we should only put origins that exist for the purposes of testing (e.g., [example.com](https://en.wikipedia.org/wiki/Example.com)) into our tests. I don't know what is at badorigin.com but we don't want people accidentally visiting places not under our control, even if they might be okay today.

It's probably okay to use microsoft.com addresses, but the example.* domains are intended to us for this purpose.

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1188,4 +1189,105 @@ describe('Testing communication', () => {
...
   });
+  describe('validateOrigin', () => {
+    it("validateOrigin returns false if the protocol of origin is not 'https:'", () => {
+      /* eslint-disable-next-line @microsoft/sdl/no-insecure-url */ /* Intentionally using http here because of what it is testing */
+      const messageOrigin = new URL('http://teams.microsoft.com');
+      const result = validateOrigin(messageOrigin);
+      expect(result).toBe(false);
+    });
+    it('validateOrigin returns true if origin is in teams pre-known allowlist', () => {
+      const messageOrigin = new URL('https://teams.microsoft.com');
```
AE-MS, line 1200: >('https://teams.microsoft.com'); [](http://example.com/codeflow?start=35&length=32)

This unit test (and the ones like it) are assuming that teams.microsoft.com is in the validOrigins list. That might be true today, but it may not be true tomorrow.

I know this might seem overly picky, but as someone who recently had to update a bunch of unit tests that were making assumptions about which capabilities worked on which platform, we should try to minimize the assumptions the unit tests make so that they continue to run and pass even when unrelated changes are made. Even in this very change, you need to update unit tests that were making assumptions about message ordering, even though your change is not related to authentication or pages, etc.

In this case, rather hardcoding teams.microsoft.com, could you just pull the first element out of the `validOrigins` list and use that? That way we are always using one that we know is valid and even if that list changes, the test should still work.

jadahiya-MSFT: I really like this idea, I'll make that change now

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Good call out, I'll make this change now!

packages/teams-js/src/internal/communication.ts
```diff
@@ -315,18 +386,16 @@ const shouldProcessMessageLogger = communicationLogger.extend('shouldProcessMess
...
+  if (isMessageComingFromCurrentWindow(messageSource)) {
     shouldProcessMessageLogger('Should not process message because it is coming from the current window');
     return false;
-  } else if (
-    Communication.currentWindow &&
-    Communication.currentWindow.location &&
-    messageOrigin &&
-    messageOrigin === Communication.currentWindow.location.origin
-  ) {
+  } else if (isMessageComingFromSelf(messageOrigin)) {
```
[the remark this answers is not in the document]
jadahiya-MSFT: I believe this check is in the case of a dialog or a child window, I could change the name of the function to represent that, but if this if were to come first it would effectively pass whenever this app sends out a message to itself. That's the only reason why I think this check is in place.

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: Sorry, ChatGPT is actually right in this situation. I'm planning to use this method as a sanitize function for the url eventually. Its current iteration simply takes a URL object and then removes any associated functions attached to it so that it is able to sent as a message, otherwise it would be considered an invalid object and throw an error. This was purely used for backwards compatibility testing of the validation methods on the Host side to ensure parity with the current methods. 

The next iteration would change the `sendAndUnwrap('validateDomains`)` call to just use the host string.

packages/teams-js/src/internal/utils.ts
```diff
@@ -37,41 +37,32 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
...
 
+export function validateAdditionalOrigins(messageOriginHost: string): boolean {
   for (const domainOrPattern of GlobalVars.additionalValidOrigins) {
     const pattern = domainOrPattern.substring(0, 8) === 'https://' ? domainOrPattern.substring(8) : domainOrPattern;
     if (validateHostAgainstPattern(pattern, messageOriginHost)) {
       return true;
     }
   }
-
-  validateOriginLogger(
```
TrevorJoelHarris, line 69: Would a modified version of this logging still be useful here? Seems like it would help to see when messages are getting rejected for this reason

packages/teams-js/src/internal/utils.ts
```diff
@@ -37,41 +37,32 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
   return false;
 }
 
-const validateOriginLogger = getLogger('validateOrigin');
+export const validateOriginLogger = getLogger('validateOrigin');
+
+export function sanitizeAndStrigifyURL(url: URL): string {
```
TrevorJoelHarris, line 42:
```suggestion
export function sanitizeAndStringifyURL(url: URL): string {
```

packages/teams-js/src/internal/communication.ts
```diff
@@ -305,6 +312,70 @@ function processMessage(evt: DOMMessageEvent): void {
...
+  );
+  return false;
+}
+
+function isMessageComingFromParentOrigin(messageOrigin: string): boolean {
+  return (
+    Communication.parentOrigin &&
+    messageOrigin &&
+    messageOrigin === Communication.parentOrigin &&
+    Communication.alreadyVerifiedDomains.size === 0
```
TrevorJoelHarris, line 363: Why is this line needed?

packages/teams-js/src/internal/communication.ts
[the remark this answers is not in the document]
TrevorJoelHarris: I believe Jay's correct, this is really checking for "message from my own domain" (which is sort of what "fromSelf" means, but maybe a little less ambiguous. I *think* Jay's scenarios are also the ones I assumed this was for (dialogs that want to use the back channel for communicating with their parent app for example)

packages/teams-js/src/internal/communication.ts
```diff
@@ -23,6 +30,7 @@ export class Communication {
   public static parentWindow: Window | any;
   public static childWindow: Window;
   public static childOrigin: string;
+  public static alreadyVerifiedDomains: Map<string, boolean> = new Map();
```
TrevorJoelHarris, line 33: The name of this threw me off a little: in my head "Verified" meant "domains that we have already checked and found valid", but what it actually means is "domains that we have already checked and what the result of that check was". I might be overthinking naming (shocking, I know), but maybe something like `alreadyCheckedDomains` instead?

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Whoops good catch, I was doing that for testing purposes, it is not needed. I'll remove it

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: I was thinking something similar, since validated and verified imply that these are the domains that passed, when the reason this exists is to not bog down calls to our service endpoint with a list of known results. i.e `example.com: false`, but `example.com` keeps sending messages to the app, we shouldn't keep having to verify it if we know that it is incorrect.

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: This specifically logs that the entire domain validation process has failed. In the new validateOrigin method we call fallbackValidateOrigin and validateAdditionalOrigin it would have to fail all 3 methods to be logged.

I could add additional logging, but messages being rejected would only occur if all 3 fail and I don't want to log failures that succeed later down the line since then essentially every message that goes through the app supplied list will get logged as a failure before succeeding. What I think might be interesting is the case where the service endpoint call fails, but the fallback passes.

PR #1996, opened 2023-10-17
Timeout install Android emulator task after 10 minutes

jekloudaMSFT:

## Description

The "Install emulator" task on the Android E2E testing pipeline is a bit flaky. When it succeeds, it usually takes under 5 minutes, but if it fails it will take the entire 60 minutes allotted to the pipeline. This PR just adds a 10 minute timeout to the task so we can fail this task faster when it doesn't succeed.

### Main changes in the PR:

1. Add 10 minute timeout to the install emulator task on the Android test pipeline

### Validation performed:

1. Android E2E tests continue to pass when emulator installs successfully

### Unit Tests added:

No, no code changes

### End-to-end tests added:

No, no code changes

### Change file added:

No, pipeline change only

### Next/remaining steps:

I think we can retry starting the emulator in a separate task after the initial task times out. I'm going to work on this as a separate PR, but I think failing the task in 10 minutes vs. 60 is enough of an improvement to check in on its own for now.

### DIFF
--- a/tools/yaml-templates/android-test.yml
+++ b/tools/yaml-templates/android-test.yml
@@ -49,8 +49,7 @@ steps:
   - bash: 'chmod u+x install_emulator.sh && ./install_emulator.sh'
     displayName: 'Install Emulator'
     workingDirectory: '$(Agent.BuildDirectory)/androidHost/devtools/ci'
-    timeoutInMinutes: 2
-    retryCountOnTaskFailure: 1
+    timeoutInMinutes: 10
 
   - task: JavaToolInstaller@0
     inputs:
--- a/tools/yaml-templates/android-test.yml
+++ b/tools/yaml-templates/android-test.yml
@@ -49,6 +49,8 @@ steps:
   - bash: 'chmod u+x install_emulator.sh && ./install_emulator.sh'
     displayName: 'Install Emulator'
     workingDirectory: '$(Agent.BuildDirectory)/androidHost/devtools/ci'
+    timeoutInMinutes: 2
+    retryCountOnTaskFailure: 1
 
   - task: JavaToolInstaller@0
     inputs:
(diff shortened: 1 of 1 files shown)


tools/yaml-templates/android-test.yml
```diff
@@ -49,6 +49,8 @@ steps:
   - bash: 'chmod u+x install_emulator.sh && ./install_emulator.sh'
     displayName: 'Install Emulator'
     workingDirectory: '$(Agent.BuildDirectory)/androidHost/devtools/ci'
+    timeoutInMinutes: 2
+    retryCountOnTaskFailure: 1
```
AE-MS, line 53: Is it worth adding a comment here that explains why these are set this way? E.g., "We were seeing failures where blah blah 60 minutes and in our experience 2 minutes was blah blah so lets fail early blah blah blah"

AE-MS (changes requested):
:clock1:

tools/yaml-templates/android-test.yml
AE-MS: (I can see someone coming through here in a year and being like "Why is this set to 2 minutes?"

jekloudaMSFT: Yeah I like that idea. The real timeout I'm going to go with is 10, but I set it to 2 so I could see what would happen it times out. (The answer is it doesn't actually retry the task, so I think I'm going to try to write a specific recovery task that runs when this task fails)

KangxuanYe: @jekloudaMSFT ADO treats timeout differently from failure btw.

AE-MS (approved):
:shipit:

shrshindeMSFT (approved):
LGTM

PR #2008, opened 2023-10-25
Extend RequestAppAudioHandling to support for audio device selection sync

bingliang-zh:

## Description

This is a continuous PR of the original [PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1931) by [shilcare](https://github.com/shilcare).

### Purpose

Give Teams apps the ability to use Teams' audio devices settings, like play audio on the device that user selected in Teams.

### Details

Add a callback method to `RequestAppAudioHandling`.
- When App first call `RequestAppAudioHandling` with this callback, the callback will be called immediately with Teams' current audio device labels. The app can use this info to initialize and select the corresponding audio devices.
- This callback will also be called when user changes audio devices in Teams, app can update its audio devices accordingly.

### Main changes in the PR:

1. add optional `audioDeviceSelectionChangedCallback` to `RequestAppAudioHandlingParams` which is used in `meeting.requestAppAudioHandling`
2. register event handler in `startAppAudioHandling`
3. unregister event handler in `stopAppAudioHandling`

### Validation performed:

To make this work properly, we also need code changes in some internal repos. We have built a Teams container with those changes.

1. Use Teams 2.1 with the custom container
2. Create a test Teams app, add some test code like
```typescript
  meeting.requestAppAudioHandling(
    {
      //...
      audioDeviceSelectionChangedCallback: (selectedDevices) => {
        console.log('audioDeviceSelectionChangedCallback');
        console.log(selectedDevices);
        return Promise.resolve();
      },
    },
   //...
  );
```
3. Replace an existing meeting app, redirect to the local test app
4. Open the modified meeting app.
5. The callback will be called after `meeting.requestAppAudioHandling` is called. And also get called after device is changed in Teams settings. The payload is correct too.

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

No. `pnpm changefile` says No change files are needed.

### Related PRs:

- <https://github.com/OfficeDev/microsoft-teams-library-js/pull/1931>

### Next/remaining steps:

Close [original PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1931) because duplicated and outdated.


packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,42 @@ export namespace meeting {
...
+   * @hidden
+   * Hide from docs.
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface AudioDeviceInfo {
+    deviceLabel: string;
```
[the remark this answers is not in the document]
bingliang-zh: Updated, added some explanations.

packages/teams-js/src/public/meeting.ts
```diff
@@ -914,6 +950,13 @@ export namespace meeting {
       };
       registerHandler('meeting.micStateChanged', micStateChangedCallback);
 
+      const audioDeviceSelectionChangedCallback = async (
+        selectedDevicesInHost: AudioDeviceSelection,
+      ): Promise<void> => {
+        await requestAppAudioHandlingParams.audioDeviceSelectionChangedCallback?.(selectedDevicesInHost);
```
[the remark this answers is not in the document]
bingliang-zh: Updated, removed promise/await/async etc.

TrevorJoelHarris (approved):

packages/teams-js/test/utils.ts
```diff
@@ -138,6 +138,20 @@ export class Utils {
...
+      if (this.messages[i].func === func) {
+        if (k === 0) {
+          return this.messages[i];
+        } else {
+          k--;
+        }
+      }
+    }
+    return null;
+  };
```
AE-MS, line 153: There's nothing wrong with the way you've written this (as far as I can see anyway) but if we can prefer using the `for..of` pattern, the core of the function will more closely match canonical JS iterating patterns. Counting up rather than down could also make this function easier to understand by future readers.

Not saying you have to make the change, but consider if this might be more readable:
```
  function findKthMessagesByFunc(func: string, k: number): MessageRequest | null {
    let countOfMatchingMessages = 0;

    for (const message of messages) {
        if (message.func === func) {
            if (countOfMatchingMessages === k) {
                return message;
            }
            ++countOfMatchingMessages;
        }
    }

    return null;
  }
```

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface AudioDeviceSelection {
+    speaker?: AudioDeviceInfo;
+    microphone?: AudioDeviceInfo;
+    error?: SdkError;
+  }
```
AE-MS, line 358: Is it correct that these could all be `undefined`? How should the developer interpret the situation where the `audioDeviceSelectionChangedCallback` is called and the object passed as the `selectedDevices` parameter has no defined properties on it?

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface AudioDeviceSelection {
+    speaker?: AudioDeviceInfo;
+    microphone?: AudioDeviceInfo;
+    error?: SdkError;
```
AE-MS, line 357: Perhaps related to my previous comment: is it ever legal for there to be an `error` value AND a defined value for `speaker` or `microphone`? What would that mean?

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+  }
+
+  /**
+   * Interface for AudioDeviceInfo, includes a device label with the same format as MediaDeviceInfo.label
+   *
+   * Hosted app can use this label to compare it with the device info fetched from `navigator.mediaDevices.enumerateDevices()`
```
AE-MS, line 363: >`navigator.mediaDevices.enumerateDevices()` [](http://example.com/codeflow?start=83&length=43)

Can you use a `{@link }` tag for this reference?

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+   * @beta
+   */
+  export interface AudioDeviceSelection {
+    speaker?: AudioDeviceInfo;
+    microphone?: AudioDeviceInfo;
+    error?: SdkError;
+  }
+
+  /**
+   * Interface for AudioDeviceInfo, includes a device label with the same format as MediaDeviceInfo.label
```
AE-MS, line 361: >MediaDeviceInfo.label [](http://example.com/codeflow?start=84&length=21)

Can you surround types like this (and a few other references below) with backticks? Thanks!

AE-MS (changes requested):
:clock1:

packages/teams-js/test/utils.ts
```diff
@@ -138,6 +138,20 @@ export class Utils {
     return null;
   };
 
+  public findKthMessagesByFunc = (func: string, _k: number): MessageRequest | null => {
```
shrshindeMSFT, line 141: Can you please explain why we need this function and why can't we use the existing `findMessageByFunc`?

shrshindeMSFT (changes requested):

shrshindeMSFT:
I don't see any validation performed here like E2E test cases. If E2E test cases are not required, can you please state the reason in the PR description?

packages/teams-js/test/utils.ts
bingliang-zh: In my case, there are two 'registerHandler' events fired in order at the same time, I can't get the second fired event use `findMessageByFunc`.

packages/teams-js/test/utils.ts
bingliang-zh: Updated

packages/teams-js/src/public/meeting.ts
bingliang-zh: Updated.

packages/teams-js/src/public/meeting.ts
bingliang-zh: Updated.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1472,6 +1472,57 @@ describe('meeting', () => {
             expect(micCallbackCalled).toBe(true);
           });
 
+          it(`should call meeting.audioDeviceSelectionChanged after meeting.requestAppAudioHandling. context: ${context}`, async () => {
```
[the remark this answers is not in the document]
bingliang-zh: Added

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
      */
     micMuteStateChangedCallback: (micState: MicState) => Promise<MicState>;
+    /**
+     * Callback for the host to tell the app to change its speaker selection
+     */
+    audioDeviceSelectionChangedCallback?: (selectedDevices: AudioDeviceSelection) => Promise<void>;
+  }
+
+  /**
+   * Interface for AudioDeviceSelection from Teams selection
```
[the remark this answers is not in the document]
bingliang-zh: Could you explain what's the difference between Teams host and other hosts? `requestAppAudioHandling` is called by a hosted app, in my case their host/origin/domain is not under teams' domain.

By now `AudioDeviceSelection` is only used in `RequestAppAudioHandlingParams` which is also a hidden and internal API, so I decide to keep these tags too.

packages/teams-js/src/public/meeting.ts
bingliang-zh: From API design aspect, `teams-js` can't be sure these will always have values, Teams might have a bug, or the user don't have any audio I/O devices. Developer can use the system's default device if needed like this example:
```ts
const devices = await navigator.mediaDevices.enumerateDevices();
const defaultSpeaker = devices.find((d) => d.deviceId === 'default' && d.kind === 'audiooutput');
const defaultMic = devices.find((d) => d.deviceId === 'default' && d.kind === 'audioinput');
```

packages/teams-js/src/public/meeting.ts
bingliang-zh: That a pattern in hub sdk. If that happens, I think the error should be respected first, speaker and mic value could be meaningless even if they existed.

packages/teams-js/src/public/meeting.ts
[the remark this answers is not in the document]
TrevorJoelHarris: This capability may be implemented by hosts other than Teams in the future, so we shouldn't mention Teams specifically in the comment.

packages/teams-js/src/public/meeting.ts
AE-MS: Even though previous designs may have unfortunately allowed for there to theoretically be both an error value and  actual data, we don't need to repeat these confusing designs. Can you update this structure to either allow for an error OR the audio device info, but not both? Unless it is actually valid to get errors and data. Thanks!

AE-MS (changes requested):

packages/teams-js/src/public/meeting.ts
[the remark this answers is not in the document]
bingliang-zh: Replaced `Teams` with `host`.

packages/teams-js/src/public/meeting.ts
bingliang-zh: Don't see a reason to change. This pattern is widely used in this repo too. For example, `ISpeakingState`, `MeetingReactionReceivedEventData`, `RaiseHandStateChangedEventData` in "meeting.ts", `ISdkResponse` in "dialog.ts". And it is also mentioned multiple times in the changelog, like

> 2.2.0: Added an optional error object to `ISpeakingState` interface to align `registerSpeakingStateChangeHandler` API with other API error handling."

> 2.0.0-beta.4: `submitHandler` callback takes a single object parameter containing both error and result

> 2.0.0: an optional `DialogSubmitHandler` callback that takes a single object parameter containing both error and result

I personally will just ignore any other data outside the error object, only respect the payload inside the Error if exist.

AE-MS: When designing a new API, it is important to ensure that it is easy for developers to use it correctly and difficult or impossible to use it incorrectly. Given the options of:
1. Requiring developers to interpret ambiguous situations (where both errors and valid values could be returned)
2. Designing the API to make it _impossible_ to use incorrectly (you either get an error, or you get a valid value)
...we need to choose option 2 when it is possible, no matter how many design mistakes previous developers may have made.

It's good to know what behavior you would personally take when confronted with this ambiguous situation, but the platform is designed for a broad range of developers and needs to be as easy to use as is practical.

AE-MS:
For boring security reasons this repo can't accept or validate PRs from forks; you'll have to move this change to branch at some point to enable validation to run and to eventually commit your changes.

packages/teams-js/src/public/meeting.ts
bingliang-zh: There is **no possible way** to force user to use it correctly **without** asking MetaOS not to send data in such a form.

This interface is used in a callback as the input data, the callback is a callback for devs to fill. After all, devs can do what they want in their code space. A simple `as any` can pass all the type checks.

Assume change it to a type union. A possible change that split the error and the payload.
```ts
  export type AudioDeviceSelection =
    | {
        speaker?: AudioDeviceInfo;
        microphone?: AudioDeviceInfo;
      }
    | {
        error?: SdkError;
      };
```
But see how's the user gonna use it.
```ts
if (Object.hasOwn(devices, 'error')) {
  // do something catch the error, like: `console.log(devices['error']);`
  return;
}
customUpdateSelectionWrittenByDev(devices['speaker'], devices['microphone']);
```
I smell bad code.

Compare to original:
```ts
const {error, speaker, microphone} = selectedDevices;
if (error) {
  // do something catch the error;
  return;
}
customUpdateSelectionWrittenByDev(speaker, microphone);
```
Much simpler and easier to read.

bingliang-zh:
Created new PR using a branch [here](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2027)

PR #2010, opened 2023-10-26
Updated default runtime for Teams Mobile to indicate that pages.appButton, pages.tabs, and stageView are not supported

AE-MS:
## Description

The default fallback runtime used by Teams (for legacy Teams clients that don't pass a runtime) indicated that the pages.appButton, pages.tabs, and stageView capabilities were universally supported across all platforms. This was incorrect as these three capabilities are not supported on Teams mobile platforms. This change fixes that.

### Main changes in the PR:

1. Remove these three capabilities from the platform agnostic fallback runtime (in runtime.ts)
2. Group the client types by mobile and non-mobile (in runtime.ts)
3. Update the structure that maps Teams versions and platforms to supported capabilities to enable these three capabilities for non-mobile platforms from "the earliest version" (which I just set to 1.0.0) (in runtime.ts).

With these changes, non-Mobile Teams clients will continue to support these capabilities (because when the `mapTeamsVersionToSupportedCapabilities` structure is applied, it will apply support for these on non-Mobile platforms for all versions of Teams > 1.0.0), while mobile Teams clients will no longer incorrectly indicate they support these capabilities because they've been removed from the baseline version and platform agnostic `versionAndPlatformAgnosticTeamsRuntimeConfig` structure.

4. This change has also revealed more assumptions that various unit tests were making so in these files:
`pages.spec.ts`, `publicAPIs.spec.ts`, and `runtime.spec.ts`, you will see many places where the tests are now explicitly setting the runtime config they expect to be true when they run.

5. There's also one other change in `runtime.spec.ts` where it hardcoded 1.4.0 as "the oldest version" and just always used that assuming it would be older than whatever version it was currently testing. I updated this test to now dynamically create an "older version" by just prepending `0.` to the front of the version, guaranteeing that it will always be the oldest.

6. I also added a box and button to the pages.appButton capability in the test app to explicitly call the `isSupported()` function so I could test it.

7. I added a handy (I think) command to the package.json file that will enable the unit tests to be run in verbose mode, which enables console logs to be seen rather than suppressed. 
### Validation performed:

1. All unit tests pass
2. All E2E tests pass
3. Validated that pages.appButton, pages.tabs, and stageView capabilities still return `isSupported() === true` on big screen Teams
4. Validated that pages.appButton, pages.tabs, and stageView capabilities now return `isSupported() === false` on mobile Teams

### Unit Tests added:

No because there are no unit tests that specifically test the default Teams capabilities for a particular version or platform. They (correctly) only test that the functions for interpreting the runtime structures work properly.

### Change file added:

Yes'm.

### Related PRs:

Other PRs that I've made that led up to this (e.g., the capability merging ones)


AE-MS:

The default Teams runtime object was updated but not the runtime interface. No new runtime version is required.

change/@microsoft-teams-js-252e0ab1-300c-4234-8b5c-76c706f3edec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
aeunms, line 2: Would we consider this a patch in that it's fixing a default runtime for mobile rather than adding a feature?

packages/teams-js/src/public/runtime.ts
```diff
@@ -317,6 +312,13 @@ export const v1HostClientTypes = [
   HostClientType.teamsDisplays,
 ];
 
+export const v1HostClientTypes = [
+  ...v1NonMobileHostClientTypes,
+  HostClientType.android,
+  HostClientType.ios,
+  HostClientType.ipados,
+];
+
```
aeunms, line 321: I think this is more of a nit suggestion but maybe also making a constant for mobile host client types would look nice

```
const v1MobileHostClientTypes = [
  HostClientType.android,
  HostClientType.ios,
  HostClientType.ipados
];

export const v1HostClientTypes = [
  ...v1NonMobileHostClientTypes,
  ...v1MobileHostClientTypes,
];
```

aeunms (commented):
Finished looking through and left some comments.

packages/teams-js/package.json
```diff
@@ -17,7 +17,8 @@
     "docs:validate": "pnpm typedoc --emit none",
     "lint": "pnpm eslint ./src ./test --max-warnings 0 --fix --ext .ts",
     "prettier": "prettier --write '**/*.{ts,js,css,html}'",
-    "test": "jest"
+    "test": "jest",
+    "test:verbose": "pnpm test -- --verbose false --silent false"
```
shrshindeMSFT, line 21: What is the purpose of this command?

shrshindeMSFT (approved):
LGTM! 🚢 Just one question related to `pnpm test:verbose` command

packages/teams-js/package.json
AE-MS: It's a handy (I think) command that will enable the unit tests to be run in verbose mode, which enables console logs to be seen rather than suppressed. It enables an existing logging statements to be seen when tests are run, providing an alternate way to debug issues besides stepping through them in VSCode (which also works). What are your thoughts on the usefulness of this command?

change/@microsoft-teams-js-252e0ab1-300c-4234-8b5c-76c706f3edec.json
AE-MS: I'm certainly open to suggestions here. For reference, this how beachball describes the difference:

Patch - bug fixes; no API changes.
Minor - small feature; backwards compatible API changes.

I will update it to "Patch"

packages/teams-js/src/public/runtime.ts
AE-MS: Sure, I can do that if it makes it clearer.

AE-MS:
> Finished looking through and left some comments.

Thank you!

aeunms (approved):
🚢

shrshindeMSFT (approved):

packages/teams-js/package.json
shrshindeMSFT: Developers have preferences I think this command will definitely be useful, especially when writing and debugging unit test case.

shrshindeMSFT:
Thank you for making this change, as it'll make developers clear that what capabilities are supported on v1 and which host client type. 🚢

PR #2012, opened 2023-10-27
Fix some locations that violate strictNullChecks

AE-MS:
## Description
The long journey to enabling the `strictNullChecks` compiler option requires incrementally fixing violations. This change fixes some of those violations.

### Main changes in the PR:
No functionality or API changes are intended; any fixes are either internal only or making explicit the existing implicit behavior. Rather than describe the general nature of the fixes here, I put comments about each type of change directly into this PR at the lines where they applied.

### Validation performed:

Ensure all existing tests pass.

### Unit Tests added:

No, as no behavior changes were made or intended.

### End-to-end tests added:

No, as no behavior changes were made or intended.

### Change file added:

Yes

### Next/remaining steps:

More fixes of this sort, and then eventually permanently enabling the `strictNullChecks` compiler option!

### DIFF
--- a/packages/teams-js/src/internal/mediaUtil.ts
+++ b/packages/teams-js/src/internal/mediaUtil.ts
@@ -207,7 +207,7 @@ export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): bo
  *
  * @internal
  */
-export function validatePeoplePickerInput(peoplePickerInputs: people.PeoplePickerInputs): boolean {
+export function validatePeoplePickerInput(peoplePickerInputs?: people.PeoplePickerInputs): boolean {
   if (peoplePickerInputs) {
     if (peoplePickerInputs.title) {
       if (typeof peoplePickerInputs.title !== 'string') {
--- a/packages/teams-js/src/public/monetization.ts
+++ b/packages/teams-js/src/public/monetization.ts
@@ -67,8 +67,7 @@ export namespace monetization {
     param2?: PlanInfo,
   ): Promise<void> {
     let callback: ((error: SdkError | null) => void) | undefined;
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-    let planInfo: PlanInfo;
+    let planInfo: PlanInfo | undefined;
     if (typeof param1 === 'function') {
       callback = param1;
       planInfo = param2;
@@ -80,6 +79,7 @@ export namespace monetization {
         if (!isSupported()) {
           throw errorNotSupportedOnPlatform;
         }
+        /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
         resolve(sendAndHandleSdkError('monetization.openPurchaseExperience', planInfo));
       });
     };
--- a/packages/teams-js/src/public/navigation.ts
+++ b/packages/teams-js/src/public/navigation.ts
@@ -33,14 +33,14 @@ export function returnFocus(navigateForward?: boolean): void {
  */
 export function navigateToTab(tabInstance: TabInstance, onComplete?: onCompleteHandlerFunctionType): void {
   ensureInitialized(runtime);
-  onComplete = onComplete ? onComplete : getGenericOnCompleteHandler();
+  const completionHandler: onCompleteHandlerFunctionType = onComplete ?? getGenericOnCompleteHandler();
   pages.tabs
     .navigateToTab(tabInstance)
     .then(() => {
-      onComplete(true);
+      completionHandler(true);
     })
     .catch((error: Error) => {
-      onComplete(false, error.message);
+      completionHandler(false, error.message);
     });
 }
 
@@ -68,14 +68,14 @@ export function navigateCrossDomain(url: string, onComplete?: onCompleteHandlerF
     FrameContexts.stage,
     FrameContexts.meetingStage,
   );
-  onComplete = onComplete ? onComplete : getGenericOnCompleteHandler();
+  const completionHandler: onCompleteHandlerFunctionType = onComplete ?? getGenericOnCompleteHandler();
   pages
     .navigateCrossDomain(url)
     .then(() => {
-      onComplete(true);
+      completionHandler(true);
     })
     .catch((error: Error) => {
-      onComplete(false, error.message);
+      completionHandler(false, error.message);
     });
 }
 
@@ -90,13 +90,13 @@ export function navigateCrossDomain(url: string, onComplete?: onCompleteHandlerF
  */
 export function navigateBack(onComplete?: onCompleteHandlerFunctionType): void {
   ensureInitialized(runtime);
-  onComplete = onComplete ? onComplete : getGenericOnCompleteHandler();
+  const completionHandler: onCompleteHandlerFunctionType = onComplete ?? getGenericOnCompleteHandler();
   pages.backStack
     .navigateBack()
     .then(() => {
-      onComplete(true);
+      completionHandler(true);
     })
     .catch((error: Error) => {
-      onComplete(false, error.message);
+      completionHandler(false, error.message);
     });
 }
--- a/packages/teams-js/src/public/pages.ts
+++ b/packages/teams-js/src/public/pages.ts
@@ -232,7 +232,7 @@ export namespace pages {
    */
   export function registerFullScreenHandler(handler: fullScreenChangeFunctionType): void {
     registerHandlerHelper('fullScreenChange', handler, [], () => {
-      if (handler && !isSupported()) {
+      if (handler !== undefined && !isSupported()) {
         throw errorNotSupportedOnPlatform;
       }
     });
@@ -405,7 +405,7 @@ export namespace pages {
      */
     export function registerOnSaveHandler(handler: saveEventType): void {
       registerOnSaveHandlerHelper(handler, () => {
-        if (handler && !isSupported()) {
+        if (handler !== undefined && !isSupported()) {
           throw errorNotSupportedOnPlatform;
         }
       });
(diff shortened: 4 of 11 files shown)


packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -207,7 +207,7 @@ export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): bo
  *
  * @internal
  */
-export function validatePeoplePickerInput(peoplePickerInputs: people.PeoplePickerInputs): boolean {
+export function validatePeoplePickerInput(peoplePickerInputs?: people.PeoplePickerInputs): boolean {
```
AE-MS, line 210: >? [](http://example.com/codeflow?start=60&length=1)

This function was already written to support an `undefined` parameter value, this change acknowledged this support (and, in fact, was already potentially called with `undefined` values)

packages/teams-js/src/public/monetization.ts
```diff
@@ -67,8 +67,7 @@ export namespace monetization {
     param2?: PlanInfo,
   ): Promise<void> {
     let callback: ((error: SdkError | null) => void) | undefined;
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-    let planInfo: PlanInfo;
+    let planInfo: PlanInfo | undefined;
```
AE-MS, line 70: >| undefined [](http://example.com/codeflow?start=27&length=11)

This variable was previously defined but not set so it already was always `undefined` to start.

packages/teams-js/src/public/navigation.ts
```diff
@@ -33,14 +33,14 @@ export function returnFocus(navigateForward?: boolean): void {
  */
 export function navigateToTab(tabInstance: TabInstance, onComplete?: onCompleteHandlerFunctionType): void {
   ensureInitialized(runtime);
-  onComplete = onComplete ? onComplete : getGenericOnCompleteHandler();
+  const completionHandler: onCompleteHandlerFunctionType = onComplete ?? getGenericOnCompleteHandler();
```
AE-MS, line 36: >const completionHandler [](http://example.com/codeflow?start=2&length=23)

The changes in this file are needed for the compiler to recognize that the completion handler function will never be undefined. It was also just generally unusual practice to redefine the value of input parameters, in the previous approach.

packages/teams-js/src/public/publicAPIs.ts
```diff
@@ -293,14 +293,14 @@ export function executeDeepLink(deepLink: string, onComplete?: executeDeepLinkOn
     FrameContexts.stage,
     FrameContexts.meetingStage,
   );
-  onComplete = onComplete ? onComplete : getGenericOnCompleteHandler();
+  const completionHandler: executeDeepLinkOnCompleteFunctionType = onComplete ?? getGenericOnCompleteHandler();
```
AE-MS, line 296: >const completionHandler [](http://example.com/codeflow?start=2&length=23)

Same explanation as navigation.ts

packages/teams-js/src/public/settings.ts
```diff
@@ -99,14 +99,14 @@ export namespace settings {
    */
   export function setSettings(instanceSettings: Settings, onComplete?: setSettingsOnCompleteFunctionType): void {
     ensureInitialized(runtime, FrameContexts.content, FrameContexts.settings, FrameContexts.sidePanel);
-    onComplete = onComplete ? onComplete : getGenericOnCompleteHandler();
+    const completionHandler: setSettingsOnCompleteFunctionType = onComplete ?? getGenericOnCompleteHandler();
```
AE-MS, line 102: >const completionHandler [](http://example.com/codeflow?start=4&length=23)

Same explanation as navigation.ts

packages/teams-js/src/public/pages.ts
```diff
@@ -232,7 +232,7 @@ export namespace pages {
    */
   export function registerFullScreenHandler(handler: fullScreenChangeFunctionType): void {
     registerHandlerHelper('fullScreenChange', handler, [], () => {
-      if (handler && !isSupported()) {
+      if (handler !== undefined && !isSupported()) {
```
AE-MS, line 235: >handler !== undefined [](http://example.com/codeflow?start=10&length=21)

Using an explicit `undefined` check rather than the implicit `if (handler)` check makes it clear to the compiler of the intent (without that check, the compiler would question whether the intent was calling `handler()`)

packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -57,7 +57,7 @@ export namespace teamsCore {
    */
   export function registerOnLoadHandler(handler: registerOnLoadHandlerFunctionType): void {
     registerOnLoadHandlerHelper(handler, () => {
-      if (handler && !isSupported()) {
+      if (handler !== undefined && !isSupported()) {
```
AE-MS, line 60: >handler !== undefined [](http://example.com/codeflow?start=10&length=21)

These are the same fixes as described in pages.ts

packages/teams-js/src/public/people.ts
```diff
@@ -48,10 +48,8 @@ export namespace people {
   ): Promise<PeoplePickerResult[]> {
     ensureInitialized(runtime, FrameContexts.content, FrameContexts.task, FrameContexts.settings);
 
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-    let callback: selectPeopleCallbackFunctionType;
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-    let peoplePickerInputs: PeoplePickerInputs;
+    let callback: selectPeopleCallbackFunctionType | undefined = undefined;
```
AE-MS, line 51: >| undefined [](http://example.com/codeflow?start=51&length=11)

These two lines were previously always set to `undefined` initially by not assigning a value explicitly.

packages/teams-js/src/public/tasks.ts
```diff
@@ -37,8 +37,7 @@ export namespace tasks {
    */
   export function startTask(taskInfo: TaskInfo, submitHandler?: startTaskSubmitHandlerFunctionType): IAppWindow {
     const dialogSubmitHandler = submitHandler
-      ? /* eslint-disable-next-line strict-null-checks/all */ /* fix tracked by 5730662 */
-        (sdkResponse: dialog.ISdkResponse) => submitHandler(sdkResponse.err, sdkResponse.result)
+      ? (sdkResponse: dialog.ISdkResponse) => submitHandler(sdkResponse.err ?? '', sdkResponse.result ?? '')
```
AE-MS, line 40: >submitHandler [](http://example.com/codeflow?start=46&length=13)

The two parameters to the `startTaskSubmitHandlerFunctionType` do not accept `undefined` values. However, the two properties on `sdkResponse` could be `undefined`. There's no good way to fix this without redesigning the API (e.g., the signature of `startTaskSubmitHandlerFunctionType` cannot be changed to accepted `undefined` because that would be a breaking API change). Consequently, to only pass legal values to the submit handler, `undefined` will get translated to the allowable empty string value.

packages/teams-js/src/public/tasks.ts
```diff
@@ -92,7 +91,10 @@ export namespace tasks {
    * @returns - Converted UrlDialogInfo object
    */
   function getUrlDialogInfoFromTaskInfo(taskInfo: TaskInfo): UrlDialogInfo {
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
+    if (taskInfo.url === undefined) {
+      throw new Error("url property of taskInfo object can't be undefined");
+    }
```
AE-MS, line 96: `TaskInfo` allows the `url` property to be undefined. But if that was ever undefined, this would eventually fail anyway since you need a URL to show a url dialog. This just makes it fail slightly earlier.

packages/teams-js/src/public/tasks.ts
```diff
@@ -111,7 +113,12 @@ export namespace tasks {
    * @returns - converted BotUrlDialogInfo object
    */
   function getBotUrlDialogInfoFromTaskInfo(taskInfo: TaskInfo): BotUrlDialogInfo {
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
+    if (taskInfo.url === undefined || taskInfo.completionBotId === undefined) {
```
AE-MS, line 116: >if (taskInfo.url === undefined || taskInfo.completionBotId === undefined) [](http://example.com/codeflow?start=4&length=73)

Same deal as the previous comment in `getUrlDialogInfoFromTaskInfo`

packages/teams-js/src/public/pages.ts
```diff
@@ -232,7 +233,7 @@ export namespace pages {
    */
   export function registerFullScreenHandler(handler: fullScreenChangeFunctionType): void {
     registerHandlerHelper('fullScreenChange', handler, [], () => {
-      if (handler && !isSupported()) {
+      if (!isNullOrUndefined(handler) && !isSupported()) {
```
AE-MS, line 236: >!isNullOrUndefined(handler) [](http://example.com/codeflow?start=10&length=27)

Using an explicit `undefined` and `null` check rather than the implicit `if (handler)` check makes it clear to the compiler of the intent (without that check, the compiler would question whether the intent was calling the `handler()` function)

packages/teams-js/src/public/teamsAPIs.ts
```diff
@@ -57,7 +58,7 @@ export namespace teamsCore {
    */
   export function registerOnLoadHandler(handler: registerOnLoadHandlerFunctionType): void {
     registerOnLoadHandlerHelper(handler, () => {
-      if (handler && !isSupported()) {
+      if (!isNullOrUndefined(handler) && !isSupported()) {
```
AE-MS, line 61: >!isNullOrUndefined(handler) [](http://example.com/codeflow?start=10&length=27)

These are the same fixes as described in pages.ts

TrevorJoelHarris (approved):

PR #2013, opened 2023-10-30
Start logging name of script currently executing when teamsjs is first loaded

TrevorJoelHarris:
## Description

Currently, teamsjs does not support being used from more than one script at the same time. Doing so will result in undefined behavior from the library. Currently, this restriction is not well publicized and can be very difficult to track down by people helping users of the library. With this change, the first time a script using teamsjs is loaded we will log out the name of the script file using teamsjs (or the fact that teamsjs is being used by script embedded directly into the html). If you see more than one of these log statements at the same time, that means that your app is using teamsjs in a way that is currently unsupported.

The log statement also includes a link to where developers can open a GitHub issues if they would like to be able to use teamsjs from more than one place at the same time. We already have that work on our backlog, but hopefully that will help us gauge the demand and importance of fixing this (which is a more intricate change)

### Main changes in the PR:

1. Start logging the name of the script using teamsjs when that script is first loaded.

### Validation performed:

1. Ran the test app and made sure the log statement was being written correctly

### Unit Tests added:

No, logging change only

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -565,6 +565,26 @@ export namespace app {
    */
   const initializationTimeoutInMs = 5000;
 
+  function logWhereTeamsJsIsBeingUsed(): void {
+    if (inServerSideRenderingEnvironment()) {
+      return;
+    }
+    const scripts = document.getElementsByTagName('script');
+    // This will always be the current script because browsers load and execute scripts in order.
+    // Whenever a script is executing for the first time it will be the last script in this array.
+    const currentScriptSrc = scripts[scripts.length - 1].src;
+    const scriptUsageWarning =
+      'Today, teamsjs can only be used from a single script or you may see undefined behavior. If you would like to use teamjs from more than one script at the same time, please open an issue at https://github.com/OfficeDev/microsoft-teams-library-js/issues';
+    if (!currentScriptSrc || currentScriptSrc.length === 0) {
+      appLogger('teamsjs is being used from a script tag embedded directly in your html. %s', scriptUsageWarning);
+    } else {
+      appLogger('teamsjs is being used from %s. %s', currentScriptSrc, scriptUsageWarning);
+    }
+  }
+
+  // This is called right away to make sure that we capture which script is being executed correctly
+  logWhereTeamsJsIsBeingUsed();
+
   /**
    * Initializes the library.
    *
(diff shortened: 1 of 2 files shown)


AE-MS:
This question is a reflection of me seeking to understand how this script loading stuff all works.

Does the code you have written here actually detect when TJS has been loaded more than once simultaneously? Or does it just always log and warn?

TrevorJoelHarris:
> This question is a reflection of me seeking to understand how this script loading stuff all works.
> 
> Does the code you have written here actually detect when TJS has been loaded more than once simultaneously? Or does it just always log and warn?

Right now, this is just logging that is intended to help people identify that teamsjs is being used more than once along with some extra information about _where_ teamsjs is being used.

We could do work to detect if teamsjs is *actually* running more than once, but I'm not sure that's a lot less work than just supporting multiple instances of teamsjs at the same time, and I think our ultimate goal should be to support that scenario.

packages/teams-js/src/public/app.ts
```diff
@@ -565,6 +565,26 @@ export namespace app {
...
+  function logWhereTeamsJsIsBeingUsed(): void {
+    if (inServerSideRenderingEnvironment()) {
+      return;
+    }
+    const scripts = document.getElementsByTagName('script');
+    // This will always be the current script because browsers load and execute scripts in order.
+    // Whenever a script is executing for the first time it will be the last script in this array.
+    const currentScriptSrc = scripts && scripts[scripts.length - 1] && scripts[scripts.length - 1].src;
+    const scriptUsageWarning =
+      'Today, teamsjs can only be used from a single script or you may see undefined behavior. If you would like to use teamjs from more than one script at the same time, please open an issue at https://github.com/OfficeDev/microsoft-teams-library-js/issues';
```
AE-MS, line 577: >Today, teamsjs can only be used from a single script or you may see undefined behavior. [](http://example.com/codeflow?start=7&length=87)

Given that this line will be logged unconditionally (and not just when it detects a multi-usage scenario), it might be slightly clearer if it was reworded to indicate that? E.g.,

"Today TJS can only be used.... This log line is used to help detect cases where TJS is loaded multiple times -- it is always written. The presence of the log itself does not indicate a multi-load situation, but multiples of these log lines will."

Or something like that?

AE-MS (changes requested):
:clock1:

AE-MS (approved):
:shipit:

lakhveerk (approved):

shrshindeMSFT (approved):
LGTM 🚢

PR #2014, opened 2023-10-30
Fix strictNullChecks violations in media.ts and mediaUtil.ts

AE-MS:
## Description

Continuing down the path of fixing more violations of `strictNullChecks` violations across teamsjs, this change fixes the violations in the media*.ts family of files.

I've annotated the PR with specific comments at the areas where the changes are made. That said, the general theme of the changes is to just update the code/types to reflect what the code is already doing -- e.g., the code is already returning `null` in places where it's not allowed, just update the code to reflect that said value is allowed and how it handles it, documenting appropriately.

### Validation performed:

Ensure all tests continue to pass.

### Change file added:

Yes.

### Related PRs:

#2012 

### Next/remaining steps:

Fix more violations of `strictNullChecks` across the next set of files.

### DIFF
--- a/packages/teams-js/src/internal/mediaUtil.ts
+++ b/packages/teams-js/src/internal/mediaUtil.ts
@@ -14,11 +14,11 @@ import { throwExceptionIfMobileApiIsNotSupported } from './internalAPIs';
  * @internal
  * Limited to Microsoft-internal use
  */
-export function createFile(assembleAttachment: media.AssembleAttachment[], mimeType: string): Blob {
+export function createFile(assembleAttachment: media.AssembleAttachment[], mimeType: string): Blob | null {
   if (assembleAttachment == null || mimeType == null || assembleAttachment.length <= 0) {
     return null;
   }
-  let file: Blob | undefined;
+  let file: Blob | null = null;
   let sequence = 1;
   assembleAttachment.sort((a, b) => (a.sequence > b.sequence ? 1 : -1));
   assembleAttachment.forEach((item) => {
@@ -42,7 +42,7 @@ export function createFile(assembleAttachment: media.AssembleAttachment[], mimeT
  * @internal
  * Limited to Microsoft-internal use
  */
-export function decodeAttachment(attachment: media.MediaChunk, mimeType: string): media.AssembleAttachment {
+export function decodeAttachment(attachment: media.MediaChunk, mimeType: string): media.AssembleAttachment | null {
   if (attachment == null || mimeType == null) {
     return null;
   }
@@ -188,10 +188,10 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
  *
  * @internal
  */
-export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): boolean {
+export function validateScanBarCodeInput(barCodeConfig?: media.BarCodeConfig): boolean {
   if (barCodeConfig) {
     if (
-      barCodeConfig.timeOutIntervalInSec === null ||
+      !barCodeConfig.timeOutIntervalInSec ||
       barCodeConfig.timeOutIntervalInSec <= 0 ||
       barCodeConfig.timeOutIntervalInSec > 60
     ) {
--- a/packages/teams-js/src/internal/mediaUtil.ts
+++ b/packages/teams-js/src/internal/mediaUtil.ts
@@ -191,9 +191,9 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
 export function validateScanBarCodeInput(barCodeConfig?: media.BarCodeConfig): boolean {
   if (barCodeConfig) {
     if (
-      !barCodeConfig.timeOutIntervalInSec ||
-      barCodeConfig.timeOutIntervalInSec <= 0 ||
-      barCodeConfig.timeOutIntervalInSec > 60
+      barCodeConfig.timeOutIntervalInSec === null ||
+      (barCodeConfig.timeOutIntervalInSec != undefined && barCodeConfig.timeOutIntervalInSec <= 0) ||
+      (barCodeConfig.timeOutIntervalInSec != undefined && barCodeConfig.timeOutIntervalInSec > 60)
     ) {
       return false;
     }
--- a/packages/teams-js/src/public/media.ts
+++ b/packages/teams-js/src/public/media.ts
@@ -25,6 +25,7 @@ import {
   validateSelectMediaInputs,
   validateViewImagesInput,
 } from '../internal/mediaUtil';
+import { isNullOrUndefined } from '../internal/typeCheckUtilities';
 import { generateGUID } from '../internal/utils';
 import { errorNotSupportedOnPlatform, FrameContexts, HostClientType } from './constants';
 import { DevicePermission, ErrorCode, SdkError } from './interfaces';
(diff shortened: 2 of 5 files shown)


packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -14,11 +14,11 @@ import { throwExceptionIfMobileApiIsNotSupported } from './internalAPIs';
  * @internal
  * Limited to Microsoft-internal use
  */
-export function createFile(assembleAttachment: media.AssembleAttachment[], mimeType: string): Blob {
+export function createFile(assembleAttachment: media.AssembleAttachment[], mimeType: string): Blob | null {
   if (assembleAttachment == null || mimeType == null || assembleAttachment.length <= 0) {
     return null;
   }
-  let file: Blob | undefined;
+  let file: Blob | null = null;
```
AE-MS, line 21: >let file: Blob | null = null; [](http://example.com/codeflow?start=2&length=29)

`null` is more clearly meant to indicate "none of this thing" than `undefined` is. Since the code was only "accidentally" using `undefined` before (by nature of just never setting a value), I switched it to `null` instead to make it clearly intentional. The actual behavior of this function and calling functions remains the same.

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -42,7 +42,7 @@ export function createFile(assembleAttachment: media.AssembleAttachment[], mimeT
  * @internal
  * Limited to Microsoft-internal use
  */
-export function decodeAttachment(attachment: media.MediaChunk, mimeType: string): media.AssembleAttachment {
+export function decodeAttachment(attachment: media.MediaChunk, mimeType: string): media.AssembleAttachment | null {
```
AE-MS, line 45: >null [](http://example.com/codeflow?start=109&length=4)

Inconsistently, this function was already using `null` in the same way the above was using `undefined`. Now they are consistent in their use of `null` to intentionally indicate `no value here.`

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -188,12 +188,12 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
  *
  * @internal
  */
-export function validateScanBarCodeInput(barCodeConfig: media.BarCodeConfig): boolean {
+export function validateScanBarCodeInput(barCodeConfig?: media.BarCodeConfig): boolean {
```
AE-MS, line 191: >? [](http://example.com/codeflow?start=54&length=1)

This code was already handling the `undefined` case for this parameter, despite not showing so in the typing. Now it clearly indicates that.

packages/teams-js/src/internal/mediaUtil.ts
```diff
@@ -188,12 +188,12 @@ export function validateViewImagesInput(uriList: media.ImageUri[]): boolean {
...
   if (barCodeConfig) {
     if (
       barCodeConfig.timeOutIntervalInSec === null ||
-      barCodeConfig.timeOutIntervalInSec <= 0 ||
-      barCodeConfig.timeOutIntervalInSec > 60
+      (barCodeConfig.timeOutIntervalInSec != undefined && barCodeConfig.timeOutIntervalInSec <= 0) ||
```
AE-MS, line 195: >barCodeConfig.timeOutIntervalInSec != undefined [](http://example.com/codeflow?start=7&length=47)

This code was previously assuming `barCodeConfig.timeOutIntervalInSec` was not `undefined` when doing this check. Now it explicitly checks for it.

Note that it is "valid" per the definition of this code for `barCodeConfig.timeOutIntervalInSec` to be `undefined`, it just should not check these timeouts in that case.

packages/teams-js/src/public/media.ts
```diff
@@ -106,14 +130,14 @@ export namespace media {
     if (!GlobalVars.isFramelessWindow) {
       const notSupportedError: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
       /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-      callback(notSupportedError, undefined);
+      callback(notSupportedError, []);
```
AE-MS, line 133: >[] [](http://example.com/codeflow?start=34&length=2)

In this and other places in this file, the parameter value passed in the case of an error was changed to a legal value, according to the defined types. The documentation for these callback types was also updated at the top to make it clear what the expected behavior is, in the case of an error.

packages/teams-js/src/public/media.ts
```diff
@@ -174,7 +198,7 @@ export namespace media {
    * Media object returned by the select Media API
    */
   export class Media extends File {
-    constructor(that: Media = null) {
```
AE-MS, line 177: > = null [](http://example.com/codeflow?start=27&length=7)

`null` was not a legal value for this. Since the value is only used immediately below (line 203) and already handles the `undefined` case the same as `null,` rather than using a default value of `null`, which would happen if `that` wasn't passed, it now instead will set it to `undefined` if `that` wasn't passed, which is handled just the same as before.

packages/teams-js/src/public/media.ts
```diff
@@ -204,14 +228,12 @@ export namespace media {
       ensureInitialized(runtime, FrameContexts.content, FrameContexts.task);
       if (!isCurrentSDKVersionAtLeast(mediaAPISupportVersion)) {
         const oldPlatformError: SdkError = { errorCode: ErrorCode.OLD_PLATFORM };
-        /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
-        callback(oldPlatformError, null);
+        callback(oldPlatformError, new Blob());
```
AE-MS, line 231: >new Blob() [](http://example.com/codeflow?start=35&length=10)

Similar to examples above, an empty `Blob` is used and now documented for the error case.

packages/teams-js/src/public/media.ts
```diff
@@ -232,23 +254,26 @@ export namespace media {
...
               // create file/blob when all chunks have arrived and we get 0/-1 as chunksequence number
               if (mediaResult.mediaChunk.chunkSequence <= 0) {
                 const file = createFile(helper.assembleAttachment, helper.mediaMimeType);
-                callback(mediaResult.error, file);
+                callback(mediaResult.error, file ?? new Blob());
               } else {
                 // Keep pushing chunks into assemble attachment
-                const assemble: AssembleAttachment = decodeAttachment(mediaResult.mediaChunk, helper.mediaMimeType);
-                helper.assembleAttachment.push(assemble);
+                const assemble: AssembleAttachment | null = decodeAttachment(
```
AE-MS, line 267: >decodeAttachment [](http://example.com/codeflow?start=60&length=16)

`decodeAttachment` can now return `null` (before it was returning `undefined`, see the comments in `mediaUtil.ts`), this handles this error case by not trying to push it into the  `assembleAttachment` array (note that previous to this, it was unconditionally pushing `undefined` into that same array in the error case, which was incorrect and not be handled by any later code)

packages/teams-js/src/public/media.ts
```diff
@@ -264,31 +289,34 @@ export namespace media {
         assembleAttachment: [],
       };
       const params = [actionName, this.content];
-      this.content && callback && sendMessageToParent('getMedia', params);
+      this.content && !isNullOrUndefined(callback) && sendMessageToParent('getMedia', params);
```
AE-MS, line 292: >isNullOrUndefined [](http://example.com/codeflow?start=23&length=17)

Without this, the violation is triggered because it thinks this code is trying to invoke/call the `callback` function

packages/teams-js/src/public/media.ts
```diff
@@ -446,7 +474,7 @@ export namespace media {
    */
   abstract class MediaController<T> {
     /** Callback that can be registered to handle events related to the playback and control of video content. */
-    protected controllerCallback: T;
+    protected controllerCallback?: T;
```
AE-MS, line 477: >? [](http://example.com/codeflow?start=32&length=1)

Despite not being declared this way previously, this value could always have been `undefined` as it is only set in the constructor (immediately below this) and the constructor clearly accepts `undefined`

shrshindeMSFT (commented):
LGTM! Just one quick question related to `createFile` function

packages/teams-js/src/internal/mediaUtil.ts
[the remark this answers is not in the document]
AE-MS: Can you say more about your concern? I want to make sure I understand. Are you suggesting that it should not be possible for `file` to ever be `null`?

In case that is what you're saying, I can at least one way that the `file` variable can be `null`:

`file` has to have an initial value. Prior to my change, the initial value was `undefined` (even though it wasn't explicitly set to `undefined`, that's the value it was getting). So `file` will be `null` to start and at least until a `Blob` is created.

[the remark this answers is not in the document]
AE-MS: Another way that `file` can remain `null` is if the `assembleAttachment` collection contains elements that have `sequence` values that don't start at 1. Perhaps that's not supposed to happen but the code allows for it.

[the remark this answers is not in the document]
shrshindeMSFT: Discussed offline, to add how the `Blob.slice` works here is the [link](https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice#start)

shrshindeMSFT (dismissed):
LGTM 🚢 !

packages/teams-js/src/public/media.ts
```diff
@@ -232,23 +254,26 @@ export namespace media {
...
                 // Keep pushing chunks into assemble attachment
-                const assemble: AssembleAttachment = decodeAttachment(mediaResult.mediaChunk, helper.mediaMimeType);
-                helper.assembleAttachment.push(assemble);
+                const assemble: AssembleAttachment | null = decodeAttachment(
+                  mediaResult.mediaChunk,
+                  helper.mediaMimeType,
+                );
+                if (assemble) {
+                  helper.assembleAttachment.push(assemble);
+                }
```
TrevorJoelHarris, line 273: since this if statement is new, does it warrant some sort of error or log statement if `assemble` is null? Seems unlikely we had cases like that that we weren't noticing, but not impossible

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/media.ts
AE-MS: Good suggestion; I have added a logging statement to indicate when this situation has arisen.

shrshindeMSFT (approved):
LGTM! 🚢

TrevorJoelHarris (approved):

PR #2020, opened 2023-11-02
Fix more "easy" strictNullChecks violations -- ones which just require adding already in-use types

AE-MS:
## Description

Continuing the journey of fixing `strictNullChecks` violations...

In this change I am mostly just updating the stated types to match the types actually being used. E.g., if the code is actually returning `null` or `undefined` in places, then the type signature should now reflect that.

There are two places (in `authentication.ts`) where I updated the return value to be set to `undefined` instead of `null` because the actual return type was already defined to allow for `undefined` -- so we should just actually use that! I added comments inline in the code review to point these out.

Because of this new, correct, types, some of the communication unit tests now started to trigger violations of `strictNullChecks` because the compiler now realizes that the now-correctly-annotated types might be being misused. For now, I have just silenced these violations. Again, nothing about how the code actually works/runs is changing, the compiler just now realizes that there are potential problems here (which is why it's important to actually have `strictNullChecks` enabled!). These new violations will be fixed as part of the overall journey to get this enabled.

### Validation performed:

### Change file added:

Yes.

### Related PRs:

#2012 
#2014 

### Next/remaining steps:

Continue to fix violations in future PRs

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -19,10 +19,10 @@ const communicationLogger = getLogger('communication');
  */
 export class Communication {
   public static currentWindow: Window | any;
-  public static parentOrigin: string;
+  public static parentOrigin: string | null;
   public static parentWindow: Window | any;
-  public static childWindow: Window;
-  public static childOrigin: string;
+  public static childWindow: Window | null;
+  public static childOrigin: string | null;
 }
 
 /**
@@ -176,7 +176,7 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
  * @internal
  * Limited to Microsoft-internal use
  */
-export function sendMessageToParentAsync<T>(actionName: string, args: any[] = undefined): Promise<T> {
+export function sendMessageToParentAsync<T>(actionName: string, args: any[] | undefined = undefined): Promise<T> {
   return new Promise((resolve) => {
     const request = sendMessageToParentHelper(actionName, args);
     /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
@@ -234,7 +234,7 @@ const sendMessageToParentHelperLogger = communicationLogger.extend('sendMessageT
  * @internal
  * Limited to Microsoft-internal use
  */
-function sendMessageToParentHelper(actionName: string, args: any[]): MessageRequest {
+function sendMessageToParentHelper(actionName: string, args: any[] | undefined): MessageRequest {
   const logger = sendMessageToParentHelperLogger;
 
   const targetWindow = Communication.parentWindow;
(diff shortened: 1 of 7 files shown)


packages/teams-js/src/public/authentication.ts
```diff
@@ -112,7 +112,7 @@ export namespace authentication {
           return value;
         } finally {
           if (!isDifferentParamsInCall) {
-            authParams = null;
+            authParams = undefined;
```
AE-MS, line 115: >undefined [](http://example.com/codeflow?start=25&length=9)

`authParams` was already defined as being possibly `undefined`, this code was just using incorrectly using `null` to indicate that "cleared" state. I checked and all the uses of `authParams` will behave the same in the case that it is `null` or `undefined`

packages/teams-js/src/public/authentication.ts
```diff
@@ -444,7 +444,7 @@ export namespace authentication {
         authHandlers.success(result);
       }
     } finally {
-      authHandlers = null;
+      authHandlers = undefined;
```
AE-MS, line 447: >undefined [](http://example.com/codeflow?start=21&length=9)

Same deal as `authParams` above

packages/teams-js/src/internal/communication.ts
```diff
@@ -176,7 +176,7 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
  * @internal
  * Limited to Microsoft-internal use
  */
-export function sendMessageToParentAsync<T>(actionName: string, args: any[] = undefined): Promise<T> {
+export function sendMessageToParentAsync<T>(actionName: string, args: any[] | undefined = undefined): Promise<T> {
```
KangxuanYe, line 179: I am curious what is the differences between `undefined = undefined` and `undefined`? Could you please explain it a little bit?

AE-MS: `args: any[] = undefined`
says: "this parameter named `args` is an array and the elements in the array can be of any type. If the caller of this function does not provide a value for the `args` parameter, then the value it gets by default is `undefined`"
In this case, `args` is declared as being an array of any type of elements. It is not defined as allowing `undefined` as a value. Perhaps an easier way to understand this if it was declared like this:
`args: any[] = 5`
You would look at that and say "Wait if `args` is an array how can you assign a number to it? A number isn't an array."

`args: any[] | undefined = undefined` says "this parameter named `args` is an array and the elements in the array can be of any type OR it can be the value `undefined`. If the caller of this function does not provide a value for the `args` parameter, then the value it gets by default is `undefined`"

Does that help?

jadahiya-MSFT (approved):

packages/teams-js/src/internal/communication.ts
KangxuanYe: Yes, that is super helpful for type check.

KangxuanYe (approved):

PR #2023, opened 2023-11-02
[Communication] Add communication for nested app auth

amehagh:
## Description

Teams-js now injects nested app authentication bridge into the app's window which will be used by MSAL to make authentication calls to the top-level window.

Ref: [spec](https://microsoft-my.sharepoint-df.com/:w:/p/ameyagholkar/EUW2j9x9rF5Fsi0cvjywqo0B5jLkUvCgD1nfIbowmSEUHw?e=HpEKfa)

### Main changes in the PR:

- Adds reference for topWindow, topOrigin and topMessageQueue.
  - For these three functions, we will default to using `parentWindow` if top and parent are the same. If they're different, we'll use top window. The existing logic will stay unchanged since in most cases, top window and parent window are the same. 
  - Furthermore, the targetWindow passed into these functions will be top window _iff_ nested auth messages are being processed/sent.
- Adds the implementation for nested app auth bridge
- Adds the handler for nested app auth response messages
- Adds the handler for sending nested app auth message to the top level window
- Refactors the existing message communication logic to accommodate sending of message to the top level window
- UTs

### Validation performed:

1. Manual checks on existing SDK messages.
2. Checks passed.
3. MSAL added to test app and validated that message is sent correctly.
4. Once HubSDK changes are in, it'll unblock us from testing the receipt of messages from the top window.

### Unit Tests added:

Yes

### End-to-end tests added:

No (MSAL is needed for these scenarios)

### Change file added:

Yes

### Images

> MSAL.js is integrated into teams-test-app and initialized _after_ the SDK is initialized. `id: 3` message is posted from teams-js and received by the teams window

Screenshot 2023-11-14 at 9 37 26 PM


packages/teams-js/src/internal/communication.ts
```diff
@@ -562,6 +737,16 @@ function createMessageRequest(func: string, args: any[] | undefined): MessageReq
   };
 }
 
+function createNestedAppAuthRequest(message: string): NestedAuthRequest {
+  return {
+    id: CommunicationPrivate.nextMessageId++,
+    func: 'nestedAppAuthRequest',
+    timestamp: Date.now(),
+    args: [],
+    data: message,
```
TrevorJoelHarris, line 746: Might be a dumb question but I will ask anyways: why does data need to be separate property? It seems like it would "fit" more with the existing contract if args contained the message

TrevorJoelHarris (commented):
General approach looks okay to me, but some questions. Happy to talk more if it's helpful

packages/teams-js/src/internal/communication.ts
amehagh: I didn't want to use `args` because I didn't want to overload it. It also reduces the probablity of errors in the HubSDK itself i.e. if HubSDK is updated to pre-process `args` in the future, NAA messages might be miscategorized or throw errors since `args` are meant to be parameters to a callback in the Hub. 

Keeping the `data` in `event.data` would be an organic approach since it is siloed for NAA requests. Specifically, in the message processors since we saw that we have access to the complete `event`, we don't need to change anything but just rely on the existing pattern and get the `data` destructured from the event in the NAA processor.

packages/teams-js/src/internal/communication.ts
```diff
@@ -460,23 +577,43 @@ function handleChildMessage(evt: DOMMessageEvent): void {
...
+    return Communication.childOrigin;
+  } else if (targetWindow === Communication.topWindow) {
+    return Communication.topOrigin;
+  } else {
+    return null;
+  }
+}
+
+function getTargetName(targetWindow: Window | null): string | null {
+  if (targetWindow === Communication.parentWindow) {
```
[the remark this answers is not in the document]
amehagh: Yeah, that's a legit miss. Let me see what I can do.

packages/teams-js/src/internal/communication.ts
```diff
@@ -307,6 +359,71 @@ function processMessage(evt: DOMMessageEvent): void {
...
+  // Validate that it is a valid auth bridge response message
+  if (!parsedData || typeof parsedData !== 'object' || parsedData.messageType !== 'NestedAppAuthResponse') {
+    logger('Unrecognized data format received by app, message being ignored. Message: %o', evt);
+    return;
+  }
+
+  // Process only if the message is coming from a different window and a valid origin
+  // valid origins are either a pre-known origin or one specified by the app developer
+  // in their call to app.initialize
+  const messageSource = evt.source || (evt as unknown as DOMMessageEvent)?.originalEvent?.source;
```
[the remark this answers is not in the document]
amehagh: I kept it independent _because_ I didn't want to touch the existing logic for the core teams-js events. 

Are you pointing at the `messsageSource` and `shouldProcess` logic to be shared out? Then I think we could do that.

packages/teams-js/src/internal/communication.ts
```diff
@@ -307,6 +359,71 @@ function processMessage(evt: DOMMessageEvent): void {
   }
 }
 
+const processAuthBridgeMessageLogger = communicationLogger.extend('processAuthBridgeMessage');
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function processAuthBridgeMessage(evt: MessageEvent, onMessageReceived: (response: string) => void): void {
```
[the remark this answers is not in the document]
amehagh: Yeah, TODO item is already added.

packages/teams-js/src/internal/communication.ts
```diff
@@ -87,6 +99,13 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
     }
   }
 
+  // Extend the window, define the NAA listener and add the NAA bridge
+  const extendedWindow = Communication.currentWindow as unknown as NestedAuthExtendedWindow;
+  const nestedAppAuthBridge = createNestedAppAuthBridge(Communication.currentWindow);
```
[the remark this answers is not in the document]
amehagh: Hmm, how about we do this:

```suggestion
// Everything related to NAA bridge is encapsulated inside this function
tryPolyfillWithNestedAppAuthBridge(Communication.currentWindow);
```

packages/teams-js/src/internal/communication.ts
```diff
@@ -87,6 +99,13 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
     }
   }
 
+  // Extend the window, define the NAA listener and add the NAA bridge
+  const extendedWindow = Communication.currentWindow as unknown as NestedAuthExtendedWindow;
```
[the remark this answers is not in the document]
amehagh: Yes, the bridge has to be added via teams-js onto the app `window`. MSAL.js is also loaded side by side with teams-js into the app `window`. MSAL will look for this bridge on the app's `window` to send NAA messages.

The bridge internally will then allow MSAL from the app's `window` to talk to the top-level frame (Teams).

packages/teams-js/src/internal/interfaces.ts
```diff
@@ -38,6 +38,24 @@ export interface ExtendedWindow extends Window {
   onNativeMessage(evt: DOMMessageEvent): void;
 }
 
+/**
+ * @hidden
+ * Marty, add comments here
+ */
+export interface NestedAppAuthBridge {
```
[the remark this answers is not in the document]
amehagh: All good. Exactly the kind of feedback I was looking for.

packages/teams-js/src/internal/communication.ts
```diff
@@ -69,6 +80,7 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
     Communication.currentWindow.parent !== Communication.currentWindow.self
       ? Communication.currentWindow.parent
       : Communication.currentWindow.opener;
+  Communication.topWindow = Communication.currentWindow.top;
```
TrevorJoelHarris, line 83: One other thing I forgot to mention: in some cases there is *no* parent window to talk to (mobile scenarios for example). In that case, the current window has a handler on it for message passing to use instead of post messages

packages/teams-js/src/internal/communication.ts
```diff
@@ -87,6 +99,13 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
     }
   }
```
amehagh, line 101: Handle NAA messages for when it's being called from a mobile platform

packages/teams-js/src/internal/communication.ts
```diff
@@ -271,6 +287,42 @@ function sendMessageToParentHelper(actionName: string, args: any[] | undefined):
...
+}
+
+const sendNestedAuthRequestToTopWindowLogger = communicationLogger.extend('sendNestedAuthRequestToTopWindow');
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function sendNestedAuthRequestToTopWindow(message: string): NestedAuthRequest {
+  const logger = sendNestedAuthRequestToTopWindowLogger;
```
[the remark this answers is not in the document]
amehagh: Yeah, I won't be able to. I should be able to add a parse and check in the bridge itself.

packages/teams-js/src/internal/communication.ts
amehagh: As discussed offline, NAA is limited to windows with a DOM. So, as of this PR, frameless windows won't be considered.

packages/teams-js/src/internal/communication.ts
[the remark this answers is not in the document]
amehagh: @TrevorJoelHarris check out the update.

amehagh:
@baljesingh You were saying this change will need the security review from M365, correct? Is there a way we can get an intermediate build so that we can test out the e2e flow with our changes?

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/communication.ts
```diff
@@ -543,6 +694,93 @@ export function sendMessageEventToChild(actionName: string, args?: any[]): void
...
+
+  if (!window) {
+    logger('Cannot polyfill nestedAppAuthBridge as current window does not exist');
+    return;
+  }
+
+  const extendedWindow = Communication.currentWindow as unknown as NestedAuthExtendedWindow;
+  const nestedAppAuthBridge = createNestedAppAuthBridge(Communication.currentWindow);
+  if (nestedAppAuthBridge) {
+    extendedWindow.nestedAppAuthBridge = nestedAppAuthBridge;
```
jekloudaMSFT, line 717: Won't this replace an existing `nestedAppAuthBridge` if called in a native context? Is there some check so that this only gets called in web contexts?

amehagh: Yeah, good catch. Let me fix that.

packages/teams-js/src/internal/nestedAppAuthInterfaces.ts
```diff
@@ -0,0 +1,51 @@
+import { MessageRequestWithRequiredProperties } from './messageObjects';
```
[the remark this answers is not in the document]
amehagh: @TrevorJoelHarris Restructuring is done 👍  Take a peek and let me know what you think.

packages/teams-js/src/internal/nestedAppAuth.ts
```diff
@@ -0,0 +1,186 @@
...
+        !parsedMessage ||
+        typeof parsedMessage !== 'object' ||
+        parsedMessage.messageType !== NestedAppAuthMessageEventNames.Request
+      ) {
+        logger('Unrecognized data format received by app, message being ignored. Message: %o', message);
+        return;
+      }
+
+      // Post the message to the top window
+      handlePostMessage(message);
```
baljesingh, line 180: Security question: How are we protecting ourselves against the possibility of a malicious child iframe sending code to trigger XSS attack?

It's crucial to verify the origin of the received messages to prevent potential cross-site scripting attacks. I'm not sure if the plan is to inject the bridge into trusted child iframes only; even in that case, we still need to consider the possibility of a compromised child iframe. Additionally, if we don't know the child iframe in advance, there might be security implications. This question can be asked in security review.

baljesingh: Thanks. It makes sense.

packages/teams-js/src/internal/communication.ts
```diff
@@ -88,6 +97,12 @@ export function initializeCommunication(validMessageOrigins: string[] | undefine
     }
   }
 
+  // Extend the window, define the NAA listener and add the NAA bridge
+  tryPolyfillWithNestedAppAuthBridge(Communication.currentWindow, {
```
baljesingh, line 101: This method is being called in the case of a frameless window; I tested it on Mac application. Similarly, it should be invoked for iOS mobile as well. I was able to successfully call the "createNestedAppAuthBridge()" method, which I believe is unexpected because mobile platforms introduce their own NAA bridge. Is it possible to add filter to avoid unnecessary introduction of bridge for mobile?

baljesingh: As discussed on chat, I was having outdated changes. Looks good to me.

baljesingh (approved):

packages/teams-js/src/internal/communication.ts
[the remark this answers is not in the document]
amehagh: Hmm, I didn't see value refactoring this logic into a separate function right now. We need to destructure source and origin in both cases so those two lines will need to stay in the process functions. Furthermore, we already have `shouldProcessMessage` which is encapsulates the common checks into one function. 

So, I'd keep it as-is for now. If needed, we can always refactor later.

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -784,6 +795,49 @@ describe('Testing communication', () => {
...
+      communication.uninitializeCommunication();
+      app._initialize(utils.mockWindow);
+    });
+
+    afterAll(() => {
+      communication.Communication.currentWindow = utils.mockWindow;
+      communication.uninitializeCommunication();
+    });
+
+    it('should send a postMessage to top window when the top window and top origin are set and are same as the parent window', () => {
```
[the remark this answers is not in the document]
amehagh: I've updated the initialization code to polyfill the bridge only if the window is _not_ frameless. We can revisit once we know the requirements for NAA in frameless contexts.

baljesingh:
> @baljesingh You were saying this change will need the security review from M365, correct? Is there a way we can get an intermediate build so that we can test out the e2e flow with our changes?

I don't know why my comment was left in draft mode somehow. I think we may need to wait for the security review to give us the green light before merging it into production. Since we already have a intermediate build, How long can we wait for this PR to be merged?

amehagh:
> > @baljesingh You were saying this change will need the security review from M365, correct? Is there a way we can get an intermediate build so that we can test out the e2e flow with our changes?
> 
> I don't know why my comment was left in draft mode somehow. I think we may need to wait for the security review to give us the green light before merging it into production. Since we already have a intermediate build, How long can we wait for this PR to be merged?

@baljesingh where can I find an intermediate build for these changes?

baljesingh:
> > > @baljesingh You were saying this change will need the security review from M365, correct? Is there a way we can get an intermediate build so that we can test out the e2e flow with our changes?
> > 
> > 
> > I don't know why my comment was left in draft mode somehow. I think we may need to wait for the security review to give us the green light before merging it into production. Since we already have a intermediate build, How long can we wait for this PR to be merged?
> 
> @baljesingh where can I find an intermediate build for these changes?

Oh, you need an intermediate build for these changes. I mistakenly assumed it was related to corresponding SDK changes. Let me check on that.

alicialu-MSFT:
Flagging this PR because we need to resolve an open question before we can approve: 

1) Teams web is implementing pairwise broker for Nested App Auth. This means we need to inject a Nestd App Auth Bridge from within the app iframe, and we're doing that through Teams JS.
2) Under the current state, this means the nested app auth bridge will always be detected by the Nested app auth MSAL.js code and the default old browser auth will not be run.
3) This doesn't bode well for other hubs, like owa, that would just do the regular browser auth/general case. 
4) Since we need hub parity for NAA to be remotely useful for MetaOS we need to come to a compromise on how to enable both pwb and the general case.

amehagh:
> Flagging this PR because we need to resolve an open question before we can approve:
> 
> 1. Teams web is implementing pairwise broker for Nested App Auth. This means we need to inject a Nested App Auth Bridge from within the app iframe, and we're doing that through Teams JS.
> 2. Under the current state, this means the nested app auth bridge will always be detected by the Nested app auth MSAL.js code and the default old browser auth will not be run.
> 3. This doesn't bode well for other hosts, like owa, that would just do the regular browser auth/general case.
> 4. Since we need host parity for NAA to be remotely useful for MetaOS we need to come to a compromise on how to enable both pwb and the general case.

@TrevorJoelHarris @alicialu-MSFT @baljesingh This concern is addressed, and changes are pushed to this PR. We're now adding the bridge _iff_ the hub supports `nestedAppAuth`. This update should add the bridge only for hubs that have defined the required nested app auth service. In our case, it'll only be Teams for now.

packages/teams-js/src/internal/communication.ts
```diff
@@ -408,22 +421,34 @@ export function sendMessageToParent(actionName: string, argsOrCallback?: any[] |
...
+ * Limited to Microsoft-internal use
+ */
+function sendRequestToTargetWindowHelper(
+  targetWindow: Window,
+  request: MessageRequestWithRequiredProperties | NestedAppAuthRequest,
+): MessageRequestWithRequiredProperties | NestedAppAuthRequest {
+  const logger = sendRequestToTargetWindowHelperLogger;
 
   if (GlobalVars.isFramelessWindow) {
     if (Communication.currentWindow && Communication.currentWindow.nativeInterface) {
```
TrevorJoelHarris, line 454: Some of the logging in this function says it's for parent, but that may or may not be true now that we're using it for parent and sometimes top. Would be good to update those messages

packages/teams-js/src/internal/communication.ts
```diff
@@ -482,6 +529,82 @@ function processMessage(evt: DOMMessageEvent): void {
...
+
+  // Process only if we received a valid message
+  if (!evt || !evt.data || typeof evt.data !== 'object') {
+    logger('Unrecognized message format received by app, message being ignored. Message: %o', evt);
+    return;
+  }
+
+  const { args } = evt.data as MessageResponse;
+  const [, message] = args ?? [];
+  const parsedData = (() => {
```
TrevorJoelHarris, line 549: Is there a type we could add for this?

packages/teams-js/src/internal/nestedAppAuth.ts
```diff
@@ -0,0 +1,209 @@
...
+  bridgeHandlers: NestedAppAuthBridgeHandlers,
+): NestedAppAuthBridge | null {
+  const logger = createNestedAppAuthBridgeLogger;
+
+  if (!window) {
+    logger('nestedAppAuthBridge cannot be created as current window does not exist');
+    return null;
+  }
+
+  const { onMessage, handlePostMessage } = bridgeHandlers;
```
TrevorJoelHarris, line 171: very minor since this is internal only, but the name `handlePostMessage` makes me thing that this function is receiving and processing postMessages, when it's actually sending them. Maybe a rename to something like `sendPostMessage`?

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/communication.ts
amehagh: Added a simple type with `messageType`. I don't want to move the entire NAA MSAL types over so best to keep it strict to the attributes we'll be using.

packages/teams-js/src/internal/communication.ts
```diff
@@ -741,6 +903,29 @@ function createMessageRequest(
...
+ *
+ * @param {string} message - The message to be included in the request. This is typically a stringified JSON object containing the details of the authentication request.
+ * The reason for using a string is to allow complex data structures to be sent as a message while avoiding potential issues with object serialization and deserialization.
+ *
+ * @returns {NestedAppAuthRequest} Returns a NestedAppAuthRequest object with a unique id, the function name set to 'nestedAppAuthRequest', the current timestamp, an empty args array, and the provided message as data.
+ */
+function createNestedAppAuthRequest(message: string): NestedAppAuthRequest {
+  return {
+    id: CommunicationPrivate.nextMessageId++,
+    func: 'nestedAppAuthRequest',
```
baljesingh, line 920: After having a discussion with the team, the func name should follow the pattern of 'capabilityName.ApiName.'. Could you please make the necessary update? 'nestedAppAuth.execute' seems good to me. Feel free to use any alternative name.

amehagh: Done.

TrevorJoelHarris (approved):

baljesingh:
Shouldn't we need to define a new capability, i.e., nestedAppAuth, in runtime.ts?

packages/teams-js/src/internal/nestedAppAuth.ts
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  if (!window) {
+    logger('Cannot polyfill nestedAppAuthBridge as current window does not exist');
+    return;
+  }
+
+  const parsedClientSupportedSDKVersion = (() => {
+    try {
+      return JSON.parse(clientSupportedSDKVersion);
```
baljesingh, line 135: Could you please confirm what the value of 'clientSupportedSDKVersion' would be? I want to validate 
```
parsedClientSupportedSDKVersion.supports?.nestedAppAuth
```

amehagh: That's what you have defined here: https://office.visualstudio.com/ISS/_git/metaos-hub-sdk/pullrequest/2481040?_a=files&path=/packages/metaos-hub-sdk/src/Services/MessageHandling/app-message-handler.ts

amehagh: The runtime config of _teams-js_ defines that the app developers can call at runtime via the teams-js APIs. teams-js runtime config is validated using `isSupported`. 

What I'm parsing in the above case is the configuration I get from the HubSDK. So if the hub has defined NAA service, then I'd get `{}` else I'd get `undefined`.

packages/teams-js/src/internal/nestedAppAuth.ts
```diff
@@ -0,0 +1,223 @@
...
+    } catch (e) {
+      return null;
+    }
+  })();
+
+  if (!parsedClientSupportedSDKVersion || !parsedClientSupportedSDKVersion.supports?.nestedAppAuth) {
```
jekloudaMSFT, line 141: The `supports` object has been used for the capability namespaces in team-js until now. Would it be more appropriate to add a new optional property to the top-level configuration like `supportsNestedAppAuth` to avoid overloading the `supports` property? cc: @baljesingh

salaman:
Quick scan, but I'm not seeing a way for an app to check if NAA is supported or not by the hub, and I think it might be important to have in the short/medium-term. OfficeJS has a similar requirement to provide a mechanism for addins to query if NAA is supported (`Office.context.requirements.isSetSupported("NestedAppAuth")`.

In the transition phase as 1P apps onboard, and in the future, we'll have potentially many old clients in the wild that won't support NAA. Outlook Desktop, for example, takes months to release changes to prod, and we have extended support channels like Semi-Annual which will guarantee many clients in prod will be on versions of Outlook without NAA support. If apps start adopting NAA then they will either break auth in prod for customers on these builds or significantly degrade the user experience.

If an app can query if NAA is supported or not then they can fallback to getAuthToken/OBO as they do today.

I'll also note that this support function cannot be as simple as return true if hub supports `nestedAppAuth` -- Outlook Desktop for example will not rely on Hub SDK/TeamsJS to inject the bridge but will support NAA, and OWA will not be using the NAA bridge at all but MSAL will work with cookie auth.

packages/teams-js/src/internal/nestedAppAuth.ts
amehagh: As far as definitions go, hubs needs to provide NAA service. That's the reason it's under the `supports` property. If the `supports` feels overloaded even for a service that the hub provides, I'm up for moving this to it's own property.

amehagh:
> Quick scan, but I'm not seeing a way for an app to check if NAA is supported or not by the hub, and I think it might be important to have in the short/medium-term. OfficeJS has a similar requirement to provide a mechanism for addins to query if NAA is supported (`Office.context.requirements.isSetSupported("NestedAppAuth")`.
> 
> In the transition phase as 1P apps onboard, and in the future, we'll have potentially many old clients in the wild that won't support NAA. Outlook Desktop, for example, takes months to release changes to prod, and we have extended support channels like Semi-Annual which will guarantee many clients in prod will be on versions of Outlook without NAA support. If apps start adopting NAA then they will either break auth in prod for customers on these builds or significantly degrade the user experience. Similarly, this will hit every hub as they gradually implement NAA functionality or have staged rollouts.
> 
> If an app can query if NAA is supported or not then they can fallback to getAuthToken/OBO as they do today.
> 
> I'll also note that this support function cannot be as simple as return true if hub supports `nestedAppAuth` -- Outlook Desktop for example will not rely on Hub SDK/TeamsJS to inject the bridge but will support NAA, and OWA will not be using the NAA bridge at all but MSAL will work with cookie auth.
> 
> Is there a plan for how to tackle these scenarios?

Thanks for the comment. If you take a deeper look at the changes, the current design injects the bridge that's necessary for MSAL to [consider using NAA](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/naa/BridgeProxy.ts#L51). NAA bridge injected via teams-js allows MSAL to send NAA requests. This is considering how the apps will work in Teams.

In case of the Teams hub, it needs to provide an implementation to the hubSDK for NAA which will proxy these messages to it's auth stack. 

Given the above information, check for `supports.nestedAppAuth` ensures that we inject the bridge only if the hub provides the implementation of NAA to HubSDK.

The above approach is with the understanding that HubSDK is integrated into other hubs as it is in Teams. If that's not the case, let me know. cc @baljesingh 

For the cases called out for Outlook, who's a good point of contact for consult? @baljesingh can you talk to partners on Outlook who can shed more light on how hubSDK is integrated into these cases so we can come up with the signal to send apps via teams-js indicating if the hub supports NAA?

To further clarify these cases:

> Outlook Desktop for example will not rely on Hub SDK/TeamsJS to inject the bridge but will support NAA

I do not understand this case. As I've linked above, MSAL needs the bridge to initialize NAA. Is the bridge injected natively for Outlook desktop? How does Outlook desktop "support NAA"? Is HubSDK/teams-js SDK _also_ used for Outlook desktop ecosystem?

> OWA will not be using the NAA bridge at all but MSAL will work with cookie auth

Same questions for OWA. What are the pieces at play? Does OWA not integrate with HubSDK/teams-js? Does an app running in OWA use MSAL as-is for auth?

It'd be great to link some examples of an app that runs across these platforms which showcases their current auth setup.

I'd lean to adding this new "signal" as a follow up PR rather than adding to this existing PR. Why? Because the bridge from this PR is added _conditionally_ and that condition will hold _true_ _only_ for Teams. It also unblocks us to prove a happy path and a foundation to iterate and extend across to other platforms.

Finally, @salaman .. I think you, me and @baljesingh need to chat over a call. GH comments aren't the best way to get such complex ideas across. I'll set something up next week.

salaman:
@amehagh I would be the contact for Outlook (desktop, OWA and "new" desktop) -- happy to discuss over a call. Just to briefly respond to a few of the points here for visibility:

>Is the bridge injected natively for Outlook desktop? How does Outlook desktop "support NAA"? Is HubSDK/teams-js SDK also used for Outlook desktop ecosystem?

Yes. The bridge is injected into frames completely outside of Hub SDK and TeamsJS in the native implementation, providing NAA for both addins and Teams apps. For Teams apps, we still use the Hub SDK, but will not be providing an implementation of the `nestedAppAuth` service so that TeamsJS does not perform its own injection, and MSAL will detect and use the bridge. My concern then is that we don't seem to have a cohesive way to let apps know this.

>Same questions for OWA. What are the pieces at play? Does OWA not integrate with HubSDK/teams-js? Does an app running in OWA use MSAL as-is for auth?

OWA also uses the Hub SDK, but won't be implementing the `nestedAppAuth` service either because cookie-based auth is enough for OWA and MSAL to be happy with no bridge at play.

>Because the bridge from this PR is added conditionally and that condition will hold true only for Teams.

I assume the condition you speak of is the presence of the `nestedAppAuth` service, which is also going to be used by the "new" Outlook native client. We are waiting, just like Teams, on the availability of this bridge work in the Hub SDK to enable the native client to execute NAA requests.

I guess my core question is, even for Teams: take any app currently relying on getAuthToken and the OBO flow, and get it to migrate over to NAA through MSAL. What happens if that app is loaded in a Teams client that doesn't have a version of the Hub SDK with `nestedAppAuth`, and thus no injected bridge within the frame? Feel free to save for the offline discussion, just wanted to provide context for my questions :)

baljesingh (approved):

packages/teams-js/src/internal/nestedAppAuth.ts
TrevorJoelHarris, line 141: @jekloudaMSFT does nestedAppAuth need to be represented in runtime.ts, even though this is the only place it is currently being checked?

jekloudaMSFT: Yes, I think so. Wasn't it already added? I thought we had discussed this before, was there a different PR that added it to runtime.ts?

TrevorJoelHarris (approved):

jekloudaMSFT (approved):

baljesingh (approved):
Changes look good to me.

PR #2024, opened 2023-11-03
Fix more violations of strictNullChecks (part 4)

AE-MS:
## Description

More `strictNullChecks` violations are being fixed in this change. This is continuing the journey from the previous PRs (but no prior knowledge needed to review this one!)

### Main changes in the PR:

I've annotated the specific changes directly on the code in the PR.

### Validation performed:

### Change file added:

Yes.

### Related PRs:

#2012, #2014, #2020

### Next/remaining steps:

Keep plugging away at these violations incrementally until we can enable `strictNullChecks` for the project.

### DIFF
--- a/packages/teams-js/src/internal/globalVars.ts
+++ b/packages/teams-js/src/internal/globalVars.ts
@@ -3,10 +3,10 @@ export class GlobalVars {
   public static initializeCalled = false;
   public static initializeCompleted = false;
   public static additionalValidOrigins: string[] = [];
-  public static initializePromise: Promise<void> | null = null;
+  public static initializePromise: Promise<void> | undefined = undefined;
   public static isFramelessWindow = false;
-  public static frameContext: FrameContexts | undefined;
-  public static hostClientType: string | undefined;
+  public static frameContext: FrameContexts | undefined = undefined;
+  public static hostClientType: string | undefined = undefined;
   public static clientSupportedSDKVersion: string;
   public static printCapabilityEnabled = false;
 }
--- a/packages/teams-js/src/internal/globalVars.ts
+++ b/packages/teams-js/src/internal/globalVars.ts
@@ -3,10 +3,10 @@ export class GlobalVars {
   public static initializeCalled = false;
   public static initializeCompleted = false;
   public static additionalValidOrigins: string[] = [];
-  public static initializePromise: Promise<void>;
+  public static initializePromise: Promise<void> | null = null;
   public static isFramelessWindow = false;
-  public static frameContext: FrameContexts;
-  public static hostClientType: string;
+  public static frameContext: FrameContexts | undefined;
+  public static hostClientType: string | undefined;
   public static clientSupportedSDKVersion: string;
   public static printCapabilityEnabled = false;
 }
--- a/packages/teams-js/src/internal/handlers.ts
+++ b/packages/teams-js/src/internal/handlers.ts
@@ -7,6 +7,7 @@ import { runtime } from '../public/runtime';
 import { Communication, sendMessageEventToChild, sendMessageToParent } from './communication';
 import { ensureInitialized } from './internalAPIs';
 import { getLogger } from './telemetry';
+import { isNullOrUndefined } from './typeCheckUtilities';
 
 const handlersLogger = getLogger('handlers');
 
@@ -155,7 +156,7 @@ export function registerHandlerHelper(
  */
 export function registerOnThemeChangeHandler(handler: (theme: string) => void): void {
   HandlersPrivate.themeChangeHandler = handler;
-  handler && sendMessageToParent('registerHandler', ['themeChange']);
+  !isNullOrUndefined(handler) && sendMessageToParent('registerHandler', ['themeChange']);
 }
 
 /**
@@ -180,7 +181,7 @@ export function handleThemeChange(theme: string): void {
  */
 export function registerOnLoadHandler(handler: (context: LoadContext) => void): void {
   HandlersPrivate.loadHandler = handler;
-  handler && sendMessageToParent('registerHandler', ['load']);
+  !isNullOrUndefined(handler) && sendMessageToParent('registerHandler', ['load']);
 }
 
 /**
@@ -207,7 +208,7 @@ function handleLoad(context: LoadContext): void {
  */
 export function registerBeforeUnloadHandler(handler: (readyToUnload: () => void) => boolean): void {
   HandlersPrivate.beforeUnloadHandler = handler;
-  handler && sendMessageToParent('registerHandler', ['beforeUnload']);
+  !isNullOrUndefined(handler) && sendMessageToParent('registerHandler', ['beforeUnload']);
 }
 
 /**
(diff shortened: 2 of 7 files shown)


AE-MS:

No changes were made to the runtime object and therefore a new runtime version was not required. Good day to you.

packages/teams-js/src/internal/globalVars.ts
```diff
@@ -3,10 +3,10 @@ export class GlobalVars {
   public static initializeCalled = false;
   public static initializeCompleted = false;
   public static additionalValidOrigins: string[] = [];
-  public static initializePromise: Promise<void>;
+  public static initializePromise: Promise<void> | null = null;
   public static isFramelessWindow = false;
-  public static frameContext: FrameContexts;
-  public static hostClientType: string;
+  public static frameContext: FrameContexts | undefined;
+  public static hostClientType: string | undefined;
```
AE-MS, line 9: Since these two variables were previously declared but not initialized, they were already being set to `undefined` by default. This change updates the type to reflect the behavior that was already occurring.

packages/teams-js/src/internal/globalVars.ts
```diff
@@ -3,10 +3,10 @@ export class GlobalVars {
   public static initializeCalled = false;
   public static initializeCompleted = false;
   public static additionalValidOrigins: string[] = [];
-  public static initializePromise: Promise<void>;
+  public static initializePromise: Promise<void> | null = null;
```
AE-MS, line 6: > public static initializePromise: Promise<void> | null = null; [](http://example.com/codeflow?start=1&length=62)

Similar to the below, this variable was already getting a value of `undefined` because it wasn't being initialized. However, various parts of code elsewhere was setting this to `null` in certain cases. To reflect those existing `null` use cases, I updated the type to accept `null` (since it was already being used) and explicitly set the default value to `null` instead of `undefined` since the existing code conventions indicated that `null` was being used for the uninitialized case rather than `undefined`.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -155,7 +156,7 @@ export function registerHandlerHelper(
  */
 export function registerOnThemeChangeHandler(handler: (theme: string) => void): void {
   HandlersPrivate.themeChangeHandler = handler;
-  handler && sendMessageToParent('registerHandler', ['themeChange']);
+  !isNullOrUndefined(handler) && sendMessageToParent('registerHandler', ['themeChange']);
```
AE-MS, line 159: >!isNullOrUndefined(handler) [](http://example.com/codeflow?start=2&length=27)

Without this, the compiler will warn you because it thinks you are trying to call `handler()` but forgot the parenthesis. In this case, the code is using the `handler` structure to validate that it has a real value.

packages/teams-js/src/private/files.ts
```diff
@@ -799,7 +799,10 @@ export namespace files {
    * @internal
    * Limited to Microsoft-internal use
    */
-  export function openDownloadFolder(fileObjectId: string = undefined, callback: (error?: SdkError) => void): void {
+  export function openDownloadFolder(
+    fileObjectId: string | undefined = undefined,
```
AE-MS, line 803: > | undefined [](http://example.com/codeflow?start=24&length=12)

Obviously the type on this parameter must include `undefined` since it's literally being set to `undefined` as the default value here.

packages/teams-js/src/private/logs.ts
```diff
@@ -25,8 +26,8 @@ export namespace logs {
    */
   export function registerGetLogHandler(handler: () => string): void {
     // allow for registration cleanup even when not finished initializing
-    handler && ensureInitialized(runtime);
-    if (handler && !isSupported()) {
+    !isNullOrUndefined(handler) && ensureInitialized(runtime);
```
AE-MS, line 29: >isNullOrUndefined [](http://example.com/codeflow?start=5&length=17)

Same as in `files.ts`

packages/teams-js/src/public/app.ts
```diff
@@ -556,7 +557,7 @@ export namespace app {
    * Gets the Frame Context that the App is running in. See {@link FrameContexts} for the list of possible values.
    * @returns the Frame Context.
    */
-  export function getFrameContext(): FrameContexts {
+  export function getFrameContext(): FrameContexts | undefined {
```
AE-MS, line 560: >| undefined [](http://example.com/codeflow?start=51&length=11)

Since `GlobalVars.frameContext` can be `undefined`, that is reflected here as well.

packages/teams-js/src/public/app.ts
```diff
@@ -699,7 +700,11 @@ export namespace app {
         processAdditionalValidOrigins(validMessageOrigins);
       }
 
-      resolve(GlobalVars.initializePromise);
+      if (GlobalVars.initializePromise !== null) {
+        resolve(GlobalVars.initializePromise);
+      } else {
+        initializeHelperLogger('GlobalVars.initializePromise is unexpectedly null');
+      }
```
AE-MS, line 707: Since the compiler now realizes that `GlobalVars.initializePromise`, it (correctly) won't let you call `resolve` with something that could be `null`. Now we are logging when we would already have been in that unexpected situation.

packages/teams-js/src/public/app.ts
```diff
@@ -732,8 +737,8 @@ export namespace app {
     GlobalVars.initializeCompleted = false;
     GlobalVars.initializePromise = null;
     GlobalVars.additionalValidOrigins = [];
-    GlobalVars.frameContext = null;
-    GlobalVars.hostClientType = null;
+    GlobalVars.frameContext = undefined;
+    GlobalVars.hostClientType = undefined;
```
AE-MS, line 741: These two were already possibly `undefined` in some cases, so I updated this code to use `undefined` as the uninitialized indicator so that there's only one way that's used to indicate "uninitialized"

packages/teams-js/src/public/app.ts
```diff
@@ -937,7 +942,7 @@ function transformLegacyContextToAppContext(legacyContext: LegacyContext): app.C
       sourceOrigin: legacyContext.sourceOrigin,
     },
     user: {
-      id: legacyContext.userObjectId,
+      id: legacyContext.userObjectId ?? '',
```
AE-MS, line 945: >userObjectId [](http://example.com/codeflow?start=24&length=12)

`userObjectId` is declared as possibly being `undefined` but `id` cannot be `undefined` so this unexpected case is now translated into the closest corresponding legal value for this property (empty string)

packages/teams-js/src/public/runtime.ts
```diff
@@ -512,7 +512,10 @@ export function generateVersionBasedTeamsRuntimeConfig(
   Object.keys(mapVersionToSupportedCapabilities).forEach((versionNumber) => {
     if (compareSDKVersions(highestSupportedVersion, versionNumber) >= 0) {
       mapVersionToSupportedCapabilities[versionNumber].forEach((capabilityReqs) => {
-        if (capabilityReqs.hostClientTypes.includes(GlobalVars.hostClientType)) {
+        if (
+          GlobalVars.hostClientType !== undefined &&
```
AE-MS, line 516: >GlobalVars.hostClientType !== undefined [](http://example.com/codeflow?start=10&length=39)

This unexpected case is now handled properly by skipping this code which would be meaningless if there was no `hostClientType` -- also many other things would blow up before this.

shrshindeMSFT (approved):
LGTM 🚢

lakhveerk (approved):
🚢

PR #2025, opened 2023-11-03
Fix the last of the remaining "easy" strictNullChecks violations

AE-MS:
## Description

This fixes the last of the "easy" fixes of `strictNullChecks` violations in the product code.

### Main changes in the PR:

The code is directly annotated in this PR to describe the individual fixes.

### Validation performed:

### Change file added:

Yes

### Related PRs:

#2012, #2014, #2020, #2024

### Next/remaining steps:

There still remain a few violations in the product code that will be more invasive to fix and will come next. They will require some refactoring or other functionality changes (or changes that are theoretical functionality changes but likely never get hit in production)).

There are also still a ton of violations in the test code that will also be fixed, but likely not until the product code is fully clean.

In case you're curious, here are the stats for the remaining violations in the product code:
```
Errors  Files
     5  src/internal/communication.ts:183
     1  src/public/app.ts:938
```

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -207,7 +207,7 @@ export function sendMessageToParent(actionName: string, callback?: Function): vo
  * @internal
  * Limited to Microsoft-internal use
  */
-export function sendMessageToParent(actionName: string, args: any[], callback?: Function): void;
+export function sendMessageToParent(actionName: string, args: any[] | undefined, callback?: Function): void;
 
 /**
  * @internal
--- a/packages/teams-js/src/internal/handlers.ts
+++ b/packages/teams-js/src/internal/handlers.ts
@@ -19,17 +19,17 @@ class HandlersPrivate {
   public static handlers: {
     [func: string]: Function;
   } = {};
-  public static themeChangeHandler: null | ((theme: string) => void);
+  public static themeChangeHandler: null | ((theme: string) => void) = null;
   /**
    * @deprecated
    */
-  public static loadHandler: null | ((context: LoadContext) => void);
+  public static loadHandler: null | ((context: LoadContext) => void) = null;
   /**
    * @deprecated
    */
-  public static beforeUnloadHandler: null | ((readyToUnload: () => void) => boolean);
-  public static beforeSuspendOrTerminateHandler: null | (() => void);
-  public static resumeHandler: null | ((context: ResumeContext) => void);
+  public static beforeUnloadHandler: null | ((readyToUnload: () => void) => boolean) = null;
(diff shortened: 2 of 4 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -207,7 +207,7 @@ export function sendMessageToParent(actionName: string, callback?: Function): vo
  * @internal
  * Limited to Microsoft-internal use
  */
-export function sendMessageToParent(actionName: string, args: any[], callback?: Function): void;
+export function sendMessageToParent(actionName: string, args: any[] | undefined, callback?: Function): void;
```
AE-MS, line 210: >undefined [](http://example.com/codeflow?start=70&length=9)

This function was already handling the case where the parameter was possibly `undefined`, and now the type indicates that.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -19,17 +19,17 @@ class HandlersPrivate {
   public static handlers: {
     [func: string]: Function;
   } = {};
-  public static themeChangeHandler: (theme: string) => void;
+  public static themeChangeHandler: null | ((theme: string) => void);
```
AE-MS, line 22: >themeChangeHandler [](http://example.com/codeflow?start=16&length=18)

All of these handlers are eventually set to `null` in the `uninitializeHandlers()` method, and this reflects that reality.

packages/teams-js/src/private/conversations.ts
```diff
@@ -155,7 +155,7 @@ export namespace conversations {
         registerHandler(
           'startConversation',
           (subEntityId: string, conversationId: string, channelId: string, entityId: string) =>
-            openConversationRequest.onStartConversation({
+            openConversationRequest.onStartConversation?.({
```
AE-MS, line 158: >openConversationRequest.onStartConversation?. [](http://example.com/codeflow?start=12&length=45)

Even though the code above this appears to try to handle this case by doing `if (openConversationRequest.onStartConversation)`, that is unfortunately not sufficient for this pattern because the in the call to `registerHandler`, an arrow function is used as the second parameter, which is creating a new inline function that is calling `openConversationRequest.onStartConversation`. Between this point when `registerHandler` is called and when that arrow function is invoked, `openConversationRequest.onStartConversation` could become `undefined` (that arrow function may not be executed until some time way in the future after many things have happened).

Using the `?.` operator ("optional chaining operator") here will cause this function to be called if it exists and will cause nothing to happen in the function is `undefined`. No other special consideration is needed since the function returns `void` and it won't matter if it doesn't execute (or rather, it won't matter if _there's nothing to execute_)

shrshindeMSFT (approved):
LGTM ! 🚢

KangxuanYe (approved):

PR #2026, opened 2023-11-03
Enable strictNullChecks for product code!

AE-MS:
## Description

After all the recent PRs to fix violations of `strictNullChecks`, the product code is now down to only 6 remaining violations. These are more invasive to fix and so not to be taken lightly. There are also still a ton of violations in the test code.

However! There are so few violations now that I am comfortable actually enabling `strictNullChecks` for the product code and just suppressing the compiler warning for the specific remaining violating lines! This is a big step forward because it will prevent any new code from adding new violations!

### Main changes in the PR:

1. Enable `strictNullChecks` in the `tsconfig.json` file for the product code -- this will ensure the compiler now enforces `strictNullChecks` for all product code.
2. Suppress the remaining violations of this in `communication.ts` and `app.ts`. Note that there is currently no way to only suppress specific TS compiler errors for a line, you have to suppress all TS errors for that line. I am comfortable with this, given how few locations this includes.
3. Suppress the eslint warning that tells you not to use `@ts-ignore` for all the lines in number 2 above
4. Disable `strictNullChecks` for the unit test files. Note that this is not fundamentally changing behavior: the unit tests, by default, inherit the same tsconfig settings from the product. Prior to this change, `strictNullChecks` was not enabled. Now that I'm enabling it for the product code, I have to override it for the unit tests to put it back to disabled (for just the unit tests).

### Validation performed:

Ensured that all existing builds and tests pass.

### Change file added:

Yes.

### Related PRs:

#2012, #2014, #2020, #2024, #2025

### Next/remaining steps:

Enabling this compiler flag is a big step forward. The next steps will be to actually start fixing some of the remaining violations that are suppressed, along with working to enable the `strictNullChecks` flag for the test code.

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -179,7 +179,8 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
 export function sendMessageToParentAsync<T>(actionName: string, args: any[] | undefined = undefined): Promise<T> {
   return new Promise((resolve) => {
     const request = sendMessageToParentHelper(actionName, args);
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
     resolve(waitForResponse<T>(request.id));
   });
 }
@@ -224,6 +225,8 @@ export function sendMessageToParent(actionName: string, argsOrCallback?: any[] |
   /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
   const request = sendMessageToParentHelper(actionName, args);
   if (callback) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
     CommunicationPrivate.callbacks[request.id] = callback;
   }
 }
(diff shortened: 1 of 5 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable strict-null-checks/all */
```
AE-MS, line 3: `strictNullChecks` is checked by both the TS compiler and eslint. Turning off the eslint checking of `strictNullChecks` here does not put us in a worse position because the compiler is still checking for it in this file. It just means that I don't have to suppress the violations BOTH for the linter and the compiler.

Eventually I will just remove the lint rule detection because the compiler detection is sufficient. But not yet.

packages/teams-js/src/internal/communication.ts
```diff
@@ -179,7 +180,8 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
 export function sendMessageToParentAsync<T>(actionName: string, args: any[] | undefined = undefined): Promise<T> {
   return new Promise((resolve) => {
     const request = sendMessageToParentHelper(actionName, args);
-    /* eslint-disable-next-line strict-null-checks/all */ /* Fix tracked by 5730662 */
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
AE-MS, line 183: >ban-ts-comment [](http://example.com/codeflow?start=51&length=14)

`ban-ts-comment` is the lint rule that tells you "don't use `@ts-ignore`!"

lakhveerk (approved):

TrevorJoelHarris (approved):

PR #2030, opened 2023-11-07
Extend RequestAppAudioHandling to support for audio device selection sync

bingliang-zh:
## Description

This is a continuous PR of the original [PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1931) by [shilcare](https://github.com/shilcare). There is another outdated PR from a fork [here](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2008), change to this branch PR because of security issues.

### Purpose

Give Teams apps the ability to use Teams' audio devices settings, like play audio on the device that user selected in Teams.

### Details

Add a callback method to `RequestAppAudioHandling`.
- When App first call `RequestAppAudioHandling` with this callback, the callback will be called immediately with Teams' current audio device labels. The app can use this info to initialize and select the corresponding audio devices.
- This callback will also be called when user changes audio devices in Teams, app can update its audio devices accordingly.

### Main changes in the PR:

1. add optional `audioDeviceSelectionChangedCallback` to `RequestAppAudioHandlingParams` which is used in `meeting.requestAppAudioHandling`
2. register event handler in `startAppAudioHandling`
3. unregister event handler in `stopAppAudioHandling`

### Validation performed:

To make this work properly, we also need code changes in some internal repos. We have built a Teams container with those changes.

1. Use Teams 2.1 with the custom container
2. Create a test Teams app, add some test code like
```typescript
  meeting.requestAppAudioHandling(
    {
      //...
      audioDeviceSelectionChangedCallback: (selectedDevices) => {
        console.log('audioDeviceSelectionChangedCallback');
        console.log(selectedDevices);
        return Promise.resolve();
      },
    },
   //...
  );
```
3. Replace an existing meeting app, redirect to the local test app
4. Open the modified meeting app.
5. The callback will be called after `meeting.requestAppAudioHandling` is called. And also get called after device is changed in Teams settings. The payload is correct too.

### Unit Tests added:

Yes

### End-to-end tests added:

`requestAppAudioHandling API Call - register audioDeviceSelectionChanged Handler`

### Change file added:

No. `pnpm changefile` says No change files are needed.

### Related PRs:

- <https://github.com/OfficeDev/microsoft-teams-library-js/pull/1931>
- <https://github.com/OfficeDev/microsoft-teams-library-js/pull/2008>

### Next/remaining steps:

Close [original PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/1931) because duplicated and outdated.
Close [fork PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2008) because duplicated and outdated.


apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -203,6 +203,18 @@
       "expectedAlertValue": "requestAppAudioHandling called",
       "expectedTestAppValue": "requestAppAudioHandling() succeeded: isHostAudioless=true"
     },
+    {
+      "title": "requestAppAudioHandling API Call - register audioDeviceSelectionChanged Handler",
```
KangxuanYe, line 207: Suggest name by following pattern: "requestAppAudioHandling API Call - register audioDeviceSelectionChanged Handle **- Success**"

(We would like to know for a case that whether we expect success or error)

KangxuanYe (changes requested):

apps/teams-test-app/e2e-test-data/meeting.json
KangxuanYe: Also I think you need to update Web Host SDK part first so that you could find this event name in side panel.

bingliang-zh: @KangxuanYe , I notice e2e pipeline fails. So I need to merge the PR in meta-hub-sdk first, then this?

bingliang-zh: > Suggest name by following pattern: "requestAppAudioHandling API Call - register audioDeviceSelectionChanged Handle **- Success**"
> 
> (We would like to know for a case that whether we expect success or error)

Updated

KangxuanYe: > @KangxuanYe , I notice e2e pipeline fails. So I need to merge the PR in metaos-hub-sdk first, then this?

Correct. Because E2E test is testing against on main branch of that repo. And I believe you have another PR on updating test app to find the corresponding event name. You need to finish that part and get it checked in and then come to this place.

bingliang-zh: Got it, pushing that PR now. Will get back to this once that is merged.

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -203,6 +203,18 @@
       "expectedAlertValue": "requestAppAudioHandling called",
       "expectedTestAppValue": "requestAppAudioHandling() succeeded: isHostAudioless=true"
     },
+    {
```
shrshindeMSFT, line 206: Do we need a failure case for this ?

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+   * @beta
+   */
+  export interface AudioDeviceSelection {
+    speaker?: AudioDeviceInfo;
+    microphone?: AudioDeviceInfo;
+    error?: SdkError;
+  }
+
+  /**
+   * Interface for AudioDeviceInfo, includes a device label with the same format as `MediaDeviceInfo.label`
```
shrshindeMSFT, line 361: Can we add @link for `MediaDeviceInfo.label` same for other mentioned in the comment.

shrshindeMSFT (changes requested):
LGTM just two minor comments.

packages/teams-js/test/utils.ts
```diff
@@ -130,9 +130,22 @@ export class Utils {
   };
 
   public findMessageByFunc = (func: string): MessageRequest | null => {
```
KangxuanYe, line 132: FindMessageByFunc and FindKthMessageByFunc could be combined to one by setting k: number as an optional parameter

KangxuanYe (changes requested):

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface AudioDeviceSelection {
+    speaker?: AudioDeviceInfo;
+    microphone?: AudioDeviceInfo;
+    error?: SdkError;
+  }
```
AE-MS, line 358: Continuing our conversation from the other PR...

I totally understand that a developer can certainly cast use `as any` and completely remove any type safety. We don't need to design or optimize for developers who are intentionally eschewing the safety provided by TypeScript. However, we should design our APIs so that developers who are trying to leverage TypeScript can do so in a way whereby it's hard or impossible to use the API/data structures in the wrong way. The design of the data should match the semantics: if it's not possible to have both an error and audio data, then the data structure should model that reality, enforced by the type system.

Here is an example of how leveraging the type system makes it impossible for a developer to use the data in the wrong way (assuming they aren't trying to intentionally subvert the type system):

```
function isSdkError(potentialError: unknown): potentialError is SdkError {
  return (potentialError as SdkError).errorCode !== undefined;
}

export interface AudioDeviceSelection {
  speaker?: AudioDeviceInfo;
  microphone?: AudioDeviceInfo;
}

interface RequestAppAudioHandlingParams {
...
  audioDeviceSelectionChangedCallback?: (resultOrError: AudioDeviceSelection | SdkError) => void;
}

function devCallback(selectedDevicesOrError: AudioDeviceSelection | SdkError): void {
  if (isSdkError(selectedDevicesOrError)) {
    // handle error
  }

  if (selectedDevicesOrError.microphone) {
    // the above line will not compile if an SdkError was returned; TypeScript recognizes that to access the microphone property, the type has to first be checked to ensure it is not SdkError
  }
}
```

Naturally there are other designs that also achieve the same goal (e.g., using a union on the object passed to the callback, rather than passing an AudioDeviceSelection | SdkError) -- but the common theme across the designs is that it enables the TypeScript compiler to prevent misuse of the data and the data models the semantics of what is and is not possible.

packages/teams-js/src/public/meeting.ts
```diff
@@ -334,6 +334,47 @@ export namespace meeting {
...
+   * Hide from docs.
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface AudioDeviceSelection {
+    speaker?: AudioDeviceInfo;
+    microphone?: AudioDeviceInfo;
```
AE-MS, line 356: Since you said in the other PR that it's valid to somethings have no error, no speaker, and no microphone, can you update the documentation so that developers will know how to handle that case? I believe you said they should get default devices instead?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/meeting.ts
bingliang-zh: Updated

apps/teams-test-app/e2e-test-data/meeting.json
bingliang-zh: Added

packages/teams-js/test/utils.ts
bingliang-zh: Updated with doc.

packages/teams-js/src/public/meeting.ts
bingliang-zh: Updated.

packages/teams-js/src/public/meeting.ts
bingliang-zh: Added

AE-MS (approved):
:shipit:

shrshindeMSFT:
Will approve once the tests are passing

bingliang-zh:
@shrshindeMSFT  @KangxuanYe The PR for meta-os has been merged. All checks have passed. Mind take a look at this again? Thanks!

shrshindeMSFT (approved):
LGTM!

KangxuanYe (approved):

PR #2033, opened 2023-11-07
Create new, alternate, MessageRequest structure with expected required properties

AE-MS:
## Description

There are some violations of `structNullChecks` remaining in `communication.ts` that are fundamentally unfixable without program flow changes. For example, the `MessageRequest` structure has some properties which are optional, but there is code in `communication.ts` that assumes they are always defined. Generally speaking, fixing these will require one-by-one incremental changes to the code structure.

This first change is the easiest. 😜 

### Main changes in the PR:

1. Create a new message request interface, `MessageRequestWithRequiredProperties`, which derives from the old interface (`MessageRequest`) that marks the previously optional properties as requred. Since this new interface derives from the old, it can be used interchangeably in any place the old interface was used.
2. Update the `createMessageRequest` function in `communication.ts` to return a `MessageRequestWithRequiredProperties` object (instead of a `MessageRequest` object) and pass that around in as many places as possible. This function already populated all the properties, so using the new type recognizes that.
3. Remove violations of `strictNullChecks` (those that are suppressed with `@ts-ignore`) which are no longer violations because they are using the new type which guarantees the properties are defined
4. I also took this opportunity to move the existing `MessageRequest` and `MessageResponse` interfaces, as well as the new interface, into a separate file used just to hold the message objects. This will help ensure that `interfaces.ts` doesn't continue to be an ever-growing file of every interface in the product. These interfaces will now be logically grouped into their own file. (It's possible that in the future these interfaces should live in `communication.ts`, but that file is also huge and complex so I didn't want to exacerbate that problem by putting them in there now).

### Validation performed:

This is a typing-only / refactoring change so I ensured that all existing validation continued to pass.

### Change file added:

Yes

### Next/remaining steps:

`MessageRequest` objects are created in many other ways besides just calling `createMessageRequest`. Future changes will have to work on updating these other ways to also create objects with all required properties. This will be less straightforward as the exact scenarios when these other use cases occur will need to be understood more deeply and will result in control flow changes in the code.

Eventually I expect that they can all be converted to always using the required properties and then `MessageRequestWithRequiredProperties` can go away and `MessageRequest` can remain the single interface used everywhere, now with the properties required.

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable strict-null-checks/all */
-/* eslint-disable strict-null-checks/all */
 
 import { FrameContexts } from '../public/constants';
 import { SdkError } from '../public/interfaces';
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable strict-null-checks/all */
+/* eslint-disable strict-null-checks/all */
 
 import { FrameContexts } from '../public/constants';
 import { SdkError } from '../public/interfaces';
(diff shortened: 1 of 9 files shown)


packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export interface MessageResponse {
+  id: number;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  args?: any[];
+  isPartialResponse?: boolean; // If the message is partial, then there will be more future responses for the given message ID.
+}
```
AE-MS, line 22: Note that these were just moved from `interfaces.ts` and have no changes.

packages/teams-js/src/internal/communication.ts
```diff
@@ -180,8 +181,6 @@ export function sendAndHandleSdkError<T>(actionName: string, ...args: any[]): Pr
 export function sendMessageToParentAsync<T>(actionName: string, args: any[] | undefined = undefined): Promise<T> {
   return new Promise((resolve) => {
     const request = sendMessageToParentHelper(actionName, args);
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
```
AE-MS, line 184: Hurray! Because of this change, this one `@ts-ignore` line can be removed!

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

Issue #2035, opened 2023-11-08
Chat deep link with one recipient throwing error in new Teams


aliblack89:
When creating a chat deep link with one recipient an error is thrown in the new Teams desktop client. If there is more than one recipient it works correctly. I have used both the `chat.openChat({...})` and `chat.openGroupChat({...})` methods, both producing the same error.

Here is an example of the function I am calling to start a chat:

```javascript
// For openChat
async function startChat() {
    await app.initialize();

    if (chat.isSupported()) {
      try {
        const result = await chat.openChat({
          user: "adelev@xyz.onmicrosoft.com",
          message: "Hello",
        });
        console.log(result);
      } catch (error) {
        console.log(error);
      }
    } else {
      console.log("openChat is not supported");
    }
  }
```

```javascript
// For openGroupChat
async function startChat() {
    await app.initialize();

    if (chat.isSupported()) {
      try {
        const result = await chat.openGroupChat({
          users: ["adelev@xyz.onmicrosoft.com"],
          message: "Hello",
        });
        console.log(result);
      } catch (error) {
        console.log(error);
      }
    } else {
      console.log("openGroupChat is not supported");
    }
  }
```

The error displayed in the console is:
<img width="513" alt="Screenshot 2023-11-08 at 10 29 51" src="https://github.com/OfficeDev/microsoft-teams-library-js/assets/592673/f1a93a48-fca9-47a2-8b44-4af07bd57929">

I am using teams-js version 2.17.0 and Teams desktop is version 23285.3804.2469.4861.

Meghana-MSFT:
Based on Teams version it seems you are on the new Teams T2.1. Could you please confirm if you are facing this issue on Teams classic as well. We will also check this at our end and get back to you.

aliblack89:
Yeah, it works on Teams classic and also the web client. It just seems to be the new Teams that I'm getting this issue.

nwojod-MSFT:
Closing issue as duplicate [#1964](https://github.com/OfficeDev/microsoft-teams-library-js/issues/1964)

PR #2050, opened 2023-11-18
Fix calls to chat capability when only a single user is specified

AE-MS:
## Description

The contract for communicating between teams-js and the host library for opening chat requests (`chat.openChat` message) is that the desired users will be passed as an array in the `members` property. The `chat.openChat` function, used when only a single user is intended to be specified, was not doing that. It was passing the single user as a string. Consequently, when the host library (and/or the host) received the message, it could not handle it correctly because it was expecting `members` to be an array of strings and not a single string.

Relatedly, the `chat.openGroupChat` function was similarly broken when a single user was passed to it, since it deferred to `chat.openChat` in that case.

As part of testing, I also discovered these issues:
- When `chat.openGroupChat` defers to `chat.openChat` in the single user case, it was ignore the promise returned from `chat.openChat`, so `chat.openGroupChat` would return immediately rather than waiting for the response from the host library
- The documentation incorrectly stated that UPNs are e-mail addresses. They are frequently e-mail addresses, but not always.
- The test app UX (`ChatAPIs.tsx`) for validating `chat.openChat` was incorrectly saying that `message` was required, when the `OpenSingleChatRequest` type passed to `chat.openChat` clearing indicates that it is optional.
- The same test app file (`ChatAPIs.tsx`) is incorrectly located the "private" directory even though the `chat` capability is public.
- There were no unit tests to cover any of these cases.

### Main changes in the PR:

1. Fix `chat.openChat` to pass the single user as an array with a single element
2. Fix `chat.openGroupChat` to leverage the `Promise` returned by `chat.openChat` when it defers to that function in the single user case
3. Update the documentation to link to the Microsoft Entra ID page describing UPNs
4. Update the test app's chat functionality to not require the `message` property
5. Add unit tests to cover all these fixed cases

Note: The incorrect location of `ChatAPIs.tsx` was not fixed in this change as it would likely have made the diff harder to read (it would have been more difficult to see that the file was both relocated and modified). I will move the file in a later change.

### Validation performed:

1. Validate that all existing unit tests pass
2. Validated that all new unit tests failed before the fixes and passed after the fixes
3. Validate that all E2E tests pass as part of this PR
4. Validate that the exact error message reported in #1964 (and #2035) could be reproed in the test host before the fix and would no longer repro after the fix

### Unit Tests added:

Yes

### End-to-end tests added:

No.

### Change file added:

Indeed.

### DIFF
--- a/packages/teams-js/src/public/chat.ts
+++ b/packages/teams-js/src/public/chat.ts
@@ -25,7 +25,8 @@ interface OpenChatRequest {
  */
 export interface OpenSingleChatRequest extends OpenChatRequest {
   /**
-   * The Microsoft Entra UPN (e-mail address) of the user to chat with
+   * The [Microsoft Entra UPNs](https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/plan-connect-userprincipalname) (usually but not always an e-mail address)
+   * of the user with whom to begin a chat
    */
   user: string;
 }
@@ -39,7 +40,8 @@ export interface OpenSingleChatRequest extends OpenChatRequest {
  */
 export interface OpenGroupChatRequest extends OpenChatRequest {
   /**
-   * Array containing Microsoft Entra UPNs (e-mail addresss) of users to open chat with
+   * Array containing [Microsoft Entra UPNs](https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/plan-connect-userprincipalname) (usually but not always an e-mail address)
+   * of users with whom to begin a chat
    */
   users: string[];
   /**
@@ -79,7 +81,7 @@ export namespace chat {
         );
       } else {
         const sendPromise = sendAndHandleStatusAndReason('chat.openChat', {
-          members: openChatRequest.user,
+          members: [openChatRequest.user],
           message: openChatRequest.message,
         });
         resolve(sendPromise);
@@ -107,7 +109,7 @@ export namespace chat {
           user: openChatRequest.users[0],
           message: openChatRequest.message,
         };
-        openChat(chatRequest);
+        resolve(openChat(chatRequest));
       } else {
         ensureInitialized(runtime, FrameContexts.content, FrameContexts.task);
         if (!isSupported()) {
--- a/packages/teams-js/test/public/chat.spec.ts
+++ b/packages/teams-js/test/public/chat.spec.ts
@@ -9,6 +9,7 @@ import {
   validateChatDeepLinkTopic,
   validateDeepLinkUsers,
 } from '../internal/deepLinkUtilities.spec';
+import { isPromiseStillPending } from '../promiseTester';
 import { Utils } from '../utils';
 
 /* eslint-disable */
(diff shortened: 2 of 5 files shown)


packages/teams-js/src/public/chat.ts
```diff
@@ -79,7 +81,7 @@ export namespace chat {
         );
       } else {
         const sendPromise = sendAndHandleStatusAndReason('chat.openChat', {
-          members: openChatRequest.user,
+          members: [openChatRequest.user],
```
AE-MS, line 84: >members: [openChatRequest.user], [](http://example.com/codeflow?start=10&length=32)

This is the change that ultimately fixes the bug: the `members` property of the `chat.openChat` message is expected to always be an array, even if there's only one user.

packages/teams-js/src/public/chat.ts
```diff
@@ -107,7 +109,7 @@ export namespace chat {
           user: openChatRequest.users[0],
           message: openChatRequest.message,
         };
-        openChat(chatRequest);
+        resolve(openChat(chatRequest));
```
AE-MS, line 112: >resolve [](http://example.com/codeflow?start=8&length=7)

Without this, calling `openGroupChat` with a single user would cause `openGroupChat` to return immediately, not waiting for the chat message to be received and acknowledged by the host.

packages/teams-js/test/public/chat.spec.ts
```diff
@@ -83,16 +84,34 @@ describe('chat', () => {
...
           };
 
+          const normalizedChatRequest = {
+            members: ['someUPN'],
+            message: 'someMessage',
+          };
+
           chat.openChat(chatRequest);
 
-          const chatResponse = {
```
AE-MS, line 88: >chatResponse [](http://example.com/codeflow?start=16&length=12)

It was incorrect for this to be named "response" because it was only checking that the request object that was being sent with the message was correct; it never checked the response (I've added a new test below to validate the response)

packages/teams-js/test/public/chat.spec.ts
```diff
@@ -83,16 +84,34 @@ describe('chat', () => {
             message: 'someMessage',
           };
 
+          const normalizedChatRequest = {
+            members: ['someUPN'],
```
AE-MS, line 88: >members: ['someUPN'], [](http://example.com/codeflow?start=12&length=21)

Note that this test now validates that the request contains the single user in an array

packages/teams-js/test/public/chat.spec.ts
```diff
@@ -184,8 +203,50 @@ describe('chat', () => {
 
           const openChatMessage = utils.findMessageByFunc('chat.openChat');
           expect(openChatMessage).not.toBeNull();
-          expect(openChatMessage.args).toEqual([chatResponse]);
+          expect(openChatMessage.args).toEqual([normalizedChatRequest]);
+        });
+
+        it(`should successfully pass chatRequest to non-legacy Teams host when only one UPN is specified - Context: ${context}`, async () => {
```
AE-MS, line 209: >it [](http://example.com/codeflow?start=8&length=2)

This is a new test to validate the existing behavior where `openGroupChat` delegates to `openChat` for a single user. This test validate that the request has the right shape (a single entry in an array)

shrshindeMSFT (approved):
🚢 LGTM 🚢

KangxuanYe:
@AE-MS I am a little bit worry that your PR will fail on a case in iOS side because it validates the output from chat.openChat API with a string for user, which is not an array. Please take a look on the link I pasted below and we might need to comment out that case to let your fix in.

https://office.visualstudio.com/ISS/_git/metaos-hub-sdk-iOS?path=/orangeios/hubSdkiOSUITests/ChatCapabilityUITests.swift&version=GBmain&line=28&lineEnd=29&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

KangxuanYe (approved):

noahdarveau-MSFT (approved):

AE-MS:
> > @AE-MS I am a little bit worry that your PR will fail on a case in iOS side because it validates the output from chat.openChat API with a string for user, which is not an array. Please take a look on the link I pasted below and we might need to comment out that case to let your fix in.
> > I am monitoring on CI pipeline but please do let me know if you need help. (TBH I will be surprise if it passes iOS E2E testing)
> > path=/orangeios/hubSdkiOSUITests/ChatCapabilityUITests.swift&version=GBmain&line=28&lineEnd=29&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents
> 
> Okay, the type is provided by moduleProvider so we will fix it later. Data should be exposed as an array even if there is one user.

Sounds good; don't forget to file a bug to remember to "fix it later."

PR #2053, opened 2023-11-20
API Telemetry - Second Batch - BarCode, Calendar, Call, Clipboard, Mail, Marketplace and Media capabilities

KangxuanYe:
## Description

Adding api names to compose api version tags for each API in BarCode, Calendar, Call, Clipboard, Mail, Marketplace and Media capabilities and let it be sent out in request to Host SDKS to make sure apiVersionTag could be added into telemetry event and then be passed to Kusto data base.

For more details, please take a look on the description in first batch PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1891

### Main changes in the PR:

1. Adding API names in telemetry files
2. update API call of each corresponding API to be able to send out request with api version tag.

### Related PRs: https://github.com/OfficeDev/microsoft-teams-library-js/pull/1891

PR #2057, opened 2023-11-21
API Telemetry - Third Batch - Chat, liveShareHost(Interactive), Meeting, Menus, Monetization, People capabilities

KangxuanYe:
## Description

Adding api names to compose api version tags for each API in Chat, LiveShareHost(Interactive), Meeting, Menus, Monetization, People capabilities and let it be sent out in request to Host SDKS to make sure apiVersionTag could be added into telemetry event and then be passed to Kusto data base.

### Main changes in the PR:


AE-MS:
Happy to review this once the conflicts have been resolved.

ndangudubiyyam (dismissed):

KangxuanYe:
> Happy to review this once the conflicts have been resolved.

Updated and now it is ready for review!  Thank you @AE-MS

packages/teams-js/src/public/monetization.ts
```diff
@@ -1,10 +1,17 @@
-import { sendAndHandleSdkError } from '../internal/communication';
+import { sendAndHandleSdkErrorWithVersion } from '../internal/communication';
 import { ensureInitialized } from '../internal/internalAPIs';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
 import { callCallbackWithErrorOrResultOrNullFromPromiseAndReturnPromise, InputFunction } from '../internal/utils';
 import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
 import { SdkError } from './interfaces';
 import { runtime } from './runtime';
 
+/**
```
KangxuanYe, line 9: @TrevorJoelHarris , here is the first time I made an update for capability mixing deprecated APIs and new(Beta) APIs in the same file so I just explicitly point it out for you to review.

packages/teams-js/src/public/monetization.ts
```diff
@@ -68,19 +75,29 @@ export namespace monetization {
   ): Promise<void> {
     let callback: ((error: SdkError | null) => void) | undefined;
     let planInfo: PlanInfo | undefined;
+    // eslint-disable-next-line strict-null-checks/all
```
KangxuanYe, line 78: @AE-MS Is this the line we talked yesterday that should be removed? (Or shouldn't be existed in TJS codebase?)

packages/teams-js/src/public/people.ts
```diff
@@ -50,21 +57,29 @@ export namespace people {
 
     let callback: selectPeopleCallbackFunctionType | undefined = undefined;
     let peoplePickerInputs: PeoplePickerInputs | undefined = undefined;
+    // eslint-disable-next-line strict-null-checks/all
```
KangxuanYe, line 60: same here for null check

packages/teams-js/src/public/people.ts
```diff
@@ -1,12 +1,19 @@
...
 import { peoplePickerRequiredVersion } from '../internal/constants';
 import { ensureInitialized, isCurrentSDKVersionAtLeast } from '../internal/internalAPIs';
 import { validatePeoplePickerInput } from '../internal/mediaUtil';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
 import { callCallbackWithErrorOrResultFromPromiseAndReturnPromise } from '../internal/utils';
 import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
 import { ErrorCode, SdkError } from './interfaces';
 import { runtime } from './runtime';
 
+/**
```
KangxuanYe, line 11: Another capability mixing deprecated APIs and new(Beta) APIs together. @TrevorJoelHarris

packages/teams-js/src/public/chat.ts
```diff
@@ -67,20 +73,22 @@ export namespace chat {
...
+    const apiVersionTag = getApiVersionTag(chatTelemetryVersionNumber, ApiName.Chat_OpenChat);
     return new Promise<void>((resolve) => {
       ensureInitialized(runtime, FrameContexts.content, FrameContexts.task);
       if (!isSupported()) {
         throw errorNotSupportedOnPlatform;
       }
       if (runtime.isLegacyTeams) {
         resolve(
-          sendAndHandleStatusAndReason(
+          sendAndHandleStatusAndReasonWithVersion(
```
AE-MS, line 84: >sendAndHandleStatusAndReasonWithVersion [](http://example.com/codeflow?start=10&length=39)

When I was looking at PR #2053, I assumed (incorrectly) that you weren't tagging the `executeDeepLink` versions of these, only the function-based versions of these. I had made this assumption in that PR because I saw that in `calendar.ts`, you didn't apply a telemetry tag to the `composeMeeting` branch that called `executeDeepLink`, only the function one.

In retrospect, it was silly of me to make that assumption because in that same PR you _did_ add a tag to the `executeDeepLink` version of the `startCall` function (as well as elsewhere).

That was a long-winded way of saying -- should the `executeDeepLink` version of `composeMeeting` from the previous PR have gotten tagged?

AE-MS:
>         resolve(openChat(chatRequest));

Do we care to use telemetry to determine how often people call `openGroupChat` but only pass a single "chatter"?

---
Refers to: packages/teams-js/src/public/chat.ts:121 in 9b2ef0e. [](commit_id = 9b2ef0e49dbc1c02c5170e51b91d2d437da9537d, deletion_comment = False)

packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -226,7 +248,13 @@ export class LiveShareHost {
   public setFluidContainerId(containerId: string): Promise<liveShare.IFluidContainerInfo> {
     ensureSupported();
     return new Promise<liveShare.IFluidContainerInfo>((resolve) => {
-      resolve(sendAndHandleSdkError('interactive.setFluidContainerId', containerId));
+      resolve(
+        sendAndHandleSdkErrorWithVersion(
+          getApiVersionTag(interactiveTelemetryVersionNumber, ApiName.Interactive_SetFluidContainerId),
+          'interactive.setFluidContainerId',
```
AE-MS, line 254: This is probably a bigger question than what should be handled in this PR, so it absolutely should not block you from checking in, but should the message being passed and the `ApiName` constant being used just be the same? In most (all?) cases, they're literally the same string. And representing them twice opens the possibility of unintentional divergence.

packages/teams-js/src/public/monetization.ts
AE-MS: If you can remove it and nothing breaks, then totally remove it. If I did it right, these errors should be caught by the tsc compiler now and so disabling the eslint check should not have any effect (and thus it could be removed). Although you don't have to do it as part of this PR, I can just go through a make a single pass across the codebase to remove them.

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/monetization.ts
TrevorJoelHarris: Looks good to me!

packages/teams-js/src/public/people.ts
TrevorJoelHarris: I assume aaron's comment on the other one applies here as well, should be safe to remove

TrevorJoelHarris (dismissed):

packages/teams-js/src/public/chat.ts
AE-MS: Sounds great.

AE-MS:
>         resolve(openChat(chatRequest));

Because as it is now, when someone calls `openGroupChat` with a single chatter, it will only get recorded as a call to `openChat` which won't necessarily help us make decisions about how frequently people are calling `openGroupChat`

---
In reply to: [1831036722](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2057#issuecomment-1831036722) [](http://example.com/codeflow?ancestors=1831036722)

---
Refers to: packages/teams-js/src/public/chat.ts:121 in 9b2ef0e. [](commit_id = 9b2ef0e49dbc1c02c5170e51b91d2d437da9537d, deletion_comment = False)

packages/teams-js/src/public/liveShareHost.ts
KangxuanYe: I have thought of this question and my answer to this is: Ideally, yes. That is to say the value of `funcName` as a parameter in SDK request should be EXACTLY the same as value of ApiVersionTag and developers don't have to write the same string twice in the future.

However, for now, due to some constraints, i.e. tech debt, openLink special scenarios, we cannot make it. But what I will suggest to do after this project is to use ApiName as the value of `funcName` to make sure that API keeps the `funcName` to be exactly the same as `apiVersionTag` so that in the future, we could remove `apiVersionTag` and use `funcName` for telemetry (at that moment the `funcName`s are API names).

Does that make sense to you? @AE-MS
(@TrevorJoelHarris tagged you for my suggestion in the future)

PR #2059, opened 2023-11-22
API Telemetry - Fifth Batch - Profile, Search, SecondaryBrowser, Settings, Sharing, StageView, VideoEffects, VisualMedia capabilites

KangxuanYe:
## Description

Adding api names to compose api version tags for each API in Profile, Search, SecondaryBrowser, Settings, Sharing, StageView, VideoEffects, VisualMedia capabilities and let it be sent out in request to Host SDKS to make sure apiVersionTag could be added into telemetry event and then be passed to Kusto data base.

### Main changes in the PR:


packages/teams-js/src/public/videoEffects.ts
```diff
@@ -1,12 +1,18 @@
...
 import { runtime } from './runtime';
 
+/**
+ * v2 APIs telemetry file: All of APIs in this capability file should send out API version v2 ONLY
+ */
+const videoEffectsTelemetryVersionNumber: ApiVersionNumber = ApiVersionNumber.V_2;
+
 /**
  * Namespace to video extensibility of the SDK
  * @beta
```
KangxuanYe, line 18: @TrevorJoelHarris, I have a question here.

If you take a look on line 23, it seems like sendMessageToParent is passed to VideoPerformanceMonitor Class to initialize an Object, which goes to file (packages/teams-js/src/internal/videoPerformanceMonitor.ts). And under that constructor, it continuously sends out request.

What is request 'video.performance.performanceDataGenerated' used for? Should I treat it as an API to assign an API version tag for it? I think so?

KangxuanYe (commented):
A question is left for this PR.

packages/teams-js/src/public/pages.ts
```diff
@@ -130,6 +130,40 @@ export function setCurrentFrameHelper(apiVersionTag: string, frameInfo: FrameInf
...
+      FrameContexts.sidePanel,
+    );
+    if (!pages.isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+    resolve(sendAndUnwrapWithVersion(apiVersionTag, 'settings.getSettings'));
+  });
+}
+
+export function configSetConfigHelper(apiVersionTag: string, instanceConfig: pages.InstanceConfig): Promise<void> {
```
TrevorJoelHarris, line 157:
```suggestion
export function setConfigHelper(apiVersionTag: string, instanceConfig: pages.InstanceConfig): Promise<void> {
```

This would make the name match with `getConfigHelper` above

packages/teams-js/src/public/videoEffects.ts
TrevorJoelHarris: You'd have to ask the author to be sure, but I believe this is to send data up to the host semi-continuously to make sure they have a good handle on the framerate the video effects are running at.

Whether you want telemetry for this or not is a good question since there won't be a 1:1 relationship between 'developer called a function' and 'this telemetry point was sent up`. Maybe worth tracking down the author, making sure you know what this is for, and then talking with PM about what they would like you to do here?

packages/teams-js/src/public/pages.ts
KangxuanYe: hmmm, I think for here, `getConfig` is an API directly under pages capability. But `setConfig` API is the one under pages.config sub capability. That's why they are different. What do you think of the naming pattern of these?

packages/teams-js/src/internal/videoPerformanceMonitor.ts
```diff
@@ -31,9 +31,11 @@ export class VideoPerformanceMonitor {
 
   private performanceStatistics: VideoPerformanceStatistics;
 
-  public constructor(private reportPerformanceEvent: (actionName: string, args: unknown[]) => void) {
+  public constructor(
+    private reportPerformanceEvent: (apiVersionTag: string, actionName: string, args: unknown[]) => void,
+  ) {
     this.performanceStatistics = new VideoPerformanceStatistics(VideoPerformanceMonitor.distributionBinSize, (result) =>
-      this.reportPerformanceEvent('video.performance.performanceDataGenerated', [result]),
+      this.reportPerformanceEvent('v0_apiVersionTag_Sample', 'video.performance.performanceDataGenerated', [result]),
```
WengHe, line 38: I don't think `videoPerformanceMonitor` needs to know what is apiVersionTag when reporting performance event, instead, you may wrap the function when creating instance of `VideoPerformanceMonitor` like:

```
new VideoPerformanceMonitor(
  (actionName: string, args: unknown[]) => sendMessageToParent('v0_apiVersionTag_Sample', actionName, args)
)

Comment (KangxuanYe) [reply packages/teams-js/src/internal/videoPerformanceMonitor.ts:38]: That's what I expect. But we would like to set the first argument / apiVersionTag to be required so that we could force developers to label apiVersionTag if it is needed for telemetry.

My concern is that, if we use new function (sendMessageToParent) with additional parameter, which is apiVersionTag, to communicate between app and host sdk or app and host, do you think it will trigger any error or exceptions based on your experience?

Comment (WengHe) [reply packages/teams-js/src/internal/videoPerformanceMonitor.ts:38]: > if we use new function (sendMessageToParent) with additional parameter, which is apiVersionTag, to communicate between app and host sdk or app and host, do you think it will trigger any error or exceptions based on your experience?

It really depends on how you implement the new function, especially, are you going to update the hub sdk as well? Our use case is to use this function to send message, and it will be received by hub sdk and forward to our front side code, as long as this process is not changed, I am good with the change.

Comment (KangxuanYe) [reply packages/teams-js/src/internal/videoPerformanceMonitor.ts:38]: Yes, we will also update corresponding change in Hub SDK side and we made demo and validated the idea that the change here will have no harm to Hub SDK / host even if they don't change anything. So I think I have gotten my answer. Thanks again! Appreciated.

File: packages/teams-js/src/internal/telemetry.ts [hunk 4]
```diff
@@ -217,11 +222,37 @@ export enum ApiName {
...
+  VideoEffects_NotifySelectedVideoEffectChanged = 'videoEffects.notifySelectedVideoEffectChanged',
+  VideoEffects_NotifyError = 'videoEffects.notifyError',
+  VideoEffects_NotifyVideoFrameProcessed = 'videoEffects.notifyVideoFrameProcessed',
+  VideoEffects_RegisterEffectParameterChangeHandler = 'videoEffects.registerEffectParameterChangeHandler',
+  VideoEffects_RegisterForVideoEffect = 'videoEffects.registerForVideoEffect',
+  VideoEffects_RegisterForVideoFrame = 'videoEffects.registerForVideoFrame',
+  VideoEffects_RegisterSetFrameProcessTimeLimitHandler = 'videoEffects.setFrameProcessTimeLimitHandler',
+  VideoEffects_RegisterStartVideoExtensibilityVideoStreamHandler = 'videoEffects.startVideoExtensibilityVideoStreamHandler',
+  VideoEffects_RegisterForVideoBufferHandler = 'videoEffects.registerForVideoBufferHandler',
+  VideoPerformanceMonitor_Constructor = 'videoPerformanceMonitor.performanceDataGenerated',
```
KangxuanYe, line 250: @TrevorJoelHarris Based on discussion with @WengHe , videoPerformanceMonitor doesn't have to send API telemetry but since we would like to treat different APIs with no difference, I would like to treat videoPerformanceMonitor as a 'capability' and put some APIs inside of telemetry list. Though apiVersionTag is composed and used for videoPerformanceMonitor's APIs, they actually are useless because they would not be sent to Kusto database.

The core concern is just to make sure we could force each API to be labelled as apiVersionTag so that we don't have two different functions and use either one of them incorrectly.

KangxuanYe (commented):
Update comments.

packages/teams-js/src/public/videoEffects.ts
TrevorJoelHarris: @KangxuanYe followed up with the owner and after that discussion feels good that this is the correct way to telemetrize this function (is telemetrize a word?)

packages/teams-js/src/public/pages.ts
TrevorJoelHarris: since they are helper functions I don't feel strongly about the naming; if you want to leave as-is you can. Go with whatever you think more clearly communicates the correct intent to people reading the code

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

PR #2067, opened 2023-11-29
externalApp capabilities

jekloudaMSFT:

## Description

Adds the 1P internal-only `externalAppAuthentication` and `externalAppCardActions` capabilities for delegating certain actions to the host.

### Main changes in the PR:

1. Add `externalAppAuthentication` capability
2. Add `externalAppCardActions` capability

### Validation performed:

1. Unit tests
2. Capabilities added to the teams test app

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


jekloudaMSFT:

Added two new capabilities. No new version required

packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Limited to Microsoft-internal use
+ */
+export namespace externalAppCardActions {
+  /**
+   * @hidden
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export enum ActionOpenUrlType {
+    DeepLinkOther,
```
jekloudaMSFT, line 19: I think it would be safer to define these as strings here and in the host, since I think by default these compile to integers, right?

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+ */
+export namespace externalAppAuthentication {
+  /*********** BEGIN REQUEST TYPE ************/
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type OriginalRequestInfo = IQueryMessageExtensionRequest | IActionExecuteInvokeRequest;
```
KangxuanYe, line 20: @erinha @jekloudaMSFT @lakhveerk  Wha t do you think of this `OriginalRequestInfo` should follow the same naming pattern, which is to be named as `IOriginalRequestInfo` as `export type IInvokeResponse = IQueryMessageExtensionResponse | IActionExecuteResponse;` because both of `IQueryMessageExtensionRequest` and `IActionExecuteInvokeRequest` starts with `I` as well.

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+  };
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type QueryMessageExtensionCard = {
+    contentType: string;
```
KangxuanYe, line 155: Could you @jekloudaMSFT @erinha double check here because I think there is a discrepancy (on parameters inside of `QueryMessageExtensionCard`) between code and doc. Not sure you two had an in-person sync but I was not there so I didn't catch up the latest update.

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    // Ask the parent window to open an authentication window with the parameters provided by the caller.
+    return sendMessageToParentAsync<[InvokeError, IInvokeResponse]>(
+      'externalAppAuthentication.authenticateAndResendRequest',
+      [
+        appId,
+        originalRequestInfo,
```
KangxuanYe, line 234: does the order matters here? this is appId, originalRequestInfo, and then Authenticate parameters but the doc shows appId, authenticate parameters and then originalRequestInfo at last.

packages/teams-js/src/private/externalAppCardActions.ts
KangxuanYe: I think by default they are integers/numbers.

packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -0,0 +1,169 @@
...
+   */
+  export function processActionOpenUrl(appId: string, url: string): Promise<ActionOpenUrlType> {
+    ensureInitialized(runtime, FrameContexts.content);
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    // TODO: confirm no issues with deserialization
+    return sendMessageToParentAsync<[ActionOpenUrlError, ActionOpenUrlType]>('externalAppCardActions.processOpenUrl', [
```
KangxuanYe, line 144: suggest to replace func value to be sent to Web Hub SDK as `externalAppCardActions.processActionOpenUrl`
cc @lakhveerk

KangxuanYe (changes requested):

KangxuanYe:
Also @jekloudaMSFT I noticed that unit test fails. Please check.

AE-MS:
I'm not as deeply involved as everyone else in this so please feel free to completely ignore my comments if they are inappropriate given all the context that y'all have

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  interface IActionExecuteInvokeRequest {
+    requestType: OriginalRequestType.ActionExecuteInvokeRequest;
+    type: string; // "invoke"
```
AE-MS, line 49: >string [](http://example.com/codeflow?start=10&length=6)

Is type really a free string? There's no enumeration of possible types?

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+    };
+  }
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  interface IActionExecuteInvokeRequest {
```
AE-MS, line 47: >IActionExecuteInvokeRequest [](http://example.com/codeflow?start=12&length=27)

If this interface is accessible externally via the exported `OriginalRequestInfo` type then this interface should also be exported

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export interface IQueryMessageExtensionRequest {
+    requestType: OriginalRequestType.QueryMessageExtensionRequest;
+    commandId: string;
+    parameters?: {
+      name?: string;
+      value?: string;
+    }[];
```
AE-MS, line 34: The current definition allows for all these:
```
interface IFoo {
    parameters?: {
      name?: string;
      value?: string;
    }[];
}

const myFoo1: IFoo = { }
const myFoo2: IFoo = { parameters: undefined }
const myFoo3: IFoo = { parameters: [] }
const myFoo4: IFoo = { parameters: [{}] }
const myFoo5: IFoo = { parameters: [{name: undefined}] }
const myFoo6: IFoo = { parameters: [{name: undefined, value: undefined}] }
```

Do each of those actually have semantically different meanings that we want to allow them? Or are some/all of them representing the same state and we're making more work for ourselves and others to have to infer the same semantic state from multiple possible values?

I'm cheating a little bit because `myFoo1` and `myFoo2` are really just syntactic sugar for the same thing. There's no way to allow `myFoo1` and prevent `myFoo2` but that's okay, they are both checked the same way. This is also true for `myFoo4` which is equivalent and not distinct from `myFoo5` and `myFoo6`. Really this is the list:

```
const myFoo1: IFoo = { }
const myFoo3: IFoo = { parameters: [] }
const myFoo4: IFoo = { parameters: [{}] }
```

Is it really correct that both `name` and `value` could be undefined? What would it mean for a `parameters` element to exist but have no member properties? And how would that semantically be different than just not having a parameters entry at all? What is the semantic difference between having a parameters array with zero elements (which is allowed by this definition) and a parameters array that was `undefined`?

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type ComposeExtensionResponse = {
+    attachmentLayout: AttachmentLayout;
+    type: ComposeResultTypes;
+    attachments: QueryMessageExtensionAttachment[];
+    suggestedActions?: QueryMessageExtensionSuggestedActions;
```
AE-MS, line 122: >suggestedActions [](http://example.com/codeflow?start=4&length=16)

This property has the same challenge I mentioned above for `parameters`. It seems to allow all these and I'm not sure what the semantic distinction is between these different states: undefined `suggestedActions`, defined `suggestedActions` with an undefined `actions` property, defined `suggestedActions` with a defined `actions` property that is an empty array

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+    actions?: Actions[];
+  };
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type Actions = {
```
AE-MS, line 142: >Actions [](http://example.com/codeflow?start=14&length=7)

Should this be `Action` singular? It doesn't seem to be allowing one to define multiple actions in it.

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+    type: string;
+  }
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type ComposeExtensionResponse = {
```
AE-MS, line 118: >type [](http://example.com/codeflow?start=9&length=4)

Why did you choose to use `type` in some places and `interface` in others? Did you intend to leverage the difference between these (that you can't extend `type` elsewhere)?

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type QueryMessageExtensionCard = {
+    contentType: string;
+    content: Record<string, unknown>; //object;
+    name?: string;
+    thumbnailUrl?: string;
```
AE-MS, line 158: >string [](http://example.com/codeflow?start=19&length=6)

Should this be a `URL`? Same question for `contentUrl`.

Note that if you do switch it, it would likely need an intentional function to serialize the value if it gets sent over to the host SDK.

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+import { authentication, FrameContexts } from '../public';
+import { errorNotSupportedOnPlatform } from '../public/constants';
+import { runtime } from '../public/runtime';
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export namespace externalAppAuthentication {
```
AE-MS, line 12: >externalAppAuthentication [](http://example.com/codeflow?start=17&length=25)

There are a lot of properties in this file which would benefit from more verbose comments describing their possible values, meaning, and intended usage.

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+   */
+  export interface IQueryMessageExtensionRequest {
+    requestType: OriginalRequestType.QueryMessageExtensionRequest;
+    commandId: string;
+    parameters?: {
+      name?: string;
+      value?: string;
+    }[];
+    queryOptions?: {
+      count: number;
```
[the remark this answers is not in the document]
jekloudaMSFT: Right

packages/teams-js/src/private/externalAppAuthentication.ts
AE-MS, line 20: >OriginalRequestInfo [](http://example.com/codeflow?start=14&length=19)

Is the intent that this type can ever be passed around and actually used without first converting to one of the discriminated union types? E.g., do we want to enable `OriginalRequestInfo` objects to be passed around and code to reason about the `requestType` property without having to first figure out if they have a `IQueryMessageExtensionRequest`  or `IActionExecuteInvokeRequest`? 

As currently defined, that's not a problem. You can pass around `OriginalRequestInfo` objects and get at the `requestType` property without figuring out the real type because both of the types have that property.

However, if the intent is that ALL the types that participate in this discriminated union (even in the future) have a `requestType` property, then that can be enforced by using a base interface instead of a discriminated union, e.g.,
```
interface OriginalRequestInfo {
    requestType: OriginalRequestType;
}
...
interface IQueryMessageExtensionRequest extends OriginalRequestInfo {
...
}
interface IActionExecuteInvokeRequest extends OriginalRequestInfo {
...
}
```

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export interface IQueryMessageExtensionRequest {
+    requestType: OriginalRequestType.QueryMessageExtensionRequest;
+    commandId: string;
```
AE-MS, line 30: >string [](http://example.com/codeflow?start=15&length=6)

Just pretend I put a comment on practically every string and said "Can this really be a free string? 'klouda' is a valid value for this?"

packages/teams-js/src/private/externalAppAuthentication.ts
AE-MS, line 49: >type [](http://example.com/codeflow?start=4&length=4)

On an object that already has a `requestType` property, another property just called `type` seems a little confusing. Type of... action? `actionType`?

packages/teams-js/src/private/externalAppAuthentication.ts
jekloudaMSFT: I agree, it would be more consistent if we used the `I` prefix here too

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export interface IQueryMessageExtensionResponse {
+    responseType: InvokeResponseType.QueryMessageExtensionResponse;
+    composeExtension?: ComposeExtensionResponse;
```
AE-MS, line 95: What would it mean to get a `IQueryMessageExtensionResponse` object that had an undefined `composeExtension` property? Essentially it would just be an empty object then, right? Does empty object mean "success" ?

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+    preview?: QueryMessageExtensionCard;
+  };
+
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type AttachmentLayout = 'list' | 'grid';
```
AE-MS, line 180: >'list' | 'grid' [](http://example.com/codeflow?start=33&length=15)

Consider alphabetizing

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+  /**
+   * @hidden
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export type ComposeResultTypes = 'result' | 'auth' | 'config' | 'message' | 'silentAuth';
```
AE-MS, line 187: >'result' | 'auth' | 'config' | 'message' | 'silentAuth'; [](http://example.com/codeflow?start=35&length=56)

Consider alphabetizing

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,329 @@
...
+      throw errorNotSupportedOnPlatform;
+    }
+
+    return sendMessageToParentAsync('externalAppAuthentication.authenticateWithSSO', [
+      appId,
+      authTokenRequest.resources,
+      authTokenRequest.claims,
+      authTokenRequest.silent,
+    ]).then(([wasSuccessful, error]: [boolean, InvokeError]) => {
+      // make sure host sdk is throwing the right type of errors
```
AE-MS, line 275: >// make sure host sdk is throwing the right type of errors [](http://example.com/codeflow?start=6&length=58)

Good idea?

packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -0,0 +1,169 @@
...
+export namespace externalAppCardActions {
+  /**
+   * @hidden
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export enum ActionOpenUrlType {
+    DeepLinkOther,
+    DeepLinkStageView,
+    DeepLinkDialog,
```
AE-MS, line 21: Consider alphabetizing

packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -0,0 +1,169 @@
...
+   * Limited to Microsoft-internal use
+   * @param appId ID of the application the request is intended for
+   * @param actionSubmitPayload The Adaptive Card Action.Submit payload
+   * @param cardActionsConfig The card actions configuration. This indicates which subtypes should be handled by this API
+   * @returns Promise that resolves when the request is completed and rejects with ActionSubmitError if the request fails
+   */
+  export function processActionSubmit(
+    appId: string,
+    actionSubmitPayload: IAdaptiveCardActionSubmit,
+    cardActionsConfig?: ICardActionsConfig,
```
AE-MS, line 107: >? [](http://example.com/codeflow?start=21&length=1)

If this parameter isn't specified, what is the default config used?

packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -0,0 +1,169 @@
...
+   * @beta
+   * @hidden
+   * Delegates an Adaptive Card Action.OpenUrl request to the host for the application with the provided app ID
+   * @internal
+   * Limited to Microsoft-internal use
+   * @param appId ID of the application the request is intended for
+   * @param url The URL to open
+   * @returns Promise that resolves to ActionOpenUrlType indicating the type of URL that was opened on success and rejects with ActionOpenUrlError if the request fails
+   */
+  export function processActionOpenUrl(appId: string, url: string): Promise<ActionOpenUrlType> {
```
AE-MS, line 136: >string [](http://example.com/codeflow?start=59&length=6)

Could this be a `URL`

packages/teams-js/src/private/externalAppAuthentication.ts
jekloudaMSFT: It just has to match the order in the host SDK, I'll double check that it does

AE-MS (dismissed):
:clock1:

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: Good question. Doc and host SDK have been updated. Should remove `name`, `thumbnailUrl` and `contentUrl`. FYI @ashishsingh2105 , this will match the host SDK side. Also please note that `content` is of type `Record<string, unknown`. (Doc has recently been updated to match.)

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: No more URLs!

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: I agree that more information is needed here. Due to timelines, I'd like to take this as a follow-up item.

To make progress on this, @ashishsingh2105 and @susanshen96 -- is it possible to get more specific information about the valid combinations of values that can be provided?

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: I agree that more information is needed here. Due to timelines, I'd like to take this as a follow-up item. @jekloudaMSFT can you please make a list (perhaps open a WI tracking input parameter validation) to keep track of follow up items on input parameters?

@ashishsingh2105 and @susanshen96 can you please provide more info on valid/expected values for these properties?

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: https://adaptivecards.io/explorer/Action.Execute.html - can only be "Action.Execute". we can add validation for this.

erinha: This reflects the Adaptive Card schema: https://adaptivecards.io/explorer/Action.Execute.html where the property is called `type`. We can add a comment in the code clarifying that.

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: @ashishsingh2105 and @susanshen96 -- can you please help clarify what valid (and invalid) input looks like here? 

Due to timelines, we can add input validation when we get more information.

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: The doc is wrong, I'll update it.

packages/teams-js/src/private/externalAppAuthentication.ts
jekloudaMSFT: Added validation

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -0,0 +1,336 @@
...
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export interface IActionExecuteInvokeRequest {
+    requestType: OriginalRequestType.ActionExecuteInvokeRequest;
+    type: string; // "invoke"
+    id: string; // "action id"
+    verb: string; // "action"
+    data: Record<string, unknown>; //object; // {}
```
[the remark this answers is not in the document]
jekloudaMSFT: Updated to string | object, but please still weigh in on this

packages/teams-js/src/private/externalAppAuthentication.ts
jekloudaMSFT: @ashishsingh2105 did you decide on most of these distinctions? Are the type/interface objects intentionally different?

packages/teams-js/src/private/externalAppCardActions.ts
jekloudaMSFT: @ashishsingh2105 @susanshen96 do you know what this would be?

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: Actually, the order of parameters, while important, isn't reflected in the doc, it's purely an implementation detail. In any case, they appear to match currently.

packages/teams-js/src/private/externalAppCardActions.ts
erinha: @ashishsingh2105 will these ever not be passed? Should this be a required parameter?

erinha (approved):

KangxuanYe (approved):

KangxuanYe (approved):

erinha (approved):

packages/teams-js/src/private/externalAppCardActions.ts
susanshen96: In the Teams codebase, it looks like the cardActionsConfig can be undefined, which means that there is no default config supplied if the action is not sent with one. @ashishsingh2105 - correct me if I'm wrong.

packages/teams-js/src/private/externalAppAuthentication.ts
susanshen96: `suggestedActions` can be `null`, `undefined`, or defined `suggestedActions` with defined `actions` array.

packages/teams-js/src/private/externalAppAuthentication.ts
susanshen96: Yes - it just means that no results were found or returned.

packages/teams-js/src/private/externalAppAuthentication.ts
jekloudaMSFT: So to confirm: defined `suggestedActions` with undefined `actions` array is not a valid input?

packages/teams-js/src/private/externalAppAuthentication.ts
ashishsingh2105: CommandId is defined by the developer in their app manifest as part of the search command.

packages/teams-js/src/private/externalAppAuthentication.ts
ashishsingh2105: We can keep it to be:
`parameters: {
    name: string;
    value: string;
}`

packages/teams-js/src/private/externalAppAuthentication.ts
susanshen96: It shouldn't be.

packages/teams-js/src/private/externalAppCardActions.ts
susanshen96: Additionally - in the Teams codebase, I'm noticing that we define it as `cardActionConfig` (without the additional s). @ashishsingh2105 - does the naming make any difference?

packages/teams-js/src/private/externalAppCardActions.ts
jekloudaMSFT: @ashishsingh2105 Could this be a URL? Will this parameter always be a URL? (@susanshen96 is this always a URL on Teams?)

packages/teams-js/src/private/externalAppAuthentication.ts
erinha: But still an array? E.g., `parameters: { name: string; value: string; }[];`

packages/teams-js/src/private/externalAppAuthentication.ts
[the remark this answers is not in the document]
erinha: @jekloudaMSFT Updated to `string | Record<string, unknown>`?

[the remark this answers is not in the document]
jekloudaMSFT: Yes, that's right: `string | Record<string, unknown>`?

PR #2076, opened 2023-12-04
Adding in validDomainss json object and publishing it as an Artifact

jadahiya-MSFT:

## Description

### Main changes in the PR:

1.  Copied validOrigins into a JSON object `validDomains.json`
2. Added copy step for Azure pipelines to publish validDomains json artifact to CDN

### DIFF
--- a/apps/teams-test-app/index_cdn.html
+++ b/apps/teams-test-app/index_cdn.html
@@ -15,8 +15,8 @@
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <script
-      src="https://res.cdn.office.net/teams-js/2.17.0/js/MicrosoftTeams.min.js"
-      integrity="sha384-xp55t/129OsN192JZYLP0rGhzjCF9aYtjY0LVtXvolkDrBe4Jchylp56NrUYJ4S2"
+      src="https://res.cdn.office.net/teams-js/2.18.0/js/MicrosoftTeams.min.js"
+      integrity="sha384-g8DoRkiR0ECQ9rwKIgY8GjQ5h0d2jp1347CmU/IRlyUHPjJZiFWEOYc+hFtT9MGL"
       crossorigin="anonymous"
     ></script>
     <div id="root"></div>
--- a/apps/teams-test-app/package.json
+++ b/apps/teams-test-app/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "author": "Microsoft Teams",
   "description": "Teams Test App utilizing Teams JavaScript client SDK to test Hosts",
-  "version": "2.17.0",
+  "version": "2.18.0",
   "scripts": {
     "build": "pnpm build:bundle",
     "build:bundle": "pnpm lint && webpack",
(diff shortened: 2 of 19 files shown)


tools/yaml-templates/build-test-publish.yml
```diff
@@ -98,6 +98,14 @@ steps:
       targetFolder: '$(Build.ArtifactStagingDirectory)\CDNFeed\$(package.version)\js'
     displayName: 'Copy TeamsJS Content for CDN'
 
+  - task: CopyFiles@2
+    inputs:
+      Contents: |
+        packages\teams-js\utils\*.json
```
AE-MS, line 104: >utils [](http://example.com/codeflow?start=26&length=5)

I'm a bit concerned about putting this json file into a "utils" directory, since it may take some people by surprise that files in that directory get published directly to a CDN (let alone all files with a .json extension). Perhaps consider a directory like `packages\teams-js\publishedToCdn`

tools/yaml-templates/build-test-publish.yml
```diff
@@ -98,6 +98,14 @@ steps:
...
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\CDNFeed\validDomains\json'
+      flattenFolders: true
+    displayName: 'Copy validOrigins for CDN'
```
AE-MS, line 107: >'Copy validOrigins for CDN [](http://example.com/codeflow?start=17&length=26)

Technically this step is copying all JSON files, not just validOrigins. As long as you are using the wildcard perhaps this description should reflect that? Or you could not use the wildcard.

jadahiya-MSFT: Sorry you're right, I thought about it while I was making this section and used the * to reflect if this gets renamed. I'm going to update it to just be validOrigins.json.

tools/yaml-templates/build-test-publish.yml
jadahiya-MSFT: I like this idea, I will update this to reflect that

packages/teams-js/utils/validOrigins.json
```diff
@@ -0,0 +1,39 @@
+{
+  "validOrigins": [
```
AE-MS, line 2: >validOrigins [](http://example.com/codeflow?start=3&length=12)

I am concerned that there are now two copies of this list, one in the code and one here. Would it be possible to keep the source of truth in this file and then import it into the JS so that the JS doesn't have to keep a copy of it? Then the fallback in the JS will be identical to what's on the CDN (at least at the time the JS package was built)?

Without that, keeping two lists will introduce the possibility of them getting out of sync.

I understand that you are operating under specific constrained timelines here, so if this suggestion will take a lot of time to implement, consider creating a WI for it to follow-up after you finish the feature. That said, I _think_ you can just easily import the json file like outlined [here](https://stackoverflow.com/a/50540064/7117308)

AE-MS (changes requested):
:clock1:

packages/teams-js/utils/validOrigins.json
jadahiya-MSFT: This is done for a specific reason, I am using this PR to push up this list as an artifact for testing pulling it from a CDN, but I can't do that until this goes through. I am planning to handle that change in a separate PR once we get the results from this test.

TrevorJoelHarris (approved):

AE-MS (approved):

PR #2078, opened 2023-12-06
Added timeout notifications (2 seconds) to video frame processing in videoEffectsEx capability

FloraLuo-MS:
## Description

We need to track the performance of the app’s video frame processing duration and record it in our telemetry dashboard. If the frame processing time exceeds the time-out limit, the SDK should notify the client of this issue.

This pr is to add a timeout to notify the client when the video app’s frame processing duration exceeds the time-out limit.

## Unit Tests added:
Yes

## Change file added:
No


change/@microsoft-teams-js-d5fe13cd-d446-4761-8abe-f733ed6aa57b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "extract timeout logic into a func & add test",
```
AE-MS, line 3: >extract timeout logic into a func & add test [](http://example.com/codeflow?start=14&length=44)

Since these notes appear in the public changelog, can you provide a little more detail about the change so that developers can better understand the change? Also please use past-tense to ensure consistency. Suggestion:

Added timeout notifications (2 seconds) to video frame processing in `videoEffectsEx` capability

packages/teams-js/test/private/videoEffectsEx.spec.ts
```diff
@@ -234,6 +237,42 @@ describe('videoEffectsEx', () => {
         expect(message?.args?.[1]).toEqual(videoEffectsEx.ErrorLevel.Warn);
       });
 
+      it('should send notifyError when frameProcessed event time outs', async () => {
+        await utils.initializeWithContext(FrameContexts.sidePanel);
+        const errorMessage = `Frame not processed in ${videoEffectsEx.frameProcessInterval}ms`;
```
AE-MS, line 242: >Frame not processed in ${videoEffectsEx.frameProcessInterval}ms [](http://example.com/codeflow?start=30&length=63)

Can this error message be exported from videoEffectsEx so it doesn't have to be duplicated here?

AE-MS (changes requested):
:clock1:

packages/teams-js/test/private/videoEffectsEx.spec.ts
FloraLuo-MS: I think it's inappropriate to export an error msg from videoEffectsEx. There is no another file that needs to use this msg.

AE-MS: I'm not sure what you mean -- the test file is another file that needs to use the message, right? Wouldn't that be barrel than duplicating the message and causing future maintenance issues where anyone who updates the string has to remember to update it twice?

WengHe (approved):

AE-MS (approved):

packages/teams-js/src/private/videoEffectsEx.ts
```diff
@@ -22,6 +22,8 @@ import { videoEffects } from '../public/videoEffects';
  * Limited to Microsoft-internal use
  */
 export namespace videoEffectsEx {
+  export const frameProcessingTimeoutInMs = 2000;
```
TrevorJoelHarris, line 25: this shouldn't be exported, no one externally is calling it

TrevorJoelHarris (changes requested):

TrevorJoelHarris (approved):

PR #2080, opened 2023-12-06
Update external app capability interfaces

jekloudaMSFT:

## Description

Minor updates to the input interfaces for the beta external app capabilities

### Main changes in the PR:

1. Updated externalAppCardActions inputs
2. Updated externalAppAuthentcation inputs

### Validation performed:

1. Unit testing

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


apps/teams-test-app/src/components/privateApis/ExternalAppCardActionsAPIs.tsx
```diff
@@ -43,7 +43,7 @@ const ProcessActionSubmit = (): React.ReactElement =>
 const ProcessActionOpenUrl = (): React.ReactElement =>
   ApiWithTextInput<{
     appId: string;
-    url: string;
+    url: URL;
```
KangxuanYe, line 46: So what should we do when we are trying to input url? Can we still use string ('') to input? if not, what should we do?

jekloudaMSFT: Hmm, yeah this should still be string, right? And then just convert to URL?

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -20,6 +20,54 @@ export namespace externalAppAuthentication {
...
+     *
+     * This flag is ignored when the host for the application is a web app (as opposed to a native application) as the behavior is unnecessary in a web-only
+     * environment without an embedded browser.
+     */
+    isExternal: boolean;
+  };
+
+  /**
+   * @hidden
+   * Parameters for SSO authentication
```
erinha, line 56: Nit: we should probably clarify that these interfaces are used exclusively with the externalApp authentication APIs.

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -243,14 +291,14 @@ export namespace externalAppAuthentication {
...
    * @internal
    * Limited to Microsoft-internal use
-   * @param appId ID of the application backend to which the request and authentication response should be sent
+   * @param appId ID of the application backend to which the request and authentication response should be sent. This must be a UUID
    * @param authenticateParameters Parameters for the authentication pop-up
    * @param originalRequestInfo Information about the original request that should be resent
    * @returns A promise that resolves to the IInvokeResponse from the application backend and rejects with InvokeError if the host encounters an error while authenticating or resending the request
    */
   export function authenticateAndResendRequest(
     appId: string,
```
erinha, line 300: Should this be a UUID then?

erinha: Same applies to all places we use `appId`

apps/teams-test-app/src/components/privateApis/ExternalAppCardActionsAPIs.tsx
KangxuanYe: That's what I think so. Otherwise the input will not be valid, I guess.

packages/teams-js/src/private/externalAppAuthentication.ts
jekloudaMSFT: Discussed offline. There doesn't seem to be a good type we can pull from the `uuid` package that we can use as an input type here, so I think it would be better to just leave the input type as string add UUID validation (uuid.validate) here. Small caveat: it looks like the app manifest allows capitalization in the ID while the UUID validation does not, but we can just cast toLower before we validate.

KangxuanYe (approved):

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -120,7 +168,7 @@ export namespace externalAppAuthentication {
 
   /**
```
erinha, line 169: On line 152 (in the new file line numbers), we have `composeExtension?: ComposeExtensionResponse` as an optional parameter. Did we get confirmation that it is indeed optional or can we make this required?

erinha (dismissed):

KangxuanYe (approved):

erinha (dismissed):

erinha (approved):

jadahiya-MSFT (approved):

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -267,14 +315,14 @@ export namespace externalAppAuthentication {
       [
         appId,
         originalRequestInfo,
-        authenticateParameters.url,
+        authenticateParameters.url.href,
```
KangxuanYe, line 318: What's the purpose of being href? Is it to make it unchangeable?

jekloudaMSFT: It's the equivalent of `toString()`. we don't want to modify it, so we can use either (but yeah if we wanted to modify the value we would have to use `href`

PR #2081, opened 2023-12-06
Move the target TypeScript platform to ES2015 (aka ES6)

AE-MS:
## Description

This change moves the target platform for all TypeScript projects to ES2015 (aka ES6). Importantly, this moves the main teams-js package to ES6 and causes the TypeScript compiler to stop emitting various glue/polyfill code into the package to support older versions (ES5). This reduces the minified teams-js package size by 15k (about 11%).

### Validation performed:

Ensured all automated validation continued to work.


apps/typed-dependency-tester/tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
   "extends": "../../tsconfig.common.json",
   "compilerOptions": {
-    "target": "es5",
```
AE-MS, line 4: This project will now inherit the target from the common file

packages/teams-js/tsconfig.json
```diff
@@ -3,8 +3,7 @@
   "compilerOptions": {
     "outDir": "./dist",
     "declarationDir": "./dts",
-    "target": "es5",
```
AE-MS, line 6: This will now inherit the target from the common config file

packages/teams-js/tsconfig.json
```diff
@@ -3,8 +3,7 @@
...
-    "lib": ["ES5", "DOM", "ES2015.promise"],
+    "lib": ["DOM", "ES2015", "ES5"],
```
AE-MS, line 6: Keeping ES5 and DOM, changing ES2015 to be all of ES2015 and not just the promise portion.

packages/teams-js/src/internal/communication.ts
```diff
@@ -676,7 +676,11 @@ function flushMessageQueue(targetWindow: Window | any): void {
  * Limited to Microsoft-internal use
  */
 export function waitForMessageQueue(targetWindow: Window, callback: () => void): void {
-  const messageQueueMonitor = Communication.currentWindow.setInterval(() => {
+  let messageQueueMonitor: ReturnType<typeof setInterval>;
+  /* const cannot be used to declare messageQueueMonitor here because of the JS temporal dead zone. In order for messageQueueMonitor to be referenced inside setInterval,
+     it has to be defined before the setInterval call. */
```
AE-MS, line 681: This is a behavior difference introduced in ES6.

In ES5, the `var` keyword is used for variable declaration, which is hoisted to the top of its scope. This means even if a variable is declared after it's used, the JavaScript engine moves the declaration to the top, so it doesn't throw an error.

However, in ES6, the `let` and `const` keywords were introduced for variable declaration. These are block-scoped and are not hoisted like `var`. More importantly, they have a temporal dead zone (TDZ), a period between entering scope and being declared where they cannot be accessed. This was causing the "Cannot access 'messageQueueMonitor' before initialization" errors I was seeing in the unit tests.

When we were compiling to ES5, TypeScript was transpiling `const` and `let` to `var`, which doesn't have a TDZ, so the error didn't occur. But when I switched to ES6, `const` and `let` are kept as is, and thus the TDZ is in effect, causing the error.

jest.config.common.js
```diff
@@ -1,9 +1,6 @@
 module.exports = {
   transform: {
-    '.(ts|tsx)': [
-      'ts-jest',
-      { tsconfig: { downlevelIteration: true, esModuleInterop: true, strictNullChecks: false } },
```
AE-MS, line 5: >downlevelIteration [](http://example.com/codeflow?start=20&length=18)

We don't need the `downlevelIteration` flag anymore because we are now running against ES6 which has the iteration behavior we were previously pushing downlevel into ES5 with this flag

PR #2087, opened 2023-12-11
Added size in FilePreviewParameters Interface

reenudeswal:

## Description

> Requirement is captured in https://microsoft-my.sharepoint-df.com/:w:/p/mashukl/EcmrfTaTdidJiX4EPqlaq7AB4JpXFKDdvlp8p1Ow8ov15Q?e=xC4JhN&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2creenulnu%40microsoft.com&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiIyOC8yMzExMjcwNTQwMCIsIkhhc0ZlZGVyYXRlZFVzZXIiOmZhbHNlfQ%3D%3D

### Main changes in the PR:

1. Added an optional size property in FilePreviewParameters interface. This can be passed by the calling app in openFilePreview API along with other metaData.
2. Updated unit tests.

### Unit Tests added:

Yes


change/@microsoft-teams-js-2982475b-f77f-44fe-8106-e56b04b27ddc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "added size property in openFilePreview API input params",
+  "packageName": "@microsoft/teams-js",
+  "email": "email not defined",
```
maglims, line 5: This seems odd. Is this changelist auto generated? If so, i thought it should set this email.

packages/teams-js/src/private/interfaces.ts
```diff
@@ -132,6 +132,15 @@ export interface FilePreviewParameters {
...
   type: string;
 
+  /**
+   * @hidden
+   * The size of the file in bytes.
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  size?: string;
```
maglims, line 142: @TrevorJoelHarris , is there any general guidance where we should put the parameter for new added one? The parameter list is not alphabet order so it doesn't look there is any pattern we can follow... maybe it doesn't matter.. The reason I asked is because sometimes I wondered the same when I make the similar changes.

packages/teams-js/test/private/privateAPIs.spec.ts
```diff
@@ -581,6 +581,7 @@ describe('AppSDK-privateAPIs', () => {
       viewerAction: ViewerActionTypes.view,
       fileOpenPreference: FileOpenPreference.Web,
       conversationId: 'someConversationId',
+      size: 'someSize',
```
maglims, line 584: since it is optional, not sure if we want to have two test cases, one is with this param(as being done here) and the other one is without this parameter (as it was before)

packages/teams-js/src/private/interfaces.ts
AE-MS, line 142: >size [](http://example.com/codeflow?start=2&length=4)

Can you name this so that the unit is in the name, e.g. `sizeOfFileInBytes`

AE-MS, line 142: >string [](http://example.com/codeflow?start=9&length=6)

Wouldn't a count of bytes be more faithfully represented in a `number` type?

packages/teams-js/test/private/privateAPIs.spec.ts
AE-MS: It's a good idea to be comprehensive but in this specific case I'm not sure that adding a test that _doesn't_ have the property is adding any new value, as it seems like practically all the properties on `FilePreviewParameters` are optional, so it seems like testing that ANY optional parameter is handled correctly would be the same as testing any other optional parameter

That said it seems like there aren't any tests that validate the right thing happens for any optional parameter, so maybe this is the time to add a test for that (although I would probably try it with one of the optional parameters in the middle of the list rather than at the end like `size`)

AE-MS (changes requested):
:clock1:

packages/teams-js/src/private/interfaces.ts
reenudeswal: Yes even I was thinking to have the unit in name. I can change it to `sizeInBytes` because I think file is implicit here as with other params.

change/@microsoft-teams-js-2982475b-f77f-44fe-8106-e56b04b27ddc.json
reenudeswal: yes it was auto generated.

packages/teams-js/test/private/privateAPIs.spec.ts
reenudeswal: Added size in the middle of the list. @AE-MS is this ok or do you want me to add more tests ?

AE-MS: This is okay for this change!

packages/teams-js/src/private/interfaces.ts
reenudeswal: With string we can handle any value without worrying for limit and null values can be checked.

reenudeswal:
@AE-MS can you approve if changes looks ok.

packages/teams-js/src/private/interfaces.ts
AE-MS: Can you say more about which limits are concerning? The maximum value that `number` can hold is 9,007,199,254,740,991, and since that is going to be in bytes in this context, it means that your interface would be limited to files that are no larger than roughly 9000 terabytes. I understand the value of future-proofing, but it doesn't seem like that developers will be handling files that large in their Teams apps anytime soon? If you were concerned with being able to handle larger files, you could instead use [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) which has no limit and still faithfully represents that this a number and not a string.

Can you elaborate on the scenarios for null-value checking that work with `string` but don't work with `number`? If this property was defined as `sizeInBytes?: number` that would allow all the same checks that I expect would commonly be used. Are you specifically concerned with checking for `null` or did you mean `undefined`?

AE-MS:
> @AE-MS can you approve if changes looks ok.

Just one open comment still. Also it looks like several files.ts tests are failing in the build pipelines with your change.

packages/teams-js/src/private/interfaces.ts
reenudeswal: I was specifically concerned for null and also where this property will get consumed is specified at string, so I used the same type here. But you are right makes more sense as a number, the consumer can convert as per there need. I will modify this to a number.

reenudeswal: @TrevorJoelHarris can you please update on this.

packages/teams-js/src/private/privateAPIs.ts
```diff
@@ -125,6 +125,7 @@ export function openFilePreview(filePreviewParameters: FilePreviewParameters): v
     filePreviewParameters.viewerAction,
     filePreviewParameters.fileOpenPreference,
     filePreviewParameters.conversationId,
+    filePreviewParameters.sizeInBytes,
```
reenudeswal, line 128: @maglims @AE-MS  need to add this is the array end in order prevent array indexing. Because Teams client is reading file title, url etc based on array index and if we add this in the middle it will change all indexes.

AE-MS (approved):

packages/teams-js/src/private/interfaces.ts
TrevorJoelHarris: It doesn't matter very much, the order doesn't affect external developers in a meaningful way. It would be nice if they were alphabetized but I don't think @reenudeswal needs to do that in this PR.

TrevorJoelHarris (approved):

maglims (approved):

PR #2089, opened 2023-12-11
Acquiring list of validDomains from CDN

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Added in function `getDomainsFromCDN()` to utils and calling it upon initialization of the SDK to acquire the list of domains from the CDN.
2. Added in fallback logic to pull in list of validDomains from `validDomains.json` file as a fallback.

### Validation performed:

1. Tested acquiring file from CDN E2E prior to caching and refreshing cache
2. Tested acquiring from CDN after cached by browser
3. Tested acquiring from incorrect CDN endpoint to ensure that fallback logic worked as expected

### Unit Tests added:

<Yes>

Added in unit tests to mock fetch calls. Added in unit tests to acquire list of domains from JSON file


packages/teams-js/src/internal/utils.ts
```diff
@@ -2,13 +2,27 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import * as uuid from 'uuid';
 
+import * as validOrigins from '../artifactsForCDN/validDomains.json';
 import { GlobalVars } from '../internal/globalVars';
 import { minAdaptiveCardVersion } from '../public/constants';
 import { AdaptiveCardVersion, SdkError } from '../public/interfaces';
 import { pages } from '../public/pages';
-import { validOrigins } from './constants';
+//import { validOrigins } from './constants';
```
jadahiya-MSFT, line 10: Commented this out for now for testing,

packages/teams-js/src/internal/utils.ts
```diff
@@ -37,6 +37,23 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
   return false;
 }
 
+export function getDomainsFromCDN(): void {
+  fetch('https://res-sdf.cdn.office.net/teams-js/validDomains/json/validDomains.json')
```
jadahiya-MSFT, line 41: Planning to switch this out for a non-sdf endpoint.

packages/teams-js/src/internal/utils.ts
```diff
@@ -37,6 +37,22 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
   return false;
 }
 
+export function getDomainsFromCDN(): void {
```
TrevorJoelHarris, line 40: nit: would be good to rename this so it doesn't sound like it returns some values. Open to naming suggestions if you have them, maybe something like `retrieveDomainsFromCDNAndStore`

packages/teams-js/src/internal/utils.ts
```diff
@@ -37,6 +37,22 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
...
+  fetch(validDomainsCdnEndpoint)
```
TrevorJoelHarris, line 41: does fetch cause any problems in SSR scenarios where this is no window object? Do we need to surround this with some sort of SSR check?

packages/teams-js/src/internal/utils.ts
```diff
@@ -54,8 +70,9 @@ export function validateOrigin(messageOrigin: URL): boolean {
     return false;
   }
   const messageOriginHost = messageOrigin.host;
+  const validDomains = GlobalVars.validOrigins.length !== 0 ? GlobalVars.validOrigins : validOriginsFallback;
```
TrevorJoelHarris, line 73: Are there scenarios where apps will be using the fallback list for a few seconds before the 'getDomainsFromCDN' call completes? If so, is that okay? Would it be better to await the getDomainsFromCDN call so that we know for sure that the correct domains have been retrieved?

packages/teams-js/src/internal/utils.ts
```diff
@@ -37,6 +37,22 @@ function validateHostAgainstPattern(pattern: string, host: string): boolean {
   return false;
 }
 
+export function getDomainsFromCDN(): void {
+  fetch(validDomainsCdnEndpoint)
+    .then(async (response) => {
+      if (!response.ok) {
+        throw response.statusText;
```
TrevorJoelHarris, line 44: Does this error bubble up to the app and cause problems? I had assumed (maybe incorrectly) that if the CDN returns an error of some sort that we would use the fallback list and log that there had been an issue retrieving the CDN list.

packages/teams-js/tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "noImplicitReturns": true,
     "sourceMap": true,
     "declaration": true,
-    "strictNullChecks": true
+    "strictNullChecks": true,
+    "resolveJsonModule": true
```
TrevorJoelHarris, line 16: why was `resolveJsonModule` needed?

jadahiya-MSFT: `resolveJsonModule` is needed because we are importing `validDomains.json` into `constants` as a single source of truth for our domains. This way whenever we need to make changes or tests we only need to update the one file.

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: Tested, this doesn't cause issues for code execution, but you did catch something important which is to set `GlobalVariables.validOrigins` to `validOriginsFallback` in the case of the `fetch` call failing. I'm making the updates now. Great catch!

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: I initially had it like that, but found that execution was able to be completed faster by handling it like this.  An inspection into why was while I was debugging, the fetch call would complete and finish translating the json into a validDomains object, but the Promise for `getValidDomainsFromCDN` would take longer to execute and notify the calling function that the domains have been acquired by a range of sometimes 50ms to 150ms longer (This is just local code execution). Translating this to using the GlobalVariable allowed us to save that execution time. This also enabled us to move the `getValidDomainsFromCDN` to earlier in our pipeline as well.

My assumption was that on really slow WiFi we would run into the issue you described, but even there where acquiring the CDN endpoint takes about "1.2 seconds" ("Simulated Slow 3G") it was still acquiring and waiting for the `getDomainsFromCDN` call to finish first. 

I do like your suggestion and I believe it may result in fewer errors, so I will experiment with using `await` again.

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: Good catch! I had this wrapped in a `if (fetch)` call previously which would've defaulted back to the fallback in SSR scenarios, I will test this new code to ensure SSR isn't broken.

packages/teams-js/src/internal/constants.ts
```diff
@@ -110,48 +112,21 @@ export const scanBarCodeAPIMobileSupportVersion = '1.9.0';
...
+export const validOriginsFallback = validOriginsJSON.validOrigins;
+
+/**
+ * @hidden
+ * CDN endpoint of the list of valid origins
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export const validDomainsCdnEndpoint = 'https://res-sdf.cdn.office.net/teams-js/validDomains/json/validDomains.json';
```
AE-MS, line 129: >validDomainsCdnEndpoint [](http://example.com/codeflow?start=13&length=23)

Can this be typed as a `URL` object instead of a string?

packages/teams-js/src/internal/utils.ts
AE-MS, line 73: >const validDomains = GlobalVars.validOrigins.length !== 0 ? GlobalVars.validOrigins : validOriginsFallback; [](http://example.com/codeflow?start=2&length=107)

Not sure if it belongs here or elsewhere (given some other feedback about the design) but it is important to write log statements to make it clear to any troubleshooters exactly where the valid origins its using came from (CDN URL or fallback)

packages/teams-js/src/internal/constants.ts
jadahiya-MSFT: Yup, I'll make that change now

packages/teams-js/src/internal/utils.ts
AE-MS: I have some feedback along these lines so I'm going to leave it in this discussion thread:

If possible, we should avoid adding a new global (makes it harder to test, leaks abstraction details to the world, potentially makes tree shaking and other optimizations more difficult in the future).

Ideally the fact that the origins are sometimes coming from a CDN and sometimes a fallback should be completely be abstract from the rest of the code. The current design seems like it mostly does this, but it's a little hard to tell because the exact componentization boundary lines are blurry since these functions live in the middle of the `utils.ts` class which has a lot of unrelated stuff in it. Is "utils" the component? That doesn't seem quite right.

I think _ideally_ the design would be something like a function that looks like this:

`getValidDomainsList(): URL[]`

...then internally, that function can asynchronously query the CDN or get it from a fallback or switch it out depending on network latency/connectivity/etc. It can also "internally" cache list it retrieves from the network. It can log when it retrieved it, when it switched from the fallback, etc. The point is that no one outside of that function has any knowledge of any of these details, all they know is "call this function to get the list."

With the current design:
- The `validateOrigin` function has to know details like there are two possible places to check for the list
- Any code in the entire app is allowed to go poking at the global list directly and know how to interpret it, what it means it if it's empty, etc.
etc.

My suggestion would be to consider how best encapsulate all the implementation details behind a single function or class or file or something to prevent inevitable misuse by future developers. Ideally in a completely separate file dedicated only to validOrigin functionality --> but I also recognize that moving code to other files can cause widespread refactoring pain so such potential changes don't have to be in the same change.

The one bit of implementation detail that I could justifiably see "leaking" outside of that component is the function you want to call early to get the CDN fetching process started, even if the list isn't needed at the time that's kicked off. You could imagine creating a `prefetchValidOrigins` function potentially or just make the early code call your single "get the list function" and then just ignore the result (since it was only calling it to start the network requests)

Hope this was helpful and not confusing. 😅

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: Adding in a separate parameter to validateOriginLogger that includes which list it is being compared to.

packages/teams-js/test/utils.ts
```diff
@@ -99,6 +99,7 @@ export class Utils {
       },
       closed: false,
     };
+    global.fetch = jest.fn(() => Promise.resolve({ status: 200, ok: true } as Response));
   }
```
AE-MS, line 103: >} [](http://example.com/codeflow?start=2&length=1)

Are there new tests to validate the fallback behavior, the list retrieval behavior, etc.?

jadahiya-MSFT: Yup, added unit tests for using the fallback when there is no supplied json object from the fetch call and if the fetch call itself fails.

packages/teams-js/src/internal/constants.ts
```diff
@@ -108,50 +110,26 @@ export const getMediaCallbackSupportVersion = '2.0.0';
  */
 export const scanBarCodeAPIMobileSupportVersion = '1.9.0';
 
+export const validDomainsFallback = validDomains;
```
jadahiya-MSFT, line 113: This is needed since importing validDomains directly from the json and using its value `validDomains.validOrigins` is soon to be deprecated. Might rename this variable `validDomainsJSON` though

packages/teams-js/src/internal/utils.ts
jadahiya-MSFT: @AE-MS I think I understood what you were trying to say and made modifications to the code to reflect it. Here's hoping I didn't get confused somewhere along the way.

packages/teams-js/src/internal/validDomains.ts
```diff
@@ -0,0 +1,104 @@
+import { GlobalVars } from '../internal/globalVars';
+import { validDomainsCdnEndpoint, validOriginsFallback } from './constants';
+import { getLogger } from './telemetry';
+import { isValidHttpsURL } from './utils';
+
+let validOrigins: string[] = [];
+export async function prefetchDomainsFromCDN(): Promise<void> {
+  if (fetch) {
```
AE-MS, line 8: >  if (fetch) [](http://example.com/codeflow?start=0&length=12)

Can you say more why you have to check for the presence of the `fetch` function? Surely any modern browser we support has implemented this.

packages/teams-js/src/internal/validDomains.ts
```diff
@@ -0,0 +1,104 @@
+import { GlobalVars } from '../internal/globalVars';
+import { validDomainsCdnEndpoint, validOriginsFallback } from './constants';
+import { getLogger } from './telemetry';
+import { isValidHttpsURL } from './utils';
+
+let validOrigins: string[] = [];
+export async function prefetchDomainsFromCDN(): Promise<void> {
```
AE-MS, line 7: >prefetchDomainsFromCDN [](http://example.com/codeflow?start=22&length=22)

Rather than writing a unique fetch call here, could this function just call retrieveDomainsFromCDNAndStore and then ignore the return value? Is only fetching the value from the CDN actually helpful? Or would the best prefetching behavior to be to fetch and then store it?

packages/teams-js/src/internal/validDomains.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (fetch) {
+    await fetch(validDomainsCdnEndpoint);
+  }
+}
+
+async function retrieveDomainsFromCDNAndStore(): Promise<string[]> {
```
AE-MS, line 13: >retrieveDomainsFromCDNAndStore [](http://example.com/codeflow?start=15&length=30)

This name is descriptive, which is GREAT. I think one thing that could be clearer is that it's not just retrieving/storing the domains, it's also returning them.

Perhaps something like `getValidDomains` might be better since, to the caller, all they are looking to do is get the valid domains. The fact that sometimes they are cached and sometimes it goes to the CDN etc. seems like a implementation detail?

I know, I know, it feels like I sometimes say "descriptive names are great" and here I am suggesting that maybe this is too descriptive. 😅

packages/teams-js/src/internal/validDomains.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+  if (fetch) {
+    return fetch(validDomainsCdnEndpoint)
+      .then((response) => {
+        if (!response.ok) {
+          validOrigins = validOriginsFallback;
+          return validOriginsFallback;
+        }
+        return response.json().then((validDomains) => {
+          validOrigins = validDomains.validOrigins;
```
AE-MS, line 25: >validDomains.validOrigins [](http://example.com/codeflow?start=25&length=25)

I have to say that the "sometimes use domains" and "sometimes use origins" naming has tripped me up a few times.

AE-MS, line 25: Should the data retrieved be validated? To make sure it looks like what we expect? At least that it's an array that can be walked through and maybe that each element could validly be converted to a `URL` type (ideally they would be stored in the array as a `URL` type but maybe that's too hard for compatibility reasons)

packages/teams-js/src/internal/validDomains.ts
```diff
@@ -0,0 +1,104 @@
...
+          return validOrigins;
+        });
+      })
+      .catch(() => {
+        validOrigins = validOriginsFallback;
+        return validOrigins;
```
AE-MS, line 31: This is notable but maybe not interesting. In this failure case you do:
```
        validOrigins = validOriginsFallback;
        return validOrigins;
```
but in the failure case above you do
```
          validOrigins = validOriginsFallback;
          return validOriginsFallback;
```

They seem equivalent but it does make me wonder if it would be possible to unify the failure cases so you aren't doing the same thing twice. E.g., maybe if `!response.ok` you `throw Error("Bad response")` which would then be caught by the same catch block and unify the failure handling?

packages/teams-js/src/internal/validDomains.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+async function retrieveDomainsFromCDNAndStore(): Promise<string[]> {
+  if (validOrigins.length !== 0) {
+    return validOrigins;
+  }
+  if (fetch) {
+    return fetch(validDomainsCdnEndpoint)
+      .then((response) => {
+        if (!response.ok) {
```
AE-MS, line 20: >if (!response.ok) { [](http://example.com/codeflow?start=8&length=19)

Can you add logging for this failure case?

packages/teams-js/src/internal/validDomains.ts
AE-MS, line 31: Can you add logging for this failure case?

packages/teams-js/test/internal/validDomains.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+      const result = await validateOrigin(messageOrigin);
+      expect(result).toBe(true);
+    });
+    it('validateOrigin returns true if origin for subdomains in teams pre-known allowlist', async () => {
+      const messageOrigin = new URL('https://subdomain.teams.microsoft.com');
+      const result = await validateOrigin(messageOrigin);
+      expect(result).toBe(true);
+    });
+    it('validateOrigin returns false if origin is not in teams pre-known allowlist', async () => {
+      const messageOrigin = new URL('https://badorigin.com');
```
AE-MS, line 38: >https://badorigin.com [](http://example.com/codeflow?start=37&length=21)

I know you are just copying this test over from the old location but in a future change consider using different domains for these tests. We shouldn't use domains in our code that we don't control. You could using something like badorigin.example.com instead.

packages/teams-js/src/internal/validDomains.ts
jadahiya-MSFT: This is for SSR/windowless scenarios

packages/teams-js/src/internal/validDomains.ts
TrevorJoelHarris: Just want to call out that this name is this way because I asked Jay to change it in a previous iteration. I'm fine with Aaron's name (especially since this now returns those domains and didn't previously).

AE-MS: Yes, @TrevorJoelHarris and I aren't fighting. 😅 I think the implementation of what the function did changed over time so the potential naming suggestions also changed.

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/constants.ts
```diff
@@ -108,50 +110,26 @@ export const getMediaCallbackSupportVersion = '2.0.0';
  */
 export const scanBarCodeAPIMobileSupportVersion = '1.9.0';
 
+export const validDomains = validDomainsJSON;
```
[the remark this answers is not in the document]
jadahiya-MSFT: Good catch, this shouldn't be exported, but I will mark it as `@hidden` and `@internal`

packages/teams-js/src/internal/validDomains.ts
jadahiya-MSFT: rewrote it to just use some variation of validOrigins to clarify things, hopefully that helps

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -0,0 +1,125 @@
...
+    return validOriginsFallback;
+  }
+}
+
+function validateValidOriginsFromCDN(validOriginsJSON: string): boolean {
+  const validOriginsCDN = JSON.parse(validOriginsJSON);
+  if (!validOriginsCDN.validOrigins) {
+    return false;
+  }
+  for (let i = 0; i < validOriginsCDN.validOrigins.length; i++) {
```
AE-MS, line 52: Nothing functionally wrong with how you've written this; for future reference I believe the more canonical JS/TS way would be to use a `for..of` loop, e.g.:
```
for (const potentiallyValidOrigin of validOriginsCDN.validOrigins) {
    try {
      new URL('https://' + potentiallyValidOrigin);
    } catch (_) {
      return false;
    }
}
```

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -0,0 +1,125 @@
...
+        validOriginsCache = validOriginsFallback;
+        return validOriginsCache;
+      });
+  } else {
+    validOriginsCache = validOriginsFallback;
+    return validOriginsFallback;
+  }
+}
+
+function validateValidOriginsFromCDN(validOriginsJSON: string): boolean {
```
AE-MS, line 47: >boolean [](http://example.com/codeflow?start=64&length=7)

It looks like you want to indicate errors by returning `false` rather than throwing.

As it is written, `JSON.parse` will throw if the string isn't valid JSON (and that's not in your narrowly-scoped try/catch block). Consider expanding the scope of your try catch to be the entire function in this case -- or add a second try/catch around JSON.parse.

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -0,0 +1,125 @@
...
+  const validOriginsCDN = JSON.parse(validOriginsJSON);
+  if (!validOriginsCDN.validOrigins) {
+    return false;
+  }
+  for (let i = 0; i < validOriginsCDN.validOrigins.length; i++) {
+    try {
+      new URL('https://' + validOriginsCDN.validOrigins[i]);
```
AE-MS, line 54: Is there value in adding logging to differentiate between what the errors were? Or at least logging the string it is trying to validate so that potential troubleshooters can see why it failed?

packages/teams-js/src/internal/validOrigins.ts
AE-MS, line 47: >validateValidOriginsFromCDN [](http://example.com/codeflow?start=9&length=27)

Since this is a boolean function, consider naming it `isValidOriginsJsonValid()`

packages/teams-js/src/internal/validOrigins.ts
jadahiya-MSFT: I blanked out on this nomenclature, thanks I wanted to use this one originally!

AE-MS (dismissed):
:shipit:

AE-MS (approved):

TrevorJoelHarris (dismissed):

AE-MS (approved):

TrevorJoelHarris (approved):

PR #2097, opened 2023-12-14
Add optional parameter to initialize call

lakhveerk:

## Description

Adding optional `customMessageIdStarter` parameter to initialize call. If there are more than one teams-js instants, and there can be a problem that both instants are making call to host with the same message IDs. customMessageIdStarter parameter gives app developer an option start the messageIDs from a different number, so they don't conflict. 

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
+++ b/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds",
+  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds. \\nHaving a high seed value for this field results in generating messageIds that have a less change of collision when there are multiple instances of teamsJS running by the application",
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
+++ b/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set teh starting number fr messageIds",
+  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds",
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
+++ b/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set teh starting number fr messageIds",
+  "packageName": "@microsoft/teams-js",
+  "email": "lakhveerkaur@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
+++ b/change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds. \\nHaving a high seed value for this field results in generating messageIds that have a less change of collision when there are multiple instances of teamsJS running by the application",
+  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds. \\nHaving a high seed value for this field results in generating messageIds that have a less chances of collision when there are multiple instances of teamsJS running by the application",
   "packageName": "@microsoft/teams-js",
   "email": "lakhveerkaur@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -63,7 +63,9 @@ interface InitializeResponse {
 export function initializeCommunication(
   validMessageOrigins: string[] | undefined,
   apiVersionTag: string,
+  customMessageIdStarter?: number,
 ): Promise<InitializeResponse> {
+  CommunicationPrivate.nextMessageId = customMessageIdStarter ?? CommunicationPrivate.nextMessageId;
   // Listen for messages post to our window
   CommunicationPrivate.messageListener = (evt: DOMMessageEvent): void => processMessage(evt);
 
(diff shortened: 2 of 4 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -63,7 +63,9 @@ interface InitializeResponse {
 export function initializeCommunication(
   validMessageOrigins: string[] | undefined,
   apiVersionTag: string,
+  customMessageIdStarter?: number,
 ): Promise<InitializeResponse> {
+  CommunicationPrivate.nextMessageId = customMessageIdStarter ?? CommunicationPrivate.nextMessageId;
```
lakhveerk, line 68: just a note: When we `ununinitializeCommunication`, we are still setting the nextMessageId back to zero instead of `customMessageIdStarter`

packages/teams-js/src/public/app.ts
```diff
@@ -727,12 +736,18 @@ export namespace app {
    *
    * @param validMessageOrigins - Optionally specify a list of cross frame message origins. They must have
    * https: protocol otherwise they will be ignored. Example: https://www.example.com
+   *
+   * @param customMessageIdStarter -
```
lakhveerk, line 740: Any help in explaining the parameter please?

packages/teams-js/src/public/app.ts
```diff
@@ -63,88 +67,93 @@ export function appInitializeHelper(apiVersionTag: string, validMessageOrigins?:
...
+        // For now, we might receive any of these possibilities:
+        // - `runtimeConfig` in `runtimeConfig` and `clientSupportedSDKVersion` in `clientSupportedSDKVersion`.
+        // - `runtimeConfig` in `clientSupportedSDKVersion` and `clientSupportedSDKVersion` in `runtimeConfig`.
+        // - `clientSupportedSDKVersion` in `runtimeConfig` and no `clientSupportedSDKVersion`.
+        // This code supports any of these possibilities
+
+        // Teams AppHost won't provide this runtime config
+        // so we assume that if we don't have it, we must be running in Teams.
+        // After Teams updates its client code, we can remove this default code.
+        try {
```
lakhveerk, line 100: I am not sure why all of this showed up as changed in this PR. Only `customMessageIdStarter` is added to some placed and that was it

packages/teams-js/src/public/app.ts
ndangudubiyyam: represents that first number in a monotonically increasing sequence of numbers we stamp the requests from the teamsJS to hubsdk. It helps in matching responses to requests.  
Having a high seed value for this field results in generating messageIds that have a less change of collision when there are multiple instances of teamsJS running by the application.

packages/teams-js/src/public/app.ts
```diff
@@ -46,10 +46,14 @@ const initializationTimeoutInMs = 5000;
 
 const appLogger = getLogger('app');
 
-export function appInitializeHelper(apiVersionTag: string, validMessageOrigins?: string[]): Promise<void> {
+export function appInitializeHelper(
+  apiVersionTag: string,
+  validMessageOrigins?: string[],
+  customMessageIdStarter?: number,
```
AE-MS, line 52: Is this something we really want to enable the entire world to do?

AE-MS:
Did my suggestion of using a peer dependency not work then, @erinha ?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/app.ts
lakhveerk: no just for 3pcxp in a beta release. (correct me if I am wrong @erinha)

packages/teams-js/src/internal/communication.ts
erinha: I think we want to set it back to `customMessageIdStarter` in `uninitializeCommunication` too. @TrevorJoelHarris thoughts?

change/@microsoft-teams-js-2105a377-282c-452d-bb03-208852fcc2dd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds. \\nHaving a high seed value for this field results in generating messageIds that have a less change of collision when there are multiple instances of teamsJS running by the application",
```
erinha, line 3:
```suggestion
  "comment": "Added optional `customeMessageIdStarter` parameter to `app.initialize` call to give an option to developers to set the starting number for messageIds. \\nHaving a high seed value for this field results in generating messageIds that have a less chance of collision when there are multiple instances of teamsJS running by the application",
```

erinha: Also, if we do need to merge this, not sure if we want to be more specific about this being only used for internal Microsoft usage, or something like that.

packages/teams-js/src/public/app.ts
erinha: Correct. This would not be something that anyone outside of this very specific use case should do. I don't think we'd ever release this in a prod version, for example. So far, we are in a prototyping world. (In fact, we should probably make this a draft PR.)

erinha: We should also look at ways to keep this parameter from showing up in docs (take a look at the resources parameter for `authentication.getAuthToken` as an example). @lakhveerk

packages/teams-js/src/public/app.ts
```diff
@@ -63,88 +67,93 @@ export function appInitializeHelper(apiVersionTag: string, validMessageOrigins?:
...
+            // If it's any error that's not a JSON parsing error, we want the program to fail.
+            throw e;
           }
+        }
 
-          GlobalVars.initializeCompleted = true;
-        },
-      );
+        GlobalVars.initializeCompleted = true;
+      });
```
erinha, line 150: This does look like it's changed. Curious if you can grab what this should be from `main` branch and copy it back into the file.

packages/teams-js/src/public/app.ts
```diff
@@ -727,12 +736,20 @@ export namespace app {
    *
    * @param validMessageOrigins - Optionally specify a list of cross frame message origins. They must have
    * https: protocol otherwise they will be ignored. Example: https://www.example.com
+   *
+   * @param customMessageIdStarter - a starting number from where the sequence of messageIds begin that we use to match the requests and responses from teams-js to host sdk.
+   * The default value is zero. A higher seed value reduces the chance of collisions in messageIds when multiple teamsJS instances are running concurrently
```
erinha, line 741: This should be hidden from public docs and labeled internal only usage for Microsoft.

packages/teams-js/src/public/app.ts
lakhveerk: I tried that twice. There must be some setting in vscode that I need to figure

lakhveerk: It looks fine locally - just the indentation had changed, but here (on github), it looks like I have changed the content

image

packages/teams-js/src/public/app.ts
```diff
@@ -727,12 +737,22 @@ export namespace app {
    *
    * @param validMessageOrigins - Optionally specify a list of cross frame message origins. They must have
    * https: protocol otherwise they will be ignored. Example: https://www.example.com
+   *
+   * @hidden
+   * @internal
+   * @param customMessageIdStarter - a starting number from where the sequence of messageIds begin that we use to match the requests and responses from teams-js to host sdk.
+   * The default value is zero. A higher seed value reduces the chance of collisions in messageIds when multiple teamsJS instances are running concurrently
```
erinha, line 744:
```suggestion
   * The default value is zero. A higher seed value reduces the chance of collisions in messageIds when multiple teamsJS instances are running concurrently (an unsupported scenario)
```

TrevorJoelHarris:
Checked with @lakhveerk, they said it was fine to close this

PR #2108, opened 2024-01-03
API Telemetry - Third Batch - Chat, liveShareHost(Interactive), Meeting, Menus, Monetization, People capabilities

KangxuanYe:
## Description

This PR is an update from https://github.com/OfficeDev/microsoft-teams-library-js/pull/2057 since there are so many changes happen on files I want to make change. So, it is easier to raise another PR and abandon previous one for review.

Adding api names to compose api version tags for each API in Chat, LiveShareHost(Interactive), Meeting, Menus, Monetization, People capabilities and let it be sent out in request to Host SDKS to make sure apiVersionTag could be added into telemetry event and then be passed to Kusto data base.

### Main changes in the PR:


packages/teams-js/src/public/liveShareHost.ts
```diff
@@ -226,7 +248,13 @@ export class LiveShareHost {
   public setFluidContainerId(containerId: string): Promise<liveShare.IFluidContainerInfo> {
     ensureSupported();
     return new Promise<liveShare.IFluidContainerInfo>((resolve) => {
-      resolve(sendAndHandleSdkError('interactive.setFluidContainerId', containerId));
+      resolve(
```
KangxuanYe, line 251: @[AE-MS](https://github.com/AE-MS) AE-MS [on Nov 28, 2023](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2108/files#discussion_r1408600340)
This is probably a bigger question than what should be handled in this PR, so it absolutely should not block you from checking in, but should the message being passed and the ApiName constant being used just be the same? In most (all?) cases, they're literally the same string. And representing them twice opens the possibility of unintentional divergence.

@[KangxuanYe](https://github.com/KangxuanYe) KangxuanYe [2 hours ago](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2108/files#discussion_r1440941709)
I have thought of this question and my answer to this is: Ideally, yes. That is to say the value of funcName as a parameter in SDK request should be EXACTLY the same as value of ApiVersionTag and developers don't have to write the same string twice in the future.

However, for now, due to some constraints, i.e. tech debt, openLink special scenarios, we cannot make it. But what I will suggest to do after this project is to use ApiName as the value of funcName to make sure that API keeps the funcName to be exactly the same as apiVersionTag so that in the future, we could remove apiVersionTag and use funcName for telemetry (at that moment the funcNames are API names).

Does that make sense to you? @AE-MS
(@TrevorJoelHarris tagged you for my suggestion in the future)

packages/teams-js/src/public/chat.ts
```diff
@@ -67,20 +74,26 @@ export namespace chat {
    * @beta
    */
   export function openChat(openChatRequest: OpenSingleChatRequest): Promise<void> {
+    const apiVersionTag = getApiVersionTag(chatTelemetryVersionNumber, ApiName.Chat_OpenChat);
+    return openChatHelper(apiVersionTag, openChatRequest);
+  }
+
+  function openChatHelper(apiVersionTag: string, openChatRequest: OpenSingleChatRequest): Promise<void> {
```
KangxuanYe, line 81: [AE-MS](https://github.com/AE-MS) commented [on Nov 28, 2023](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2108/files#issuecomment-1831036722)

Do we care to use telemetry to determine how often people call openGroupChat but only pass a single "chatter"?Because as it is now, when someone calls openGroupChat with a single chatter, it will only get recorded as a call to openChat which won't necessarily help us make decisions about how frequently people are calling openGroupChat

KangxuanYe: @AE-MS I think this is the same edge case as we have for openLink and we try to be as accurate as possible. So I create a helper function and move logics to that function to make sure even if `openChat` is called from `openGroupChat`, we could still know that it is from `openGroupChat`.

KangxuanYe (commented):
Update comments from abandoned PR.

packages/teams-js/src/public/liveShareHost.ts
TrevorJoelHarris: It's a tricky question. Most functions (and possibly all *new* functions should probably be the same string value. We have back compat scenarios though (tasks vs. dialog for example) where we want different telemetry points for the same func values very explicitly.

I don't think aaron is wrong that having separate string values here opens us up to potentially having different values we don't mean to have, but at least in my head I'd rather have too *much* detail that we then have to combine again in the telemetry layer ("these two tags are actually the same call") than not enough detail ("uh oh, this single tag is actually two different function calls") so erring on this side seems okay to me. Just my vote though!

TrevorJoelHarris (approved):

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -70,6 +70,8 @@ export enum ApiName {
   Call_StartCall = 'call.startCall',
   Clipboard_Read = 'clipboard.read',
   Clipboard_Write = 'clipboard.write',
+  Chat_OpenChat = 'chat.openChat',
+  Chat_OpenGroupChat = 'chat.openGroupChat',
   Dialog_AdaptiveCard_Bot_Open = 'dialog.adaptiveCard.bot.open',
```
maglims, line 75: these two lines go above Clipboard lines?

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -99,6 +109,36 @@ export enum ApiName {
...
+  Meeting_RegisterMicStateChangeHandler = 'meeting.registerMicStateChangeHandler',
+  Meeting_RegisterRaiseHandStateChangedHandler = 'meeting.registerRaiseHandStateChangedHandler',
+  Meeting_RegisterSpeakingStateChangeHandler = 'meeting.registerSpeakingStateChangeHandler',
+  Meeting_RequestAppAudioHandling = 'meeting.requestAppAudioHandling',
+  Meeting_RequestStartLiveStreaming = 'meeting.requestStartLiveStreaming',
+  Meeting_RequestStopLiveStreaming = 'meeting.requestStopLiveStreaming',
+  Meeting_ShareAppContentToStage = 'meeting.shareAppContentToStage',
+  Meeting_StopSharingAppContentToStage = 'meeting.stopSharingAppContentToStage',
+  Meeting_ToggleIncomingClientAudio = 'meeting.toggleIncomingClientAudio',
+  Meeting_SetMicStateWithReason = 'meeting.setMicStateWithReason',
```
AE-MS, line 131: >Meeting_SetMicStateWithReason [](http://example.com/codeflow?start=2&length=29)

Alphabetically this seems like it should go above `Meeting_ShareAppContentToStage`

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -99,6 +109,36 @@ export enum ApiName {
...
+  Menus_HandleActionMenuItemPress = 'menus.handleActionMenuItemPress',
+  Menus_HandleNavBarMenuItemPress = 'menus.handleNavBarMenuItemPress',
+  Menus_HandleViewConfigItemPress = 'menus.handleViewConfigItemPress',
+  Menus_RegisterActionMenuItemPressHandler = 'menus.registerActionMenuItemPressHandler',
+  Menus_RegisterNavBarMenuItemPressHandler = 'menus.registerNavBarMenuItemPressHandler',
+  Menus_RegisterSetModuleViewHandler = 'menus.registerSetModuleViewHandler',
+  Menus_SetUpViews = 'menus.setUpViews',
```
AE-MS, line 138: >Menus_SetUpViews [](http://example.com/codeflow?start=2&length=16)

This seems like it should alphabetically be below `Menus_SetNavBarMenu`

packages/teams-js/src/internal/telemetry.ts
KangxuanYe: Oh good catch. I would write it down and adjust in 4th batch. Thank you.

packages/teams-js/src/public/chat.ts
```diff
@@ -67,20 +74,26 @@ export namespace chat {
    * @beta
    */
   export function openChat(openChatRequest: OpenSingleChatRequest): Promise<void> {
+    const apiVersionTag = getApiVersionTag(chatTelemetryVersionNumber, ApiName.Chat_OpenChat);
```
maglims, line 77: Does this need to be unique?  

If yes, from the function  "getApiVersionTag" itself, it doesnt seem that it will guarantee to return a unique string...
 return `${apiVersionNumber}_${functionName}`;

for example, two functions with same apiVersionNumber and same function names(function overload)

maglims: taking a further look, since the functionName is a "controlled" name, which we define it in telemetry.ts, it may not be an issue if we update telemetry.ts file correctly for new updates, so I don't know if we ever run into the issue I called out above.
The comment is  just from purely coding perspective, if apiVersionTag needs to be unique, that function itself cannot guarantee that..

packages/teams-js/src/internal/telemetry.ts
KangxuanYe: updated

packages/teams-js/src/public/chat.ts
KangxuanYe: That's a good thought. I think the biggest concern is what if a person add the same value to different keys and composed tags used the same versions. This could be a human mistake though very little chance would happen. I will take a look on it and to see if we could figure it out leveraging on ESLint check.

packages/teams-js/src/internal/telemetry.ts
KangxuanYe: fixed.

packages/teams-js/src/internal/telemetry.ts
KangxuanYe: fixed.

TrevorJoelHarris (approved):

maglims (approved):

AE-MS (approved):
:shipit:

PR #2114, opened 2024-01-06
Fix and enhance readability (hopefully) of unit tests for versioned runtime 

AE-MS:
## Description

The unit tests for versioned runtime application function (`generateVersionBasedTeamsRuntimeConfig`) were confusing to understand, wrong, and incomplete.

- There was a lot of nesting both inside and outside of the tests, making it hard to read and understand.
- The tests only tested for one version lower, and not all versions in the map
- The tests which tried to validate that unsupported platforms were always unsupported were broken in the case where one capability was supported on some platforms and then later that same capability was added for other platforms.
 
### Main changes in the PR:

1. Refactor and rewrite the tests to, hopefully, be more readable.
2. Write a new test that properly validates the unsupportedness of platforms

Reviewers may find it easier to ignore the diff and just look at the "right side" (e.g. the contents of the updated file without reference to what it used to look like)

### Validation performed:

1. Tests pass
2. Tests pass even if I make @maglims's change (which was failing the tests in the PR #2112 and alerted me to these issues)
4. Tests fail when I intentionally introduce bugs into `generateVersionBasedTeamsRuntimeConfig`

### DIFF
--- a/packages/teams-js/src/public/runtime.ts
+++ b/packages/teams-js/src/public/runtime.ts
@@ -437,7 +437,7 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
   '2.0.5': [
     {
       capability: { webStorage: {} },
-      hostClientTypes: [HostClientType.android, HostClientType.desktop, HostClientType.ios],
+      hostClientTypes: [HostClientType.android, HostClientType.ios],
     },
   ],
 };
@@ -513,6 +513,11 @@ export function generateVersionBasedTeamsRuntimeConfig(
           GlobalVars.hostClientType !== undefined &&
           capabilityReqs.hostClientTypes.includes(GlobalVars.hostClientType)
         ) {
+          console.log(
+            `Adding capabilities from version ${versionNumber} for host client ${
+              GlobalVars.hostClientType
+            }:  ${JSON.stringify(capabilityReqs.capability)}`,
+          );
           newSupports = mergeRuntimeCapabilities(newSupports, capabilityReqs.capability);
         }
       });
--- a/packages/teams-js/src/public/runtime.ts
+++ b/packages/teams-js/src/public/runtime.ts
@@ -513,11 +513,6 @@ export function generateVersionBasedTeamsRuntimeConfig(
           GlobalVars.hostClientType !== undefined &&
           capabilityReqs.hostClientTypes.includes(GlobalVars.hostClientType)
         ) {
-          console.log(
-            `Adding capabilities from version ${versionNumber} for host client ${
-              GlobalVars.hostClientType
-            }:  ${JSON.stringify(capabilityReqs.capability)}`,
-          );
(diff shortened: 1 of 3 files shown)


AE-MS:

The changes I made to the runtime file do not require a new runtime version.

packages/teams-js/src/public/runtime.ts
```diff
@@ -437,7 +437,7 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
   '2.0.5': [
     {
       capability: { webStorage: {} },
-      hostClientTypes: [HostClientType.android, HostClientType.desktop, HostClientType.ios],
```
AE-MS, line 440: >HostClientType.desktop [](http://example.com/codeflow?start=48&length=22)

I suspect that `HostClientType.desktop` was added here to workaround the bug I'm fixing in these tests.

This map should only represent the version when a capability first started working on a platform; that platform doesn't need to then be added to all newer versions. Support going forward is presumed.

maglims (commented):
Thank you for the quick workaround!

maglims (dismissed):

packages/teams-js/test/public/runtime.spec.ts
```diff
@@ -190,107 +190,226 @@ describe('runtime', () => {
...
-
   describe('generateVersionBasedTeamsRuntimeConfig tests based on Teams default configuration', () => {
-    Object.entries(mapTeamsVersionToSupportedCapabilities).forEach(([version, capabilityAdditionsForEachVersion]) => {
-      capabilityAdditionsForEachVersion.forEach((capabilityAdditionsForClientTypesInASpecificVersion) => {
-        const capabilityAdditionsForThisVersion = capabilityAdditionsForClientTypesInASpecificVersion.capability;
-        capabilityAdditionsForClientTypesInASpecificVersion.hostClientTypes.forEach((clientType) => {
-          it(`Back compat host client type ${clientType} supporting up to ${version} should support ${JSON.stringify(
-            capabilityAdditionsForThisVersion,
-          )}`, async () => {
+    it('Validate that all client types where a capability began to be supported report that capability IS supported in that version', async () => {
```
TrevorJoelHarris, line 194: Do you think it would be worth restructuring this so instead of one big "super" test for all versions/capabilities/etc. we had more discrete tests for smaller numbers of combinations? We do something likes this for testing framecontexts:

image

Doing this might make the tests take longer (depending on how many they are), but could also make test failures easier to understand and investigate

TrevorJoelHarris (approved):

maglims (approved):

packages/teams-js/test/public/runtime.spec.ts
AE-MS: Your suggestion is interesting; I would like to explore it further. If it's okay with you, I'd like to check in the change as-is to unblock @maglims to unblock Teams Mobile. I will follow-up in a separate PR on the possibilities of further restructuring.

PR #1903, opened 2023-08-30
User/mahimashukla/3 p drag and drop

mahimaMSFT:
For more information about how to contribute to this repo, visit [this page](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md).

## Description

> Drag and drop of files in compose should upload file to 3P storage providers ( Box, Google Drive, Egnyte 
etc) when default provider is not ODSP.

> If this Pull Request should close/resolve any issues when merged, use [the special syntax for that](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) here.

### Main changes in the PR:

1. Added the API which can be used by the 3p apps to get the files which were dragged and dropped in compose, when user wants to upload in 3P apps.
2. Unit test to test the new API's

### Validation performed:

1. Validated the code with unit test cases.
2. Created a small local app to check if the new API's are working as expected.

### Unit Tests added:

> Unit tests are required for all changes. If no unit tests were added as part of this change, please explain why they aren't necessary.

Yes, unit test cases are added.

### End-to-end tests added:

Yes

### Change file added:

> Ensure the change file meets the [formatting requirements](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball).

Yes

### Next/remaining steps:

> List the next or remaining steps in implementing the overall feature in subsequent PRs (or is the feature 100% complete after this?).

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

PR #2125, opened 2024-01-11
Minor change to set the helper values in ThirdPartyCloudStorageAPIs

mahimaMSFT:

## Description

[https://github.com/OfficeDev/microsoft-teams-library-js/pull/1903/files#diff-ec0b521b72add73ea784045aa2a4fd63089f0352ada767a1cac53e707089e0ca](url)
In above PR, in apps/teams-test-app/src/components/ThirdPartyCloudStorageAPIs.tsx helper was not having the fileType value because of which createFile function was failing.
Have assigned the fileType value to the helper

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/change/@microsoft-teams-js-6a7d3368-138b-40f7-86c3-0d9e78441ff5.json
+++ b/change/@microsoft-teams-js-6a7d3368-138b-40f7-86c3-0d9e78441ff5.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Minor change in helper values",
+  "packageName": "@microsoft/teams-js",
+  "email": "mahimashuklaMSFT@gmail.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/public/thirdPartyCloudStorage.ts
+++ b/packages/teams-js/src/public/thirdPartyCloudStorage.ts
@@ -193,6 +193,7 @@ export namespace thirdPartyCloudStorage {
             const assemble: AssembleAttachment | null = decodeAttachment(fileResult.fileChunk, fileResult.fileType);
             if (assemble) {
               helper.assembleAttachment.push(assemble);
+              helper.fileType = fileResult.fileType;
             } else {
               Files3PLogger(
                 `Received a null assemble attachment for when decoding chunk sequence ${fileResult.fileChunk.chunkSequence}; not including the chunk in the assembled file.`,


packages/teams-js/src/public/thirdPartyCloudStorage.ts
```diff
@@ -193,6 +193,7 @@ export namespace thirdPartyCloudStorage {
             const assemble: AssembleAttachment | null = decodeAttachment(fileResult.fileChunk, fileResult.fileType);
             if (assemble) {
               helper.assembleAttachment.push(assemble);
+              helper.fileType = fileResult.fileType;
```
AE-MS, line 196: While this seems like it would fix the problem you observed for this specific usage (`fileType` property not getting set), I would suggesting considering a different approach to fixing this:

Since it looks like it is never valid for there to be an `AttachmentListHelper` without a valid `fileType` property, If `AttachmentListHelper` was a `class` instead of an `interface`, and the constructor for it required passing the `fileType` as a parameter, that would prevent `AttachmentListHelper` objects from ever being able to exist without a valid `fileType` (since you couldn't create one without providing it). With a class-based design, not only would it fix this specific instance of the problem, it would also prevent this problem ever occurring again, even as the code continued to be modified, since no one could create `helper` objects that didn't have a `fileType` set.

The `helper` variable could be initially be set to `null`, then in this code on line 195 it could check if `helper` was `null`, and if it was, create a `new AttachmentListHelper` with the appropriate `fileType`. If it wasn't `null`, it could just use it (and confidently know that the `fileType` had already been set).

Then on line 229 it could set `helper` back to `null`, if needed.

I'd love to hear your thoughts on this alternate approach to fixing this issue, which would not only fix this instance but prevent similar problems in the future.

mahimaMSFT: Agreed with this approach, but have some points.

While implementing this we tried to make it consistence with existing media implementation: packages\teams-js\src\public\media.ts

We are setting FileResult fileType value to helper's fileType,  since FileResult's fileType is a mandatory field so there won't be any chance that it is null, and helper will set null value to fileType.

Since, we are using interface-based approach for all the other values in this file, so I think we should use that for all the values.

In this code, it is receiving data in chunks and for every chunk we are pushing its value in assembleAttachment array. 
If there are total 10 or 15 files it will create one object for each new file.

AE-MS: Using the media implementation as a model is not the best approach. Is has a design which confuses developers and is in the process of being replaced with capabilities that have clearer interfaces, separation of responsibilities, and cross platform support. Even using interfaces elsewhere in this file shouldn't be the basis for the design here: we should generally strive to use the right tool for the job -- that is, choose a design that matches the problem requirements, not bias towards existing designs that are built under different constraints.

I'm not sure what you mean by "there won't be any chance that is is null" -- there literally was a chance that it was null, that problem is what precipitated this bug fix, yes?

I applaud the desire to ensure we write optimized code that doesn't unnecessarily allocate objects. However, in this case there's a tradeoff between ease-of-use and performance. Creating new objects is usually quite cheap, particularly in this case when the new object that I'm suggesting you allocate will literally have no data at the time it is allocated. Is there profile data suggesting that optimizing the number of objects allocated in this scenario significant contributes to the performance of the workflow? We should optimize for correctness, ease-of-use, and prevention of bugs. Only once we find that there's a clear tension (i.e., data) between those design principles and performance should we loosen our commitment to that design in favor of something that is more brittle but makes a significant difference in the runtime or memory usage.

mahimaMSFT: sure, I have implemented this approach.

mahimaMSFT: I would like to request further code refactoring as a swift follow-up, as it will not affect the end user utilizing this API. We are behind schedule for our feature commitment date, and we can only verify if the feature is working end-to-end once this change is implemented in the beta version.

AE-MS: Okay, I'm totally sure what you mean by "further code refactoring." Should I not leave comments on any of the new code you've written? Or do you mean that you aren't going to write the unit tests for it? Or...?

change/@microsoft-teams-js-6a7d3368-138b-40f7-86c3-0d9e78441ff5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Minor change in helper values",
```
TrevorJoelHarris, line 3: The comment you have here is insufficient; it gets turned patch notes for internal and external developers. Try something like:

```suggestion
  "comment": "Fixed bug where `thirdPartyCloudStorage.getDragAndDropFiles` failed deterministically",
```

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/thirdPartyCloudStorage.ts
mahimaMSFT: Hi @AE-MS ,
I noticed I missed adding a comment for the unit test case. I want to clarify that I appreciate comments on the code, and I'm open to your suggestions. I'd be happy to incorporate your inputs as a quick follow-up task, which we are tracking here: https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/3496318

Could you please help me with one inputs,
currently in the code, the initial value of the helper will be null, when we will receive the first chunk of the first file we are creating and object of helper and after it for all the other chunk I am adding the assemble values in helper object assembleAttachment array.

Need some help on how to write a test case which is specific to helper function, currently in test case where we are sending multiple files will cover the helper object.

AE-MS: It seems like what you want to do is write a unit test that will trigger the callback (`handleGetDragAndDropFilesCallbackRequest`) and then verify that it calls `createFile` with the correct parameters (e.g, that the `mimeType` parameter to that method is properly populated)

mahimaMSFT: Thankyou @AE-MS ,
I have modified the unit test, verified the files which we are receiving is having proper fileType/mimetype and fileName.

AE-MS (approved):

TrevorJoelHarris (approved):

PR #2134, opened 2024-01-18
"Remove" unused callbackUrl parameter from authentication.notifySuccess/Failure

AE-MS:
## Description

The `callbackUrl` parameter to `authentication.notifySuccess` and `authentication.notifyFailure` was a mystery. 
This was originally added with PR #[80](https://github.com/OfficeDev/microsoft-teams-library-js/pull/80) in 2017. However, the documentation about the usage of this parameter is unclear. Moreover, neither our team, nor the Teams team, nor the Outlook Win32 team, nor the graph connectors team has any knowledge about the usage of this parameter and say it's unused/unnecessary.

### Main changes in the PR:

"Remove" this parameter. Since theoretically no one in the world should be using this parameter (since it served no purpose and supposedly supports a scenario that doesn't exist (graph connector configuration in Outlook Win32), actually removing it should not be a problem. However, just in case -- and because we don't want to make an API-breaking change -- I'm making an overload of `notifySuccess` and `notifyFailure` that don't have the parameter at all. Then, I'm marking the version with the optional parameter as `@deprecated`. In TJS3.0, we can actually remove it.

Note: TypeDoc seems inconsistent about how it handles overloads. Supposedly, it supports documenting all the overloads. However, there are a number of issues that have been filed about how it doesn't actually show them properly. _Some_ of the issues turned out to be that the UI for showing all the overloads ("a little green arrow") was hard to find. I generated the docs for this change and, as far as I can tell, there's no way to see the deprecated overload. I couldn't find any UI anywhere that showed it. This is actually okay because the one that does show up the docs (the new version without the parameter) is the one we want people to see anyway. However, I did properly document/deprecate the version with the optional parameter in case it shows up in any documentation (e.g., a future version of TypeDoc or the special custom rendering that docs.microsoft.com does for TypeDoc-based code).

I also removed the unit tests that were using this parameter

### Validation performed:

1. Validate that all tests pass
2. Make sure no references or usage of this parameter was found on the Internet

### Change file added:

Yes.


jadahiya-MSFT (approved):

packages/teams-js/src/public/authentication.ts
```diff
@@ -428,10 +426,9 @@ export namespace authentication {
    *
    * @param result - Specifies a result for the authentication. If specified, the frame that initiated the authentication pop-up receives
    * this value in its callback or via the `Promise` return value
-   * @param callbackUrl - Specifies the url to redirect back to if the client is Win32 Outlook.
+   * @param _callbackUrl - This parameter is deprecated and unused
    */
-  export function notifyFailure(reason?: string, callbackUrl?: string): void {
-    redirectIfWin32Outlook(callbackUrl, 'reason', reason);
+  export function notifyFailure(reason?: string, _callbackUrl?: string): void {
```
TrevorJoelHarris, line 431: This change looks fine, but how would you feel about hiding this function from docs and introducing a new alternate signature that doesn't allow someone to pass in a callback url? Seems like that would lead to less confusion for new people

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/authentication.ts
AE-MS: I had considered that when recognizing that I couldn't really "deprecate" the parameter. I'm trying to think about why I didn't take that route. This is not a good reason but I think it was because I didn't know what to call the new functions. 😅

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

PR #2137, opened 2024-01-19
Update eslint packages and fix subsequently triggered linting errors

AE-MS:
## Description

Did you ever see this error and/or get tired of seeing it when building?

```
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0

YOUR TYPESCRIPT VERSION: 4.9.5

Please only submit bug reports when using the officially supported version.
```

I did! The repo was using an older version of the `@typescript-eslint/typescript-estree` package, which it was only using indirectly because of the version of eslint it was using.

Consequently, I updated eslint and a few other packages that were needed as a part of it. This lead to some new lint errors being detected, which I fixed here (including one that was actually a bug!)

I also unified the tsconfig being used by eslint with the actual product tsconfig, so we don't have to maintain two parallel configurations and remember to keep them in sync, etc.

Note: I put PR comments directly on the code where I felt that further detail was warranted.

### Validation performed:

1. Ensure all linting passes
2. Ensure all tests pass
3. Ensure that intentionally introducing a lint error in both the test and production code still produces a lint error
4. Ensure that the original WARNING about the TypeScript version is no longer present.


apps/teams-test-app/src/components/utils/ApiWithTextInput.tsx
```diff
@@ -21,7 +21,7 @@ export interface ApiWithTextInputProps<T> {
   defaultInput?: string;
 }
 
-export const ApiWithTextInput = <T extends unknown>(props: ApiWithTextInputProps<T>): React.ReactElement => {
```
AE-MS, line 24: > extends unknown [](http://example.com/codeflow?start=34&length=16)

`extends unknown` does nothing and is unnecessary

apps/teams-test-app/src/components/utils/ApiWithTextInput.tsx
```diff
@@ -21,7 +21,7 @@ export interface ApiWithTextInputProps<T> {
...
+export const ApiWithTextInput = <T,>(props: ApiWithTextInputProps<T>): React.ReactElement => {
```
AE-MS, line 24: >, [](http://example.com/codeflow?start=34&length=1)

The type parameter needs to be declared with a comma here because this is a TSX file (which has ambiguous angle bracket parsing in some contexts, where it can't tell if you're reference an HTML element/control or if you're using a generic). In this case, it would have been ambiguous so the comma is required to indicate that this is a type parameter.

packages/teams-js/.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   ignorePatterns: ['.eslintrc.js'],
   parserOptions: {
-    project: './tsconfig.strictNullChecks.json',
+    project: './tsconfig.eslint.json',
```
AE-MS, line 4: Eslint will now share the same tsconfig as the used when compiling (with one minor change that you will see in the eslint tsconfig file)

packages/teams-js/tsconfig.eslint.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["node_modules", "./webpack.config.ts"]
```
AE-MS, line 3: Eslint shares the main tsconfig except that it does NOT exclude the test directory, since we want eslint to run on the test directory as well.

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -106,7 +106,7 @@ export enum ApiName {
   GeoLocation_GetCurrentLocation = 'geoLocation.getCurrentLocation',
   GeoLocation_HasPermission = 'geoLocation.hasPermission',
   GeoLocation_Map_ChooseLocation = 'geoLocation.map.chooseLocation',
-  GeoLocation_RequestPermission = 'geoLocation.hasPermission',
```
AE-MS, line 109: This was a human error that eslint now detected as a duplicate enum entry.

packages/teams-js/test/utils.ts
```diff
@@ -196,7 +196,10 @@ export class Utils {
     return null;
   };
 
-  public respondToMessage = async (message: MessageRequest | NestedAppAuthRequest, ...args: unknown[]): Promise<void> => {
+  public respondToMessage = async (
+    message: MessageRequest | NestedAppAuthRequest,
+    ...args: unknown[]
+  ): Promise<void> => {
```
AE-MS, line 202: VSCode just really wanted to format this line differently

tools/bundle-size-tools/src/ADO/PrCommentsUtils.ts
```diff
@@ -118,7 +118,7 @@ export class prCommentsUtils {
     await gitApi.updateThread(thread, this.repoId, this.pullRequestId, existingThread.id);
   }
 
-  private async getThreadByType(threadType: string): Promise<GitPullRequestCommentThread> {
+  private async getThreadByType(threadType: string): Promise<GitPullRequestCommentThread | undefined> {
```
AE-MS, line 121: > | undefined [](http://example.com/codeflow?start=88&length=12)

This function can return undefined and all callers were already handling that possibility.

apps/teams-test-app/src/components/utils/ApiWithTextInput.tsx
erinha: Thanks for the explanation, I had no idea!

erinha (approved):
Looks good! Thank you for the explanations in comments.

jadahiya-MSFT (approved):

PR #2140, opened 2024-01-19
Private MessageChannels.getTelemetryPort

kerrynf:
## Description
Adds support for a new private API to accept a MessagePort from the Host to allow the app to reuse the host application's internal telemetry batching infrastructure.

### Main changes in the PR:

1. Add support for receiving callbacks with ports in communication.ts
2. Add new private messageChannels namespace with the getTelemetryPort API

### Validation performed:

1. Updated the test app to expose the new API
2. Tested against an internal hub to confirm E2E functionality

### Unit Tests added:
Yes

### End-to-end tests added:
No - these will be fast follow once we have E2E support for the API.

### Change file added:
Yes


packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,51 @@
...
+          return telemetryPort;
+        } else {
+          throw new Error('MessageChannels.getTelemetryPort: Host did not return a MessagePort.');
+        }
+      } catch (e) {
+        throw new Error('MessageChannels.getTelemetryPort: Error thrown from message promise.');
+      }
+    }
+
+    throw new Error('MessageChannels.getTelemetryPort: SDK not initialized.');
```
kerrynf, line 49: What are the SDK best practices around communicating API errors to the calling app? I've seen exceptions in some handlers, so using that pattern to start with.

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,51 @@
+import { requestPortFromParent } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { runtime } from '../public/runtime';
+
+let telemetryPort: MessagePort | undefined;
```
kerrynf, line 5: We don't want to have more than one channel open for a particular iFrame, so we should cache the received port in case we are asked for it again.

Is there better approach for this?

TrevorJoelHarris: I think the only change I'd ask for is to move this into the messageChannels namespace so that telemetryPort isn't in the global namespace. That could cause some weird problems down the line.

I don't remember all the details of your feature right now (sorry) but since you said something about this being for iframes we should also make sure that whatever you're doing here will still make sense in cases where the app isn't hosted inside of an iframe (mobile platforms for example)

kerrynf: This feature is currently intended for the Teams desktop/web code to allow delegation of the telemetry from the iFrame to the worker thread. 

Based on the documentation I've found, if there is a similar need for Android, we would be able to fulfil this API using the [WebMessagePort APIs](https://developer.android.com/reference/android/webkit/WebMessagePort) in Android.

There doesn't look to be an equivalent API for iOS however.

We could potentially mock something on the teams-js SDK side if needed, but unless we believe we need to improve 1P telemetry performance for the mobile apps, I would suggest we just fail to return a port on those hosts, and fallback to the frame doing its own telemetry?

packages/teams-js/src/internal/communication.ts
```diff
@@ -52,6 +52,9 @@ class CommunicationPrivate {
   public static promiseCallbacks: {
     [id: number]: Function; // (args[]) => void
   } = {};
+  public static portCallbacks: {
```
TrevorJoelHarris, line 55: Why do you need to store portCallbacks (which are still promise based) separately from the normal promiseCallbacks? Is it just to help keep straight that for promiseCallbacks you pass `message.args` only and for portCallbacks you pass only evt.ports[0]?

change/@microsoft-teams-js-5681fcee-94b0-459c-ac25-ec6377394b5d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add private MessageChannels capability",
```
shrshindeMSFT, line 3: Please use past tense in the comment. `Added`

packages/teams-js/src/internal/communication.ts
```diff
@@ -712,6 +744,18 @@ function handleParentMessage(evt: DOMMessageEvent): void {
       logger('Removing registered promise callback for message %i', message.id);
       delete CommunicationPrivate.promiseCallbacks[message.id];
     }
+    const portCallback = CommunicationPrivate.portCallbacks[message.id];
+    if (portCallback) {
+      logger('Invoking the registered port callback for message %i with arguments %o', message.id, message.args);
+      let port: MessagePort | null = null;
+      if (evt.ports && evt.ports[0] instanceof MessagePort) {
+        port = evt.ports[0];
```
TrevorJoelHarris, line 752: It would be helpful to throw an exception here if no port is passed down, rather than where you are doing it in `messageChannel.ts`. I'm don't anticipate a lot of functions will be requesting ports from the host, but just in case it would be good to check for that common error and throw here rather than depend on all of the callers to handle it consistently.

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: (fwiw I think I like keeping them separate because it reduces the odds of someone passing down ports from the host in ways we aren't expecting and makes sure to keep port passing walled off in its own section of communication, just trying to make sure I'm not missing anything

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @throws Error if {@linkcode app.initialize} has not successfully completed,
+   * if the host does not support the feature, or if an error is thrown in message handling.
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export async function getTelemetryPort(): Promise<MessagePort> {
+    // If the port has already been initialized, return it.
+    if (telemetryPort) {
```
shrshindeMSFT, line 28: Why telemetry port is returned before ensureInitialized call? Won't it be problematic in case of multiple calls to this function ?

packages/teams-js/src/internal/communication.ts
```diff
@@ -320,6 +324,34 @@ export function sendMessageToParentAsync<T>(actionName: string, args: any[] | un
...
 }
 
+/**
+ * @hidden
+ * Send a message to parent requesting a MessageChannel Port.
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function requestPortFromParent(actionName: string, args: any[] | undefined = undefined): Promise<MessagePort> {
```
TrevorJoelHarris, line 334:
```suggestion
export function requestPortFromParent(apiVersionTag: string, actionName: string, args: any[] | undefined = undefined): Promise<MessagePort> {
```

We have another engineer on our team currently working on converting all of our communication functionality to take in apiVersionTag here as well, so that it can be passed down from the caller. See `sendMessageToParentWithVersion` for an example of how that works. Function should do something similar to that (verify tag format, etc.) 

The tag that `messageChannels` passes in should be a combination of ApiVersionNumber.V_2 (because this function was first called in v2 of TeamsJS) and the actionName.

You can look at callers of `sendMessageToParentWithVersion` to see how to build tags correctly.

If we introduce new functionality or alternate signatures of this in the future we can update the tags to keep them straight in our telemetry.

packages/teams-js/test/private/messageChannels.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  describe('Testing messageChannels.getTelemetryPort', () => {
+    it('should throw if called before initialization', async () => {
+      expect.assertions(1);
+      utils.uninitializeRuntimeConfig();
+      await expect(messageChannels.getTelemetryPort()).rejects.toThrowError(new Error(errorLibraryNotInitialized));
+    });
+
+    it('should return port from message and then from local variable', async () => {
+      expect.assertions(2);
+      await utils.initializeWithContext('content');
```
shrshindeMSFT, line 43: Can you please add other contexts as well and there respective results?

shrshindeMSFT (changes requested):

packages/teams-js/src/private/messageChannels.ts
kerrynf: If there is a cached port inside the iFrame, it returns immediately. If not, it will call the parent to fetch the port, and store it for any future calls.

The cached port will never be present before the initialization is complete, because we cannot fetch it from the parent until after initialization is complete.

Can you explain further what flow you are envisioning causing issues?

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,59 @@
...
+import { runtime } from '../public/runtime';
+
+/**
+ * @hidden
+ * Namespace to request message ports from the host application.
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export namespace messageChannels {
```
TrevorJoelHarris, line 12: This namespace and all of the exported functions should be marked with the `@beta` tag, which will let people know that the signatures here may change with little or no notice. Once you have this in production and are happy with it / feel confident you won't need to make breaking changes you can come back and remove the `@beta` tag.

packages/teams-js/src/private/messageChannels.ts
TrevorJoelHarris, line 28: Is it safe to assume that the telemetry port never changes? Would be good to add a short log statement here saying that the cached telemetry port is being used. Would also be good to add that information to the docs above the function.

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    ensureInitialized(runtime);
+
+    // Send request for telemetry port
+    let response: MessagePort | undefined;
+    try {
+      response = await requestPortFromParent('messageChannels.getTelemetryPort');
+    } catch (e) {
+      throw new Error('MessageChannels.getTelemetryPort: Error thrown from message promise.');
```
TrevorJoelHarris, line 39: Would be good to capture the original error message from `e` here

packages/teams-js/test/private/messageChannels.spec.ts
kerrynf: This API is not restricted to any particular contexts and will return the same result in all cases.

It didn't seem that the other tests I referenced were testing different contexts unless the behavior differed - is the expected test pattern to test all contexts to demonstrate that the same result is returned?

change/@microsoft-teams-js-5681fcee-94b0-459c-ac25-ec6377394b5d.json
kerrynf: Updated.

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * @hidden
+   * Undocumented function used to clear state between unit tests
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export function _clearTelemetryPort(): void {
```
TrevorJoelHarris, line 56: Please call this function from `app._uninitialize`, which we use to clear all persistent state between tests.

packages/teams-js/src/private/messageChannels.ts
TrevorJoelHarris, line 12: This namespace should have an `isSupported` function like (almost) all of the other namespaces in TeamsJS. That function makes sure the host supports this capability. Developers can call it directly if they wish, but each function also checks the value of `isSupported` before sending messages to the host. (This is why @shrshindeMSFT asked you to call ensureInitialized, isSupported won't return the correct value before the sdk has been initialized).

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -656,6 +656,173 @@ describe('Testing communication', () => {
...
     });
   });
+  describe('requestPortFromParent', () => {
+    let utils: Utils = new Utils();
+    const actionName = 'test';
+    beforeEach(() => {
+      class MockMessagePort {}
+      global.MessagePort = MockMessagePort as unknown as typeof MessagePort;
+      utils = new Utils();
+      communication.uninitializeCommunication();
```
TrevorJoelHarris, line 666: nit: these calls should be able to be replaced with a call to `app._uninitialize`

packages/teams-js/test/private/messageChannels.spec.ts
TrevorJoelHarris: @kerrynf I definitely believe we haven't always been good about testing every context in some of these cases, but in general it's good to test them all because if you were to come back and restrict this function to a smaller set of contexts in the future it would be a breaking change for developers who depend on this function, which we would want to catch in the test layer.

TrevorJoelHarris (changes requested):

apps/teams-test-app/src/components/privateApis/MessageChannelAPIs.tsx
```diff
@@ -0,0 +1,25 @@
+import { messageChannels } from '@microsoft/teams-js';
+import React from 'react';
+
+import { ApiWithoutInput } from '../utils';
+import { ModuleWrapper } from '../utils/ModuleWrapper';
+
+const GetTelemetryPort = (): React.ReactElement =>
```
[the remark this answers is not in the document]
kerrynf: Fast follow will be good here I think, given the dependencies on the internal repo for testing.

packages/teams-js/src/internal/communication.ts
kerrynf: Yes, the separation was to make sure that we can't leak ports unexpectedly in other call types, keeping it very explicit.

packages/teams-js/src/internal/communication.ts
kerrynf: Makes sense.

packages/teams-js/src/private/messageChannels.ts
kerrynf: Yes, for a single iFrame, the intent is to have a single port that is requested during start up. Ideally the iFrame should cache the port itself within its telemetry framework, but if for some reason it is rerequested, we don't want a new message channel to be created.

packages/teams-js/test/private/messageChannels.spec.ts
kerrynf: Sounds good. :)

packages/teams-js/src/private/messageChannels.ts
kerrynf: Sounds good.

kerrynf: Doing this breaks the global initialization flow - all the tests start failing with the error "app.Messages is undefined". Propose leaving this one here as it should only be needed in this namespace?

packages/teams-js/src/private/messageChannels.ts
kerrynf: image

The pattern that I see in the other APIs is calling ensureInitialized twice - once directly, and then again as part of the isSupported check. This seems redundant?

Are we fine to just call isSupported with the assumption it is also calling ensureInitialized?

kerrynf: Done

packages/teams-js/src/private/messageChannels.ts
shrshindeMSFT: Discussed offline. As every app has it's own iFrame there won't be issues of accessing telemetry port from another iframe.

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Limited to Microsoft-internal use
+   */
+  export async function getTelemetryPort(): Promise<MessagePort> {
+    // If the port has already been initialized, return it.
+    if (telemetryPort) {
+      logger('Returning telemetry port from cache');
+      return telemetryPort;
+    }
+
+    isSupported();
```
shrshindeMSFT, line 42:
```suggestion
          if (!isSupported()) {
        throw errorNotSupportedOnPlatform;
      }
```

shrshindeMSFT (changes requested):

packages/teams-js/src/internal/communication.ts
kerrynf: Realized that throwing it here wouldn't actually trickle back to the caller, have put it in the callback itself.

change/@microsoft-teams-js-5681fcee-94b0-459c-ac25-ec6377394b5d.json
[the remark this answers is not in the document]
kerrynf: This comment format looks like it should apply for public capabilities? I don't think we can document this in the public capmatrix. (Maybe your link would help here, but it isn't working for me, seems to take me back to this PR :))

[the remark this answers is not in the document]
TrevorJoelHarris: Ah, yeah that's fair. Here's [what the link resolves to](https://learn.microsoft.com/en-us/microsoftteams/platform/m365-apps/teamsjs-support-m365), not sure why it is taking you back to this PR.

packages/teams-js/test/internal/communication.spec.ts
kerrynf: Hmm, I tried this, and 3 of the tests started failing - this is the pattern used by the existing sendMessageToParentAsync tests (where I copied it from :)). 

image

I looked at app._uninit, which does look like it calls to unitializeCommunication, so I don't really know what is happening here...

packages/teams-js/src/private/messageChannels.ts
kerrynf: Oops I had that before, and obviously messed it up in refactoring! Good catch.

packages/teams-js/src/private/messageChannels.ts
TrevorJoelHarris: Okay yeah that's *weird*. I will open a bug to have someone investigate why this is happening, but definitely not your fault so okay to leave this as-is for now.

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: Same as above, I looked into this and it's very weird and not your fault. Okay to leave this as-is in your PR. I'll open a bug to investigate on our end.

TrevorJoelHarris (changes requested):

apps/teams-test-app/public/manifest.json
```diff
@@ -8,9 +8,9 @@
   "developer": {
     "name": "M365",
     "mpnId": "",
-    "websiteUrl": "https://<REPLACE WITH URL OF THE APP>",
```
kerrynf, line 11: Accidental change, will revert.

packages/teams-js/src/public/constants.ts
```diff
@@ -165,7 +165,10 @@ export enum ChannelType {
 /** An error object indicating that the requested operation or feature is not supported on the current platform or device.
  * @typedef {Object} SdkError
  */
-export const errorNotSupportedOnPlatform: SdkError = { errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM };
+export const errorNotSupportedOnPlatform: SdkError = {
+  errorCode: ErrorCode.NOT_SUPPORTED_ON_PLATFORM,
+  message: 'Not supported on current platform',
```
TrevorJoelHarris, line 170: This new message is causing some E2E tests to fail because they were expecting only an error code in this case. I think we would be fine to add a message here, but updating all of those tests is probably outside the scope of the change you're making. Feel free to go back to error code only (which matches what the rest of our repo does right now) and open a bug for us to add a message here in the future.

TrevorJoelHarris (dismissed):

packages/teams-js/src/internal/communication.ts
```diff
@@ -320,6 +324,45 @@ export function sendMessageToParentAsync<T>(actionName: string, args: any[] | un
...
+  });
+}
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function waitForPort(requestId: number): Promise<MessagePort> {
+  return new Promise<MessagePort>((resolve, reject) => {
+    CommunicationPrivate.portCallbacks[requestId] = (port: MessagePort | undefined, args?: any[]) => {
```
AE-MS, line 355: >any [](http://example.com/codeflow?start=91&length=3)

Could this be `unknown` instead of `any`? Since you aren't actually making any assumptions about the type in your code it doesn't matter in practice today, but it would make it easier for us to eventually enable the `no-explicit-any` rule in the future.  😜

shrshindeMSFT (dismissed):
LGTM 🚢

packages/teams-js/src/internal/communication.ts
```diff
@@ -320,6 +324,45 @@ export function sendMessageToParentAsync<T>(actionName: string, args: any[] | un
...
+      if (port instanceof MessagePort) {
+        resolve(port);
+      } else {
+        // First arg is the error message, if present
+        reject(args && args.length > 0 ? args[0] : new Error('Host responded without port or error details.'));
```
AE-MS, line 360: >reject(args && args.length > 0 ? args[0] : new Error('Host responded without port or error details.')); [](http://example.com/codeflow?start=8&length=103)

It seems... notable? that if someone wanted to handle the error from here, they'd have to be prepared to handle two different types of error patterns: either the promise rejection will contain a string or it will contain an `Error` object containing a string. Is this an important distinction for the caller to make (i.e., should they take different actions depending on which type of error pattern is being used)?

You are clearly handling this in your unit tests, so it's obviously possible to handle it (I also tried writing some [sample code](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAtgTwAoCcRwJYQKYwLwxjYDuMq6W2APGAK5wBG2KAfABRsrYQgA2AbtgA0MLgCtswKAEp8LGAG8AsACgYo7BKlsipAKIo0KNgCIAZiBAnp0gNyqAvndWqAhhARhgMM7S9QMcBgAExAAFQALDDAAczZpAC4yNEwcKn4QDGD5ZTUYKBQERVV1dVBIWC4IWl5YAldiVwxYRHJU7Hs8svAeXmwAOl4QOIADACVuGqgkgBIFKqmHEec8hxhgVyhgCJg2bCS-AGswEGIwWVzSmAwzXagEAAdsEDM92TwPmAByaBRomK+FxKV3WPT4AyGowAwq5aDEIrBmEZZgpsEsVlc1theDhrrc9tcKq4vM9bgYjECuqVyr0IcM2CMYXCETAkSAUCiBnBuBBXDE0ctOpjgQ5HC4VKFIv94rYgA) in the playground just to see how each case would be handled)

Anyway I just wondered whether this is a distinction without a difference and the "Host responded without port or error details." string should just be plainly in the rejection without being wrapped in an `Error` object, since in all these cases the caller is.... likely just going to print/log the message? Maybe you aren't really expecting the caller to do handle this rejection anyway?

Anyway I'm not sure I'm recommending anything change here but this did make me stop and think. Which I guess is a good thing to do on a Friday!

packages/teams-js/src/internal/communication.ts
```diff
@@ -52,6 +52,9 @@ class CommunicationPrivate {
   public static promiseCallbacks: {
     [id: number]: Function; // (args[]) => void
   } = {};
+  public static portCallbacks: {
+    [id: number]: Function; // (args[], port) => void
```
AE-MS, line 56: >Function; // (args[], port) => void [](http://example.com/codeflow?start=18&length=35)

Rather than declaring the value as type `Function` with the expectation in the comment (are the expected parameters reversed in the comment?), could this be declaration be more explicit about the expected function type, e.g.,

```
public static portCallbacks: {
  [id: number]: (port: MessagePort, args?: unknown[]) => void;
} = {};
```

packages/teams-js/src/internal/communication.ts
AE-MS, line 355: >undefined [](http://example.com/codeflow?start=73&length=9)

This callback accepts a type of `MessagePort | undefined` as its first parameter, but in the usage on line 765, it looks like the possible values being passed are `MessagePort | null`

(I think this could be fixed in a spirit similar to my earlier comment about the type used when declaring `portCallbacks` ?)

AE-MS (changes requested):
Thank you for considering my feedback!

packages/teams-js/src/internal/communication.ts
```diff
@@ -320,6 +324,45 @@ export function sendMessageToParentAsync<T>(actionName: string, args: any[] | un
...
+): Promise<MessagePort> {
+  if (!isFollowingApiVersionTagFormat(apiVersionTag)) {
+    throw Error(
+      `apiVersionTag: ${apiVersionTag} passed in doesn't follow the pattern starting with 'v' followed by digits, then underscore with words, please check.`,
+    );
+  }
+  return new Promise((resolve) => {
+    const request = sendMessageToParentHelper(apiVersionTag, actionName, args);
+    resolve(waitForPort(request.id));
+  });
```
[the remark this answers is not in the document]
kerrynf: That is a good question - the pattern above is the one followed by all the other functions in the communication.ts file, which is what I based this on:
image

According to Copilot they are probably equivalent, but if we want to change the pattern, I would recommend changing all the instances in the file, rather than just one?

packages/teams-js/src/internal/communication.ts
kerrynf: I'm definitely not clear on the correct 'promise rejection' pattern for use in the SDK, or what callers should expect in terms of error handling. :)

Here we have an example I looked at for throwing/rejecting with the error payload from args:
image

In other places in communication.ts, we are throwing 'normal' new Errors:
image

I admit I was disappointed to find that the 'error from host' didn't fit the 'proper' error format and needed special handling, but it contains more useful information generally that the locally thrown ones.

Thoughts?

packages/teams-js/src/internal/communication.ts
kerrynf: That is better than the example I followed above. :) Updated for explicit typing.

packages/teams-js/src/internal/communication.ts
kerrynf: That seems to work (args is always typed any[] elsewhere in this file, so you will still have more work to do :))

packages/teams-js/src/internal/communication.ts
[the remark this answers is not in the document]
kerrynf: Ok, as much as I like consistency, I've agreed to clear improvements on typing and other things that break with the patterns in the file. :) I'll update my function.

packages/teams-js/src/internal/communication.ts
kerrynf: Yep, this is proof that loosely typing the functions is a bad idea. Updated the typing and this instance to be undefined in all places.

packages/teams-js/src/internal/communication.ts
AE-MS: I don't really feel strongly one way or another about this and I don't know what kind of precedent @TrevorJoelHarris wants to set going forward so I'll just sign-off from my perspective and defer to him if he wants to do any follow-up here.

AE-MS (approved):
:shipit:

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: I definitely prefer option 1, but also accept that our codebase isn't currently very consistent about how errors are handled. I prefer wrapping the string in an error when possible rather than just rejecting with a plain string.

Long term, I'd like all of the errors to be combination code and message, so if throwing something like that here for the second case is possible that would be great. I won't hold up the PR for it though, given the sorry state we're in right now when it comes to error consistency. @kerrynf I will leave it up to you: if you don't mind throwing more consistent errors out of this function as part of this PR, that would be great! Every little bit helps.

TrevorJoelHarris (approved):

packages/teams-js/src/internal/communication.ts
kerrynf: Pasting my message image here, since I crashed the page earlier and it didn't save... TLDR: I think this PR should complete as is, but we should tackle error handling ASAP. :)

image

PR #2147, opened 2024-01-26
Add Loop domains to the list of valid origins

dotnetjunky:
I'm from the Microsoft Loop team. We're working on standing up Loop App as a MetaOS Hub. According to the onboarding documentation, we need to add our app domains to the list of valid origins.


TrevorJoelHarris:
@dotnetjunky I believe that you're in touch with people on my team already. Once you've been approved to be a host as Loop we can talk about which origins to add here. If you just need to unblock testing, origins can be added to the `app.initialize` call to let certain apps run inside of Loop.

PR #2151, opened 2024-01-31
Adding join meeting api for m365 calendar

aditya-darji:

## Description

Adding join meeting api for m365 calendar

### Main changes in the PR:

Adding join meeting api for m365 calendar

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


apps/teams-test-app/src/components/CalendarAPIs.tsx
```diff
@@ -37,12 +37,27 @@ const OpenCalendarItem = (): React.ReactElement =>
...
+  ApiWithTextInput<string>({
+    name: 'joinMeetingItem',
+    title: 'Join Meeting',
+    onClick: {
+      submit: async (input) => {
+        await calendar.joinMeeting(input);
+        return 'Completed';
+      },
+      validateInput: (x) => {
+        console.log(x);
```
RishabBindra, line 50: We shouldn't add console.log

packages/teams-js/src/public/calendar.ts
```diff
@@ -75,6 +75,27 @@ export namespace calendar {
...
+  /**
+   * Checks if the calendar capability is supported by the host
+   * @returns boolean to represent whether the calendar capability is supported
+   *
+   * @throws Error if {@linkcode app.initialize} has not successfully completed
+   */
+  export function joinMeeting(joinMeetingParams: JoinMeetingParams): Promise<void> {
+    return new Promise<void>((resolve) => {
+      ensureInitialized(runtime, FrameContexts.content);
+      if (!isSupported()) {
```
RishabBindra, line 87: It is probably worth understanding what this does, and how it works.
For our case, we only want to show joinMeeting is supported, but other hubs may not support joinMeeting.

packages/teams-js/src/public/calendar.ts
```diff
@@ -106,4 +127,55 @@ export namespace calendar {
...
+    subject: string;
+    source: string;
+    conversationId: string;
+    messageId: string;
+    rootMessageId: string;
+    isBroadcastMeeting?: boolean | undefined;
+    t2BroadcastParticipantRoleDetails?: IT2BroadcastParticipantRoleDetails | undefined;
+  }
+
+  export interface IT2BroadcastParticipantRoleDetails {
```
RishabBindra, line 142: I don't think all this is needed.
Remove all the broadcast related items.

packages/teams-js/src/public/calendar.ts
```diff
@@ -106,4 +127,55 @@ export namespace calendar {
...
+  }
+
+  export enum BroadcastParticipantRole {
+    Attendee = 'Attendee',
+    Contributor = 'Contributor',
+    None = 'None',
+    Producer = 'Producer',
+  }
+
+  export enum ClientType {
```
RishabBindra, line 170: How is clientType used?
For our case, wouldn't it be OutlookWeb only?
We should confirm this.

aditya-darji: Removed it. Was being used for broadcast role details.

apps/teams-test-app/src/components/CalendarAPIs.tsx
aditya-darji: Removed the console.log

packages/teams-js/src/public/calendar.ts
aditya-darji: Removed all the broadcast related details.

TrevorJoelHarris:
There are a few obvious issues with this PR (incorrect documentation on your new function, your new function should be in the `meeting.ts` file instead of the `calendar.ts file`), but you won't be able to get an in depth PR review until your change is passing the build.

After a quick glance, it looks like your build is failing very early because you haven't made a changefile. The process for making a changefile and the expectations for how it should be formatted are listed here: https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball

change/@microsoft-teams-js-16f5e9b8-9162-4db5-bdd1-bebbda9ab9d4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added joinMeeting api",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Added `meeting.joinMeeting` function",
```

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
     );
   }
 
+  /**
+   * This function is used to join a meeting
```
TrevorJoelHarris, line 930: More detail on what this actually does would be useful to developers. Does it open the meeting in a pop out? Does it navigate away from the application?

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+   * @param joinMeetingParams
```
TrevorJoelHarris, line 931: Please use `@link` in this comment to connect this param to the `JoinMeetingParams` interface. That will create a hyperlink from this parameter to its type in the dev docs.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+   */
+  export function joinMeeting(joinMeetingParams: JoinMeetingParams): Promise<string | object> {
```
TrevorJoelHarris, line 933: Please add an `@returns` section to your function comment detailing what this function returns and how developers should use that return value.

TrevorJoelHarris: Actually, why does this return anything at all other than `Promise<void>`? It looks like you're not calling a function that returns a value inside of the promise resolution

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+      );
+    });
+  }
+
+  /** Join meeting parameters */
+  export interface JoinMeetingParams {
+    /** The meeting url to join. */
+    skypeTeamsMeetingUrl: string;
+    /** The subject of the meeting. */
+    subject: string;
```
TrevorJoelHarris, line 951: Why does the user need to pass in a meeting's subject to join a meeting?

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris, line 933: Your function should have a full suite of unit tests added to `meeting.spec.ts`

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+    /** The conversation id of the meeting. */
```
TrevorJoelHarris, line 952: This comment seems wrong.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+          joinMeetingParams,
+        ),
+      );
+    });
+  }
+
+  /** Join meeting parameters */
+  export interface JoinMeetingParams {
+    /** The meeting url to join. */
+    skypeTeamsMeetingUrl: string;
```
TrevorJoelHarris, line 949: Given the name of this, I'm going to assume that this is a publicly documented property from somewhere like the graph. Can you please link to documentation for where this property comes from in the comment?

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+    /** The subject of the meeting. */
+    subject: string;
+    /** The conversation id of the meeting. */
+    source: EventActionSource;
+    /** The conversation id of the meeting. */
+    conversationId: string;
+    /** The message id of the meeting. */
+    messageId: string;
```
TrevorJoelHarris, line 957: What does it mean for a meeting to have a `messageId`? How would a developer retrieve the messageId of a meeting they wanted to join? Is there some existing public documentation I can read about how messageIds are tied to meetings?

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+  /** Join meeting parameters */
+  export interface JoinMeetingParams {
+    /** The meeting url to join. */
+    skypeTeamsMeetingUrl: string;
+    /** The subject of the meeting. */
+    subject: string;
+    /** The conversation id of the meeting. */
+    source: EventActionSource;
+    /** The conversation id of the meeting. */
+    conversationId: string;
```
TrevorJoelHarris, line 955: Why is a `conversationId` needed to join a meeting? Neither `conversationId` nor `messageId` were in the original list of CalendarEvent data from your proposal

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+    /** The message id of the meeting. */
+    messageId: string;
+    /** The root message id of the meeting. */
+    rootMessageId: string;
+  }
+
+  /** Event action source */
+  export type EventActionSource =
```
TrevorJoelHarris, line 963: This would be easier for developers to use as an `enum` (see the `MeetingReactionType` enum in this file for an example)

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+    /** The subject of the meeting. */
+    subject: string;
+    /** The conversation id of the meeting. */
+    source: EventActionSource;
+    /** The conversation id of the meeting. */
+    conversationId: string;
+    /** The message id of the meeting. */
+    messageId: string;
+    /** The root message id of the meeting. */
+    rootMessageId: string;
```
TrevorJoelHarris, line 959: What is the difference between a `rootMessageId` and a `messageId`? Are both actually needed for joining a meeting?

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris, line 949: `skypeTeamsMeetingUrl` should be a URL type to guarantee that developers are passing in a valid URL. You'll then have to take that URL and serialize it into a string before passing it to `sendAndHandleStatusAndReasonWithVersion`. For an example, see `marketplace.Item.imageURL` and the corresponding `serializeCartItems` function in `marketplaceUtils.ts`

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,47 @@ export namespace meeting {
...
+  }
+
+  /** Event action source */
+  export type EventActionSource =
+    | 'calendar_grid_context_menu'
```
TrevorJoelHarris, line 964: These are *all* very calendar specific, but I believe that the function you are adding is also available for 3P developers to call. Should there be at least one Source type that is not calendar specific for those developers? Something like `other`?

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris, line 955: Are all of the various ids here actually just strings, or would they be more correct to be Guids / UUIDs? Even if they have to be just strings, I assume you will want to do *some* validation in the function above to guarantee that they are non-null, non-empty, etc.

TrevorJoelHarris: If the ids should be uuid, see `marketplaceUtils.validateUuid` for an example on how to make sure the UUID is a legal format.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris, line 933: There should be new e2e tests for your function. Please add those to `meeting.json`. That file defines the e2e tests, and the test app that the e2e tests use has its existing meeting components defined in `MeetingAPIs.ts`.

In order for your e2e tests to work correctly you'll need to check in the other half of your changes to the host-sdk before these changes are checked in.

TrevorJoelHarris (changes requested):

change/@microsoft-teams-js-16f5e9b8-9162-4db5-bdd1-bebbda9ab9d4.json
TrevorJoelHarris: test

aditya-darji: Done the changes

packages/teams-js/src/public/meeting.ts
aditya-darji: @TrevorJoelHarris, Added the details of the function.

packages/teams-js/src/public/meeting.ts
aditya-darji: Added the @link to JoinMeetingParams

packages/teams-js/src/public/meeting.ts
aditya-darji: @TrevorJoelHarris, changed the return type to Promise<void>

packages/teams-js/src/public/meeting.ts
aditya-darji: We made the meeting subject optional.

packages/teams-js/src/public/meeting.ts
aditya-darji: Changed the type to URL and accommodated it accordingly.

packages/teams-js/src/public/meeting.ts
aditya-darji: Added another source as 'other' to EventActionSource.

packages/teams-js/src/public/meeting.ts
aditya-darji: @TrevorJoelHarris, added the public documentation in the comment and made a few changes in parameters accordingly.

packages/teams-js/src/public/meeting.ts
aditya-darji: Corrected the comment

packages/teams-js/src/public/meeting.ts
aditya-darji: Changed it to enum

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: What is the scenario where a developer passing in the subject would be helpful? What would that accomplish?

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -340,6 +340,40 @@
...
+      "inputValue": {
+        "joinWebUrl": "https://www.example.com/",
+        "subject": "subject",
+        "source": "Other",
+        "chatInfo": {
+          "messageId": "messageId",
+          "replyChainMessageId": "replyChainMessageId"
+        }
+      },
+      "expectedTestAppValue": "Error: Error: JoinUrl or chatInfo not passed"
```
KangxuanYe, line 376: @TrevorJoelHarris , I am not sure if this "Error: Error: ..." prompts are what we expect but the test-app-value does match the output.

TrevorJoelHarris: @aditya-darji I'm sure that we have other tests that look for something similar to this, but we shouldn't and I'd like to hold new tests to a higher quality standard. Why does the string "Error" show up twice here?

packages/teams-js/src/public/calendar.ts
TrevorJoelHarris: I see that you don't have an isSupported function in the new location of this code (meeting.ts) which is good. Most of our namespaces have isSupported in them, but meeting doesn't yet because it is very complicated. Your approach is correct for where the function is located now.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,124 @@ export namespace meeting {
...
+   * More details can be found at:
+   * [Online Meeting Base - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0)
+   */
+  export interface JoinMeetingParams {
+    /** The join URL of the online meeting. */
+    joinWebUrl: URL | string;
+    /** The chat information associated with this online meeting. {@link ChatInfo}*/
+    chatInfo: ChatInfo;
+    /** The subject of the meeting. */
+    subject?: string;
```
TrevorJoelHarris, line 1000: You didn't answer my question that I left earlier on this, so I'll ask it again here: What is the scenario where a developer would need to specify a subject of a meeting in order to join it? The meeting should already have a subject. We should never ask developers to pass up information that isn't needed to perform functionality.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: You didn't answer here so I just rewrote the question where this code lives in the PR. Closing this thread in favor of the new one.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: You didn't answer my question, but it looks like you removed `rootMessageId`. I assume that is because it was not actually needed.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: In a side chat, you told me that these ids are definitly not UUIDs, and can take formats like `19:-D5VrQ35l94Drye2JbdTaiHiizqBAb4o_YnHlYtq5XM1@thread.tacv2`.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +926,123 @@ export namespace meeting {
...
+    joinMeetingParams.joinWebUrl = new URL(joinMeetingParams.joinWebUrl);
+    const serializedJoinMeetingParams = {
+      joinWebUrl: joinMeetingParams.joinWebUrl.href,
+      source: joinMeetingParams.source || EventActionSource.Other,
+      chatInfo: {
+        threadId,
+        messageId: messageId || '0',
+        replyChainMessageId: replyChainMessageId || '0',
+      },
+    };
```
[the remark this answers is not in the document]
TrevorJoelHarris: They added this field.

apps/teams-test-app/e2e-test-data/meeting.json
RishabBindra: @TrevorJoelHarris @KangxuanYe 
If you notice [MeetingAPIs.tsx](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2151/files#diff-112d449943532744b73c00724f5cc14ad68b6b17088d5f3e2efe22ad146ee3e9:~:text=throw%20new%20Error(%27JoinUrl%20or%20chatInfo%20not%20passed%27)%3B) in this PR, you will notice when we try to validate input, in case the input is incorrect, we `throw new Error('JoinUrl or chatInfo not passed');` in line 424.
This is only used in the e2e test.

When we checked the call stack, we noticed that the calling method has a catch block, where it will append the string "Error: " with the error it caught. Line 70 in [this file](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/apps/teams-test-app/src/components/utils/ApiWithTextInput.tsx).
Hence we see two errors.

packages/teams-js/src/public/meeting.ts
RishabBindra: @TrevorJoelHarris we noticed that presently the meeting subject is being passed to start a meeting. However, when we tried to start a meeting without the subject too, it seemed to work.
There may be some specific scenario or corner case where this is required that we are not aware of, as this is handled by the calling and meeting team. However, presently the join meeting command demands it.

It is possible that subject is used for a pre warming scenario, or landing page scenario that we are unaware of. In case you would like me to find out more about this, I can do that.

TrevorJoelHarris: @RishabBindra looked into this more, and here's what this value is being used for: 

"`subject` is needed to populate the title in the pre join screen. When we join a meeting, we do not immediately have all the information related to the meeting. Hence when we land on the pre join screen, the title is needed for the new window. If the title is not passed, calling has a fallback to fetch it and populate it, however it is more latent."

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
...
+  /**
+   * Contains information associated with Microsoft Teams meetings.
+   * More details can be found at:
+   * [Chat Info - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/chatinfo?view=graph-rest-1.0)
+   */
+  export interface ChatInfo {
+    /** The unique identifier for a thread in Microsoft Teams. */
+    threadId: string;
+    /** The unique identifier of a message in a Microsoft Teams channel. */
+    messageId?: string;
```
TrevorJoelHarris, line 1013: In what circumstances would a developer set the `messageId` and `replyChainMessageId`? In a side chat you told me this:

"The threadId (and other information) is needed to open the right side chat during a meeting."

`threadid` I get, it makes sense that you would have to specify what chat thread to open in the sidepane. `messageId` and `replyChainMessageId` I don't understand. Both of them seem like they are specifically used for identifying single messages in that thread, which doesn't seem like something a developer would need to do in order to open the sidepane.

Does that sound correct?

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
...
+    }
+    return true;
+  }
+
+  /**
+   * Contains information associated with parameters required for joining the Microsoft Teams meetings.
+   * More details can be found at:
+   * [Online Meeting Base - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0)
+   */
+  export interface JoinMeetingParams {
```
TrevorJoelHarris, line 993: Based on what we've learned about `subject` and the comment I left in `ChatInfo`, here's what I would suggest for a new structure for this interface:

```typescript
export interface JoinMeetingParams {
  /** The join URL of the online meeting. */
  joinWebUrl: URL | string;
  /** The unique identifier for a thread in Microsoft Teams. Should be the thread id of the chat associated with the meeting being joined*/
  threadId: string;
  /** The source of the join button click. {@link EventActionSource} */
  source: EventActionSource;
  /** The subject to show on the pre-join screen of the meeting. If this value is omitted the meeting's actual subject will be displayed after a brief delay. */
  customPreJoinSubject?: string;
}
```

I've removed the ChatInfo structure entirely because I assume threadId is the only meaningful parameter a developer will specify from it. I've also changed the name of `subject` and added a more detailed comment explaining what it's for and what happens if it isn't specified. I've also made `source` a required parameter (you listed it as optional). If it is *truly* optional, then the comment should explain what happens if the source is not specified.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1372,6 +1372,82 @@ describe('meeting', () => {
...
+          ...mockJoinMeetingParams,
+          chatInfo: {
+            ...chatInfoParams,
+            threadId: '',
+          },
+        });
+        await expect(response).rejects.toEqual('Invalid joinMeetingParams');
+      });
+
+      it('should reject if joinWebUrl is empty', async () => {
```
TrevorJoelHarris, line 1432: You should have a unit test that verifies this is rejected if joinWebUrl has a value of 'this is not a valid url' since you have code that will throw an exception in that case.

apps/teams-test-app/e2e-test-data/meeting.json
TrevorJoelHarris: Thanks @RishabBindra ! I'll open a bug for someone to look into fixing this for everyone. Appreciate the due diligence.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
...
+   */
+  function validateJoinMeetingParams(joinMeetingParams: JoinMeetingParams): boolean {
+    if (
+      !joinMeetingParams?.chatInfo?.threadId ||
+      !joinMeetingParams?.joinWebUrl ||
+      joinMeetingParams.chatInfo.threadId === ''
+    ) {
+      return false;
+    }
+    if (typeof joinMeetingParams.joinWebUrl === 'string' && joinMeetingParams.joinWebUrl === '') {
```
TrevorJoelHarris, line 978: Would it make sense to also check that the string is a valid URL here?

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
     );
   }
 
+  /**
+   * This function is used to join a meeting.
+   * This opens a meeting in a new window for the desktop app.
+   * In case of a web app, it will close the current app and open the meeting in the same tab.
```
AE-MS, line 931: What is the expected behavior on mobile platforms?

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
...
+  }
+
+  /**
+   * Contains information associated with parameters required for joining the Microsoft Teams meetings.
+   * More details can be found at:
+   * [Online Meeting Base - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0)
+   */
+  export interface JoinMeetingParams {
+    /** The join URL of the online meeting. */
+    joinWebUrl: URL | string;
```
AE-MS, line 995: >string [](http://example.com/codeflow?start=22&length=6)

If this should always be a URL then consider updating the type to only accept `URL` and not also `string` to make it clear to the caller what the expectation is and prevent any issues with accepting invalid URLs via a a string.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
...
+    /** The chat information associated with this online meeting. {@link ChatInfo}*/
+    chatInfo: ChatInfo;
+    /** The subject of the meeting. */
+    subject?: string;
```
AE-MS, line 999: Based on the other comment you wrote, I recommend updating the property name and description. Calling this `subject` is likely to make a caller think they need to find the right subject for the meeting before populating this (or worse, given that it's now optional, just never populate it at all because it's too much work to find the subject, at which point why even allow this to be passed).

Consider calling it something like `temporaryMeetingWindowTitle`

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/meeting.ts
aditya-darji: meeting.joinMeeting is not supported for mobile platform.

packages/teams-js/src/public/meeting.ts
aditya-darji: @TrevorJoelHarris, we are converting the string to a url type in line 942:

> joinMeetingParams.joinWebUrl = new URL(joinMeetingParams.joinWebUrl); 

This essentially removes the possibility of string being an invalid url.

packages/teams-js/src/public/meeting.ts
aditya-darji: @TrevorJoelHarris,
1. In chatInfo, the replyChainMessageId and messageId are used for the channel meetings and hence are important for joining/opening chat for channel meetings.
2. Will use your suggestion for replacing subject with customPreJoinSubject.
3. We had made source as an optional parameter and if nothing was passed in we assume it to be 'other'. Hence there's no need for source to be optional.

packages/teams-js/src/public/meeting.ts
aditya-darji: @AE-MS, we have kept the type as URL | string and then we are checking if it is a valid string on line 942:
> joinMeetingParams.joinWebUrl = new URL(joinMeetingParams.joinWebUrl);
This is done to provide more functionality.

packages/teams-js/src/public/meeting.ts
aditya-darji: @AE-MS, we have been following the public documentation: https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0, on @TrevorJoelHarris suggestions and found that subject is the field name which we would have to go with. Hence the name and description.

packages/teams-js/src/public/meeting.ts
aditya-darji: @TrevorJoelHarris, developer will have to pass messageId and replyChainMessageId in case of a channel meeting. In case of channel meetings, messageId and replyChannelMessageId are used to find the post in a channel where this meeting resides.

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,124 @@ export namespace meeting {
...
+   * This function is used to join a meeting.
+   * This opens a meeting in a new window for the desktop app.
+   * In case of a web app, it will close the current app and open the meeting in the same tab.
+   * @param joinMeetingParams The parameters for joining the meeting. {@link JoinMeetingParams}
+   * @throws error if the meeting join fails, the promise will reject to an object with the error message.
+   */
+
+  export function joinMeeting(joinMeetingParams: JoinMeetingParams): Promise<void> {
+    if (!validateJoinMeetingParams(joinMeetingParams)) {
+      return Promise.reject('Invalid joinMeetingParams');
```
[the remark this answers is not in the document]
aditya-darji: Done

packages/teams-js/test/public/meeting.spec.ts
aditya-darji: Added a UT for this.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: A function called `validateJoinMeetingParams` should validate all of the parameters correctness, including that the joinUrl is a valid URL. If someone adds another function to this file that uses this helper function, they should be able to rely on this function validating *all* parameter correctness.

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: In response to 2, you can definitely change the name of subject. There's no technical reason that the names have to match, and in this case the subject has important extra context. The developer doesn't have to pass in the actual value of the meeting's subject, they can pass in anything they want. Right?

TrevorJoelHarris: In response to 3: okay, that's your team's call. If you want to leave it optional with a default of other you can, just keep in mind that once you check this in and people start using it you can't change your mind and make the parameter required later.

change/@microsoft-teams-js-16f5e9b8-9162-4db5-bdd1-bebbda9ab9d4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `meeting.joinPrivateMeeting` and `meeting.joinChannelMeeting` function",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Added `meeting.joinPrivateMeeting` and `meeting.joinChannelMeeting`",
```

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,173 @@ export namespace meeting {
...
+   * More details regarding parameters can be found at:
+   * [Online Meeting Base - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0)
+   * [threadId - Chat Info - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/chatinfo?view=graph-rest-1.0)
+   */
+  export interface JoinPrivateMeetingParams {
+    /** The join URL of the online meeting. */
+    joinWebUrl: URL | string;
+    /** The unique identifier for a thread in Microsoft Teams. */
+    threadId: string;
+    /** The subject of the meeting. */
```
TrevorJoelHarris, line 1048: All optional parameters should have comments detailing what happens if the parameter is omitted

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,173 @@ export namespace meeting {
...
+    subject?: string;
+    /** The source of the join button click. {@link EventActionSource} */
```
TrevorJoelHarris, line 1050: All optional parameters should have comments detailing what happens if the parameter is omitted

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: @aditya-darji fixed this

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,173 @@ export namespace meeting {
...
+    }
+
+    joinPrivateMeetingParams.joinWebUrl = new URL(joinPrivateMeetingParams.joinWebUrl);
+
+    const serializedJoinPrivateMeetingParams = {
+      joinWebUrl: joinPrivateMeetingParams.joinWebUrl.href,
+      subject: joinPrivateMeetingParams.subject,
+      source: joinPrivateMeetingParams.source || EventActionSource.Other,
+      threadId: joinPrivateMeetingParams.threadId,
+      messageId: '0',
```
TrevorJoelHarris, line 947: Could you put a comment inside of this function explaining why it's okay to set `messageId` and `replyChainMessageId` to 0? (It's because those values aren't used when joining private meetings, and I assume 0 is sort of a default "ignore this value" value in the Teams backend?)

packages/teams-js/src/public/meeting.ts
TrevorJoelHarris: Only open issue is 2, going to resolve this and open a new item tracking that

packages/teams-js/src/public/meeting.ts
```diff
@@ -925,6 +925,173 @@ export namespace meeting {
...
+   * [Online Meeting Base - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0)
+   * [threadId - Chat Info - Microsoft Graph v1.0](https://learn.microsoft.com/en-us/graph/api/resources/chatinfo?view=graph-rest-1.0)
+   */
+  export interface JoinPrivateMeetingParams {
+    /** The join URL of the online meeting. */
+    joinWebUrl: URL | string;
+    /** The unique identifier for a thread in Microsoft Teams. */
+    threadId: string;
+    /** The subject of the meeting. */
+    subject?: string;
```
TrevorJoelHarris, line 1049: I believe that @RishabBindra is still investigating whether or not `subject` is something that devs should be allowed to pass in. Please respond to this with your findings when you have them.

packages/teams-js/src/public/meeting.ts
aditya-darji: Removed subject, messageId, replyChainMessageId and threadId after consulting CMD team.

packages/teams-js/src/public/meeting.ts
aditya-darji: Removed all other optional parameters but source.

packages/teams-js/src/public/meeting.ts
aditya-darji: Hi @TrevorJoelHarris, we discussed things with CMD team and got to conclusion that we would only take joinWebUrl and source as parameters for meeting.joinMeeting.

packages/teams-js/src/public/meeting.ts
aditya-darji: Added the comment.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1522,17 +1403,17 @@ describe('meeting', () => {
...
-          meeting.joinPrivateMeeting({
-            ...mockJoinPrivateMeetingParams,
+          meeting.joinMeeting({
+            ...mockjoinMeetingParams,
             joinWebUrl: 'this is incorrect url in string format',
           }),
         ).toThrowError('this is incorrect url in string format');
       });
 
       it(`should successfully joinMeeting`, async () => {
```
TrevorJoelHarris, line 1413: This unit test looks like a duplicate of the unit test on line 1394 (`https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetingbase?view=graph-rest-1.0`)

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1372,6 +1372,53 @@ describe('meeting', () => {
       });
     });
 
+    describe('joinMeeting', () => {
```
TrevorJoelHarris, line 1375: You are missing some unit tests for your code, specifically it doesn't look like you are testing that the right parameters are being sent up to the host sdk in the message and it doesn't look like you are having the unit tests respond to the message (which would resolve the Promise), but are instead letting the unit test timeout and then resolve the promise that way (which isn't actually testing your code correctly).

[This file](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/packages/teams-js/test/public/search.spec.ts) is a good example of how unit tests should be put together. `search.CloseSearch` is similar to the function you are adding in that the return type is `Promise<void>`.

You should have unit tests for your function in both the 'framed' and 'frameless` scenarios. If you look in `meeting.spec.ts` you can see that there are sections of unit tests for each of these already (look for `describe('framed', ()` and `describe('frameless', ()`

For both framed and frameless tests, you can see the message being passed out of TeamsJS and up to the host by writing code in your test like this:

```typescript
const joinMeetingPromise = meeting.joinMeeting(args);
const joinMeetingMessage = utils.findMessageByFunc('meeting.joinMeeting');
expect(joinMeetingMessage).not.toBeNull();
```

`joinMeetingMessage.args` will have whatever parameters are passed up to the host sdk in it, so you can verify that the right values are there (for example, source should be Other when no source is passed into the function).

For *framed* tests, you can respond to the message using code like this:

```typescript
utils.respondToMessage(joinMeetingMessage, args); // args is whatever arguments the host sdk should pass back to respond to the promise and get it to resolve
await joinMeetingPromise;
```

This will make sure that the promise resolves correctly in the framed state. For frameless tests, instead do this:

```typescript
utils.respondToFramelessMessage({
  data: {
    id: joinMeetingMessage?.id,
    args: [], // whatever args are passed down from the host sdk go here
  },
} as DOMMessageEvent);
await joinMeetingPromise;
```

Let me know if that gives you any trouble

TrevorJoelHarris (changes requested):
Please update your unit tests, otherwise this looks okay to me

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -1372,6 +1476,155 @@ describe('meeting', () => {
...
+      const dataError = 'Something went wrong...';
+
+      const mockjoinMeetingParams: meeting.JoinMeetingParams = {
+        joinWebUrl: new URL('https://example.com'),
+        source: meeting.EventActionSource.Other,
```
TrevorJoelHarris, line 1484: These tests look like a big improvement to me, thanks! The only tests I see missing (both framed and frameless) is that you never validate passing a source _except_ for `meeting.EventActionSource.Other`. That will make it possible for bugs to creep in in the future because `Other` is also the default value for `source` in your code. 2 options here:

1) You make the default value of Source here in your mock something than Other (for both framed and frameless tests). Then your `should successfully send the joinMeeting message` tests should test this case.

2) You just add another test that sets the source to a non default value.

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -340,6 +340,29 @@
...
+      "title": "joinMeeting API Call - Success",
+      "type": "callResponse",
+      "version": ">2.20.0",
+      "boxSelector": "#box_joinMeeting",
+      "inputValue": {
+        "joinWebUrl": "https://www.example.com/",
+        "source": "Other"
+      },
+      "expectedAlertValue": "joinMeeting called with joinWebUrl: \"https://www.example.com/\"",
+      "expectedTestAppValue": "joinMeeting called with joinWebUrl: \"https://www.example.com/\""
```
KangxuanYe, line 354: actual input: `joinMeeting called with joinWebUrl: https://www.example.com/`, missing double quotes for URL.

KangxuanYe: Assert failure from Cypress:

Timed out retrying after 30000ms: expected '<span#text_joinMeeting>' to have text 'joinMeeting called with joinWebUrl: "https://www.example.com/"', but the text was 'joinMeeting called with joinWebUrl: https://www.example.com/'

KangxuanYe: My hypothesis is that, the differences are on `expectedTestAppvalue`. It means that maybe somewhere in your test app works improperly when displaying strings.

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -340,6 +340,29 @@
...
+    },
+    {
+      "title": "joinMeeting API Call - Error",
+      "type": "callResponse",
+      "version": ">2.20.0",
+      "boxSelector": "#box_joinMeeting",
+      "inputValue": {
+        "joinWebUrl": "https://www.example.com/",
+        "source": "Other"
```
KangxuanYe, line 363: Could you please explain what's the difference between successful case and this error case? Why it should show the alert that joinWebUrl or threadId is not passed?

KangxuanYe (changes requested):
Left comments on changes need to be edited and error cases input.

apps/teams-test-app/e2e-test-data/meeting.json
aditya-darji: Oh, okay! made the changes accordingly. Thanks!

apps/teams-test-app/e2e-test-data/meeting.json
aditya-darji: This test required a few changes after removal of threadId from parameters list. Made those, now if joinWebUrl is not passed (which is a required field), it should throw an error.

apps/teams-test-app/src/components/MeetingAPIs.tsx
```diff
@@ -414,6 +414,28 @@ const UpdateMicState = (): React.ReactElement =>
...
+    onClick: {
+      validateInput: (input) => {
+        if (!input?.joinWebUrl) {
+          throw new Error('joinWebUrl not passed');
+        }
+      },
+      submit: async (input, setResult) => {
+        meeting.joinMeeting(input);
+        setResult('joinMeeting() succeeded');
+        return `joinMeeting called with joinWebUrl: ${input.joinWebUrl}`;
```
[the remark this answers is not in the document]
aditya-darji: Thanks! Made the changes accordingly.

packages/teams-js/test/public/meeting.spec.ts
aditya-darji: Made the changes in the e2e tests to check for different source values.

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

packages/teams-js/src/public/meeting.ts
AE-MS: If URLs are the only type of data you expect, please type it appropriately to indicate clearly to the consumer your expectations.

packages/teams-js/src/public/meeting.ts
AE-MS: That seems like important information to document here, rather than leaving developers guessing for the behavior on other platforms.

TrevorJoelHarris:
@AE-MS finished this work in a [different PR](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2203), so closing this one.

PR #2154, opened 2024-02-01
Teams mobile doc improvements

JoshuaPartlow:
## Description

Documentation update to add links for Teams mobile clarifications

### Main changes in the PR:

1. Adds links to clarifying articles to API documentation comments.
2. Corrects some minor formatting and linking issues.

## Validation

n/a

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/src/public/app.ts
```diff
@@ -814,6 +814,10 @@ export namespace app {
 
   /**
    * Notifies the frame that app has loaded and to hide the loading indicator if one is shown.
+   * 
+   * @remarks
+   * For information about mobile behavior, see the [Content pages](https://learn.microsoft.com/microsoftteams/platform/tabs/how-to/create-tab-pages/content-page?tabs=teamsjs-v2#show-a-native-loading-indicator) article.
```
AE-MS, line 819: >For information about mobile behavior [](http://example.com/codeflow?start=5&length=37)

What should we expect developers to see when they go to that URL? Here is where I think I would be confused: is that page telling me that... I shouldn't call methods like `notifyAppLoaded()` when I'm on mobile?

I think the answer is "No, your app should always call these methods and you don't need to write special code, but just know that the loading indicator behavior may be different on mobile."  Is that correct? If so it's not totally clear from these comments or that doc page.

packages/teams-js/src/public/app.ts
```diff
@@ -814,6 +814,10 @@ export namespace app {
 
   /**
    * Notifies the frame that app has loaded and to hide the loading indicator if one is shown.
+   * 
+   * @remarks
```
AE-MS, line 818: >@remarks [](http://example.com/codeflow?start=5&length=8)

Should these remarks be specifically on the `notifyAppLoaded` method like this? It seems like the documentation is generally about how loading indicator is handled and not specific to this one function.

packages/teams-js/src/public/barCode.ts
```diff
@@ -63,7 +63,12 @@ export namespace barCode {
   /**
    * Checks whether or not media has user permission
    *
-   * @returns true if the user has granted the app permission to media information, false otherwise
+   * @remarks 
+   * For additional information about deivce permissions, see the Learn articles:
+   *  - [Integrate QR or barcode scanner capability](https://learn.microsoft.com/microsoftteams/platform/concepts/device-capabilities/qr-barcode-scanner-capability)
```
AE-MS, line 68: >Integrate QR or barcode scanner capability [](http://example.com/codeflow?start=9&length=42)

Following this link takes me to a page which provides great information about barcode scanning -- but that page then links to a function in the older (1.x) version of teamsjs and a different capability than this one (it links to `media`) -- is it weird that this capability and function links to docs that don't mention this function at all?

packages/teams-js/src/public/barCode.ts
```diff
@@ -63,7 +63,12 @@ export namespace barCode {
   /**
    * Checks whether or not media has user permission
    *
-   * @returns true if the user has granted the app permission to media information, false otherwise
+   * @remarks 
+   * For additional information about deivce permissions, see the Learn articles:
```
AE-MS, line 67: Would it be more appropriate if this documentation was on the `barCode` namespace rather than each function in it? Since the documentation is an overview of the functionality and not specific to each function?

packages/teams-js/src/public/media.ts
```diff
@@ -756,6 +756,9 @@ export namespace media {
   /**
    * Select an attachment using camera/gallery
    *
+   * @remarks
```
AE-MS, line 759: >remarks [](http://example.com/codeflow?start=6&length=7)

Would it be more appropriate to put this on the `media` namespace rather than this one function? Presumably this applies to other functions in this file?

AE-MS (changes requested):
:clock1:

change/@microsoft-teams-js-795fcb3d-57f7-436b-b7e2-b3d3c2df6bad.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
TrevorJoelHarris, line 2: This should be type "none" since the docs only changes won't affect the behavior of the library at all.

PR #2157, opened 2024-02-02
[WIP][Task 8700131] Implementing pages.responseButton sub-capabilities

MengyiGong:

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+       *
+       * @beta
+       */
+      responseId: string;
+      /**
+       * Id of File on OneDrive
+       *
+       * @beta
+       */
+      oneDriveFileId: string;
```
TrevorJoelHarris, line 1130: oneDriveFileId, messageId, and conversationId are probably values that should be passed into the app as part of the ActionInfo object in the context that is defined in `interfaces.ts`:

```typescript
/**
 * Information common to all actions
 *
 * @beta
 */
export interface ActionInfo {
  /**
   * Maps to the action id supplied inside the manifest
   */
  actionId: string;
  /**
   * Array of corresponding action objects
   */
  actionObjects: BaseActionObject<ActionObjectType>[];
}
```

In ActionInfo, the `actionObjects` array contains one or more items than an action has been taken on. In this case, there will be at least one item representing the file that the action has been taken on:

```typescript
/**
 * Stores information needed to represent M365 Content stored
 * in OneDrive or Sharepoint
 *
 * @beta
 */
export interface M365ContentAction extends BaseActionObject<ActionObjectType.M365Content> {
  /**
   * Only office content IDs are passed to the app. Apps should use these ids
   * to query the Microsoft graph for more details.
   */
  itemId: string;
  /** Represents an optional secondary identifier for an action in a Microsoft 365 content item. */
  secondaryId?: SecondaryId;
}
```

In this case, `M365ContentAction.itemId` should be the oneDriveId of the file, and the `secondaryId` is an optional second id that may be needed in order to retrieve that file from the MS Graph (for example, the host can pass down a `driveId` if one is interact with the file in the graph). In general, this item being an `M365ContentAction` is also used as a signal to app developers that they can use the information specified here to download the file from a particular MS Graph Endpoint, detailed here: https://learn.microsoft.com/en-us/graph/api/resources/onedrive?view=graph-rest-1.0

If I remember correctly, the host will be taking this file and moving it to OneDrive before opening the app, so the host passing in an M365ContentAction still feels correct. However, based on what you've specified below, the host will also need to pass in a messageId and conversationId in the M365ContentAction so that the developer can pass them back to the host in order to initialize the ResponseButton.

If I have all of that right, M365ContentAction should probably have some sort of new optional property named something like 'originalSource` or `originalOwner` that contains that information. If you had that, then `ResponseInfo` could be simplified a lot to look like this:

```typescript
/**
     * The required information that the app needs to let host apps know.
     *
     * @beta
     */
    export interface ResponseInfo {
      /**
       * The unique ID helps the host and apps identify which button it refers to.
       *
       * @beta
       */
      responseId: string;
      /**
       * This should be one of the items passed to the app in the ActionInfo. Please pass the item back to initialize the
       * ResponseButton for the corresponding item that an action was taken on. Note just for maggie: this text is not final, more an 
       * explanation for you. You may want to change it to make it more meaningful for external developers
       * @beta
       */
      item: BaseActionObject<ActionObjectType>;
    }
```

packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+     * @param params - Parameters sent to host apps contain the required information.
+     *
+     * @beta
+     */
+    export function showResponseButton(params: ResponseInfo): Promise<void> {
+      ensureInitialized(runtime, FrameContexts.content);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+      if (!params) {
```
TrevorJoelHarris, line 1157: would also be good to make sure sure each param is not null, along with any other validation that we can do (no empty strings, etc.)

packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+       * The conversation where the action was invoked from
+       *
+       * @beta
+       */
+      conversionId: string;
+    }
+
+    /**
+     * Notify host apps to show the response button.
+     * @param params - Parameters sent to host apps contain the required information.
```
TrevorJoelHarris, line 1148: Use @link to connect this to ResponseInfo above. That will make a hyperlink that developers can follow to read more details

packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+     *
```
TrevorJoelHarris, line 1149: This should also have a `@returns` section in the comment talking about how it returns a `Promise<void>` that will resolve when the ResponseButton is visible in the host.

packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+      return sendAndHandleSdkErrorWithVersion(
+        getApiVersionTag(pagesTelemetryVersionNumber, ApiName.Pages_ResponseButton_ShowResponseButton),
+        'pages.responseButton.showResponseButton',
+        params,
+      );
+    }
+
+    /**
+     * Notify host apps to hide the response button
+     *
```
TrevorJoelHarris, line 1169: same comment about `@returns` that I made above

packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+      );
+    }
+
+    /**
+     * Registers a handler for when the button in the host apps is clicked by the user
+     * @param appEventHandler - The handler to invoke when the user clicks on the button in host apps.
+     *
+     * @beta
+     */
+    export function responseButtonEventHandler(appEventHandler: handlerFunctionType): void {
```
TrevorJoelHarris, line 1189: Is there no data the host has to pass down to the app developer when this handler is triggered? For example, if the response button is a button with a dropdown that presents multiple options (forward, replyAll, etc.) do you need to make an enum or something similar available here?

packages/teams-js/src/public/pages.ts
```diff
@@ -14,7 +14,14 @@ import { isNullOrUndefined } from '../internal/typeCheckUtilities';
 import { createTeamsAppLink } from '../internal/utils';
 import { app } from './app';
 import { errorNotSupportedOnPlatform, FrameContexts } from './constants';
-import { FrameInfo, ShareDeepLinkParameters, TabInformation, TabInstance, TabInstanceParameters } from './interfaces';
+import {
```
TrevorJoelHarris, line 17: Your new capability should also have e2e tests written for it. The pages e2e tests are in `pages.json`

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/pages.ts
MengyiGong: Yes, will do, and this is because those compat tests should add after the web hub-sdk messageHandler or those tests will fail.

packages/teams-js/src/public/pages.ts
```diff
@@ -1096,4 +1103,117 @@ export namespace pages {
...
+     *
+     * @beta
+     */
+    export interface ResponseInfo {
+      /**
+       * The unique ID helps the host and apps identify which button it refers to.
+       *
+       * @beta
+       */
+      responseId: string;
```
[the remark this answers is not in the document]
MengyiGong: Thanks for your suggestion! I updated with enum type and also make sure it accepts the string type if the response id has not yet been added to the enum to provide the easy extension for hosts and app developers.

packages/teams-js/src/public/pages.ts
MengyiGong: Thanks for your suggestion, those are all really helpful. 
After talking with PMs @irenewang101 and @Deepthi, we've decided to add originalSource under M365ContentAction. This allows every actionObject to have its own originalSource, making it easier to extend for our hosts.

packages/teams-js/src/public/pages.ts
irenewang101: I remember from earlier discussion(like a year ago) we did not to want to maintain a list of enums (like: reply, replyall, forward), due to the reason that if hubs decide to add more dropdown options, we will need to add to this list and needs SDK changes. So, it is hard to scale and maintain in the long term. 

If we want to allow developers know which button got clicked, and able to handle different logic based on which button got clicked. That sounds good to me. My only ask is that developer don't need to register a handler for each of the dropdown item if they share the same handler logic.

TrevorJoelHarris:
This work has been postponed, so closing this PR for now

PR #2158, opened 2024-02-03
Add default test app input to make it easier to use manually

TrevorJoelHarris:
## Description

- The main change is that I have started setting `defaultInput` on every `ApiWithTextInput` component in the test app and populating it with as-reasonable-as-I-can-make-it input. This means that generally if you want to use the teams test app in orange now, you can just open it and push the buttons and all the right stuff will happen. If you want to use the test app in "real" hosts you may need to make changes to the defaults, but at least you'll be able to see the correct structure.
- I also noticed that many of our tests don't seem to be checking alerts correctly, so I've changed the `expectedAlertValue` fields to match the actual alert box output. Someone else will be making the changes to the framework to actually verify the alert contents correctly in cypress

### Main changes in the PR:

1. See above

### Validation performed:

1. Used all of the test app by hand, and ran one e2e test locally to make sure the default text wasn't interfering with the e2e tests.

### Unit Tests added:

No

### End-to-end tests added:

No but boy howdy did I edit them

### Change file added:

None needed, test only


TrevorJoelHarris:
Here is an example of what the test app looks like now when run in orange: 
image
I didn't type any of those strings, they are there on boot now.

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -172,7 +172,7 @@
       "inputValue": {
         "appContentUrl": "www.someUrl.com"
       },
-      "expectedAlertValue": "shareAppContentToStage called",
+      "expectedAlertValue": "shareAppContentToStage called with [object Object] + [object Object]",
```
erinha, line 175: This makes me sad and happy because it highlights that we probably have some parsing to do to actually validate the results (i.e., stringify them).

TrevorJoelHarris: Yeah, there are definitely some things we should clean up in here but for now I focused on making things work as-is

maglims (commented):
From the test app UI page, are we able to see what the default value? something like: can I hover on the button to see its value?

TrevorJoelHarris:
> From the test app UI page, are we able to see what the default value? something like: can I hover on the button to see its value?

Currently you can only see the value by pressing the "Default" button, which will put it in the text field.

pnpm-lock.yaml
```diff
@@ -1,9 +1,5 @@
 lockfileVersion: '6.0'
 
-settings:
```
erinha, line 3: Do you know why this file changed?

erinha (approved):
Looks good, thank you again for doing this, it will help make debugging/local testing much easier!

Also just have one question about the lock file changes, not sure what caused any changes there based on the changes in this PR.

pnpm-lock.yaml
TrevorJoelHarris: I tried to add sass/scss support at one point, and then changed my mind to not overcomplicate this change. Re-running pnpm install again after removing the sass stuff caused it to bump this file

apps/teams-test-app/src/components/privateApis/ChatAPIs.tsx
```diff
@@ -50,6 +51,10 @@ const OpenChat = (): React.ReactElement =>
         return 'chat.openChat() was called';
       },
     },
+    defaultInput: JSON.stringify({
+      user: ['testUpn'],
```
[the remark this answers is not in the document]
TrevorJoelHarris: Okay, I'll update the default to be not an array. I'm going to leave the actual tests alone though. Those tests are expecting hosts to say that they saw arrays in their alerts, so just updating the tests without updating each platform's host code will cause test breaks. I will leave that to someone else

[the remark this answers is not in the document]
KangxuanYe: @TrevorJoelHarris That's correct for changing only default value. 

I am suspecting this test input for single-user chat API is different based on Teams-Js versions and Hub SDK versions now.

apps/teams-test-app/e2e-test-data/chat.json
```diff
@@ -48,7 +48,7 @@
       "version": ">2.2.0 && <=2.17.0",
       "boxSelector": "#box_openChat2",
       "inputValue": {
-        "user": ["testUpn"],
+        "user": "testUpn",
```
maglims, line 51: Instead of here, it looks the changes should be made in line 63 and 66?

where the input is "testUpn", and expectedAlertValue is [[\"testUpn\"]] instead of [\"testUpn\"]?

TrevorJoelHarris:
```suggestion
        "user": ["testUpn"],
```

KangxuanYe (approved):

maglims (approved):

PR #2170, opened 2024-02-09
Update meeting.json to account for different alert formats in host SDK versions

AE-MS:
## Description

When the meeting tests are run against older host SDKs, the alerts are in a different format. This change accounts for that.

### DIFF
--- a/apps/teams-test-app/e2e-test-data/meeting.json
+++ b/apps/teams-test-app/e2e-test-data/meeting.json
@@ -193,7 +193,7 @@
       "expectedTestAppValue": "shareAppContentToStage() succeeded"
     },
     {
-      "title": "shareAppContentToStage API Call (>=2.19.0, host SDK>=2.8.0) - Success",
+      "title": "shareAppContentToStage API Call (TJS >=2.19.0, host SDK>=2.8.0) - Success",
       "type": "callResponse",
       "version": ">=2.19.0",
       "hostSdkVersion": {
--- a/apps/teams-test-app/e2e-test-data/meeting.json
+++ b/apps/teams-test-app/e2e-test-data/meeting.json
@@ -166,9 +166,25 @@
       "expectedTestAppValue": "Receieved error {\"errorCode\":1000,\"message\":\"Encountered an error\"}"
     },
     {
-      "title": "shareAppContentToStage API Call (<2.19.0) - Success",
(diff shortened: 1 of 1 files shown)


lakhveerk (approved):
I left a tiny comment. Otherwise changes look good!

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -177,9 +193,12 @@
       "expectedTestAppValue": "shareAppContentToStage() succeeded"
     },
     {
-      "title": "shareAppContentToStage API Call (>=2.19.0) - Success",
+      "title": "shareAppContentToStage API Call (>=2.19.0, host SDK>=2.8.0) - Success",
```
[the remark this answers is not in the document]
AE-MS: Fixed

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -166,9 +166,25 @@
       "expectedTestAppValue": "Receieved error {\"errorCode\":1000,\"message\":\"Encountered an error\"}"
     },
     {
-      "title": "shareAppContentToStage API Call (<2.19.0) - Success",
+      "title": "shareAppContentToStage API Call (host SDK <2.8.0) - Success",
```
erinha, line 169: Not really sure you need to update the title. The `hostSdkVersion.web` value is already prepended in the test title by the framework. This is how it appears in the logs:

"Target Web Host SDK v-<2.8.0 - shareAppContentToStage API Call (host SDK <2.8.0) - Success"

AE-MS: I was updating the title for the benefit of people reading the JSON and not really because of the logging output (though obviously it could help there too, but not in this case)

I'm just trying to fit into whatever the existing system is here: should I remove the version from the title?

erinha (approved):

apps/teams-test-app/e2e-test-data/meeting.json
erinha: I think it's helpful to folks reading the code, so I'm fine with leaving it in. It doesn't exactly match convention (if one previous change can be considered "convention"), but I think it's an improvement.

jadahiya-MSFT (approved):

PR #2177, opened 2024-02-15
Updated socks package to a version that doesn't use the IP package

shrshindeMSFT:

## Description

This repo has a CG violation for the `ip` project. The `ip` project has a published security vulnerability:
https://github.com/indutny/node-ip/issues/136 
https://github.com/advisories/GHSA-78xj-cgh5-2h22 

Unfortunately, there's no fixed version of the `ip` project. Moreover, there are questions in the "Issues" section of that project asking whether there's even a maintainer for it.

We indirectly use `ip` because we use lerna, which has a whole chain of dependencies which eventually leads to `ip`
The package in that chain which depends on `ip` is `socks.` The socks package [HAS been updated](https://github.com/JoshGlazebrook/socks/pull/94)  to stop depending on "ip" given the uncertainty of getting a fix in `ip.`

This change forces us to use the latest version of socks that removes the `ip` dependency. Eventually, over probably a while, the whole `lerna` dependency change will update to use this newer version of socks, but who knows how long that will take. In the meantime, this override ensures that with our current set of dependencies we no longer consume the vulnerable version of `ip.`

I also added a section to our package.json file that provides room to explain what each override is for, as a way of hopefully helping our future selves to know when we might be able to remove any of them. As the explanation the file says, it's a hacky way of putting comments in the file since comments are not allowed in JSON.

### Main changes in the PR:

Override `socks` package in `package.json` file.

### Unit Tests added:

N/A

### End-to-end tests added:

N/A

### Change file added:

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,9 @@
   "pnpm": {
     "overrides": {
       "minimist": "^1.2.8",
-      "socks": "^2.7.3"
+      "semver": "^7.5.4",
+      "socks": "^2.7.3",
+      "string_decoder": "^1.3.0"
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -120,6 +120,10 @@
       "url-parse": "^1.5.0",
       "word-wrap": "^1.2.4",
       "y18n": "^4.0.1"
+    },
+    "overrides-explanation": {
+      "WHAT IS THIS SECTION": "pnpm ignores this section and comments aren't allowed in JSON files. This section documents why the above overrides have been put in place. If you add an override, describe it in this section.",
+      "socks": "There is a vulnerability in the ip package which has no fix. We consume ip via socks (eventually via lerna). Socks released a new version that removed the ip dependency. We are using this newer version of socks to avoid the vulnerability. If ip is ever updated or lerna (or any package in the chain) eventually updates to a version of socks that doesn't depend on ip, we can remove this override"
     }
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -100,34 +100,29 @@
     "webpack-subresource-integrity": "^5.1.0",
     "yargs": "^17.7.2"
   },
-  "resolutions": {
-    "axios": "^1.6.0",
-    "dns-packet": "^1.3.2",
-    "follow-redirects": "^1.15.4",
-    "glob-parent": "^5.1.2",
-    "lodash": "^4.17.21",
-    "merge": "^2.1.1",
-    "minimist": "^0.2.4",
-    "postcss": "^8.4.31",
-    "semver": "^7.5.2",
-    "serialize-javascript": "^3.1.0",
-    "set-value": "^2.0.1",
-    "string_decoder": "^1.3.0",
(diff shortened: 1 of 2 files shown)


jadahiya-MSFT (approved):

pnpm-lock.yaml
```diff
@@ -1,23 +1,7 @@
 lockfileVersion: '6.0'
 
 overrides:
-  axios: ^1.6.0
-  dns-packet: ^1.3.2
-  follow-redirects: ^1.15.4
-  glob-parent: ^5.1.2
```
jadahiya-MSFT, line 7: Removing some of these will reintroduce old dependabot/CG issues

jadahiya-MSFT: I believe string-decoder and semver are the key culprits, but I'll double check and get back to you

jadahiya-MSFT: minimist version can't be below 0.2.4

shrshindeMSFT: Updated `minimist` to use latest version.

package.json
```diff
@@ -121,5 +121,13 @@
   },
   "dependencies": {
     "uuid": "^9.0.0"
+  },
+  "pnpm": {
+    "overrides": {
+      "minimist": "^1.2.8",
+      "semver": "^7.5.4",
+      "socks": "^2.7.3",
+      "string_decoder": "^1.3.0"
```
AE-MS, line 130: It looks like you are adding these here but I see from the change in the lock file that some were already being overridden there -- do you know what causes that? How did there get to be overrides in the lock file that weren't listed in package.json?

package.json
```diff
@@ -121,5 +121,13 @@
   },
   "dependencies": {
     "uuid": "^9.0.0"
+  },
+  "pnpm": {
+    "overrides": {
```
AE-MS, line 126: >overrides [](http://example.com/codeflow?start=5&length=9)

Is it worth considering doing something to document the justification for these overrides so people in the future know why they are here and/or when they can remove them?

AE-MS (changes requested):
A few questions

jadahiya-MSFT (approved):

package.json
shrshindeMSFT: When migrating from `yarn` to `pnpm` the resolutions field is useful to override packages as `pnpm import` commands generates `pnpm-lock.yaml` file. Now when I added `"pnpm": {"overrides": }` field in `package.json` then `pnpm.overrides` takes precedence over `resolutions` field as mentioned in the [pnpm resolutions](https://pnpm.io/next/package_json#resolutions) documentation. To resolve this, I've updated `resolutions` to `pnpm.overrides`.

pnpm-lock.yaml
shrshindeMSFT: This shouldn't be an issue anymore as `resolutions` field is updated to `pnpm.overrides`

AE-MS (approved):

jadahiya-MSFT (approved):

PR #2178, opened 2024-02-16
Update openChat API test case

KangxuanYe:

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


apps/teams-test-app/e2e-test-data/chat.json
```diff
@@ -24,10 +24,10 @@
       "version": ">2.17.0",
       "boxSelector": "#box_openChat",
       "inputValue": {
-        "user": ["testUpn"],
+        "user": "testUpn",
```
AE-MS, line 27: >"user": "testUpn", [](http://example.com/codeflow?start=8&length=18)

This looks like the right change to make given that `OpenChatSingleRequest` contains a single `string` value for the `user` property. How was this working before your change? Shouldn't something have failed when trying to input an array into a string value?

AE-MS (changes requested):
:clock1:

apps/teams-test-app/e2e-test-data/chat.json
KangxuanYe: This is a great great great question because I took a lot of time to figure out how it works with incorrect input in different repos.

AE-MS: Is there an improvement we should be making in the web hub SDK to be more sensitive to these message formats? It seems wrong that the same test can fail in one platform and not in another. All hub SDKs should have the same expectations on how they parse messages coming from teamsjs, no?

AE-MS: Note: I'm fine with not blocking this PR for that, but please be sure to follow-up. We can't have inconsistent parsing/treatment across platforms.

AE-MS (approved):
Please note my comment about inconsistent parsing across platforms.

TrevorJoelHarris (approved):

PR #2183, opened 2024-02-21
Fix a few doc bugs I discovered

AE-MS:
### Main changes in the PR:

1. Fix doc URL links to remove hardcoded locale (en-us) and allow the docs site to redirect to the current user's locale. This is a best practice from the docs team
2. Fix the documentation on the `user` property of `OpenSingleChatRequest` to reflect that it's only a single UPN
3. Fix the markdown on the URL in the documentation for `SecondaryM365ContentIdName` which had reversed brackets/parentheses

### Change file added:

Yuppers


shrshindeMSFT (approved):
Thanks for removing the language tags from the documentation.

KangxuanYe (approved):

KangxuanYe:
Errr, I didn't realize I need to update those things when bug gets fixed. Will try to remember it next time to (ask someone) to fix those docs.

PR #2187, opened 2024-02-27
Update base64ToBlob logic to accept special characters

shrshindeMSFT:

## Description

This PR fixes the `atob` function issues by using `Buffer.from(str)`. In `atob` function some of the special characters are not completely processed for ex. `Update...` will `Updateâ€¦`  where `...` are 3 different characters. To resolve this issue we use `Buffer.from(str)` for texts and `atob` for images in `base64ToBlob` function. Another notable change here is Polyfill for `buffer` in webpack as it was removed from `webpack 5` here is the link for the web pack [changelog](https://github.com/webpack/changelog-v5).

### Main changes in the PR:

1. Use `Buffer.from(str)` for string instead of `atob`
2. Add `Buffer` as polyfill for nodejs

### Validation performed:

Added unit test cases

### Unit Tests added:

Yes

### End-to-end tests added:

N/A

### Change file added:
Yes

### DIFF
--- a/packages/teams-js/src/internal/utils.ts
+++ b/packages/teams-js/src/internal/utils.ts
@@ -331,7 +331,6 @@ export function base64ToBlob(mimeType: string, base64String: string): Promise<Bl
     if (!base64String) {
       reject('Base64 string cannot be null or empty.');
     }
-    const byteCharacters = atob(base64String);
     /**
      * For images we need to convert binary data to image to achieve that:
      *   1. A new Uint8Array is created with a length equal to the length of byteCharacters.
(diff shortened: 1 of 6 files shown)


lakhveerk (dismissed):
lgtm

packages/teams-js/src/internal/utils.ts
```diff
@@ -342,12 +341,14 @@ export function base64ToBlob(mimeType: string, base64String: string): Promise<Bl
...
      */
     if (mimeType.startsWith('image/')) {
+      const byteCharacters = atob(base64String);
       const byteArray = new Uint8Array(byteCharacters.length);
       for (let i = 0; i < byteCharacters.length; i++) {
         byteArray[i] = byteCharacters.charCodeAt(i);
       }
       resolve(new Blob([byteArray], { type: mimeType }));
     }
+    const byteCharacters = Buffer.from(base64String, 'base64').toString();
```
baljesingh, line 351: Is there any validation other than a null check to ensure it is a valid base64 string? If not, it would be better to enclose the entire transformation within a try-catch block and reject it with an error in the catch block.

noahdarveau-MSFT (approved):

packages/teams-js/src/internal/utils.ts
shrshindeMSFT: We already have a unit test case for that in place.

jadahiya-MSFT (dismissed):

noahdarveau-MSFT (approved):

PR #2197, opened 2024-03-04
Remove valid origin that is no longer needed

TrevorJoelHarris:
## Description

No longer need one of the valid origins.

### Main changes in the PR:

1. Removed `*.teams.microsoft.com` from the list of valid origins we store. (`teams.microsoft.com` is still there though)
2. Updated unit tests.

### Validation performed:

1. Ran unit tests

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/test/internal/validOrigins.spec.ts
+++ b/packages/teams-js/test/internal/validOrigins.spec.ts
@@ -30,7 +30,7 @@ describe('validOrigins', () => {
       expect(result).toBe(true);
     });
     it('validateOrigin returns true if origin for subdomains in teams pre-known allowlist', async () => {
-      const messageOrigin = new URL('https://subdomain.teams.microsoft.com');
+      const messageOrigin = new URL('https://test.www.office.com');
       const result = await validateOrigin(messageOrigin);
       expect(result).toBe(true);
     });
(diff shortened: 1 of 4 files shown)


ydogandjiev (approved):

jadahiya-MSFT (approved):

packages/teams-js/test/internal/validOrigins.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('validOrigins', () => {
       expect(result).toBe(true);
     });
     it('validateOrigin returns true if origin for subdomains in teams pre-known allowlist', async () => {
-      const messageOrigin = new URL('https://subdomain.teams.microsoft.com');
```
jadahiya-MSFT, line 33: Shouldn't we change these tests to now expect false instead as well, to ensure that this is working as intended?

TrevorJoelHarris: We still allow for wildcard domains for a couple other origins, so it seemed like retrofitting the tests that tested for this case working correctly to use different valid defaults made sense. Does that seem okay?

jadahiya-MSFT: Yup that seems ok to me

Issue #2196, opened 2024-03-04
Clipboard api throws 'NotAllowedError: Failed to execute 'write' on 'Clipboard': Document is not focused'


ZimnitskiYury:
**Clipboard write() method not working in desktop Teams**

I'm developing a personal tab for Microsoft Teams. Part of the functionality of my app is copying some data to the clipboard. For this, I use the write() method from the clipboard namespace. **It works fine in the web version of Teams** but throws an error in the desktop Teams.
`Failed to copy to clipboard. NotAllowedError: Failed to execute 'write' on 'Clipboard': Document is not focused.`

Here is the code snippets that I use to copy data to the clipboard:
1.
```
const content = ‘Hello, world!’; 
const blob = new Blob([content], { type: ‘text/plain’ });
clipboard.write(blob).catch((error) => { throwError(error); });
```

2. I also tried to use window.focus() and this approach, but neither of them worked:
```
const content = ‘Hello, world!’;
const blob = new Blob([content], {type: ‘text/plain’});

const button = document.createElement(‘button’); 
button.innerHTML = ‘Copy’; 
button.style.display = ‘none’; 
button.addEventListener(‘click’, async () => { 
    clipboard.write(blob)
                   .catch((error) => { 
                       throwError(error); 
                   })
                  .finally(() => { 
                     button.remove(); 
                  });
}); 
document.body.append(button); 
pages.returnFocus(); 
button.focus(); 
button.click();
```

Prasad-MSFT:
The PR #2201   is a WIP to fix this issue, so please follow that PR for tracking when the fix is available!

ZimnitskiYury:
actual for #2220

shrshindeMSFT:
Thank you for reporting. I'll look into it and see if I can repro this issue.

shrshindeMSFT:
Thank you for your patience. The fix for this issue is deployed and is available for everyone to use.

ZimnitskiYury:
@shrshindeMSFT issue still exist. 
microsoft/teams-js: 2.24.0
Microsoft Teams version: 24137.2216.2931.2440
The client version is 49/24050307617.

if i write to clipboard image:

- web teams works fine
- desktop throws error `Failed to copy to clipboard. NotAllowedError: Failed to execute 'write' on 'Clipboard': Document is not focused.`

if I write to clipboard text/plain or text/html:

- web and desktop teams throw Internal Error 
[...]

PR #2201, opened 2024-03-05
Bug Fix:  Clipboard capability throwing error `DOMException: Document not focused` 

shrshindeMSFT:

## Description

Currently developers are facing [issue ](https://github.com/OfficeDev/microsoft-teams-library-js/issues/2196) in desktop clients on clipboard capability. This PR adds changes to how `App` interacts with parent window over JSON string rather than `blob`.

### Main changes in the PR:

1. Added desktop to use JSON string to read response from clipboard.
2. Added desktop to unit tests.

### Unit Tests added:

Yes

### End-to-end tests added:

N/A

### Change file added:

Yes

Issue Link:[ Clipboard api throws 'NotAllowedError: Failed to execute 'write' on 'Clipboard': Document is not focused](https://github.com/OfficeDev/microsoft-teams-library-js/issues/2196)

### DIFF
--- a/change/@microsoft-teams-js-1f05f3e0-286f-4831-8747-649c0e47e551.json
+++ b/change/@microsoft-teams-js-1f05f3e0-286f-4831-8747-649c0e47e551.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed clipboard issue for desktop client to resolve DOMExecption:Document not focused",
+  "packageName": "@microsoft/teams-js",
+  "email": "98348000+shrshindeMSFT@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-1f05f3e0-286f-4831-8747-649c0e47e551.json
+++ b/change/@microsoft-teams-js-1f05f3e0-286f-4831-8747-649c0e47e551.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fixed clipboard issue for desktop client to resolve DOMExecption:Document not focused",
+  "comment": "Fixed `clipboard` issue for desktop client to resolve 'DOMExecption: Document not focused' error",
   "packageName": "@microsoft/teams-js",
   "email": "98348000+shrshindeMSFT@users.noreply.github.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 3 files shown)


AE-MS:
Is there an E2E test that can catch this problem?

packages/teams-js/src/public/clipboard.ts
```diff
@@ -82,7 +82,11 @@ export namespace clipboard {
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
-    if (isHostClientMobile() || GlobalVars.hostClientType === HostClientType.macos) {
+    if (
+      isHostClientMobile() ||
+      GlobalVars.hostClientType === HostClientType.macos ||
```
TrevorJoelHarris, line 87: Can you remind me why we need to do this for desktop and macos? In general, it's good to avoid adding platform specific checks like this if we can because it can be hard to maintain or change in the future once it is checked into TeamsJS and removed from `@beta`. Is there some way we could call `sendAndHandleSdkError` and then look at the contents of what is returned and decide whether or not to call `base64ToBlob` based on just that without looking at `hostClientType`?

shrshindeMSFT: I have updated as you mentioned. Now depending on the response received we decide to call `base64ToBlob` without relying on `hostClientType`

change/@microsoft-teams-js-1f05f3e0-286f-4831-8747-649c0e47e551.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed clipboard issue for desktop client to resolve DOMExecption:Document not focused",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Fixed `clipboard` issue for desktop client to resolve 'DOMExecption: Document not focused' error",
```

shrshindeMSFT:
> Is there an E2E test that can catch this problem?

Right now, we don't have infra to catch this e2e test case problem, but as a follow up I'll add something to pipelines which will test against desktop clients.

noahdarveau-MSFT (approved):

jadahiya-MSFT (approved):

PR #2211, opened 2024-03-07
Trharris/removevo

TrevorJoelHarris:
## Description

Removed one default valid origin that is no longer needed

### Main changes in the PR:

1. Removed single default valid origin
2. Updated tests

### Validation performed:

1. Ran unit tests

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ microsoft-teams-library-js.sln
 artifacts/
 bin/
 obj/
+
+#VS Code Local Files
+*.sln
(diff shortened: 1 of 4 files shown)


jadahiya-MSFT (approved):

.gitignore
```diff
@@ -43,3 +43,6 @@ microsoft-teams-library-js.sln
 artifacts/
 bin/
 obj/
+
+#VS Code Local Files
+*.sln
```
TrevorJoelHarris, line 48: For some reason, VS code has started to create soln files that git then wants to check in. I don't want sln files checked in, and therefore I have added them to our .gitignore file.

AE-MS: I don't really care that much about this but _normally_ sln files would be checked in so that everyone on the team builds from the same solution. On our team I think this only matters for the blazor app so it's probably not terribly important.

AE-MS (approved):
:shipit:

PR #2221, opened 2024-03-14
Add getContent API

ShanmathiMayuramKrithivasan:
## Description

This PR introduces `sharing.history.getContent` API to fetch content shared in a Teams context.

### Main changes in the PR:

1. Added a new sub-namespace `history` in `sharing` namespace.
2. Added interface `IContentResponse` for the content details params that would be returned.
3. Added support for 'getContent' API.
4. Updated `sharing.spec.ts` with unit tests for this new API.

### Unit Tests added:
Yes

### End-to-end tests added:
Existing Teams repo end-to-end test to be updated.

### Change file added:
Yes

### Next/remaining steps:

- [ ] Teams changes
- [ ] Hosts SDK changes


vikramtha:
Hello, I am the PM assigned to review this. I have just reached out to Shanmathi Mayuram Krithivasan on teams asking about timelines

AE-MS:
Please make sure you add a changefile using the guidelines in the [contributor's guide](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball). The build will not pass until it is present. Thanks!

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+      /** Id of the app where the content was shared from */
+      appId: string;
+      /** Title of the shared content */
+      title: string;
+      /** Reference of the shared content */
+      contentReference: string;
+      /** Id of the thread where the content was shared */
+      threadId: string;
+      /** Id of the user who shared the content */
+      author: string;
```
AE-MS, line 219: What format is this in?

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+      /** Type of the shared content */
+      contentType: string;
```
AE-MS, line 221: Can this really be any string? Or is there a list of possible content types? Is it a MIME type?

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+     */
+    export interface IContentResponse {
+      /** Id of the app where the content was shared from */
+      appId: string;
+      /** Title of the shared content */
+      title: string;
+      /** Reference of the shared content */
+      contentReference: string;
+      /** Id of the thread where the content was shared */
+      threadId: string;
```
AE-MS, line 217: Can thread IDs really be any string or is there a format to them? Is "Shanmathi" a valid thread ID?

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+
+    /**
+     * Get the list of content shared in a context
+     * @param callback - Callback contains 2 parameters, `error` and `contentDetails`.
+     * `error` can either contain an error of type `SdkError`, in case of an error, or null when fetch is successful.
+     * `contentDetails` will be the list of contents {@link IContentResponse} that was shared in the context, or null when the request fails.
+     *
+     * @beta
+     */
+    export function getContent(callback: getContentCallbackFunctionType): void {
```
AE-MS, line 242: >callback: getContentCallbackFunctionType [](http://example.com/codeflow?start=31&length=40)

Please return a `Promise` instead of using a callback. Thanks!

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+      if (!callback) {
+        throw new Error('[get content] Callback cannot be null');
+      }
+      ensureInitialized(runtime, FrameContexts.sidePanel, FrameContexts.meetingStage);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+      sendMessageToParent(
+        getApiVersionTag(sharingTelemetryVersionNumber_v1, ApiName.Sharing_History_GetContent),
```
AE-MS, line 251: >sharingTelemetryVersionNumber_v1 [](http://example.com/codeflow?start=25&length=32)

I actually don't know how these version numbers are supposed to be set. @KangxuanYe what's the right constant to use for new functions? Is it v1 like this?

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: author is the userID here and will be GUID.

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+     * @throws Error if {@linkcode app.initialize} has not successfully completed
+     *
+     * @beta
+     */
+    export function isSupported(): boolean {
+      return ensureInitialized(runtime) && runtime.supports.sharing
+        ? runtime.supports.sharing.history
+          ? true
+          : false
+        : false;
```
AE-MS, line 270: This is a bit difficult to follow when written like this. Consider instead something like:
`return ensureInitialized(runtime) && runtime.supports.sharing?.history !== undefined`

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: These are string types. Right now on client the contentType is "ShareToStage" for this scenario.

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: No, threadIds are GUIDs that are returned from Teams side.

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
   }
+
+  /**
+   * Namespace to get the list of content shared in a context
+   *
+   * @beta
+   */
+  export namespace history {
+    /**
+     * Represents IContentResponse parameters.
```
AE-MS, line 205: >Represents IContentResponse parameters. [](http://example.com/codeflow?start=7&length=39)

Consider:

"Represents the data returned when calling {@link sharing.history.getContent}"

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,80 @@ export namespace sharing {
...
+     *
+     * @beta
+     */
+    export type getContentCallbackFunctionType = (
+      error: SdkError | null,
+      contentDetails: IContentResponse[] | null,
+    ) => void;
+
+    /**
+     * Get the list of content shared in a context
```
AE-MS, line 235: >shared in a context [](http://example.com/codeflow?start=31&length=19)

Can you elaborate on what "shared in a context" means? I'm not sure that developers reading this would know how to interpret that phrase.

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -290,6 +290,7 @@ export const enum ApiName {
   Settings_Save_Success = 'settings.save.success',
   Settings_SetSettings = 'settings.setSettings',
   Settings_SetValidityState = 'settings.setValidityState',
+  Sharing_History_GetContent = 'sharing.getContent',
```
AE-MS, line 293: >sharing.getContent [](http://example.com/codeflow?start=32&length=18)

Should this be `sharing.history.getContent` ?

AE-MS (changes requested):
:clock1:

AE-MS:
I haven't looked at the new unit tests yet since one of the changes I requested will change the function signature of `getContent` -- I will look at them on the next iteration. Thanks for considering my feedback!

AE-MS:
Please don't forget to response to the earlier comment from the github action bot about the runtime change you made. Thanks!

AE-MS:
Your build is failing because it thinks you are missing a changefile. I see you do have a changefile in here, so I'm not totally sure why it is complaining. Did you create the file using the `pnpm changefile` command? One thing I notice is that you list the change type as `beta` which isn't one of the four valid types. Perhaps that is what is causing the problem?
See more detail in the [contributor's guide](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball).

packages/teams-js/src/public/sharing.ts
AE-MS: Cool; can you use a type that only allows for UUIDs then, rather than `string`? Or if you really must use `string`, can you validate that any values you receive are valid UUIDs?

Please also document the format. Thanks!

packages/teams-js/src/public/sharing.ts
AE-MS: Same comment as for `author`, then. Thanks!

packages/teams-js/src/public/sharing.ts
AE-MS: So there's only one possible value? Can you document the valid values so that developers know what to expect?

ShanmathiMayuramKrithivasan: Added validation and updated documentation.

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: Added validation and updated documentation.

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: Updated to return a promise

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: Updated

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: Updated comment

packages/teams-js/src/internal/telemetry.ts
ShanmathiMayuramKrithivasan: yes updated

ShanmathiMayuramKrithivasan:


Based on the RUNTIME.md file, in this PR we have added a new optional property or supported capability. So will not require a new version.

ShanmathiMayuramKrithivasan:
> Your build is failing because it thinks you are missing a changefile. I see you do have a changefile in here, so I'm not totally sure why it is complaining. Did you create the file using the `pnpm changefile` command? One thing I notice is that you list the change type as `beta` which isn't one of the four valid types. Perhaps that is what is causing the problem? See more detail in the [contributor's guide](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md#change-log-using-beachball).

Created a new change file. This should fix the build errors.

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +198,68 @@ export namespace sharing {
...
+      author: string;
+      /** Type of the shared content.
+       * For sharing to Teams stage scenarios, this value would be ShareToStage
+       */
+      contentType: string;
+    }
+
+    /**
+     * Get the list of content shared in a Teams meeting
+     *
```
AE-MS, line 232: >* [](http://example.com/codeflow?start=5&length=1)

Please document the return type using the `@returns` doc format. You can find some other examples in this file and elsewhere.

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +198,68 @@ export namespace sharing {
...
+    export function getContent(): Promise<IContentResponse[]> {
+      ensureInitialized(runtime, FrameContexts.sidePanel, FrameContexts.meetingStage);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+
+      return sendAndHandleSdkError(
+        getApiVersionTag(sharingTelemetryVersionNumber_v1, ApiName.Sharing_History_GetContent),
+        'sharing.history.getContent',
+      ).then((contentDetails: IContentResponse[]) => {
```
AE-MS, line 244: >then [](http://example.com/codeflow?start=8&length=4)

There's nothing wrong with how you wrote this but it would likely be easier to read if you use `await` instead of `.then`. Your call, though.

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +198,68 @@ export namespace sharing {
...
+        contentDetails.map((contentDetails) => {
+          validateUuid(contentDetails.author);
```
AE-MS, line 246: >validateUuid [](http://example.com/codeflow?start=10&length=12)

I looked into the implementation of these functions and they seem mostly correct but they weirdly "allow" for the input parameter to be `undefined` or `null` without complaining. This seems wrong to me (in general, not specifically for your scenario), since `undefined` and `null` are both invalid UUIDs.

To be sure your validation is comprehensive, can I suggest you check here to ensure they aren't `null` or `undefined`? I know it's lame that the shared function that someone wrote doesn't do the right thing and that you can't just use that shared function, but I'd rather you have the correct validation and not suffer because of a suboptimal shared function. You could even help the whole codebase by adding a new function next to `validateUuid` in the shared location that doesn't tolerate `undefined` or `null`, e.g. `validateDefinedUuid` or something. Whatever works for you.

(unless it is actually valid for these two values to be `null` or `undefined` for you, in which case you should update their types on the `IContentResponse` interface)

packages/teams-js/src/internal/utils.ts
```diff
@@ -400,3 +400,22 @@ export function ssrSafeWindow(): Window {
...
+ * @param id The ID to validate against the UUID format
+ * @throws Error if ID is not a valid UUID
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function validateUuid(id: string | undefined | null): void {
+  if (id === undefined || id === null) {
+    return;
+  }
```
AE-MS, line 414: Is it correct that `undefined` and `null` are "valid" UUIDs? Aren't those invalid?

packages/teams-js/test/internal/utils.spec.ts
```diff
@@ -185,4 +186,22 @@ describe('utils', () => {
       }
     });
   });
+
+  describe('validateUuid', () => {
```
AE-MS, line 190: >validateUuid [](http://example.com/codeflow?start=12&length=12)

Can you please add a unit test for the empty string case (it should be invalid)?

AE-MS: (the answer to this depends on the answer to my other comment in `utils.ts` about why `undefined` and `null` are considered valid. There should also be unit tests here for those cases)

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +198,68 @@ export namespace sharing {
...
+      /** Title of the shared content */
+      title: string;
+      /** Reference of the shared content */
+      contentReference: string;
+      /** Id of the thread where the content was shared. This is a UUID */
+      threadId: string;
+      /** Id of the user who shared the content. This is a UUID */
+      author: string;
+      /** Type of the shared content.
+       * For sharing to Teams stage scenarios, this value would be ShareToStage
```
AE-MS, line 225: >ShareToStage [](http://example.com/codeflow?start=67&length=12)

Please surround this with backticks, e.g.,

For sharing to Teams stage scenarios, this value would be `ShareToStage`

And also add:
Other `contentType` values will be added here over time
(or some other text that indicates that there are currently no other known values but it will expand in the future)

packages/teams-js/test/public/sharing.spec.ts
```diff
@@ -694,4 +696,104 @@ describe('sharing_v2', () => {
         }
       });
   });
+
+  describe('Testing getContent', () => {
```
AE-MS, line 700: >describe [](http://example.com/codeflow?start=2&length=8)

Can you please add two tests:
- one that ensures that an error is thrown if the authorid is not a valid uuid
- one that ensures that an error is thrown if the threadid is not a valid uuid

AE-MS:
I didn't see any changes made to the test app to validate this new capability, and similarly there are no new E2E tests added. I know that you can't actually run those E2E scenarios or tests until the corresponding change is made in the host SDK. Please ensure you come back and add them after you make that host SDK change. Thanks!

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +198,68 @@ export namespace sharing {
...
+       */
+      contentType: string;
+    }
+
+    /**
+     * Get the list of content shared in a Teams meeting
```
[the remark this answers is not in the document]
ShanmathiMayuramKrithivasan: Updated

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: Updated to use `await`.

packages/teams-js/src/internal/utils.ts
ShanmathiMayuramKrithivasan: Updated to make all these cases invalid.

packages/teams-js/test/internal/utils.spec.ts
ShanmathiMayuramKrithivasan: Added tests for `undefined`, `null` and empty string.

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +198,68 @@ export namespace sharing {
   export function isSupported(): boolean {
     return ensureInitialized(runtime) && runtime.supports.sharing ? true : false;
   }
+
+  /**
+   * Namespace to get the list of content shared in a context
```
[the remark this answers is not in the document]
ShanmathiMayuramKrithivasan: Updated description.

packages/teams-js/src/public/sharing.ts
ShanmathiMayuramKrithivasan: Updated and added the comments.

packages/teams-js/test/public/sharing.spec.ts
ShanmathiMayuramKrithivasan: Updated tests for validating valid uuid,

ShanmathiMayuramKrithivasan:
> I didn't see any changes made to the test app to validate this new capability, and similarly there are no new E2E tests added. I know that you can't actually run those E2E scenarios or tests until the corresponding change is made in the host SDK. Please ensure you come back and add them after you make that host SDK change. Thanks!

Yes will add E2E tests once the host SDK changes are in.

AE-MS (dismissed):

packages/teams-js/src/public/sharing.ts
KangxuanYe: This is V2 rather than V1. Only API exists in Teams-JS library v1 and does NOT exist in Teams-JS library v2 will be labelled as V1. Vice versa.

This new API is designed to be in Teams-JS library V2 so it should fix here from v1 to v2.

KangxuanYe (changes requested):
Need to fix telemetry version

packages/teams-js/src/public/sharing.ts
```diff
@@ -194,4 +194,73 @@ export namespace sharing {
...
+     * @beta
+     */
+    export async function getContent(): Promise<IContentResponse[]> {
+      ensureInitialized(runtime, FrameContexts.sidePanel, FrameContexts.meetingStage);
+      if (!isSupported()) {
+        throw errorNotSupportedOnPlatform;
+      }
+
+      const contentDetails: IContentResponse[] = await sendAndHandleSdkError(
+        getApiVersionTag(sharingTelemetryVersionNumber_v1, ApiName.Sharing_History_GetContent),
```
KangxuanYe, line 243:
```suggestion
        getApiVersionTag(sharingTelemetryVersionNumber_v2, ApiName.Sharing_History_GetContent),
```

KangxuanYe (changes requested):
Proposed a suggestion.

KangxuanYe (approved):

AE-MS (approved):

PR #2228, opened 2024-03-19
Added new capability where apps can register for whether other applications have been installed while your app is running

TrevorJoelHarris:
## Description

Added a new capability that allows limited 1P apps to register to receive events whenever an app is installed on the host.

### Main changes in the PR:

1. Added `OtherAppStateChange` capability
2. Added unit tests
3. Added e2e tests

### Validation performed:

1. Ran unit tests locally
2. Ran e2e tests locally

### Unit Tests added:

Yes

### End-to-end tests added:

Yes

### Change file added:

Yes

### DIFF
--- a/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
+++ b/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
@@ -1,26 +1,24 @@
 {
   "name": "OtherAppStateChange",
   "platforms": "Web",
+  "version": ">2.21.0",
   "hostSdkVersion": {
     "web": ">2.13.1"
   },
   "testCases": [
     {
-      "version": ">2.21.0",
       "title": "isSupported - Success",
       "type": "callResponse",
       "boxSelector": "#box_otherAppStateChange_isSupported",
       "expectedTestAppValue": "OtherAppStateChanged module is supported"
     },
     {
-      "version": ">2.21.0",
       "title": "registerAppInstallationHandler - Success",
       "type": "callResponse",
       "boxSelector": "#box_otherAppStateChange_registerInstallHandler",
       "expectedTestAppValue": "received"
     },
     {
-      "version": ">2.21.0",
       "title": "registerAppInstallationHandler should not receive install events if app is not approved",
       "type": "registerAndRaiseEvent",
       "boxSelector": "#box_otherAppStateChange_registerInstallHandler",
@@ -33,7 +31,6 @@
       "expectedTestAppValue": "received"
     },
     {
-      "version": ">2.21.0",
       "title": "unregisterAppInstallationHandler - Sends but not processed if app is not approved",
       "type": "callResponse",
       "boxSelector": "#box_otherAppStateChange_unregisterInstallHandler",
--- a/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
+++ b/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
@@ -3,21 +3,21 @@
   "platforms": "Web",
   "testCases": [
     {
-      "version": ">=2.21.0",
+      "version": ">2.21.0",
       "title": "isSupported - Success",
       "type": "callResponse",
       "boxSelector": "#box_otherAppStateChange_isSupported",
       "expectedTestAppValue": "OtherAppStateChanged module is supported"
     },
     {
-      "version": ">=2.21.0",
+      "version": ">2.21.0",
       "title": "registerAppInstallationHandler - Success",
       "type": "callResponse",
       "boxSelector": "#box_otherAppStateChange_registerInstallHandler",
       "expectedTestAppValue": "received"
     },
     {
-      "version": ">=2.21.0",
+      "version": ">2.21.0",
       "title": "registerAppInstallationHandler should not receive install events if app is not approved",
       "type": "registerAndRaiseEvent",
       "boxSelector": "#box_otherAppStateChange_registerInstallHandler",
@@ -30,7 +30,7 @@
       "expectedTestAppValue": "received"
     },
     {
-      "version": ">=2.21.0",
+      "version": ">2.21.0",
       "title": "unregisterAppInstallationHandler - Sends but not processed if app is not approved",
       "type": "callResponse",
       "boxSelector": "#box_otherAppStateChange_unregisterInstallHandler",
--- a/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
+++ b/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
@@ -1,6 +1,9 @@
 {
   "name": "OtherAppStateChange",
   "platforms": "Web",
+  "hostSdkVersion": {
+    "web": ">2.13.1"
+  },
   "testCases": [
     {
       "version": ">2.21.0",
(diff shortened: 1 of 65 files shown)


TrevorJoelHarris:

Added one new capability to runtime, not a breaking change. No need to make version changes.

packages/teams-js/src/private/otherAppStateChange.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @beta
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * Represent an event that has happened with other number of applications installed on this host.
+   * (e.g. a new app has been installed)
+   */
+  export interface OtherAppStateChangeEvent {
+    /** An array of app ids that this event applies to */
+    appIds: string[];
```
AE-MS, line 36: >string [](http://example.com/codeflow?start=12&length=6)

Can this be a...

just kidding.

packages/teams-js/src/private/otherAppStateChange.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @hidden
+   * @beta
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * This function can be called so that the handler passed to {@link registerAppInstallationHandler}
+   * will no longer receive app installation events
+   *
```
AE-MS, line 96: Is it worth documenting what happens if you call this function without first having registered a handler (or if you call this function multiple times in a row)?

I think the answer is "no op"

packages/teams-js/src/private/otherAppStateChange.ts
TrevorJoelHarris: they don't have the emoji I am looking for

packages/teams-js/src/private/otherAppStateChange.ts
TrevorJoelHarris: You are correct, the answer is no-op. I will add that comment

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -202,6 +202,8 @@ export enum ApiName {
   Navigation_NavigateToTab = 'navigation.navigateToTab',
   Navigation_ReturnFocus = 'navigation.returnFocus',
   Notifications_ShowNotification = 'notifications.showNotification',
+  OtherAppStateChange_Install = 'otherApp.install',
+  OtherAppStateChange_UnregisterInstall = 'otherApp.unregisterInstall',
```
AE-MS, line 206: The lack of symmetry here seems notable; intentional?

"install" vs "unregisterInstall"

Should it be "registerInstall" vs "unregisterInstall"?

apps/teams-test-app/e2e-test-data/otherAppStateChange.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "OtherAppStateChange",
```
AE-MS, line 2: >name [](http://example.com/codeflow?start=3&length=4)

Can you add a hub SDK version restriction to these that the hub SDK must be >= v2.13.1 ? That's obviously not right because your hub SDK changes aren't going to be in v2.13.1 but until we fix the problem where we can't say "latest or greater" we should at least do v2.13.1 then come back and update it once we actually release the version with your changes.

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/telemetry.ts
TrevorJoelHarris: This is a great question! Let me explain my reasoning.

This is where the `OtherAppStateChange_Install` telemetry string is used in `otherAppStateChange.ts`:

image

The two instances of this string are used for different things:

The first line, all we really need there is a string that uniquely identifies this function call from all other function calls we are tracking. That way we can track whether or not functions are getting used by developers. This string would, you are correct, be more correctly written out to be something like `'otherApp.registerInstall'`

The second line though needs to be the name of the event that this function will receive from the host. So if the function registers for install events, the second string needs to be `'otherApp.install'` because we are receiving this event every time one or more apps are installed.

This leaves us with two options.

Option 1: Make two separate string literals, 'otherApp.registerInstall' and 'otherApp.install'. The code is easier to read but we now have two strings we need to keep in our minified js and devs have to pay the cost of downloading both.

Option 2: Use the same string for both. The code is a little more confusing but we only make one string literal.

Ultimately, I thought it was keep the package a little smaller which is why I landed on this solution. Good catch, and hopefully my explanation makes sense. Let me know if you disagree or want me to change it.

apps/teams-test-app/e2e-test-data/otherAppStateChange.json
TrevorJoelHarris: @Kangxuan-Ye do I still need to do this? I know you are literally in the process of making changes here right now. Just let me know what the best practice is and I will do it.

packages/teams-js/src/private/otherAppStateChange.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * @hidden
+ * @internal
+ * @beta
+ * Limited to Microsoft-internal use
+ *
+ * This capability contains the APIs for handling events that happen to other applications on the host
+ * *while* the developer's application is running. For example, if the developer wants to be notified
+ * when another application has been installed.
```
maglims, line 17: just curious, will the event trigger also applies to other cases, like: app uninstall, or app updates/

TrevorJoelHarris: For now, no, but if we add those in the future they will be added here. I wanted to keep this limited to just what is needed for the current scenario.

AE-MS (dismissed):
:shipit:

apps/teams-test-app/e2e-test-data/otherAppStateChange.json
```diff
@@ -0,0 +1,43 @@
...
+      "eventName": "otherApp.install",
+      "eventData": [
+        {
+          "appIds": ["123", "456"]
+        }
+      ],
+      "expectedTestAppValue": "received"
+    },
+    {
+      "version": ">2.21.0",
```
KangxuanYe, line 36: Forget to mention, if you want, you could have version over the file as a capability-wide field to control test cases.

apps/teams-test-app/src/components/OtherAppStateChangeAPIs.tsx
```diff
@@ -0,0 +1,53 @@
...
+    onClick: async () => `OtherAppStateChanged module ${otherAppStateChange.isSupported() ? 'is' : 'is not'} supported`,
+  });
+
+const RegisterAppInstallHandler = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'otherAppStateChange_registerInstallHandler',
+    title: 'Register App Install Handler',
+    onClick: async (setResult) => {
+      const handler = (event: otherAppStateChange.OtherAppStateChangeEvent): void => {
+        console.log(event);
```
KangxuanYe, line 20: Is this line on purpose to console the log?

TrevorJoelHarris: It was because it seemed useful for debugging, but I can definitely remove it.

erinha (approved):
🚀

AE-MS (approved):

PR #2230, opened 2024-03-20
Implementing the externalAppCommands API

MengyiGong:

## Description

Adds the 1P internal-only externalAppCommands API for delegating certain actions to the host.

### Main changes in the PR:

1. Add externalAppCommands capability
2. Add externalAppCommands.processActionCommand API

### Validation performed:

1. Unit tests
2. Capabilities added to the teams test app

### Unit Tests added:

<Yes>

### End-to-end tests added:

<No>

### Change file added:

<Yes>

### DIFF
--- a/apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
@@ -1,0 +1,58 @@
+import { externalAppCommands } from '@microsoft/teams-js';
+import React from 'react';
+
+import { ApiWithoutInput, ApiWithTextInput } from '../utils';
+import { ModuleWrapper } from '../utils/ModuleWrapper';
+
+const CheckExternalAppCommandsCapability = (): React.ReactElement =>
+  ApiWithoutInput({
+    name: 'checkExternalAppCommandsCapability',
+    title: 'Check External App Commands Capability',
+    onClick: async () =>
+      `External App Commands module ${externalAppCommands.isSupported() ? 'is' : 'is not'} supported`,
+  });
+
+const ProcessActionCommand = (): React.ReactElement =>
+  ApiWithTextInput<{
+    appId: string;
+    commandId: string;
+    extractedParameters: Record<string, string>;
+  }>({
+    name: 'processActionCommand',
+    title: 'Process Action Command',
+    onClick: {
+      validateInput: (input) => {
+        if (!input.appId) {
+          throw new Error('appId is required');
+        }
+        if (!input.commandId) {
+          throw new Error('commandId is required');
+        }
+        if (!input.extractedParameters) {
+          throw new Error('extractedParameters is required');
+        }
+      },
+      submit: async (input) => {
+        const result = await externalAppCommands.processActionCommand(
+          input.appId,
+          input.commandId,
+          input.extractedParameters,
+        );
+        return JSON.stringify(result);
+      },
+    },
+    defaultInput: JSON.stringify({
+      appId: 'b7f8c0a0-6c1d-4a9a-9c0a-2c3f1c0a3b0a',
+      commandId: 'testCommandId',
+      extractedParameters: { key: 'value-string' },
+    }),
+  });
+
+const ExternalAppCommandsAPIs = (): React.ReactElement => (
+  <ModuleWrapper title="External App Commands">
+    <CheckExternalAppCommandsCapability />
+    <ProcessActionCommand />
+  </ModuleWrapper>
+);
+
+export default ExternalAppCommandsAPIs;
--- a/packages/teams-js/src/private/externalAppCommands.ts
+++ b/packages/teams-js/src/private/externalAppCommands.ts
@@ -1,0 +1,166 @@
+import { sendMessageToParentAsync } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
+import { FrameContexts } from '../public';
+import { errorNotSupportedOnPlatform } from '../public/constants';
+import { runtime } from '../public/runtime';
+import { ExternalAppErrorCode } from './constants';
+import { externalAppAuthentication } from './externalAppAuthentication';
+
+/**
+ * v2 APIs telemetry file: All of APIs in this capability file should send out API version v2 ONLY
+ */
+const externalAppCommandsTelemetryVersionNumber: ApiVersionNumber = ApiVersionNumber.V_2;
+
+/**
+ * @hidden
+ * Namespace to delegate the ActionCommand to the host
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * @beta
+ */
+export namespace externalAppCommands {
+  /**
+   * @hidden
+   * The payload of IActionCommandResponse
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export type IActionCommandResponse = ITextActionCommandResponse | ICardActionCommandResponse;
+
+  /**
+   * @hidden
+   * The payload of IBaseActionCommandResponse
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface IBaseActionCommandResponse {
+    taskModuleClosedReason: TaskModuleClosedReason;
+  }
+
+  /**
+   * @hidden
+   * The text result type
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface ITextActionCommandResponse extends IBaseActionCommandResponse {
+    resultType: 'text';
+    text: string | undefined;
+  }
+
+  /**
+   * @hidden
+   * The card result type
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface ICardActionCommandResponse extends IBaseActionCommandResponse {
+    resultType: 'card';
+    attachmentLayout: externalAppAuthentication.AttachmentLayout;
+    attachments: externalAppAuthentication.QueryMessageExtensionAttachment[];
+  }
+
+  /**
+   * @hidden
+   * The result type for the ActionCommandResultType
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export type ActionCommandResultType = 'card' | 'text';
+
+  /**
+   * @hidden
+   * The reason for the TaskModuleClosedReason
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export type TaskModuleClosedReason = 'Done' | 'CancelledByUser';
+
+  /**
+   *
+   * @hidden
+   * Error that can be thrown from IExternalAppCommandsService.processActionCommand
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface ActionCommandError {
+    errorCode: ExternalAppErrorCode;
+    message?: string;
+  }
+
+  /**
+   * @internal
+   * Limited to Microsoft-internal use
+   * @hidden
+   * This API delegates an ActionCommand request to the host for the application with the provided following parameters:
+   *
+   * @param appId ID of the application the request is intended for. This must be a UUID
+   * @param commandId extensibilityProvider use this ID to look up the command declared by ActionME
+   * @param extractedParameters are the key-value pairs that the dialog will be prepopulated with
+   *
+   * @returns Promise that resolves with the {@link IActionCommandResponse} when the request is completed and rejects with {@link ActionCommandError} if the request fails
+   *
+   * @beta
+   */
+  export async function processActionCommand(
+    appId: string,
+    commandId: string,
+    extractedParameters: Record<string, string>,
+  ): Promise<IActionCommandResponse> {
+    ensureInitialized(runtime, FrameContexts.content);
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    const [error, response] = await sendMessageToParentAsync<[ActionCommandError, IActionCommandResponse]>(
+      getApiVersionTag(externalAppCommandsTelemetryVersionNumber, ApiName.ExternalAppCommands_ProcessActionCommands),
+      ApiName.ExternalAppCommands_ProcessActionCommands,
+      [appId, commandId, extractedParameters],
+    );
+    if (error) {
+      throw error;
+    } else {
+      return response;
+    }
+  }
+
+  /**
+   * @hidden
+   * Checks if the externalAppCommands capability is supported by the host
+   * @returns boolean to represent whether externalAppCommands capability is supported
+   *
+   * @throws Error if {@linkcode app.initialize} has not successfully completed
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export function isSupported(): boolean {
+    return ensureInitialized(runtime) && runtime.supports.externalAppCommands ? true : false;
+  }
+}
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -701,6 +701,14 @@ export namespace app {
      * Will be `undefined` when not running in Teams.
      */
     team?: TeamInfo;
+
+    /**
+     * When `processActionCommand` activates a dialog, this dialog should automatically fill in some fields with information. This information comes from M365 and is given to `processActionCommand` as `extractedParameters`.
+     * App developers need to use these `extractedParameters` in their dialog.
+     * They help pre-fill the dialog with necessary information (`dialogParameters`) along with other details.
+     * If there's no key/value pairs passed, the object will be empty in the case
+     */
+    dialogParameters: Record<string, string>;
   }
 
   /**
@@ -1038,6 +1046,7 @@ function transformLegacyContextToAppContext(legacyContext: LegacyContext): app.C
             mySiteDomain: legacyContext.mySiteDomain,
           }
         : undefined,
+    dialogParameters: legacyContext.dialogParameters || {},
   };
 
   return context;
--- a/apps/teams-test-app/src/App.tsx
+++ b/apps/teams-test-app/src/App.tsx
@@ -38,6 +38,7 @@ import PeopleAPIs from './components/PeopleAPIs';
 import ChatAPIs from './components/privateApis/ChatAPIs';
 import ExternalAppAuthenticationAPIs from './components/privateApis/ExternalAppAuthenticationAPIs';
 import ExternalAppCardActionsAPIs from './components/privateApis/ExternalAppCardActionsAPIs';
+import ExternalAppCommandsAPIs from './components/privateApis/ExternalAppCommandsAPIs';
 import FilesAPIs from './components/privateApis/FilesAPIs';
 import FullTrustAPIs from './components/privateApis/FullTrustAPIs';
 import MeetingRoomAPIs from './components/privateApis/MeetingRoomAPIs';
@@ -146,6 +147,7 @@ const App = (): ReactElement => {
         <DialogUrlBotAPIs />
         <ExternalAppAuthenticationAPIs />
         <ExternalAppCardActionsAPIs />
+        <ExternalAppCommandsAPIs />
         <FilesAPIs />
         <FullTrustAPIs />
         <GeoLocationAPIs />
--- a/change/@microsoft-teams-js-2759cb8f-9b8e-42d5-91f1-9328285f396e.json
+++ b/change/@microsoft-teams-js-2759cb8f-9b8e-42d5-91f1-9328285f396e.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `externalAppCommands` 1P internal-only capability",
+  "packageName": "@microsoft/teams-js",
+  "email": "maggiegong@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/internal/telemetry.ts
+++ b/packages/teams-js/src/internal/telemetry.ts
@@ -111,6 +111,7 @@ export enum ApiName {
   ExternalAppAuthentication_AuthenticateWithSSOAndResendRequest = 'externalAppAuthentication.authenticateWithSSOAndResendRequest',
   ExternalAppCardActions_ProcessActionOpenUrl = 'externalAppCardActions.processActionOpenUrl',
   ExternalAppCardActions_ProcessActionSubmit = 'externalAppCardActions.processActionSubmit',
+  ExternalAppCommands_ProcessActionCommands = 'externalAppCommands.processActionCommand',
   Files_AddCloudStorageFolder = 'files.addCloudStorageFolder',
   Files_AddCloudStorageProvider = 'files.addCloudStorageProvider',
   Files_AddCloudStorageProviderFile = 'files.addCloudStorageProviderFile',
--- a/packages/teams-js/src/private/constants.ts
+++ b/packages/teams-js/src/private/constants.ts
@@ -1,0 +1,9 @@
+/**
+ * @hidden
+ * Error codes that can be thrown from externalAppCommands and externalAppCardCommands Action Submit specifically
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export enum ExternalAppErrorCode {
+  INTERNAL_ERROR = 'INTERNAL_ERROR', // Generic error
+}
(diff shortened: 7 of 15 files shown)


packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+   * @beta
+   */
+  export interface MessageExtensionItem {
+    card: Card;
+    previewCard?: Card | null;
+  }
+
+  /**
+   * @hidden
+   * The payload of Card
```
lakhveerk, line 73: @MengyiGong, Arun updated us with a new interface. Check the wave 2 doc please

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+  export interface ActionCommandError {
+    errorCode: string;
+    message?: string;
+  }
+
+  /**
+   * @internal
+   * Limited to Microsoft-internal use
+   * @hidden
+   * This API will delegate an ActionCommand request to the extensibility provider to send the request to the backend service of the application with provided:
```
lakhveerk, line 143: with provided Id? With provided parameters? The sentence seemed incomplete. 
Also, should we say request is delegated to 'host' following the same pattern from externalAppCardActions?

lakhveerk:
We also need to edit the context interface for getContext to add dialogParameters. Is that coming in a separate PR?

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * @param appId ID of the application the request is intended for. This must be a UUID
+   * @param commandId extensibilityProvider use this ID to look up the command declared by ActionME
+   * @param extractedParameters is the parameters that the dialog will be prepopulated with
+   * @param fromElement if `fromElemetn` is undefined, send the botME as a default value
```
AE-MS, line 148: >fromElement [](http://example.com/codeflow?start=12&length=11)

This parameter doesn't seem to exist.

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+   * @returns Promise that resolves when the request is completed and rejects with ActionCommandError if the request fails
```
AE-MS, line 149: >ActionCommandError [](http://example.com/codeflow?start=83&length=18)

Consider using `{@link ActionCommandError}` here

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+   * @hidden
+   * Error that can be thrown from IExternalAppCommandsService.processActionCommand
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface ActionCommandError {
+    errorCode: string;
```
AE-MS, line 135: >string [](http://example.com/codeflow?start=15&length=6)

An error "code" sounds like a number. Is it really a string?

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * @internal
+   * Limited to Microsoft-internal use
+   * @hidden
+   * This API will delegate an ActionCommand request to the extensibility provider to send the request to the backend service of the application with provided:
+   *
+   * @param appId ID of the application the request is intended for. This must be a UUID
+   * @param commandId extensibilityProvider use this ID to look up the command declared by ActionME
+   * @param extractedParameters is the parameters that the dialog will be prepopulated with
```
AE-MS, line 147: >is the parameters [](http://example.com/codeflow?start=32&length=17)

Consider "are the key-value pairs" to make it clearer what the values are

packages/teams-js/src/private/externalAppCommands.ts
AE-MS, line 149: >that resolves [](http://example.com/codeflow?start=22&length=13)

Consider "that resolves with the {@link IActionCommandResponse}"

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * @hidden
+   * The payload of IActionCommandResponse
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface IActionCommandResponse {
```
AE-MS, line 31: >IActionCommandResponse [](http://example.com/codeflow?start=19&length=22)

Why are all the properties on this interface both able to be `undefined` and `null`? What is the sematic difference between when one is `undefined` and one is `null`?

AE-MS, line 31: >IActionCommandResponse [](http://example.com/codeflow?start=19&length=22)

All the properties on this interface can be undefined. Does that mean that a valid return value from `processActionCommand` is an empty object? What would that mean?

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * type will be result or message
+     */
+    type?: ResultTypes | null;
+    /**
+     * attachment items will exist only when type is result
```
AE-MS, line 37: >attachment items will exist only when type is result [](http://example.com/codeflow?start=7&length=52)

If this is true, please restructure the interface so that there's no `items` property when the `type` isn't `result`. Consider using a type union or inheritance to achieve this.

AE-MS: Same for `text` property but for the opposite `type`  😁

packages/teams-js/src/private/externalAppCommands.ts
AE-MS: (this same question applies to all the properties in this file that can be `undefined` or `null`)

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * @hidden
+   * The payload of Card
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface Card {
```
AE-MS, line 80: >Card [](http://example.com/codeflow?start=19&length=4)

Can you provide more documentation on what each of these properties means? Or a link to some Card documentation that describes them?

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * @beta
+   */
+  export type CardSourceType =
+    | 'ZeroInstallAppInstalledAppLinkUnfurl'
+    | 'NonZeroInstallAppInstalledAppLinkUnfurl'
+    | 'ZeroInstallAppNotInstalledAppLinkUnfurl'
+    | 'URLP'
+    | 'URLPRichInfo'
+    | 'NonLinkUnfurling';
```
AE-MS, line 122: Please alphabetize these

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    return sendMessageToParentAsync<[ActionCommandError, IActionCommandResponse]>(
+      getApiVersionTag(externalAppCommandsTelemetryVersionNumber, ApiName.ExternalAppCommands_ProcessActionCommands),
+      'externalAppCommands.processActionCommand',
+      [appId, commandId, extractedParameters],
+    ).then(([error, response]: [ActionCommandError, IActionCommandResponse]) => {
```
AE-MS, line 168: >.then [](http://example.com/codeflow?start=5&length=5)

There's nothing functionally wrong with using `.then` but consider using `await` instead to make the code clearer.

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+  ): Promise<IActionCommandResponse> {
+    ensureInitialized(runtime, FrameContexts.content);
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    return sendMessageToParentAsync<[ActionCommandError, IActionCommandResponse]>(
+      getApiVersionTag(externalAppCommandsTelemetryVersionNumber, ApiName.ExternalAppCommands_ProcessActionCommands),
+      'externalAppCommands.processActionCommand',
```
AE-MS, line 166: >'externalAppCommands.processActionCommand' [](http://example.com/codeflow?start=6&length=42)

Can you just reuse `ApiName.ExternalAppCommands_ProcessActionCommands` here rather than introducing a duplicate string literal?

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,192 @@
...
+   * @param commandId extensibilityProvider use this ID to look up the command declared by ActionME
+   * @param extractedParameters is the parameters that the dialog will be prepopulated with
+   * @param fromElement if `fromElemetn` is undefined, send the botME as a default value
+   * @returns Promise that resolves when the request is completed and rejects with ActionCommandError if the request fails
+   *
+   * @beta
+   */
+  export function processActionCommand(
+    appId: string,
+    commandId: string,
```
AE-MS, line 155: >string [](http://example.com/codeflow?start=15&length=6)

Can this command id really be any string? Is "Maggie" a valid command id?

AE-MS (changes requested):
:clock1:

MengyiGong:
> We also need to edit the context interface for getContext to add dialogParameters. Is that coming in a separate PR?

Yes, that was my planning. But considering of the deadline is due the end of this week, I will update those changes in this PR as well.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Thanks for helping double check with the comments, I will rephase it to make sure it makes more sense.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: That's a great question, and the errorCode is string in the design docs: https://microsoft-my.sharepoint-df.com/:w:/r/personal/erinha_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7Be3ed58ec-b5d5-4c11-8942-182bba9a0c7e%7D&action=edit&wdPid=17fd8a17&share=IQHsWO3j1bURTIlCGCu6mgx-AWrjzLcPCv_GPhxBpYPA1Iw @lakhveerk could you help confirming the type of the errorCode for ActionCommandError? Thanks a lot!

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Updated with the latest docs. Thanks for reminding me the updates.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Thanks for pointing this out. Updated!

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Docs updated and we don't need this data model anymore.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Updated this data model by following the latest docs. And it makes more sense that we will not have a valid return value from processActionCommand is an empty object.

MengyiGong: Thanks for pointing this out, I updated `null` to `undefinded`

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Thanks for your suggestion, updated the comment.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Thanks for your suggestion, updated!

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Thanks for your suggestion, updated the comment with the link.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Out-dated data model in the latest docs.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Thanks for pointing this out! Updated with await.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Talked with @lakhveerk offline, `Error code is INTERNAL_ERROR for everything and yes, it is string type. It just followed a pattern that we had for Cards.` 
And I updated with a generic errorCode type named `ExternalAppErrorCode` in constants file for externalAppCommands api and external-app-card for Action Submit specifically.

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: Commands are defined by an external developer. We validate it in web-SDK by making sure the commandId from this API is defined in manifest. If the commandId comes in that is not defined in manifest (and hence in app definition), web-SDK throws an error. Also, we cannot valid the commandId here, because we don't have access to the external app definition in teams-js.

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * @hidden
+   * The payload of IActionCommandResponse
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export interface IActionCommandResponse {
```
lakhveerk, line 31: Maggie you use this interface structure

```
interface IBaseActionCommandResponse {
  taskModuleClosedReason: TaskModuleClosedReason;
}

interface ITextActionCommandResponse extends IBaseActionCommandResponse {
  resultType: 'text';
  text: string | null;
}

interface ICardActionCommandResponse extends IBaseActionCommandResponse {
  resultType: 'card';
  attachmentLayout: AttachmentLayout;
  attachments: QueryMessageExtensionAttachment[];
}

export type IActionCommandResponse = ITextActionCommandResponse | ICardActionCommandResponse;

export type TaskModuleClosedReason = 'Done' | 'CancelledByUser';
```

packages/teams-js/src/public/constants.ts
```diff
@@ -201,3 +201,13 @@ export const errorInvalidCount = new Error('Invalid input count: Must supply a v
 export const errorInvalidResponse = new Error(
   'Invalid response: Received more images than the specified max limit in the response.',
 );
+
+/**
+ * @hidden
+ * Error codes that can be thrown from externalAppCommands and externalAppCardCommands Action Submit specifically
```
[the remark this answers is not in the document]
lakhveerk: oh, better idea, you can add it to existing, private/interfaces.ts file.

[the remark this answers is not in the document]
MengyiGong: Thanks! That makes sense, updated with a new constants.ts file under the private folder!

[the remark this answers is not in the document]
MengyiGong: @lakhveerk that's also a great idea. I have updated it with a new constants.ts file in the private folder for now. Because everything in the private/interface.ts file are interfaces. It would be great to keep this file following the interfaces pattern. What do you think?  I'm open-minded about both approaches!

apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
```diff
@@ -0,0 +1,58 @@
...
+      `External App Commands module ${externalAppCommands.isSupported() ? 'is' : 'is not'} supported`,
+  });
+
+const ProcessActionCommand = (): React.ReactElement =>
+  ApiWithTextInput<{
+    appId: string;
+    commandId: string;
+    extractedParameters: Record<string, string>;
+  }>({
+    name: 'ProcessActionCommand',
```
KangxuanYe, line 21: processActionCommand

KangxuanYe (changes requested):

apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
MengyiGong: updated!

apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
```diff
@@ -0,0 +1,58 @@
...
+          input.appId,
+          input.commandId,
+          input.extractedParameters,
+        );
+        return JSON.stringify(result);
+      },
+    },
+    defaultInput: JSON.stringify({
+      appId: 'b7f8c0a0-6c1d-4a9a-9c0a-2c3f1c0a3b0a',
+      commandId: 'command-Id',
```
KangxuanYe, line 46: Please update this to commandId field with value `testCommandId` to be more meaningful and get aligned with change I suggested to Lakhveer in web host sdk side.

MengyiGong: Sure, updated!

packages/teams-js/src/private/externalAppCommands.ts
```diff
@@ -0,0 +1,208 @@
...
+   * The data model of QueryMessageExtensionAttachment
+   *
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * @beta
+   */
+  export type QueryMessageExtensionAttachment = QueryMessageExtensionCard & {
+    preview?: QueryMessageExtensionCard;
+  };
```
AE-MS, line 139: This definition of `QueryMessageExtensionAttachment` allows for `preview` to be `undefined`. Isn't that just a `QueryMessageExtensionCard` then? Would the only value of having a new `QueryMessageExtensionAttachment` type be if the `preview` property was actually present (i.e., not `undefined`)?

packages/teams-js/src/public/app.ts
```diff
@@ -701,6 +701,13 @@ export namespace app {
      * Will be `undefined` when not running in Teams.
      */
     team?: TeamInfo;
+
+    /**
+     * When `processActionCommand` activates a dialog, this dialog should automatically fill in some fields with information. This information comes from M365 and is given to `processActionCommand` as `extractedParameters`.
+     * App developers need to use these `extractedParameters` in their dialog.
+     * They help pre-fill the dialog with necessary information (`dialogParameters`) along with other details.
```
AE-MS, line 708: How should developers treat the case where `dialogParameters` is `undefined` differently from where it is defined but it's an empty object?

packages/teams-js/src/private/externalAppCommands.ts
MengyiGong: `This definition of QueryMessageExtensionAttachment allows for preview to be undefined. Isn't that just a QueryMessageExtensionCard then?` Yes, it is. When the preview is defined, the type is QueryMessageExtensionAttachment; otherwise the type is QueryMessageExtensionCard. Thanks @lakhveerk for confirming the logic!

MengyiGong: I updated the comment to make sure well-explained the logic to app developers.

packages/teams-js/src/public/app.ts
MengyiGong: That's a really good question, and here's the confirmation I received: We don't do anything differently when there's an empty object for dialog parameters here. Even if it's an empty array, it indicates that no key/value was passed by M365.

packages/teams-js/src/public/app.ts
```diff
@@ -701,6 +701,13 @@ export namespace app {
...
+     */
+    dialogParameters?: Record<string, string>;
```
lakhveerk, line 710: We also need to add the parameter in legacyContext. hubSDK sends over the legacy context, and teams-js change it to the app context. I will show you an example in hubSDk

packages/teams-js/src/public/app.ts
AE-MS: Does that mean that in order for developers to know that there are no key/value parameters, they have to check BOTH that this is defined AND that the object is not empty?

Could we make their job easier by just _always_ having a defined object here and making the object empty in the case that there's no key/value pairs passed?

MengyiGong: `Could we make their job easier by just always having a defined object here and making the object empty in the case that there's no key/value pairs passed` that totally makes sense. 
How about this: 
we will have `dialogParameters?: Record<string, string>;` in the legacy context which send over by hubSDK, and
`dialogParameters: Record<string, string>;` in the app.context. So that we can always do `dialogParameters: legacyContext.dialogParameters || {},` in the`transformLegacyContextToAppContext` function to reduce the checking logic from app.context?

packages/teams-js/src/public/app.ts
MengyiGong: Updated! Thanks for letting me know the legacyContext!!

packages/teams-js/src/public/app.ts
lakhveerk: I think there are downsides of defining the dialogParameters all the time - 
1. App developers will think they can make use of these parameters when they cannot. The parameters will be available only when M365 open a dialog for ME
2. dialogParameters should only be present in the 'task' frameContext and no where else. I am adding some logic in hubSDk to ensure this. 

I think, I would rather have developers check for undefined and empty, but also open for other opinions. Bring you votes everybody! @erinha @MengyiGong @AE-MS

AE-MS: Can you say more about your concern for #1? If the object is empty, won't that be the signal to app developers that the object can't be used? If that's not what a defined empty object means, then what does it mean?

lakhveerk: > we will have dialogParameters?: Record<string, string>; in the legacy context which send over by hubSDK, and
> dialogParameters: Record<string, string>; in the app.context. So that we can always do dialogParameters: legacyContext.dialogParameters || {}, in thetransformLegacyContextToAppContext function to reduce the checking logic from app.context?

I like the idea Maggie, let's go with that. I also agree with Aaron's point that external app developers won't do anything differently when the object is empty vs undefined. It does not give them any extra info.

MengyiGong: > `Could we make their job easier by just always having a defined object here and making the object empty in the case that there's no key/value pairs passed` that totally makes sense. How about this: we will have `dialogParameters?: Record<string, string>;` in the legacy context which send over by hubSDK, and `dialogParameters: Record<string, string>;` in the app.context. So that we can always do `dialogParameters: legacyContext.dialogParameters || {},` in the`transformLegacyContextToAppContext` function to reduce the checking logic from app.context?
> 
> Thoughts? @lakhveerk

Thanks for both of your great suggestions! @AE-MS @lakhveerk I updated the `dialogParameters` to defined value in app.Context and undefined `dialogParameters` in `legacyContext`. The undefine dialogParameters from hubSDK mapped as `dialogParameters: legacyContext.dialogParameters || {}`, in thetransformLegacyContextToAppContext function. Let me know if this sounds good. :)

AE-MS (approved):
:shipit:

apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
```diff
@@ -0,0 +1,58 @@
...
+          input.commandId,
+          input.extractedParameters,
+        );
+        return JSON.stringify(result);
+      },
+    },
+    defaultInput: JSON.stringify({
+      appId: 'b7f8c0a0-6c1d-4a9a-9c0a-2c3f1c0a3b0a',
+      commandId: 'testCommandId',
+      extractedParameters: { key: 'value-string' },
```
lakhveerk, line 47: the default value can be 'testParamName1' instead of key. This is what @KangxuanYe suggested for the app definition

KangxuanYe (approved):
Approved with e2e test verified in local

apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
KangxuanYe: Oh yes, didn't I mention it here? Please use `extractedParameters: { testParamName1: 'testValue1' }` instead.

KangxuanYe: I am okay with this PR gets checked in because there will be another PR adding testing json file and you can raise another PR to fix it. Not a big deal and depends on you.

lakhveerk (approved):

apps/teams-test-app/src/components/privateApis/ExternalAppCommandsAPIs.tsx
MengyiGong: Thanks @lakhveerk  @KangxuanYe ! I will make sure I address this testing strings in the testing json file PR.

Issue #2223, opened 2024-03-14
Error loading Teams app in Microsoft365.com (errorCode: 5600, "AppID is not valid. Must be GUID format")


pablodiazgutierrez:
We have 3 nearly-identical apps published on Microsoft Teams, with only different market targets. When we load them in https://microsoft365.com/, only one works, and the other two fail with a console error message stating "AppID is not valid. Must be GUID format", even though the AppID is a unique, valid GUID.

When we sideload the apps, they work fine, as confirmed by MSFT engineering staff. But once it's gone through validation and it's served through AppSource, the problems begin.

Can anyone provide any hints on what to look for?

ChetanSharma-msft:
Hello @pablodiazgutierrez - Thanks for raising your query.
Could you please share the manifest packages so that we can test it from our end?

pablodiazgutierrez:
> Hello @pablodiazgutierrez - Thanks for raising your query. Could you please share the manifest packages so that we can test it from our end?

Absolutely, thanks @ChetanSharma-msft, please see attached.

[pm-hipaa-1.1.27.zip](https://github.com/OfficeDev/microsoft-teams-library-js/files/14616071/pm-hipaa-1.1.27.zip)

TrevorJoelHarris:
Hi @pablodiazgutierrez, I'm really sorry you hit this issue. I thought this was interesting and so I dug into it a little bit. I am *pretty* sure that I know what the problem is:

The app id you gave in your manifest for the app that is failing is `2b248bdc-b79d-35eb-ece8-e81c1ba9fcef`.

While that does *look* like a valid UUID it actually is not. I did some digging and found someone who actually wrote a pretty clear regex of what a valid UUID needs to be: 

`/^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i` (from: https://stackoverflow.com/a/13653180)

In your case, the 17th character of any UUID has to be in the set `[89ab]`, but the 17th character of your UUID is `e`:

2b248bdc-b79d-35eb-**E**ce8-e81c1ba9fcef

Can you try uploading your app again with a valid UUID and see if that fixes the problem? Let me know!

pablodiazgutierrez:
[...]

PR #2233, opened 2024-03-25
Removed all of GUID checks

KangxuanYe:

## Description

GUID checks are removed for app Id because apps are not promised to have Id in GUID format.

We had a bug ticket saying that appID works in Teams but failed in Teams-JS library because it is not GUID format. That is to say we are too strict to App Id format so we need to loose it a little bit to be aligned with Teams.
https://github.com/OfficeDev/microsoft-teams-library-js/issues/2223

### Main changes in the PR:

1. Remove everything related to GUID check.

### Validation performed:

1. isValidAppId function is removed so it should be compiled successfully and no where is using that function.

### DIFF
--- a/packages/teams-js/src/internal/utils.ts
+++ b/packages/teams-js/src/internal/utils.ts
@@ -402,16 +402,3 @@ export function inServerSideRenderingEnvironment(): boolean {
 }
 
 const appIdRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
-
-/**
- * @param appID The app ID to validate against the GUID format
- * @throws Error if appID is not a valid GUID
- *
- * @internal
- * Limited to Microsoft-internal use
- */
-export function validateAppIdIsGuid(appId: string): void {
-  if (!appIdRegex.test(appId)) {
-    throw new Error('App ID is not valid. Must be GUID format. App ID: ' + appId);
-  }
-}
--- a/packages/teams-js/src/internal/utils.ts
+++ b/packages/teams-js/src/internal/utils.ts
@@ -400,5 +400,3 @@ export function ssrSafeWindow(): Window {
 export function inServerSideRenderingEnvironment(): boolean {
   return typeof window === 'undefined';
 }
-
-const appIdRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
--- a/change/@microsoft-teams-js-66c6dadf-b3af-4de0-b855-89376c0f873a.json
+++ b/change/@microsoft-teams-js-66c6dadf-b3af-4de0-b855-89376c0f873a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Removed all of GUID checks on app Id",
+  "packageName": "@microsoft/teams-js",
+  "email": "kangxuanye@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-66c6dadf-b3af-4de0-b855-89376c0f873a.json
+++ b/change/@microsoft-teams-js-66c6dadf-b3af-4de0-b855-89376c0f873a.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Removed all of GUID checks on app Id",
+  "comment": "Removed validation that appIds are UUIDs since some very old published apps have IDs that are not UUIDs (they were published before the manifest schema specified they had to be UUIDs)",
   "packageName": "@microsoft/teams-js",
   "email": "kangxuanye@microsoft.com",
   "dependentChangeType": "patch"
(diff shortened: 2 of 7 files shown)


jadahiya-MSFT:
Can you add to the description why this change is needed?

packages/teams-js/src/internal/utils.ts
```diff
@@ -402,16 +402,3 @@ export function inServerSideRenderingEnvironment(): boolean {
 }
 
 const appIdRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
jekloudaMSFT: Is this still used somewhere?

KangxuanYe:
> Can you add to the description why this change is needed?

Just updated.

AE-MS (approved):

jekloudaMSFT (approved):

jadahiya-MSFT (approved):

packages/teams-js/src/internal/utils.ts
KangxuanYe: Nope. So I removed it. Thanks for pointing it out.

pablodiazgutierrez:
Hello, thank you for this, and sorry about the delay in sharing this. We just got notice from the "Teams Store App Suspension Services" that one of our apps still receives this UUID validation error, and is slated for removal from the store because of that reason. From what I can see, there are no changes to what apps are able to load or not on microsoft365.com. Specifically, this is our situation:

- 5be2b320-a5b7-4221-893c-dee506e4e365 is the original app, which has always worked fine
- ffca97e5-0569-c88c-ac30-7c5717e2e5ae continues to show the "AppID is not valid. Must be GUID format" message in the console
- 2b248bdc-b79d-35eb-ece8-e81c1ba9fcef also breaks, same error message

Do you know if these changes have been deployed, or if perhaps they were somehow reverted elsewhere?
CC @TrevorJoelHarris who figured out the root cause in the 1st place

AE-MS:
> Hello, thank you for this, and sorry about the delay in sharing this. We just got notice from the "Teams Store App Suspension Services" that one of our apps still receives this UUID validation error, and is slated for removal from the store because of that reason. From what I can see, there are no changes to what apps are able to load or not on microsoft365.com. Specifically, this is our situation:
> 
> * 5be2b320-a5b7-4221-893c-dee506e4e365 is the original app, which has always worked fine
> * ffca97e5-0569-c88c-ac30-7c5717e2e5ae continues to show the "AppID is not valid. Must be GUID format" message in the console
> * 2b248bdc-b79d-35eb-ece8-e81c1ba9fcef also breaks, same error message
> 
> Do you know if these changes have been deployed, or if perhaps they were somehow reverted elsewhere? CC @TrevorJoelHarris who figured out the root cause in the 1st place

These changes were first in the 2.22 release of teamsjs and have remained until the current release and have not been reverted. We can look to see if there are other parts of the app platform that are enforcing UUID validity but it seems like removing/suppressing all valid UUID checks throughout the full stack may be a longer road (especially since the validation appears to be accurate according to the standard).

Were you able to update the UUID (2b248bdc-b79d-35eb-ece8-e81c1ba9fcef) so that it conforms to the valid format that @TrevorJoelHarris outlined in #2223? That would seem like the most direct way to workaround this issue.

pablodiazgutierrez:
Thank you @AE-MS. Unfortunately, updating the app ID would require losing all our registered users and subscribers, requiring them to install a new add-on, which is highly disruptive from the business standpoint.
I see that the validation is done via the [uuid](https://github.com/uuidjs/uuid) library, which is quite strict and doesn't provide a "loose" option.
I'm going to ask the store team for assistance, to see if they've dealt with this before.

AE-MS:
> Thank you @AE-MS. Unfortunately, updating the app ID would require losing all our registered users and subscribers, requiring them to install a new add-on, which is highly disruptive from the business standpoint.
> 
> I see that the validation is done via the [uuid](https://github.com/uuidjs/uuid) library, which is quite strict and doesn't provide a "loose" option.
> 
> I'm going to ask the store team for assistance, to see if they've dealt with this before.

Do you know if the aforementioned console log statement is what the store team is using as a justification for potential removal? Or are there other ways the UUID issue manifests itself?

pablodiazgutierrez:
> Do you know if the aforementioned console log statement is what the store team is using as a justification for potential removal? Or are there other ways the UUID issue manifests itself that the store team is looking at?

The main issue is that the app (which was originally a Teams app) won't load in certain hosts like microsoft365.com. But we're also encountering issues failing silently which we strongly suspect are related to this root cause, because when we sideload an identical manifest with a valid UUID, they magically just work. Some examples off the top of my head:

- The Outlook compose message extension won't show up in the "apps" popup even though it's installed
- The "edit" action of adaptive cards in Teams fails without attempting to send us a request
- Adaptive cards also fail to reload

It might be too much of a hassle to get all those validations match the original algorithm that Teams still uses, but it's also unfortunate that our apps, being among the first in AppSource, are being penalized for being early. (Sorry, I'm just frustrated; I don't mean to seem as if I'm looking for someone to blame.)

PR #2235, opened 2024-03-26
Adding optional parameter 'fromElement' to openUrl

lakhveerk:

## Description

Hosts have a dependency on the element and cannot open a url without knowing what type (composeExtension vs plugin), the request has come for. The openUrl is supported for both elements.

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/packages/teams-js/src/private/externalAppCardActions.ts
+++ b/packages/teams-js/src/private/externalAppCardActions.ts
@@ -1,7 +1,6 @@
 import { sendMessageToParentAsync } from '../internal/communication';
 import { ensureInitialized } from '../internal/internalAPIs';
 import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
-import { validateAppIdIsGuid } from '../internal/utils';
 import { FrameContexts } from '../public';
 import { errorNotSupportedOnPlatform } from '../public/constants';
 import { runtime } from '../public/runtime';
@@ -106,9 +105,6 @@ export namespace externalAppCardActions {
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
-
-    validateAppIdIsGuid(appId);
-
     return sendMessageToParentAsync<[boolean, ActionSubmitError]>(
       getApiVersionTag(
         externalAppCardActionsTelemetryVersionNumber,
@@ -139,8 +135,6 @@ export namespace externalAppCardActions {
     if (!isSupported()) {
       throw errorNotSupportedOnPlatform;
     }
-
-    validateAppIdIsGuid(appId);
 
     return sendMessageToParentAsync<[ActionOpenUrlError, ActionOpenUrlType]>(
       getApiVersionTag(
--- a/packages/teams-js/src/private/externalAppCardActions.ts
+++ b/packages/teams-js/src/private/externalAppCardActions.ts
@@ -114,12 +114,13 @@ export namespace externalAppCardActions {
    * @beta
    * @hidden
    * Delegates an Adaptive Card Action.OpenUrl request to the host for the application with the provided app ID.
-   * Currently `processActionOpenUrl` is supported only for composeExtensions and not for plugins
+   * If `fromElement` is not provided, the information from the manifest is used to determine whether the URL can
+   * be processed by the host. Deep link URLs for plugins are not supported and will result in an error.
    * @internal
    * Limited to Microsoft-internal use
    * @param appId ID of the application the request is intended for. This must be a UUID
    * @param url The URL to open
-   * @param fromElement The element on behalf of which the M365 app is making the request
+   * @param fromElement The element on behalf of which the M365 app is making the request.
    * @returns Promise that resolves to ActionOpenUrlType indicating the type of URL that was opened on success and rejects with ActionOpenUrlError if the request fails
    */
   export function processActionOpenUrl(
--- a/packages/teams-js/src/private/externalAppCardActions.ts
+++ b/packages/teams-js/src/private/externalAppCardActions.ts
@@ -127,9 +127,14 @@ export namespace externalAppCardActions {
    * Limited to Microsoft-internal use
    * @param appId ID of the application the request is intended for. This must be a UUID
    * @param url The URL to open
+   * @param fromElement The element on behalf of which the M365 app is making the request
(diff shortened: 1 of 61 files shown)


packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -127,9 +127,14 @@ export namespace externalAppCardActions {
...
    * @param appId ID of the application the request is intended for. This must be a UUID
    * @param url The URL to open
+   * @param fromElement The element on behalf of which the M365 app is making the request
    * @returns Promise that resolves to ActionOpenUrlType indicating the type of URL that was opened on success and rejects with ActionOpenUrlError if the request fails
    */
-  export function processActionOpenUrl(appId: string, url: URL): Promise<ActionOpenUrlType> {
+  export function processActionOpenUrl(
+    appId: string,
+    url: URL,
+    fromElement?: { name: 'composeExtensions' | 'plugins' },
```
AE-MS, line 136: >? [](http://example.com/codeflow?start=15&length=1)

Is there a reason that this has to be optional? Can't we just make the app that needs this always specify?

At minimum we should document the default behavior when this isn't specified.

AE-MS (changes requested):
:clock1:

packages/teams-js/src/private/externalAppCardActions.ts
lakhveerk: We have to make it optional because M365 cannot send the info just yet. On hostSDK side, Hosts are not ready to consume a new parameter yet either. Adding an optional will not break them. It would let them consume the new parameter when they are ready.

Hosts have a dependency on element type, and they need to do things differently based on composeExtension vs plugin. Right now, they can handle just composeExtensions.  Even if they could handle plugins today, sending a default value won't be beneficial because they would not be able to trust the value and execute the request.

lakhveerk: fromElement will be useful only when M365 sends the value. 

Tagging Erin @erinha and Saurabh @sbtron

AE-MS: Does that mean that this parameter will eventually be made required?

Can you please document the behavior in the description of the `fromElement` parameter on line 130 for what happens if a consumer doesn't provide a value?

AE-MS (approved):

packages/teams-js/test/private/externalAppCardActions.spec.ts
```diff
@@ -142,7 +144,7 @@ describe('externalAppCardActions', () => {
           const message = utils.findMessageByFunc('externalAppCardActions.processActionOpenUrl');
           if (message && message.args) {
             expect(message).not.toBeNull();
-            expect(message.args).toEqual([testAppId, testUrl.href]);
+            expect(message.args).toEqual([testAppId, testUrl.href, null]);
```
jekloudaMSFT, line 147: Why is this `null` and not `undefined`?

jekloudaMSFT (dismissed):

packages/teams-js/src/private/externalAppCardActions.ts
```diff
@@ -113,14 +113,20 @@ export namespace externalAppCardActions {
   /**
    * @beta
    * @hidden
-   * Delegates an Adaptive Card Action.OpenUrl request to the host for the application with the provided app ID
+   * Delegates an Adaptive Card Action.OpenUrl request to the host for the application with the provided app ID.
+   * Currently `processActionOpenUrl` is supported only for composeExtensions and not for plugins
```
erinha, line 117: That isn't quite true. Generic URLs are supported for plugins.

packages/teams-js/src/private/externalAppCardActions.ts
erinha: Does this accurately describe the logic?

If `fromElement` is not provided, the information from the manifest is used to determine whether the URL can be processed by the host. Deep link URLs for plugins are not supported and will result in an error.

erinha (changes requested):
Question about documentation comments.

packages/teams-js/src/private/externalAppCardActions.ts
lakhveerk: But can Teams process them without knowing the 'fromElement'? From my understanding, Teams cannot process any urls for plugins without knowing the url is for plugin. And they are not going to know plugin vs composeExtension until M365 chat send the info

packages/teams-js/test/private/externalAppCardActions.spec.ts
lakhveerk: I spent a lot of time here, but could not find where exactly the undefined is changed to null. I think it is happening somewhere in /test/utils.ts. The type of fromElement is undefined everywhere in communication file but when I look it up from utils.findMessageByFunc, the arg come as null instead of undefined. 

But from teams-js to hubsdk 'undefined' is sent. I think that matters more. If I find more time I will investigate more

jekloudaMSFT: Yeah I agree, as long as what we send to host matches what we expect there it should be fine

erinha (approved):
Looks good to me, thank you!

AE-MS (approved):

packages/teams-js/test/private/externalAppCardActions.spec.ts
lakhveerk: I found that in communication file the message is stringified before sending across the wire. 
 ```
      logger('Sending message %i to ${targetWindowName} via framelessPostMessage interface', request.id);
      (Communication.currentWindow as ExtendedWindow).nativeInterface.framelessPostMessage(JSON.stringify(request));
    }
```

When we stringify a undefined value in an array, it is converted to null
image

on host sdk we still receive undefined - that is still mystery to me

PR #2237, opened 2024-03-26
Telemetry string size reduction

noahdarveau-MSFT:

## Description

### Main changes in the PR:

1. Made the telemetry string enum `const` to reduce package size. By making the enum `const` the minified code will symbolize the enum, reducing the package size by ~10%. In typescript, enums by default, have reverse mapping capabilities.

For example (reverse mapping capabilites):
````
enum Enum {
  A,
}
let a = Enum.A;
let nameOfA = Enum[a]; // "A"
````

So given the following enum:
````
enum Enum {
  foo = "bar",
  foo_two = "bar_two"
}
````
When a regular enum is minified, both the left and right sides of the enum are not reduced, resulting in something resembling the following when minified:
````
a.foo="bar",a.foo_two="bar_two"
````
However, a const enum removes the reverse mapping capabilites, which allows for the left side of an enum to be symbolized.

So if the above enum were `const`, it becomes something resembling the following when minified:
````
a.a="bar",a.b="bar_two"
````

Since the ApiName enum has so many values, with many of the values containing a large number of characters on the left side, making it const results in the left size symbolization being a substantial size decrease to the minified file.
### Validation performed:

1. Manually went through `ApiName` usages and ensured there were no reverse mapping usages, as well as confirmed with Kangxuan

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/change/@microsoft-teams-js-889e1105-9a8b-468a-94d7-ed8253cdf569.json
+++ b/change/@microsoft-teams-js-889e1105-9a8b-468a-94d7-ed8253cdf569.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Made ApiName enum const to reduce package size",
+  "comment": "Made enum const to reduce package size",
   "packageName": "@microsoft/teams-js",
   "email": "noahdarveau@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-889e1105-9a8b-468a-94d7-ed8253cdf569.json
+++ b/change/@microsoft-teams-js-889e1105-9a8b-468a-94d7-ed8253cdf569.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Made ApiName enum const to reduce package size",
+  "packageName": "@microsoft/teams-js",
+  "email": "noahdarveau@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/internal/telemetry.ts
+++ b/packages/teams-js/src/internal/telemetry.ts
@@ -50,7 +50,7 @@ export enum ApiVersionNumber {
   V_3 = 'v3',
 }
 
-export enum ApiName {
+export const enum ApiName {
   App_GetContext = 'app.getContext',
   App_Initialize = 'app.initialize',
   App_NotifyAppLoaded = 'app.notifyAppLoaded',

### CODE REVIEW

**Review by MengyiGong — DISMISSED** [review]

**Comment by maglims** [comment]
wow, 10% size reduction is very impressing! Great work!

File: packages/teams-js/src/internal/telemetry.ts [hunk 1]
```diff
@@ -50,7 +50,7 @@ export enum ApiVersionNumber {
   V_3 = 'v3',
```
maglims, line 50: Will making ApiVersionNumber a const also help? (if it can be made as a const)

change/@microsoft-teams-js-889e1105-9a8b-468a-94d7-ed8253cdf569.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Made ApiName enum const to reduce package size",
```
TrevorJoelHarris, line 3: Might make this more generic; since the ApiName enum isn't exported for customers this is maybe more detail than they need.

```suggestion
  "comment": "Made enum const to reduce package size",
```

TrevorJoelHarris (dismissed):

packages/teams-js/src/internal/telemetry.ts
TrevorJoelHarris: Maybe one other question I will attach here: is there something like a linting rule we could enable to prevent people from making enums that don't use the const keyword?

noahdarveau-MSFT: Technically yes, making any enum a const enum that can be should theoretically reduce package size, but since it's such a small enum it would be negligible change. (I think it would saves 12 bytes). The reason making the `ApiName` telemetry strings `const` saves so much space is because it is such a big enum with both a lot of values, and the left side names of the values are a lot of characters as well. In typescript, enums by default have reverse mappings, but const enums do not, so making it const allows all of the left side names to be symbolized in the minified code.

noahdarveau-MSFT: > Maybe one other question I will attach here: is there something like a linting rule we could enable to prevent people from making enums that don't use the const keyword?

We could definitely make a lint rule, however keep in mind that there are times when a regular, non-const enum is the right move, so if we did make a rule, it should just be a warning/reminder rather than a blocking error.

shrshindeMSFT:
Great work. I have just one tiny comment, can you please update the PR description on why we are making this change and const enum vs enum.

AE-MS:
Thanks for writing a super-clear PR description!

AE-MS (approved):

jadahiya-MSFT (approved):

maglims (approved):

MengyiGong (approved):

PR #2240, opened 2024-03-28
Adding UUIDs to MessageRequests and MessageResponses

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Added in `uuid` parameter to `MessageRequest` and `MessageResponse` objects
2. Modified `id` to be either a string or number now.


AE-MS:
Just to let you know my philosophy as I review this:
- The specific message interfaces are fundamentally wire protocol structures (e.g. serialized structures) so I would expect that they would use `string` to represent UUIDs, since UUIDs would have to be a type that isn't an serialized type (it can be serialized into a `string`)
- All other uses of UUID throughout the code should use a strongly-typed UUID object to ensure validity and proper usage
- The UUID objects should be created when the messages are deserialized and then turned into a `string` when serialized into a message.

I haven't started looking at the code yet (that's next!) but this is how I'm thinking about reviewing it. Would love to hear more on your thinking about this philosophy and how @TrevorJoelHarris feels about it as well. If we disagree on the philosophy or discuss and decide it should be something different, some of my comments may no longer be valid. 😁

packages/teams-js/src/internal/communication.ts
```diff
@@ -47,15 +47,18 @@ class CommunicationPrivate {
   public static topMessageQueue: MessageRequest[] = [];
   public static nextMessageId = 0;
   public static callbacks: {
-    [id: number]: Function; // (arg1, arg2, ...etc) => void
+    [id: number | string]: Function; // (arg1, arg2, ...etc) => void
```
AE-MS, line 50: >string [](http://example.com/codeflow?start=18&length=6)

Can you create a UUID type that can be used to hold UUIDs and be passed around throughout the code?

jadahiya-MSFT:
> Just to let you know my philosophy as I review this:
> 
> * The specific message interfaces are fundamentally wire protocol structures (e.g. serialized structures) so I would expect that they would use `string` to represent UUIDs, since UUIDs would have to be a type that isn't an serialized type (it can be serialized into a `string`)
> * All other uses of UUID throughout the code should use a strongly-typed UUID object to ensure validity and proper usage
> * The UUID objects should be created when the messages are deserialized and then turned into a `string` when serialized into a message.
> 
> I haven't started looking at the code yet but this is how I'm thinking about the shape of these changes. Would love to hear more on your thinking about this philosophy and how @TrevorJoelHarris feels about it as well. I don't want to start reviewing until we agree on this because the way I look at the code will vary wildly if our perspectives differ here.

That's an interesting perspective, some of the design considerations were made to not break communication between older versions of the Hosts with newer versions of this SDK and vice versa, newer versions of the Hosts with older versions of this SDK, but I believe that can be reflected in the serialization/deserialization part of the code. I will look into incorporating some of this feedback in the meantime.

AE-MS:
> > Just to let you know my philosophy as I review this:
> > 
> > * The specific message interfaces are fundamentally wire protocol structures (e.g. serialized structures) so I would expect that they would use `string` to represent UUIDs, since UUIDs would have to be a type that isn't an serialized type (it can be serialized into a `string`)
> > * All other uses of UUID throughout the code should use a strongly-typed UUID object to ensure validity and proper usage
> > * The UUID objects should be created when the messages are deserialized and then turned into a `string` when serialized into a message.
> > 
> > I haven't started looking at the code yet but this is how I'm thinking about the shape of these changes. Would love to hear more on your thinking about this philosophy and how @TrevorJoelHarris feels about it as well. I don't want to start reviewing until we agree on this because the way I look at the code will vary wildly if our perspectives differ here.
> 
> That's an interesting perspective, some of the design considerations were made to not break communication between older versions of the Hosts with newer versions of this SDK and vice versa, newer versions of the Hosts with older versions of this SDK, but I believe that can be reflected in the serialization/deserialization part of the code. I will look into incorporating some of this feedback in the meantime.

I don't know if having an "interesting" perspective is a good or bad thing? 😅

My desire to use strongly-typed UUID object should only affect the internal implementation within teams-js and not affect back-compat since the serialized message types wouldn't be affected by the internal-to-teamsjs usage.

jadahiya-MSFT:
> > > Just to let you know my philosophy as I review this:
> > > 
> > > * The specific message interfaces are fundamentally wire protocol structures (e.g. serialized structures) so I would expect that they would use `string` to represent UUIDs, since UUIDs would have to be a type that isn't an serialized type (it can be serialized into a `string`)
> > > * All other uses of UUID throughout the code should use a strongly-typed UUID object to ensure validity and proper usage
> > > * The UUID objects should be created when the messages are deserialized and then turned into a `string` when serialized into a message.
> > > 
> > > I haven't started looking at the code yet but this is how I'm thinking about the shape of these changes. Would love to hear more on your thinking about this philosophy and how @TrevorJoelHarris feels about it as well. I don't want to start reviewing until we agree on this because the way I look at the code will vary wildly if our perspectives differ here.
> > 
> > 
> > That's an interesting perspective, some of the design considerations were made to not break communication between older versions of the Hosts with newer versions of this SDK and vice versa, newer versions of the Hosts with older versions of this SDK, but I believe that can be reflected in the serialization/deserialization part of the code. I will look into incorporating some of this feedback in the meantime.
> 
> I don't know if having an "interesting" perspective is a good or bad thing? 😅
> 
> My desire to use strongly-typed UUID object should only affect the internal implementation within teams-js and not affect back-compat since the serialized message types wouldn't be affected by the internal-to-teamsjs usage.

Poor choice of words on my part. By interesting idea, I meant that I liked it and wanted to look into making the changes to the code to reflect it. 😅 And my second sentence was more to say that using a strongly-typed UUID object wouldn't affect back compat since we can always serialize it to a string or number (for back compat) before putting the message on the wire as you brought up above.

AE-MS:
> > > > Just to let you know my philosophy as I review this:
> > > > 
> > > > * The specific message interfaces are fundamentally wire protocol structures (e.g. serialized structures) so I would expect that they would use `string` to represent UUIDs, since UUIDs would have to be a type that isn't an serialized type (it can be serialized into a `string`)
> > > > * All other uses of UUID throughout the code should use a strongly-typed UUID object to ensure validity and proper usage
> > > > * The UUID objects should be created when the messages are deserialized and then turned into a `string` when serialized into a message.
> > > > 
> > > > I haven't started looking at the code yet but this is how I'm thinking about the shape of these changes. Would love to hear more on your thinking about this philosophy and how @TrevorJoelHarris feels about it as well. I don't want to start reviewing until we agree on this because the way I look at the code will vary wildly if our perspectives differ here.
> > > 
> > > 
> > > That's an interesting perspective, some of the design considerations were made to not break communication between older versions of the Hosts with newer versions of this SDK and vice versa, newer versions of the Hosts with older versions of this SDK, but I believe that can be reflected in the serialization/deserialization part of the code. I will look into incorporating some of this feedback in the meantime.
> > 
> > 
> > I don't know if having an "interesting" perspective is a good or bad thing? 😅
> > My desire to use strongly-typed UUID object should only affect the internal implementation within teams-js and not affect back-compat since the serialized message types wouldn't be affected by the internal-to-teamsjs usage.
> 
> Poor choice of words on my part. By interesting idea, I meant that I liked it and wanted to look into making the changes to the code to reflect it. 😅 And my second sentence was more to say that using a strongly-typed UUID object wouldn't affect back compat since we can always serialize it to a string or number (for back compat) before putting the message on the wire as you brought up above.

Sweet and thank you for considering my input and the value of type-safety across our codebase. And won't it be fun to define a custom class to hold a UUID? And then we can start using it elsewhere in the code, too!

packages/teams-js/src/internal/communication.ts
```diff
@@ -862,8 +883,12 @@ function createMessageRequest(
  * @returns {NestedAppAuthRequest} Returns a NestedAppAuthRequest object with a unique id, the function name set to 'nestedAppAuthRequest', the current timestamp, an empty args array, and the provided message as data.
  */
 function createNestedAppAuthRequest(message: string): NestedAppAuthRequest {
+  const messageId: MessageID = CommunicationPrivate.nextMessageId++;
```
TrevorJoelHarris, line 886: Do you think these changes would warrant some new e2e tests?

packages/teams-js/test/private/teams.spec.ts
```diff
@@ -250,7 +250,6 @@ describe('Testing teams capabillity', () => {
     });
 
     describe('joinedTeams.isSupported function', () => {
-      const utils = new Utils();
```
TrevorJoelHarris, line 253: Why was this removed?

packages/teams-js/test/public/people.spec.ts
```diff
@@ -155,7 +155,9 @@ describe('people', () => {
   });
 
   describe('Testing people.isSupported function', () => {
+    let utils: Utils = new Utils();
```
TrevorJoelHarris, line 158: Similarly to my question in `teams.spec.ts`: what about your change made this required / helpful?

packages/teams-js/src/internal/communication.ts
```diff
@@ -611,42 +621,49 @@ const handleParentMessageLogger = communicationLogger.extend('handleParentMessag
 function handleParentMessage(evt: DOMMessageEvent): void {
   const logger = handleParentMessageLogger;
 
-  if ('id' in evt.data && typeof evt.data.id === 'number') {
+  if ('id' in evt.data && (typeof evt.data.id === 'number' || typeof evt.data.id === 'string')) {
```
TrevorJoelHarris, line 624: Seems like here (and in the other repo) we should have code that verifies that the uuid is an actual uuid of the correct format. Definitely needs to be a string to be transferred between the host and app, but we should code defensively and validate that format. Would also be good, if possible, to convert the string to a "real" UUID type of some sort after validation

TrevorJoelHarris (changes requested):

packages/teams-js/test/private/teams.spec.ts
jadahiya-MSFT: We are setting const utils = new Utils() above in the beforeEach(), it seems unnecessary to do it at the beginning of this describe when it's already being reset.

packages/teams-js/test/public/people.spec.ts
jadahiya-MSFT: This is the same as the above

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Aah you're right, I had it going through the `validateUUID` function, I forgot to push that change back up. Let me do that now!

packages/teams-js/src/internal/communication.ts
```diff
@@ -841,8 +858,12 @@ function createMessageRequest(
   func: string,
   args: any[] | undefined,
 ): MessageRequestWithRequiredProperties {
+  const messageId: MessageID = CommunicationPrivate.nextMessageId++;
```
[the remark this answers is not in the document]
jadahiya-MSFT: Sort of, this is mostly so that if we have a Host that's using an older version of their codebase this change wouldn't impact them. Internally we are using UUIDs, but upon sending we still need to account for how the host will process these messages and have to add on the number as a messageID

packages/teams-js/src/internal/communication.ts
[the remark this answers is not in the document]
jadahiya-MSFT: Hmm good point, my thought process is since this change affects all of our codebase we should have to run the unit tests twice, once in a legacy mode that sends numbers as it used to and the other that sends uuids.

I should add some unit tests in communication though that verifies that the `handleParentMessage` works correctly though.

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,9 +1,28 @@
...
+export type MessageID = number | MessageUUID;
+
 /**
  * @internal
  * Limited to Microsoft-internal use
  */
 export interface MessageRequest {
-  id?: number;
+  id?: MessageID;
+  uuid?: MessageUUID;
```
AE-MS, line 25: Can you add a comment describing the difference between and proper usage of these two properties?

packages/teams-js/src/internal/communication.ts
```diff
@@ -47,15 +53,18 @@ class CommunicationPrivate {
...
+    [id: MessageID]: Function; // (args[]) => void
   } = {};
   public static portCallbacks: {
-    [id: number]: (port?: MessagePort, args?: unknown[]) => void;
+    [id: MessageID]: (port?: MessagePort, args?: unknown[]) => void;
   } = {};
   public static messageListener: Function;
+  public static legacyMessageIdsToUuidMap: {
+    [legacyId: number]: MessageUUID;
+  } = {};
```
AE-MS, line 67: It looks like this structure is used to map messages that have ids as numbers ("legacy id") to the new UUID. Based on that, I would expect the type of the property to be an actual UUID object and not a `string`. What are the benefits of using a `string` here over a strongly typed UUID object?

packages/teams-js/src/internal/communication.ts
```diff
@@ -47,15 +53,18 @@ class CommunicationPrivate {
   public static topMessageQueue: MessageRequest[] = [];
   public static nextMessageId = 0;
   public static callbacks: {
-    [id: number]: Function; // (arg1, arg2, ...etc) => void
+    [id: MessageID]: Function; // (arg1, arg2, ...etc) => void
```
AE-MS, line 56: >MessageID [](http://example.com/codeflow?start=9&length=9)

It looks like these maps are still using a `string` (which is the serialized format for the UUID) rather than an actual UUID type. Can we use an actual UUID type for our maps, rather than relying on the implementation detail of how a UUID is serialized?

packages/teams-js/src/internal/communication.ts
```diff
@@ -611,42 +621,52 @@ const handleParentMessageLogger = communicationLogger.extend('handleParentMessag
...
-  if ('id' in evt.data && typeof evt.data.id === 'number') {
+  if ('id' in evt.data && (typeof evt.data.id === 'number' || typeof evt.data.id === 'string')) {
     // Call any associated Communication.callbacks
     const message = evt.data as MessageResponse;
-    const callback = CommunicationPrivate.callbacks[message.id];
-    logger('Received a response from parent for message %i', message.id);
+    const callbackId =
+      typeof message.id === 'string' ? message.id : CommunicationPrivate.legacyMessageIdsToUuidMap[message.id];
+    if (callbackId) {
+      validateUuid(callbackId);
```
AE-MS, line 630: >validateUuid(callbackId); [](http://example.com/codeflow?start=6&length=25)

Rather than just validating the ID, can we instead use a UUID type that validates the data in its constructor and then throws if its not a real UUID?

packages/teams-js/src/internal/communication.ts
```diff
@@ -611,42 +621,52 @@ const handleParentMessageLogger = communicationLogger.extend('handleParentMessag
...
   const logger = handleParentMessageLogger;
 
-  if ('id' in evt.data && typeof evt.data.id === 'number') {
+  if ('id' in evt.data && (typeof evt.data.id === 'number' || typeof evt.data.id === 'string')) {
     // Call any associated Communication.callbacks
     const message = evt.data as MessageResponse;
-    const callback = CommunicationPrivate.callbacks[message.id];
-    logger('Received a response from parent for message %i', message.id);
+    const callbackId =
+      typeof message.id === 'string' ? message.id : CommunicationPrivate.legacyMessageIdsToUuidMap[message.id];
```
AE-MS, line 628: >typeof message.id === 'string' [](http://example.com/codeflow?start=6&length=30)

I think you are trying to determine whether the id is a UUID or a number here. If this code "knows" that a UUID is just a string format, then there really isn't any value in using the custom type aliases you defined in `messageObjects.ts` (`MessageUUID` and `BaseUUID`) since other parts of the code are ignoring that alias and just going to the underlying data.

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: I really like this idea, amongst to the other ones suggested. I'll make those changes now

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Sorry, I'm a bit confused by this question. The ids are using the type `MessageID`. Unfortunately, I can't use an object as an index and have to use either a template literal, number or string.

AE-MS: Is there a reason you are restricted to using this data structure to store the callbacks? You could consider using a JavaScript `Map` (although I'm guessing that won't work because you probably want custom comparison semantics on your keys). It seems like using the [Map structure in the immutable-js library](https://github.com/immutable-js/immutable-js/#equality-treats-collections-as-values) would suit this application.

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,9 +1,55 @@
...
+      this.uuid = generateGUID();
+    }
+  }
+
+  public getUuidValue(): string {
+    return this.uuid;
+  }
+}
+
+export class MessageUUID extends BaseUUID {}
```
TrevorJoelHarris, line 24: out of curiosity, what does having this extend BaseUUID get you? It seems like you could delete this line and rename BaseUUID to MessageUUID and you would be at basically the same place.

TrevorJoelHarris: Maybe you are thinking there would be ones that are for things other than Message, although that still might just warrant renaming BaseUUID to something like UUID.

packages/teams-js/src/internal/communication.ts
```diff
@@ -264,33 +269,33 @@ export function requestPortFromParentWithVersion(
...
+    });
   });
 }
 
 /**
  * @internal
  * Limited to Microsoft-internal use
  */
-function waitForResponse<T>(requestId: number): Promise<T> {
+function waitForResponse<T>(requestId: MessageUUID): Promise<T> {
```
TrevorJoelHarris, line 296:
```suggestion
function waitForResponse<T>(requestuuId: MessageUUID): Promise<T> {
```
because the old number value is still called messageId I had to think for a moment to figure out if this was correct. A namechange away from id to uuid might help?

packages/teams-js/src/internal/communication.ts
```diff
@@ -615,39 +664,52 @@ function handleParentMessage(evt: DOMMessageEvent): void {
...
       let port: MessagePort | undefined;
       if (evt.ports && evt.ports[0] instanceof MessagePort) {
         port = evt.ports[0];
       }
       portCallback(port, message.args);
 
-      logger('Removing registered port callback for message %i', message.id);
-      delete CommunicationPrivate.portCallbacks[message.id];
+      logger('Removing registered port callback for message %i', callbackId);
+      CommunicationPrivate.portCallbacks.delete(callbackId);
```
TrevorJoelHarris, line 706: Will calls to delete run into the same problem with objects as keys? Do we need a delete function that also looks up the keys by value?

packages/teams-js/src/internal/communication.ts
```diff
@@ -605,6 +614,46 @@ function updateRelationships(messageSource: Window, messageOrigin: string): void
...
+    }
+  }
+  return CommunicationPrivate.legacyMessageIdsToUuidMap[response.id];
+}
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function retrieveMessageUUIDFromCallback(
```
TrevorJoelHarris, line 644: A comment explaining why this function is needed would be helpful. It's not immediate obvious to people who aren't familiar with Javascript Maps and using objects as keys in them why something like this would be needed instead of just accessing the map the "normal" way.

packages/teams-js/src/internal/communication.ts
```diff
@@ -615,39 +664,52 @@ function handleParentMessage(evt: DOMMessageEvent): void {
...
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
       callback.apply(null, [...message.args, message.isPartialResponse]);
 
       // Remove the callback to ensure that the callback is called only once and to free up memory if response is a complete response
       if (!isPartialResponse(evt)) {
-        logger('Removing registered callback for message %i', message.id);
-        delete CommunicationPrivate.callbacks[message.id];
+        logger('Removing registered callback for message %i', callbackId);
+        CommunicationPrivate.callbacks.delete(callbackId);
```
TrevorJoelHarris, line 679: nit: I wonder if it would be helpful to take lines 679 -681 and turn them into a helper function that looked something like:

```typescript
function removeMessageHandlers(message: MessageResponse, map: Map<MessageUUID, unknown>): void {
  const callbackId = retrieveMessageUUIDFromResponse(message);
  map.delete(callbackId); // need to make sure delete works since this might have by-value problems
  if (!message.uuid) {
    delete CommunicationPrivate.legacyMessageIdsToUuidMap[message.id];
  }
}
```

TrevorJoelHarris (commented):
Left some minor comments but I think this is really looking good.

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: I modified the callback above to actually return rather than the value, this way we only need to do the map lookup once and then we can both acquire and delete the values using the same key value. So I don't believe this to be an issue

packages/teams-js/src/internal/messageObjects.ts
jadahiya-MSFT: Ya I was thinking about this being extensible down the line, but might as well address that in the future instead of trying to make an empty object that handles it now.

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: good catch, I meant to do this

packages/teams-js/src/internal/communication.ts
```diff
@@ -654,6 +658,21 @@ function retrieveMessageUUIDFromCallback(
...
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function removeMessageHandlers(
+  message: MessageResponse,
+  callbackId: MessageUUID,
+  map: Map<MessageUUID, unknown>,
+): void {
+  map.delete(callbackId); // need to make sure delete works since this might have by-value problems
```
TrevorJoelHarris, line 670: This comment suggests that you still need to look into this

jadahiya-MSFT: Actually, looked into this. I forgot to remove it. Thanks for catching that!

TrevorJoelHarris (approved):

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,9 +1,57 @@
+import { generateGUID, validateUuid } from './utils';
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * MessageUUID object that holds the uuid used for processing MessageRequests and MessageResponses
+ */
+export class MessageUUID {
```
AE-MS, line 9: >MessageUUID [](http://example.com/codeflow?start=13&length=11)

This seems like a valuable and useful general purpose class that the library should use for holding UUIDs across the codebase. Would you be willing to rename it to something less-specific to messages and put it in a general shared location?

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,9 +1,57 @@
...
+  public constructor(uuid?: string) {
+    if (uuid) {
+      validateUuid(uuid);
+      this.uuid = uuid;
+    } else {
+      this.uuid = generateGUID();
+    }
+  }
+
+  public getUuidValue(): string {
```
AE-MS, line 21: >getUuidValue [](http://example.com/codeflow?start=9&length=12)

Could we just call this `toString()` ?

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,9 +1,57 @@
+import { generateGUID, validateUuid } from './utils';
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * MessageUUID object that holds the uuid used for processing MessageRequests and MessageResponses
+ */
+export class MessageUUID {
+  private uuid: string;
```
AE-MS, line 10: Could you instead declare this in the constructor like this:

```
public constructor(private readonly uuid: string = generateGUID()) {
    validateUuid(uuid);
}
```

jadahiya-MSFT: Aah yes, that is much cleaner. I'll do that :)

packages/teams-js/src/internal/communication.ts
```diff
@@ -46,16 +53,13 @@ class CommunicationPrivate {
...
-    [id: number]: Function; // (arg1, arg2, ...etc) => void
-  } = {};
-  public static promiseCallbacks: {
-    [id: number]: Function; // (args[]) => void
-  } = {};
-  public static portCallbacks: {
-    [id: number]: (port?: MessagePort, args?: unknown[]) => void;
-  } = {};
+  public static callbacks: Map<MessageUUID, Function> = new Map();
+  public static promiseCallbacks: Map<MessageUUID, Function> = new Map();
```
AE-MS, line 56: >Function [](http://example.com/codeflow?start=51&length=8)

Since `Function` is general and generally disallowed, could this be updated to use a type more specific to the type of function being expected? Maybe:

```
type PromiseCallback<T> = (value: T) => void;

const promiseCallbacks: Map<number, PromiseCallback<unknown>> = new Map();
```


With this solution we can lean on the type system to help ensure no one accidentally puts any other type of function into the `Map`

AE-MS, line 56: >Function [](http://example.com/codeflow?start=51&length=8)

Since `Function` is general and generally disallowed, could this be updated to use a type more specific to the type of function being expected? Maybe:

```
const promiseCallbacks: Map<number, ((value: unknown) => void)> = new Map();
```


With this solution we can lean on the type system to help ensure no one accidentally puts any other type of function into the `Map`

packages/teams-js/src/internal/communication.ts
```diff
@@ -373,14 +378,18 @@ const sendRequestToTargetWindowHelperLogger = communicationLogger.extend('sendRe
...
   targetWindow: Window,
-  request: MessageRequestWithRequiredProperties | NestedAppAuthRequest,
+  messageRequest: MessageRequestWithRequiredProperties | NestedAppAuthRequest,
 ): MessageRequestWithRequiredProperties | NestedAppAuthRequest {
   const logger = sendRequestToTargetWindowHelperLogger;
   const targetWindowName = getTargetName(targetWindow);
+  const request: SerializedMessageRequest = {
+    ...messageRequest,
+    uuid: messageRequest.uuid.getUuidValue(),
+  };
```
AE-MS, line 387: I am concerned about having the code that knows how to serialize a message request living in the middle of this function in this file. If new properties that need serialization get added, it won't necessarily be obvious to future ENG that they have to go find this code to update it. Would you consider writing a function that lives right next to the message request interfaces (something like `serializeMessageRequest(request: MessageRequestWithRequiredProperties): SerializedMessageRequest`) so that way all the code for dealing with the implementation of the message structures lives together, rather than spread throughout the code?

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -16,7 +64,8 @@ export interface MessageRequest {
  * Limited to Microsoft-internal use
  */
 export interface MessageResponse {
```
AE-MS, line 43: >MessageResponse [](http://example.com/codeflow?start=17&length=15)

I would have expected to see a matching `SerializedMessageResponse` object as well -- would it be possible to have one for all the same type safety reasons as we have above?

packages/teams-js/src/internal/communication.ts
```diff
@@ -615,39 +683,43 @@ function handleParentMessage(evt: DOMMessageEvent): void {
   if ('id' in evt.data && typeof evt.data.id === 'number') {
     // Call any associated Communication.callbacks
     const message = evt.data as MessageResponse;
```
AE-MS, line 684: >MessageResponse [](http://example.com/codeflow?start=32&length=15)

Would the data coming back over this pipe actually be a `SerializedMessageResponse`? And we should deserialize it first, before examining its data or manipulating it?

packages/teams-js/src/internal/communication.ts
```diff
@@ -605,6 +614,65 @@ function updateRelationships(messageSource: Window, messageOrigin: string): void
 
 const handleParentMessageLogger = communicationLogger.extend('handleParentMessage');
 
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function retrieveMessageUUIDFromResponse(response: MessageResponse): MessageUUID {
```
AE-MS, line 620: >retrieveMessageUUIDFromResponse [](http://example.com/codeflow?start=9&length=31)

What are the semantics for this function if the ID that comes in the response isn't present in any of the maps?

packages/teams-js/test/utils.ts
```diff
@@ -2,20 +2,34 @@ import { validOriginsFallback as validOrigins } from '../src/internal/constants'
...
 import { GlobalVars } from '../src/internal/globalVars';
 import { DOMMessageEvent, ExtendedWindow } from '../src/internal/interfaces';
-import { MessageResponse } from '../src/internal/messageObjects';
+import { MessageID, MessageResponse, MessageUUID, SerializedMessageRequest } from '../src/internal/messageObjects';
 import { NestedAppAuthRequest } from '../src/internal/nestedAppAuthUtils';
 import { HostClientType } from '../src/public';
 import { app } from '../src/public/app';
 import { applyRuntimeConfig, IBaseRuntime, setUnitializedRuntime } from '../src/public/runtime';
 
 export interface MessageRequest {
```
AE-MS, line 11: >MessageRequest [](http://example.com/codeflow?start=17&length=14)

How is this interface different than the one `messageObjects.ts`? Should this one just be importing and using that one, instead of duplicating it?

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: I thought about the deserialization, but we don't pass along the `MessageResponse` object anywhere in the code. It is mostly used during either data processing on our end or creation of the object right before sending. I do think that the naming for this object should change to `SerializedMessageResponse` since it is the response received from the host or sent to the host.

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,9 +1,57 @@
...
+  apiVersionTag?: string;
+}
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export interface SerializedMessageRequest {
+  id?: MessageID;
+  uuid?: string;
```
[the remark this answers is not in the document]
jadahiya-MSFT: yup, I can make the change. I kept it as `uuid` purely to match processing on the host side, but I can add logic that changes this to `uuidAsString` there.

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Sorry, I thought I already made a function like this to be extensible in the future. I might have reverted that change by mistake when I got my repo into a bad state. I'll add this in.

packages/teams-js/src/internal/communication.ts
```diff
@@ -344,7 +349,7 @@ export function sendMessageToParent(
   if (callback) {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
```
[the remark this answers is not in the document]
jadahiya-MSFT: This causes issues now that I've changed callback to be of type `((value?: unknown) => void)`

packages/teams-js/test/utils.ts
jadahiya-MSFT: I agree. Let me remove this.

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: I'm not sure what you mean by this. What this does is a `byValue` comparison of the uuid coming in from the host to the callback maps we have and returns their `MessageUUID` key for use. This way we have the correct key to perform operations on the map rather than one that is equal in value but not considered the same object.

If there is no uuid being passed in by the `host` due to older versioning on the Host side, we default back to using the `id` as a number to map to the correct UUIDs, since UUIDs are being passed internally.

packages/teams-js/src/internal/messageObjects.ts
jadahiya-MSFT: Since the data in the `MessageResponse` object is being used immediately, I elected to not make a `SerializedMessageResponse` and `MessageResponse` object since the UUID for example is being sent as a UUID object across our code rather than as two separate `MessageResponse` objects.

For extensibility though, I think I will go back and change this into a serialized and deserialized object.

packages/teams-js/test/utils.ts
jadahiya-MSFT: Looks like MessageRequest doesn't have the property isPartialResponse which is used in the PrivateAPI spec. I will rework those tests to make sure that they are done correctly. Since `isPartialResponse` should be only for `MessageResponses`

packages/teams-js/src/internal/interfaces.ts
```diff
@@ -50,3 +52,19 @@ export interface DOMMessageEvent {
...
+ */
+export class UUID {
+  public constructor(private readonly uuid: string = generateGUID()) {
+    validateUuid(uuid);
+  }
+
+  public toString(): string {
+    return this.uuid;
+  }
+}
```
AE-MS, line 70: Thanks for putting this here and enabling easy shared usage! Can you write a small set of unit tests to validate its functionality (e.g. that it doesn't accept invalid uuids as input, etc.)?

packages/teams-js/src/internal/communication.ts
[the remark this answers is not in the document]
AE-MS: While I believe it is correct to have removed the `Function` type usage from the `promiseCallbacks`, I don't think you can remove it from the `callbacks` map because it's being populated from this function, and this function takes `callback` as a parameter that's typed as `Function` -- so the `callbacks` map where it is stored has to stay as `Function` (unless you want to also change the parameter type on this function and keep walking up the chain for all callers of this function and so on -- a way more invasive change than I think is appropriate to do in this PR).

In summary: I think you have to leave `callbacks` as a map to `Function`

And then maybe you can remove these suppressions. But if not, that's okay, too.

packages/teams-js/src/internal/communication.ts
AE-MS: I wrote something more elaborate down below but I think that you can't change this for `callbacks` and only can change it for `promiseCallbacks` without making a much more invasive change to update all the places you are describing. I think `callbacks` has to stay `Function` because it's used in all those other places that also use `Function`. It's possible to improve here but it would require changing all the upstream dependences of this which is more work than is appropriate for this PR.

packages/teams-js/src/internal/communication.ts
AE-MS: What happens if the incoming message has a uuid but it's not one that we have in any of the maps?

packages/teams-js/src/internal/communication.ts
```diff
@@ -605,6 +614,65 @@ function updateRelationships(messageSource: Window, messageOrigin: string): void
...
+function removeMessageHandlers(
+  message: MessageResponse,
+  callbackId: MessageUUID,
+  map: Map<MessageUUID, unknown>,
+): void {
+  map.delete(callbackId);
+  if (!message.uuid) {
+    delete CommunicationPrivate.legacyMessageIdsToUuidMap[message.id];
+  }
+}
```
AE-MS, line 674: Can you add some unit tests for these functions?

packages/teams-js/src/internal/communication.ts
```diff
@@ -605,6 +614,65 @@ function updateRelationships(messageSource: Window, messageOrigin: string): void
...
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function removeMessageHandlers(
+  message: MessageResponse,
+  callbackId: MessageUUID,
+  map: Map<MessageUUID, unknown>,
+): void {
+  map.delete(callbackId);
```
AE-MS, line 670: Since special handling is needed to "find" a uuid in the map (E.g. that function above here that compares their string values), would similar special handling be needed here? Otherwise how could it reliably find which entry in the map to delete?

packages/teams-js/test/utils.ts
```diff
@@ -1,19 +1,24 @@
...
+    ...serializedMessage,
+    uuid: serializedMessage.uuidAsString ? new MessageUUID(serializedMessage.uuidAsString) : undefined,
+  };
+  return message;
+}
+
+function getMessageUUIDString(message: MessageRequest): string | undefined {
+  const messageUuid = message.uuid ? message.uuid.toString() : undefined;
+  return messageUuid;
 }
```
AE-MS, line 23: Not a big deal but I think you could write this as an understandable one-liner:

```
const getMessageUUIDString(message: MessageRequest): string | undefined => message.uuid?.toString();
```

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: The behavior expectation in that case is that there is no callback associated with the MessageRequest, but it still could be a processable Message.

packages/teams-js/src/internal/interfaces.ts
```diff
@@ -50,3 +52,19 @@ export interface DOMMessageEvent {
...
   args?: any[];
 }
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * UUID object
+ */
+export class UUID {
```
TrevorJoelHarris, line 62: This started out as an interface which is why it is in the `interfaces.ts` file, but now that it is a class with functionality it would probably make sense to give it its own file and test file

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -1,22 +1,61 @@
...
 
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export interface SerializedMessageResponse {
+  id: MessageID;
+  uuidAsString?: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  args?: any[];
```
AE-MS, line 48: >any [](http://example.com/codeflow?start=9&length=3)

Can `unknown` be used here, instead of having to disable the any warning?

packages/teams-js/src/internal/communication.ts
AE-MS: So what value will this function return in that case? It's typed as always returning a `MessageUUID` but what does it return if it can't find a matching message in one of the maps?

change/@microsoft-teams-js-acd75970-b390-4cfe-b488-4bd6f51f723e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added uuid parameter to MessageRequest interface and enabled MessageResponse ids to also be uuids",
```
AE-MS, line 3: >Added uuid parameter to MessageRequest interface and enabled MessageResponse ids to also be uuids [](http://example.com/codeflow?start=14&length=97)

Consider:

Added uuid parameter to `MessageRequest` interface and enabled `MessageResponse` ids to also be uuids

packages/teams-js/src/internal/communication.ts
AE-MS: I see this comment was resolved but I don't see the unit tests for these -- maybe I missed it?

packages/teams-js/src/internal/messageObjects.ts
jadahiya-MSFT: Looked into this and this would cause issues with some unit tests, I'll make a work item to look into making this change in the future, but for now will keep it as a `any[]`

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: I see your point, I'll update this function so that it can return an undefined. 

Because we don't know if the Host is using an older version of their SDK or a newer version with uuids, this function should always map to a MessageUUID object through the legacyMessageIdsToUuidMap[response.id]` since they are generated together.

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Good catch, I think I accidentally resolved this thinking it was the UUID object test

jadahiya-MSFT: I've added these test in now

TrevorJoelHarris (approved):

packages/teams-js/src/internal/communication.ts
```diff
@@ -614,40 +683,47 @@ function handleParentMessage(evt: DOMMessageEvent): void {
...
-      // @ts-ignore
-      callback.apply(null, [...message.args, message.isPartialResponse]);
-
-      // Remove the callback to ensure that the callback is called only once and to free up memory if response is a complete response
-      if (!isPartialResponse(evt)) {
-        logger('Removing registered callback for message %i', message.id);
-        delete CommunicationPrivate.callbacks[message.id];
+    const serializedResponse = evt.data as SerializedMessageResponse;
+    const message: MessageResponse = deserializeMessageResponse(serializedResponse);
+    const callbackId = retrieveMessageUUIDFromResponse(message);
```
AE-MS, line 688: >retrieveMessageUUIDFromResponse [](http://example.com/codeflow?start=23&length=31)

Should this log in the case where `retrieveMessageUUIDFromResponse` returns `undefined` so that when we diagnose issues we can be alerted to the fact that we're receiving messages we don't recognize?

packages/teams-js/src/internal/messageObjects.ts
AE-MS: Thank you!

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: That's a good call, I'll add that now

TrevorJoelHarris (approved):

AE-MS (approved):
:shipit:

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

packages/teams-js/src/internal/communication.ts
```diff
@@ -614,40 +689,47 @@ function handleParentMessage(evt: DOMMessageEvent): void {
...
+        if (evt.ports && evt.ports[0] instanceof MessagePort) {
+          port = evt.ports[0];
+        }
+        portCallback(port, message.args);
+
+        logger('Removing registered port callback for message %i', callbackId);
+        removeMessageHandlers(message, CommunicationPrivate.portCallbacks);
+      }
+      if (message.uuid) {
+        CommunicationPrivate.legacyMessageIdsToUuidMap = {};
```
jekloudaMSFT, line 731: This is effectively clearing the legacyMessageIdsToUuidMap because we know the host is using uuids?

jekloudaMSFT (approved):

jekloudaMSFT (approved):

packages/teams-js/src/internal/uuidObject.ts
```diff
@@ -0,0 +1,17 @@
+import { generateGUID, validateUuid } from './utils';
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * UUID object
+ */
+export class UUID {
+  public constructor(public readonly uuid: string = generateGUID()) {
```
AE-MS, line 10: >public [](http://example.com/codeflow?start=21&length=6)

Can you say more about why this needs to be `public` now? Enabling direct access to the underlying data defeats the abstraction provided by the class.

jadahiya-MSFT: Sorry, this shouldn't be public, That was just used for testing purposes. Good catch! Changing it back to private

AE-MS (approved):
:shipit:

PR #2246, opened 2024-04-01
Const enum recommendation Eslint plugin

noahdarveau-MSFT:

## Description

### Main changes in the PR:

1. I wrote a custom rule plugin for eslint that will give a warning to the developer to consider if the enum they wrote could be `const`. The plugin will give a warning, but since the repo is configured to treat warnings as errors, they will need to add an eslint ignore line to the line just before their enum if it should not be `const`. The warning text when eslint detects a non-const enum is `Please consider if you can use a const enum for ${enumName} to minimize bundle size. If not, add "/* eslint-disable-next-line recommend-const-enums/recommend-const-enums */" to the line above to disable this warning.`

How do people feel about adding this plugin to the repo? I wrote it with the intention of it just being a reminder to people so they stop and think about what they're writing with a little bit more intentionality. On the other hand, I could definitely see it as an unnecessary pain, that could definitely lead to people just blindly following the recommendation without actually thinking about it. So I figured I'd just put this PR out here and see what people thought about it, so some opinions on the matter would be greatly appreciated.
 
### Validation performed:

1. <Step 1>
3. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


jadahiya-MSFT (approved):

jadahiya-MSFT:
I personally don't see anything wrong with adding this in. I definitely agree with the sentiment that people will probably just follow the guidance given rather than thinking about it. I think something that could be helpful is if we also were able to show package size increases in the warning. 

Just out of curiosity, is there any scenarios that going forward we wouldn't want our enums to be given the `const` tag?

noahdarveau-MSFT:
> Just out of curiosity, is there any scenarios that going forward we wouldn't want our enums to be given the `const` tag?

Generally speaking, enums should be `const` if we're just using them to hold values nice and neatly, but there are some edge cases, like if you wanted to use the reverse mapping functionality or need to access the enum object at runtime, that you would want to keep it a regular enum.

AE-MS:
I think this should be something we enable in the repo because the default behavior for enums is going to cause the package to grow unnecessarily. Most engineers who add enums will not realize this (myself included). Enabling a rule that causes every engineer who adds an enum to learn about the impact and then make an intentional decision about whether or not it can be `const` is the right behavior.

Once this rule is enabled, I assume the way that an engineer would "make the decision" is to either add `const` to their enum or add a rule suppression to the enum line that says "I know what I'm doing, this should really not be const." ?

packages/teams-js/src/private/remoteCamera.ts
```diff
@@ -58,6 +58,7 @@ export namespace remoteCamera {
    * @internal
    * Limited to Microsoft-internal use
    */
+  /* eslint-disable-next-line recommend-const-enums/recommend-const-enums */
```
AE-MS, line 61: >eslint-disable-next-line recommend-const-enums/recommend-const-enums [](http://example.com/codeflow?start=5&length=68)

For all the places where this is now being suppressed, can we set a good example and also provide an explanation for why its being suppressed? Without a reason listed I worry that any future engineer will look at these existing instances and think that the standard is just to always suppress without being thoughtful about why

packages/teams-js/test/public/runtime.spec.ts
```diff
@@ -282,6 +282,7 @@ describe('runtime', () => {
     return `${version}.1`;
   }
 
+  /* eslint-disable-next-line recommend-const-enums/recommend-const-enums */
```
AE-MS, line 285: > /* eslint-disable-next-line recommend-const-enums/recommend-const-enums */ [](http://example.com/codeflow?start=1&length=75)

Can you say more about why this one needs to be suppressed?

packages/teams-js/test/resultValidation.ts
```diff
@@ -1,5 +1,6 @@
 import { MessageRequest } from './utils';
 
+/* eslint-disable-next-line recommend-const-enums/recommend-const-enums */
```
AE-MS, line 3: Can you say more about why this one needs to be suppressed?

packages/teams-js/test/promiseTester.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable-next-line recommend-const-enums/recommend-const-enums */
```
AE-MS, line 1: Can you say more about why this one needs to be suppressed?

packages/teams-js/eslint-rules/eslint-plugin-recommend-const-enums/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "eslint-plugin-recommend-const-enums",
+  "version": "0.0.0",
+  "license": "MIT",
+  "author": "Noah",
```
AE-MS, line 5: >Noah [](http://example.com/codeflow?start=13&length=4)

lol

packages/teams-js/eslint-rules/eslint-plugin-recommend-const-enums/recommendConstEnums.js
```diff
@@ -0,0 +1,29 @@
...
+  create: function (context) {
+    return {
+      TSEnumDeclaration: function (node) {
+        const enumName = node.id.name;
+        const enumType = node.const ? 'const' : 'regular';
+
+        if (enumType === 'regular') {
+          context.report({
+            node: node,
+            message: `Please consider if you can use a const enum for ${enumName} to minimize bundle size. If not, add "/* eslint-disable-next-line recommend-const-enums/recommend-const-enums */" to the line above to disable this warning.`,
```
AE-MS, line 22: >Please consider if you can use a const enum for ${enumName} to minimize bundle size. [](http://example.com/codeflow?start=22&length=84)

Is there a documentation page we can point people at to learn about the impact and ramifications about using `const` enum? E.g., I learned a bunch from everything you've written about it recently, but any random person who makes a change probably won't know anything when they see this warning and if they bother to do any research they likely won't find anything as comprehensive as what you've shared.

Even if you just document what you've learned on a new page on [the wiki](https://github.com/OfficeDev/microsoft-teams-library-js/wiki), and then link to that, that would be great. In addition to explaining the impact on package size and why, providing guidance like:

"You should default to using a const enum unless you know that you need one of these things:
- Reverse mapping blah blah"
- etc.
- etc.
"

AE-MS, line 22: >  [](http://example.com/codeflow?start=195&length=1)

Can you add "And a comment explaining why this can't be const" to the recommendation?

AE-MS (changes requested):
:clock1:

packages/teams-js/test/public/runtime.spec.ts
noahdarveau-MSFT: It doesn't. It's a leftover from when I was developing the plugin. Since the test folder doesn't impact the bundle size, I forgot to make them `const` too. I went ahead and made them `const`

packages/teams-js/test/resultValidation.ts
noahdarveau-MSFT: See above

packages/teams-js/test/promiseTester.ts
noahdarveau-MSFT: See above

packages/teams-js/eslint-rules/eslint-plugin-recommend-const-enums/recommendConstEnums.js
noahdarveau-MSFT: I added a wiki page [here](https://github.com/OfficeDev/microsoft-teams-library-js/wiki/Notes-on-Enum-Usage), and linked to it in the warning message. Give the wiki page a read and let me know if you think there's anything I should add/remove/change etc.

AE-MS: Looks great; thanks for writing a good explanation!

AE-MS (approved):
:shipit:

PR #2247, opened 2024-04-02
Adding nested app auth support check api for app developers

baljesingh:
## Description

> Introduced NAA support check api for app developers, allowing them to determine whether utilizing the NAA channel through MSAL is advisable.

### Main changes in the PR:

1. Renamed the existing `nestedAppAuth` to `nestedAppAuthUtils` to ensure consistency with other utility files and to avoid confusion with the newly introduced `nestedAppAuth` namespace.
2. Added isNAAChannelRecommended to both IRuntimeV3 and IRuntimeV4.
3. Introduced a new namespace nestedAppAuth.
4. Implemented isChannelRecommended() API to check for NAA channel recommendation.

### Validation performed:

1. Manually tested in local.

### Unit Tests added:

Yes

### End-to-end tests added:

Yes - provided e2e test data. 

### Change file added:

Yes

### Next/remaining steps:

- [ ] Host SDK changes - Web/Mobile


baljesingh:

Added an optional property and hence no new version is required.

apps/teams-test-app/e2e-test-data/nestedAppAuth.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "nestedAppAuth",
+  "platforms": "*",
```
baljesingh, line 3: Validating whether this '*' breaks the pipeline on mobile platforms if the UI tests have not yet been implemented there.

KangxuanYe: No, it will not.

jadahiya-MSFT (dismissed):

apps/teams-test-app/e2e-test-data/nestedAppAuth.json
baljesingh: Closing this validation question as it will not cause any breakage. Thanks @KangxuanYe

apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const CheckisNAAChannelRecommended = (): ReactElement =>
+    ApiWithoutInput({
+      name: 'checkisNAAChannelRecommended',
+      title: 'Check NAA Channel Recommended',
+      onClick: async () => `NAA channel ${nestedAppAuth.isChannelRecommended() ? 'is' : 'is not'} recommended`,
+    });
+
+  return (
+    <ModuleWrapper title="NestedAppAuth">
+      <CheckisNAAChannelRecommended />
```
shrshindeMSFT, line 17:
```suggestion
      <CheckIfNAAChannelRecommended />
```

packages/teams-js/test/public/nestedAppAuth.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Utils } from '../utils';
+
+/* eslint-disable */
+/* As part of enabling eslint on test files, we need to disable eslint checking on the specific files with
+   large numbers of errors. Then, over time, we can fix the errors and reenable eslint on a per file basis. */
+
+/**
+ * Test cases for nested app auth APIs
+ */
+describe('nestedAppAuth', () => {
```
shrshindeMSFT, line 13: Can you verify if we need framed or frameless options for UTs here?

jadahiya-MSFT: Since we are just setting  runtimeConfig in the tests and not sending any messages, I think this shouldn't be needed.

apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
jadahiya-MSFT: If you make the above suggestion, make sure to update the above objects to have the same nomenclature.

packages/teams-js/src/public/nestedAppAuth.ts
```diff
@@ -0,0 +1,20 @@
...
+export namespace nestedAppAuth {
+  /**
+   * Checks if MSAL-NAA channel recommended by the host
+   * @returns true if host is recommending NAA channel and false otherwise
+   *
+   * @throws Error if {@linkcode app.initialize} has not successfully completed
+   *
+   * @beta
+   */
+  export function isChannelRecommended(): boolean {
```
maglims, line 17: Is this same as isSupported() function as other modules? If so, is there any reason we name it differently? 
(as far as I am aware, isSupported() is a well established function name for checking if a capability is supported)

packages/teams-js/test/public/nestedAppAuth.spec.ts
baljesingh: Thanks for confirming @jadahiya-MSFT. As discussed yesterday @shrshindeMSFT, we should be good to close this conversation. Feel free to open it if you have any further concerns.

packages/teams-js/src/public/runtime.ts
```diff
@@ -148,6 +148,7 @@ interface IRuntimeV2 extends IBaseRuntime {
 interface IRuntimeV3 extends IBaseRuntime {
   readonly apiVersion: 3;
   readonly hostVersionsInfo?: HostVersionsInfo;
+  readonly isNAAChannelRecommended?: boolean;
```
maglims, line 151: this is mostly related to my other question on isChannelRecommended, so we are introducing a new param in RunTime for supportivity check for NAA? Is that by design that we are not leveraging existing "Suppports"check for NAA?

packages/teams-js/src/public/runtime.ts
```diff
@@ -219,6 +220,7 @@ interface IRuntimeV3 extends IBaseRuntime {
 interface IRuntimeV4 extends IBaseRuntime {
   readonly apiVersion: 4;
   readonly hostVersionsInfo?: HostVersionsInfo;
+  readonly isNAAChannelRecommended?: boolean;
```
maglims, line 223: I think there is a MOS app called Channel(or Channels). This may cause slightly confusion depending on the audience.. also I am not sure about "Recommended" vs "Supported"... "Recommended" feels a little bit strong to recommend what app developers should do.. They can just query and see if NAA is supported, then decide if they want to leverage or not. 

So based on the above, I would vote for isNAASupported over isNAARecommended or isNAAChannelRecommended

packages/teams-js/src/public/nestedAppAuth.ts
maglims: ok, it looks your PR on web hub sdk explains this in some details. Let me look at that first to see if It resolves my question. Thanks.

apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
baljesingh: Thanks @jadahiya-MSFT. I don't have a strong inclination to change it from 'is' to 'If' because we are not aiming to form a grammatically correct sentence. It actually reflects the actual parameter name, 'isNAAChannelRecommended.' Consider this similar to the existing 'isSupported,' where we have 'CheckIsSupported.' However, the casing of 'is' should be changed to align with PascalCase/UpperCamelCase, i.e., Check**I**sNAAChannelRecommended. I can make the changes as proposed if @shrshindeMSFT also aligns with my thinking and has no concerns; otherwise, I'm looking forward to continuing this discussion if I've missed any crucial points.

packages/teams-js/src/public/nestedAppAuth.ts
baljesingh: No. It serves differently than isSupported() , please take a look at the reason in SDK PR because I'm not sure if we can explain it here due to some wording restrictions. Let me know if you have any concerns.

packages/teams-js/src/public/runtime.ts
baljesingh: The 'isNAASupported' serves a different purpose. It indicates whether NAA is supported or not, whereas 'isNAAChannelRecommended' indicates whether the NAA **channel** is recommended, which doesn't necessarily mean NAA is supported. We have certain scenario where NAA is not available but we still want to send true to this variable. Hence 'isNAASupported' may cause more confusion. As of now, there haven't been any concerns raised internally regarding this variable. If you believe the reasoning provided isn't sufficiently strong, I may need to initiate a separate discussion with stakeholders and internal team members on this matter.

baljesingh: Yes, you're correct in pointing out that 'Recommended' may come across as a strong word choice. However, it accurately reflects our intention. While 'Supported' implies that developers have the option to choose alternative paths, we specifically want to discourage considering the alternative path and hence we're labeling as 'Recommended.' The reason for this emphasis is due to security concerns associated with the alternative path. We can discuss more offline.

maglims (dismissed):

packages/teams-js/src/public/nestedAppAuth.ts
baljesingh: As discussed offline, there is no more concern regarding the name.

packages/teams-js/src/public/runtime.ts
baljesingh: As discussed offline, there are no more concerns. Feel free to open it again if you have any other questions.

maglims (approved):

shrshindeMSFT (approved):

PR #2255, opened 2024-04-04
Added third party cookies test component in teams test app

Ella-ly:

## Description

Added third party test component in teams test app. This component is for validating `allow-storage-access-by-user-activation` attribute which allows app develops to use storage access API (Web API). 

### Main changes in the PR:

1. Added third party cookies test component in teams test app.

### Change file added:

No change files are needed.

### Screenshots:
image

image


apps/teams-test-app/src/components/ThirdPatryCookies.tsx
```diff
@@ -0,0 +1,81 @@
...
+    onClick: refreshCookies,
+  });
+};
+
+const hasCookieAccess = async (): Promise<boolean> => {
+  if (!document.requestStorageAccess) {
+    console.log(
+      'Storage Access API not supported. Assume it is an older browser that doesn not block 3P cookies and we have access.',
+    );
+    return true;
```
maglims, line 55: oh, is it supposed to return true on storage access api not supported?

apps/teams-test-app/src/App.tsx
```diff
@@ -182,6 +183,7 @@ const App = (): ReactElement => {
         <VideoAPIs />
         <VideoExAPIs />
         <VisualMediaAPIs />
+        <CookieAccessComponent />
```
MengyiGong, line 186: It would be great to follow the alphabetical order.

Ella-ly: Updated.

apps/teams-test-app/src/components/ThirdPatryCookies.tsx
Ella-ly: Basically, if it is an older browser that does not block 3P cookies then we will have access and do not need to call Storage Access APIs to request access. We can use and set the document.cookie directly, for example, `document.cookie = 'testCookies=testvalue; path=/; Secure; SameSite=None';`. Please see the [Checking and requesting storage access](https://developer.mozilla.org/en-US/docs/Web/API/Storage_Access_API/Using#checking_and_requesting_storage_access) section in Using the Storage Access API document. I simplified the code to better align with the test need.

MengyiGong (approved):

AE-MS (approved):
:shipit:

PR #2260, opened 2024-04-09
MessageChannels.getTelemetryPort: Add E2E Test JSON

kerrynf:
## Description

This change updates the test app and adds JSON for E2E tests for the MessageChannels.getTelemetryPort function.

A follow up PR will add the capability test for MessageChannels once the test app changes are tied to a published version.

### Main changes in the PR:

1. Update test app to have capability check box
2. Create test JSON

### Validation performed:

1. Test JSON was validated with an external test runner

### Unit Tests added:
No, this is follow up to the main change [here](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2140), where the unit tests were added.

### End-to-end tests added:

Yes

### Change file added:

Not required for adding an E2E (no changes to published packages)

### Related PRs:

[https://github.com/OfficeDev/microsoft-teams-library-js/pull/2140](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2140)

### DIFF
--- a/apps/teams-test-app/e2e-test-data/messageChannels.json
+++ b/apps/teams-test-app/e2e-test-data/messageChannels.json
@@ -1,16 +1,11 @@
 {
   "name": "Message Channels",
-  "checkIsSupported": {
-    "capabilityName": "MessageChannels",
-    "expectedOutput": "MessageChannels is not supported",
-    "testUrlParams": []
-  },
   "platforms": "Web",
   "testCases": [
     {
       "title": "getTelemetryPort API Call - Success",
       "type": "callResponse",
-      "version": ">2.20.0",
+      "version": ">=2.20.0",
       "hostSdkVersion": {
         "web": ">=2.12.0"
       },
@@ -20,7 +15,7 @@
     {
       "title": "getTelemetryPort API Call - Failure (Not MS Owned)",
       "type": "callResponse",
-      "version": ">2.20.0",
+      "version": ">=2.20.0",
       "hostSdkVersion": {
         "web": ">=2.12.0"
       },
--- a/apps/teams-test-app/e2e-test-data/messageChannels.json
+++ b/apps/teams-test-app/e2e-test-data/messageChannels.json
@@ -1,0 +1,26 @@
+{
+  "name": "Message Channels",
+  "checkIsSupported": {
+    "capabilityName": "MessageChannels",
+    "expectedOutput": "MessageChannels is not supported",
+    "testUrlParams": []
+  },
+  "platforms": "Web",
+  "testCases": [
+    {
+      "title": "getTelemetryPort API Call - Success",
+      "type": "callResponse",
+      "version": ">2.10.0",
+      "boxSelector": "#box_checkTelemetryPort",
+      "expectedTestAppValue": "Telemetry port: [object MessagePort]"
+    },
+    {
+      "title": "getTelemetryPort API Call - Failure (Not MS Owned)",
+      "type": "callResponse",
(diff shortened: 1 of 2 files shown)


erinha (approved):

apps/teams-test-app/e2e-test-data/messageChannels.json
```diff
@@ -0,0 +1,26 @@
...
+  "name": "Message Channels",
+  "checkIsSupported": {
+    "capabilityName": "MessageChannels",
+    "expectedOutput": "MessageChannels is not supported",
+    "testUrlParams": []
+  },
+  "platforms": "Web",
+  "testCases": [
+    {
+      "title": "getTelemetryPort API Call - Success",
```
erinha, line 11: There should be a host SDK version restriction on this test. Let me find an example.

erinha: Here's an example. I'm pretty sure the message channel feature went out in v2.12.0
"hostSdkVersion": {
        "web": ">=2.12.0"
 },

erinha (changes requested):
Left a hostSdkVersion filtering comment

apps/teams-test-app/e2e-test-data/messageChannels.json
kerrynf: Thank you! I forgot I was still trying to figure that bit out. Let me see if this fixes the build as well.

KangxuanYe: One question, when is getTelemetryPort API is supported? You can keep version field but just make sure it is labelled correctly (which is to say, don't include a version of TJS (test app) that doesn't have that API

kerrynf: Does the version in the test JSON represent the Test app version or the teams-js SDK version? I would have assumed the latter?

apps/teams-test-app/e2e-test-data/messageChannels.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Message Channels",
+  "checkIsSupported": {
+    "capabilityName": "MessageChannels",
+    "expectedOutput": "MessageChannels is not supported",
+    "testUrlParams": []
```
KangxuanYe, line 6: This line needs to be removed because it isn't supported in support check.

KangxuanYe (commented):
Another comment

apps/teams-test-app/e2e-test-data/messageChannels.json
KangxuanYe: You could understand it as both. They are tied to each other. You will never have a chance to run test app V2.20.0 on TJS V2.21.0 for example. The test pulls TJS V2.21.0 release branch and use test app under that branch's folder.

kerrynf: Right, so with the fix for the test app in this PR, we can't run the capability test on isSupported until we can target the test app version with this UI feature.

I've updated the PR to remove the isSupported test block and we will need to add that back later.

apps/teams-test-app/e2e-test-data/messageChannels.json
kerrynf: Had to remove the whole block here, but when I add this back in the next PR I'll exclude the testUrlParams section since I don't need to set anything here.

KangxuanYe (approved):

AE-MS (approved):
:shipit:

PR #2270, opened 2024-04-17
Trharris/authenticate https

TrevorJoelHarris:

## Description

`authentication.authenticate` will now throw an error if any url that is not https is passed into it. This should have already been the case and it not being enforced in TeamsJS is a bug.

### Main changes in the PR:

1. Updated `authentication.authenticate` to throw an error if a non-https url is passed to it
2. Restructured `authentication.authenticate` to make it clear that there were only 2 case: HostClientType.web || !HostClientType.web
3. Removed unnecessary and duplicative unit tests from authentication code
4. Fixed a bug in `utils.initializeWithContext` (a test only function) where many people were passing in an undefined HostClientType which can cause various weirdness during test execution. That function now has a default HostClientType value if one is not passed in.

### Validation performed:

1. Ran all unit tests

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


ydogandjiev (approved):

change/@microsoft-teams-js-40968a44-8220-4162-a289-038e99f5bee3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated `authention.authenticate` so that it only accepts https URLs.",
```
AE-MS, line 3: >authention [](http://example.com/codeflow?start=23&length=10)

authentication

packages/teams-js/src/public/authentication.ts
```diff
@@ -202,6 +195,12 @@ export namespace authentication {
     });
   }
 
+  function throwIfStringNotValidHttpsUrl(link: string): void {
+    if (!isValidHttpsURL(new URL(link))) {
+      throw new Error(`${ErrorCode.INVALID_ARGUMENTS}: url must be https`);
```
AE-MS, line 200: >url must be http [](http://example.com/codeflow?start=55&length=16)

I probably wouldn't say you _have_ to do this but since I already made one suggestion that I think you'll want to make, I will make this one too:
Since you do have a string here maybe it would be a little friendlier if you included the URL? Could help someone debug the problem if it ever happens. E.g.,

`${ErrorCode.INVALID_ARGUMENTS}: url must be https and given URL (${link}) is not`

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/authentication.ts
```diff
@@ -202,6 +195,12 @@ export namespace authentication {
     });
   }
 
+  function throwIfStringNotValidHttpsUrl(link: string): void {
```
erinha, line 198: I see that in Niharika's PR she's also adding URL validation that is slightly different than this one. Is it possible to reconcile them?

https://github.com/OfficeDev/microsoft-teams-library-js/pull/2272/files#diff-6deac44c9539303a6e3b3d416f349dc7109be94374522c76e78161acd98ece25

packages/teams-js/src/public/authentication.ts
erinha, line 200: Would it be appropriate to throw an SdkError and include the `ErrorCode.INVALID_ARGUMENTS` as the ErrorCode parameter?

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -893,55 +914,65 @@ describe('Testing authentication capability', () => {
...
-            it(`authentication.authenticate should handle auth failure with parameters in the ${hostClientType} client from ${context} context`, async () => {
-              await utils.initializeWithContext(context, hostClientType);
-              const authenticationParams: authentication.AuthenticatePopUpParameters = {
-                url: 'https://someurl',
-                width: 100,
-                height: 200,
-                isExternal: true,
-              };
-              const promise = authentication.authenticate(authenticationParams);
+          it(`authentication.authenticate should throw an error if non-https URL passed in`, async () => {
```
erinha, line 944: Should this include that it's a non-web client?

packages/teams-js/test/public/location.spec.ts
```diff
@@ -293,8 +293,8 @@ describe('location', () => {
     });
 
     it('Frameless - getLocation should throw error when not supported in the runtime config', () => {
-      utils.setRuntimeConfig({ apiVersion: 1, supports: {} });
```
erinha, line 296: How did these work before?

packages/teams-js/src/public/authentication.ts
TrevorJoelHarris: No, honestly I don't think we should keep throwing SdkErrors. When we throw Error objects they come along with extra info like callstack that we don't get when throwing SdkErrors.

packages/teams-js/src/public/authentication.ts
TrevorJoelHarris: Yes, we can reconcile them. It seems like her PR should either be updaing the existing `isValidHttpsUrl` function, or it should be introducing a new function that throws errors instead of returning a boolean, but share the underlying "is this URL valid" logic with the logic in `isValidHttpsURL`

packages/teams-js/src/public/authentication.ts
erinha: Good to know, thanks!

packages/teams-js/test/public/location.spec.ts
TrevorJoelHarris: Long story, but essentially initializeWithContext used to behave very oddly because the majority of people were passing no HostClientType into it. That then tried to respond to the initialize message with an undefined HostClientType in the response which caused assorted "weirdness" because a supposedly required parameter in the response (HostClientType) was missing. Runtime was also in that response but after the messed up HostClientType value, which (I think) was causing this `setRuntimeConfig` call to not get wiped out by processing the initialize request.

Now that `initializeWithContext` *always* sets HostClientType, you need to set runtime config after that call is finished.

packages/teams-js/src/public/authentication.ts
TrevorJoelHarris: I will wait for Niharika to check in and then use her new `validateUrl` function for consistency.

packages/teams-js/src/public/authentication.ts
```diff
@@ -202,6 +195,12 @@ export namespace authentication {
...
+    if (!isValidHttpsURL(new URL(link))) {
+      throw new Error(`${ErrorCode.INVALID_ARGUMENTS}: ${link} must be valid https URL`);
```
ndangudubiyyam, line 200: What if the URL is a https URL but while casting the string to new URL() failed?. Do we want to throw that error.
This can happen when the input is an encoded URI and the new URL(endcodedURL) fails.

TrevorJoelHarris: Can you give me an example of the type of URL you are talking about? Not sure I follow.

ndangudubiyyam: ex: 'https%3A%2F%2Faccounts.microsoft.com%2F' is a valid url but it will fail the casting to new URL () because it is encoded.

TrevorJoelHarris: Talked about this with @ndangudubiyyam offline. This function is in a bit of a weird state because it allows strings to be passed in as URLs (strong typing is our friend and we should use it whenever we can).

What I have here (no use of decode) is actually correct relative to how this function is already working. Within the function there is code like this:

image

This code is intended to turn relative URLs into absolute ones by appending them to the end of current path (which is something you get for free when inserting a string into an <a> tag. That means that _before this change_ the type of URL that @ndangudubiyyam is describing would have already not been legal to pass in (an url encoded absolute URL) because it would have been identified as a relative URL and appended to the current path, giving you something like:

`https://accounts.microsoft.com/https%3A%2F%2Faccounts.microsoft.com%2F`

However, a url encoded _relative_ url would have worked just fine. If the url string `hello%20world` was passed into this function you'd get something like:

`https://accounts.microsoft.com/hello%20world`

Which is a legal URL!

I actually don't think I would want full URLs that are encoded to be allowed to be passed in here anyways, so I'll make sure we maintain the current behavior.

AE-MS (approved):
:shipit:

packages/teams-js/src/internal/utils.ts
```diff
@@ -429,6 +429,12 @@ export function validateUrl(url: URL, errorToThrow?: Error): void {
   }
 }
 
+export function fullyQualifyUrlString(urlString: string): URL {
+  const link = document.createElement('a');
+  link.href = urlString;
```
erinha, line 434: FWIW, I recently learned that `link.toString()` returns a read-only version of `link.href` -- not actually sure if there's any valid reason to prefer it in this case, but thought I'd share my recent awareness.

TrevorJoelHarris: That's interesting! I think we still want href here since we do want to overwrite href with a new value, but still might be useful someday.

erinha (approved):

packages/teams-js/src/internal/utils.ts
erinha: Oh duh. Yes of course!

packages/teams-js/src/internal/utils.ts
```diff
@@ -429,6 +429,12 @@ export function validateUrl(url: URL, errorToThrow?: Error): void {
   }
 }
 
+export function fullyQualifyUrlString(urlString: string): URL {
```
AE-MS, line 432: >fullyQualifyUrlString [](http://example.com/codeflow?start=16&length=21)

Can you add a comment describing what this function does with an example? I don't think it's obvious to people who haven't been staring at this code for a while why it's creating a new DOM element and what goal that achieves.

maglims, line 432: Interesting to notice: 
fullyQualifyUrlString is converting a string to Url type, then in validateUrl, it converts Url type back to string, because it needs to validate the tag and length etc on..

so wondering if it worths to update validateUrl method to accept a string, then after tag/length validation, we convert to url for https validation. i.e, it might have a chance to merge  fullyQualifyUrlString logics into validateUrl.

AE-MS, line 432: >urlString [](http://example.com/codeflow?start=38&length=9)

Would `fullUrlOrRelativePath` be a more accurate name for this?

packages/teams-js/src/public/authentication.ts
```diff
@@ -358,9 +348,11 @@ export namespace authentication {
     // Ensure that the new window is always smaller than our app's window so that it never fully covers up our app
     width = Math.min(width, Communication.currentWindow.outerWidth - 400);
     height = Math.min(height, Communication.currentWindow.outerHeight - 200);
+
     // Convert any relative URLs into absolute URLs before sending them over to the parent window
-    const link = document.createElement('a');
-    link.href = authenticateParameters.url.replace('{oauthRedirectMethod}', 'web');
+    const qualifiedURL = fullyQualifyUrlString(authenticateParameters.url.replace('{oauthRedirectMethod}', 'web'));
```
AE-MS, line 353: >qualifiedURL [](http://example.com/codeflow?start=10&length=12)

I don't actually care about this but it's kind of funny that you called this `fullyQualifiedURL` above but just `qualifiedUrl` here.

TrevorJoelHarris: Sometimes I like to leave little traps in the PR for you to find

packages/teams-js/src/internal/utils.ts
TrevorJoelHarris: It is interesting, but I think I want to leave this as-is. I would actually do a couple of "unnecessary" conversions within in internal functions in exchange for using the URL type as much as possible. There might be future improvements we can make there, but for now we have small, pretty understandable functions and have started to increase the number of URL objects we're using, which is good!

AE-MS (approved):
:shipit:

maglims (approved):

ndangudubiyyam (approved):

erinha (approved):

PR #2271, opened 2024-04-17
Defer auth window to Hosts.

shrshindeMSFT:

## Description

There is code in teamsjs to display and communicate with an auth window. It only executes in certain circumstances (when the host environment is web). In all other case, this function defers to the host. In the recent Copilot work, it came to an understanding that unconditionally deferring to the host always seems to work, even in web environments. Based on this, removing the authentication code in teamsjs and just always defer to the host in all situations. Hence this PR addresses this removal of code from `authentication.authenticate` function. 

### Main changes in the PR:

1. Update the `authentication.authenticate` code to defer to host in all host client type. 
2. Update unit test cases accordingly.

### Validation performed:

1. Updated unit test cases and verified in E2E test cases.

### Unit Tests added:

Yes

### End-to-end tests added:

N/A

### Change file added:

Yes

### DIFF
--- a/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
+++ b/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Defer auth window to host.",
+  "packageName": "@microsoft/teams-js",
+  "email": "98348000+shrshindeMSFT@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
--- a/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
+++ b/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
@@ -1,5 +1,5 @@
 {
-  "type": "none",
+  "type": "patch",
   "comment": "Updated `authentication.authenticate` method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)",
   "packageName": "@microsoft/teams-js",
   "email": "98348000+shrshindeMSFT@users.noreply.github.com",
--- a/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
+++ b/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "Updated authentication.authenticate method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)",
+  "comment": "Updated `authentication.authenticate` method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)",
   "packageName": "@microsoft/teams-js",
   "email": "98348000+shrshindeMSFT@users.noreply.github.com",
   "dependentChangeType": "none"
--- a/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
+++ b/change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "Defer auth window to host.",
+  "comment": "Updated authentication.authenticate method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)",
   "packageName": "@microsoft/teams-js",
   "email": "98348000+shrshindeMSFT@users.noreply.github.com",
   "dependentChangeType": "none"
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -4,7 +4,7 @@ import {
   sendMessageToParentAsync,
   waitForMessageQueue,
 } from '../internal/communication';
-import { registerHandler, removeHandler } from '../internal/handlers';
+// import { registerHandler, removeHandler } from '../internal/handlers';
 import { ensureInitializeCalled, ensureInitialized } from '../internal/internalAPIs';
 import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
 import { FrameContexts } from './constants';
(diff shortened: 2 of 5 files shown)


packages/teams-js/src/public/authentication.ts
```diff
@@ -395,54 +327,6 @@ export namespace authentication {
     removeHandler('navigateCrossDomain');
   }
```
AE-MS, line 328: If we are deleting the start code presumably we can also delete the stop code?

packages/teams-js/src/public/authentication.ts
```diff
@@ -341,51 +318,6 @@ export namespace authentication {
     }
   }
```
AE-MS, line 319: Is this code still needed?

AE-MS:
>   let authWindowMonitor: number | undefined;

Do we need these?

---
Refers to: packages/teams-js/src/public/authentication.ts:27 in a05902d. [](commit_id = a05902dd082b428c33fae7c42efc1632e56cbaea, deletion_comment = False)

AE-MS:
>   }

Is this code still needed? I would think that since teamjs isn't creating an auth window anymore and not needing to answer calls from it anymore, it won't need to handle success/failure since those messages would just go directly to the host library now?

---
Refers to: packages/teams-js/src/public/authentication.ts:53 in a05902d. [](commit_id = a05902dd082b428c33fae7c42efc1632e56cbaea, deletion_comment = False)

AE-MS:
>   }

Are these still needed if the success/failure messages should go to the host library now?

---
Refers to: packages/teams-js/src/public/authentication.ts:410 in a05902d. [](commit_id = a05902dd082b428c33fae7c42efc1632e56cbaea, deletion_comment = False)

change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Defer auth window to host.",
```
AE-MS, line 3: >Defer auth window to host. [](http://example.com/codeflow?start=14&length=26)

Suggestion:

Updated `authentication.authenticate` method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/authentication.ts
shrshindeMSFT: We cannot delete this code as it's used in `initialize` function indirectly. When teamsJS is `initialized()` we register two handlers `authentication.authenticate.success` and `authentication.authenticate.failure` with `handleSuccess` and `handleFailure` respectively. Even if the auth window is deferred to the host, we need to monitor that the popped up window is closed and what was the message returned(success/failure). This is my understanding, if you need more explanation, I'm happy to talk.

change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
shrshindeMSFT: Updated.

packages/teams-js/src/public/authentication.ts
shrshindeMSFT: I tried it out and apparently, we don't need this code anymore as it'll be handled by hosts.

packages/teams-js/src/public/authentication.ts
shrshindeMSFT: Removed in the latest commit

change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
TrevorJoelHarris, line 2: this probably isn't a "none" change, since the behavior will be different for some applications before and after, right? Maybe patch instead?

change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Updated authentication.authenticate method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Updated `authentication.authenticate` method to defer authentication requests in web hosts to the host, rather than handling internally (requests on all other types of hosts were already being deferred)",
```

packages/teams-js/src/public/authentication.ts
```diff
@@ -159,46 +128,25 @@ export namespace authentication {
...
-            }
-          }),
-        );
-      } else {
-        // Open an authentication window with the parameters provided by the caller.
-        authHandlers = {
-          success: resolve,
-          fail: reject,
-        };
-        openAuthenticationWindow(authenticateParameters);
```
TrevorJoelHarris, line 200: Please make sure @ydogandjiev and Teams know about this. Yuri mentioned that in the past this code path was needed on some browsers and not others because some browsers decided to block the popup from happening on web if it was launched from Teams instead of from inside TeamsJS. He mentioned that Firefox and Safari specifically were a problem, but would be good to do a thorough test pass of all major browsers to make sure this change isn't going to break anything.

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/authentication.ts
shrshindeMSFT: Sure, that's the plan before checking in the code.

change/@microsoft-teams-js-5f6c016e-96a3-464a-95e1-67d1f38531b0.json
shrshindeMSFT: Updated

packages/teams-js/src/public/authentication.ts
shrshindeMSFT: I have created a document and share it with Yuri.

packages/teams-js/src/public/authentication.ts
```diff
@@ -1,15 +1,13 @@
 import {
   Communication,
-  sendMessageEventToChild,
   sendMessageToParent,
   sendMessageToParentAsync,
   waitForMessageQueue,
 } from '../internal/communication';
-import { GlobalVars } from '../internal/globalVars';
-import { registerHandler, removeHandler } from '../internal/handlers';
+// import { registerHandler, removeHandler } from '../internal/handlers';
```
AE-MS, line 7: Please remove

AE-MS:
>           allowedHostClientType.forEach((hostClientType) => {

Do the below two tests still need to be run for every possible hostClientType, given that their behavior no longer varies for any host client type? It seems like just testing it for one of them should be sufficient.

---
Refers to: packages/teams-js/test/public/authentication.spec.ts:633 in 370cca9. [](commit_id = 370cca99784243f0c9a78469905bb2e7848ae49c, deletion_comment = False)

packages/teams-js/src/public/authentication.ts
shrshindeMSFT: Updated.

AE-MS:
>         allowedHostClientType.forEach((hostClientType) => {

I'm sorry I didn't see this after leaving my previous comment. It seems like this can also be removed for the same reason as above?

---
Refers to: packages/teams-js/test/public/authentication.spec.ts:687 in 5435101. [](commit_id = 5435101453753aa0ca06574f77a8031cbe8fa574, deletion_comment = False)

packages/teams-js/src/public/authentication.ts
```diff
@@ -159,46 +127,25 @@ export namespace authentication {
...
-        );
-      } else {
-        // Open an authentication window with the parameters provided by the caller.
-        authHandlers = {
-          success: resolve,
-          fail: reject,
-        };
-        openAuthenticationWindow(authenticateParameters);
-      }
+    return new Promise<string>((resolve) => {
```
TrevorJoelHarris, line 130: Now that we know this code is needed for some specific browsers, it would be good to change this PR to instead put in a thorough comment documenting why the `window.open` flow is needed and how it works.

PR #2274, opened 2024-04-17
Add release yaml

jekloudaMSFT:

## Description

To migrate our release pipeline to OneBranch releases, we need to define our release pipeline in yaml. This is that yaml.
I'm starting with the beta release pipeline, and I'll move the others over once we confirm this works

### Main changes in the PR:

1. Added sdf-release pipeline

### Validation performed:

1. None yet, need to set up the pipeline on Azure to validate

### Unit Tests added:

No

### End-to-end tests added:
No

### Change file added:

No


tools/releases/sdf-release.yml
```diff
@@ -0,0 +1,98 @@
...
+        - task: NodeTool@0
+          displayName: Use Node 18.x
+          inputs:
+            versionSpec: 18.x
+        - task: PowerShell@2
+          displayName: Update npmrc with NPM-TOKEN
+          inputs:
+            filePath: $(PIPELINE.WORKSPACE)\_OfficeDev.microsoft-teams-library-js\scripts\add-npmrc-npmtoken.ps1
+            script: >
+              $version = Get-ChildItem -Path $(System.DefaultWorkingDirectory)/_OfficeDev.microsoft-teams-library-js/CDNFeed -Directory -Name
```
jadahiya-MSFT, line 62: This should be a separate Powershell task that happens at the very end.

jadahiya-MSFT (changes requested):
Just waiting on the above

jadahiya-MSFT (approved):

tools/releases/sdf-release.yml
```diff
@@ -0,0 +1,80 @@
...
+      project: ISS
+variables:
+  - name: System.Debug
+    value: true
+extends:
+  template: v2/Microsoft.Official.yml@GovernedTemplates
+  parameters:
+    platform:
+      name: m365
+      workload: IC3
```
AE-MS, line 20: >workload: IC3 [](http://example.com/codeflow?start=6&length=13)

I'm going to assume this is correct even though we aren't part of IC3?

tools/releases/sdf-release.yml
```diff
@@ -0,0 +1,80 @@
...
+            steps:
+              - download: _OfficeDev.microsoft-teams-library-js
+              - task: prepare-deployment@1
+                displayName:
+                inputs:
+                  taskType: credentialFetchTaskAzureRM
+                  armserviceconnection: $(serviceConnectionId)
+                  subscriptionid: $(subscriptionId)
+              - task: AssetRetention@3
+                displayName: ARtifact Retention Orchestrator Workflow (ARROW)
```
AE-MS, line 43: >ARtifact [](http://example.com/codeflow?start=29&length=8)

Artifact

AE-MS (changes requested):
I don't really know how to do a good job reviewing this but if Jay is okay with it and it works, seems okay to me with the tiny suggestion and question.

tools/releases/sdf-release.yml
jadahiya-MSFT: Strangely enough this is actually on purpose to hit the ARROW acronym. Otherwise it would've been Arow (not controlled by our team).

jadahiya-MSFT:
> I don't really know how to do a good job reviewing this but if Jay is okay with it and it works, seems okay to me with the tiny suggestion and question. Thank you for doing this!
> 
> BTW if our release pipeline today isn't in YAML, where is it? Is it just in some ADO UI somewhere?

Currently our release pipeline gets translated to YAML but it is configured through ADO UI, without any allowed YAML changes.

tools/releases/sdf-release.yml
AE-MS: lol

tools/releases/sdf-release.yml
jekloudaMSFT: I can ask the support team for confirmation. This was generated by the OneBranch tooling, but the allowed options are Substrate, SPO, OfficeOnline, IC3, Teams, Outlook, and Yammer. None of which seem like an obvious choice for our releases.

jekloudaMSFT: Asked and answered. We should use Substrate as a placeholder for now, and this value will be optional in the future

AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

PR #2276, opened 2024-04-17
Add ability for teams test app to remember scroll position on reload

alexneyman-MSFT:

https://github.com/OfficeDev/microsoft-teams-library-js/assets/112656629/6010ddf5-a342-429c-9195-b09c7bcb5d9b

## Description

Our test app is very large and sometimes we're using it to manually test a specific capability which requires scrolling or ctrl-F to the right spot every reload. This change optionally sets the scroll position of the window and stores it in session storage to enable returning to the same spot after a reload. I've also added a scroll-to-top button.
### Main changes in the PR:

1. Save last scroll position to session storage and return to that spot on reload (if toggled on)
2. add scroll-to-top button for easily getting back to the top

### Validation performed:

See attached video


apps/teams-test-app/src/components/RememberScrollToggle.tsx
```diff
@@ -0,0 +1,54 @@
...
+  // Initialize the toggle state from sessionStorage
+  const [isScrollFeatureActive, setIsScrollFeatureActive] = useState(
+    () => JSON.parse(sessionStorage.getItem('isScrollFeatureActive') || '{}') ?? true,
+  );
+
+  useEffect(() => {
+    // Save the toggle state to sessionStorage whenever it changes
+    sessionStorage.setItem('isScrollFeatureActive', JSON.stringify(isScrollFeatureActive));
+
+    console.log('scrolling');
```
jadahiya-MSFT, line 14: Do we want to keep the `console.log` statements in here?

apps/teams-test-app/src/components/RememberScrollToggle.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    if (!isScrollFeatureActive) {
+      return;
+    }
+
+    const handleScroll = throttle(() => {
+      const scrollPosition = window.scrollY;
+      sessionStorage.setItem('iframeScrollPosition', JSON.stringify(scrollPosition));
+      console.log('scrolling');
```
jadahiya-MSFT, line 23: Same for here

apps/teams-test-app/src/components/ScrollToTop.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useEffect, useState } from 'react';
+
+const ScrollToTopButton = () => {
+  const [isVisible, setIsVisible] = useState(false);
+
+  // Function to check scroll position and set visibility
+  const toggleVisibility = () => {
+    if (window.scrollY > 500) {
```
jadahiya-MSFT, line 8: Whats your thoughts on if we instead used a variable constant for the 500? Something like `const MIN_SCROLL_POSITION`.

apps/teams-test-app/src/components/RememberScrollToggle.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }, 2000);
```
jadahiya-MSFT, line 24: What's your thoughts on changing the 2000 here to `const THROTTLE_TIME = 2000`?

apps/teams-test-app/src/components/RememberScrollToggle.tsx
```diff
@@ -0,0 +1,54 @@
+import throttle from 'lodash.throttle';
```
jadahiya-MSFT, line 1: What's the difference between using throttle and wait if you don't mind me asking? Just genuinely curious

TrevorJoelHarris: Not sure what `wait` function you mean, but `throttle` will make sure that a function only gets called once every X milliseconds (2000 in this case). Will prevent a function from being called on every scroll event for example.

apps/teams-test-app/src/components/RememberScrollToggle.tsx
TrevorJoelHarris: Probably not, this looks like it was just used during development to make sure the function was getting called and also to make sure it wasn't getting called too frequently. Sound correct @alexneyman-MSFT ?

apps/teams-test-app/src/components/RememberScrollToggle.tsx
TrevorJoelHarris: Agreed, anytime we need to use specific numbers like this it's always my preference to put them in something like a named const. Makes it much easier to find and change in the future, as well as document what the const is for, why it has that value, etc. Good to avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) whenever possible!

apps/teams-test-app/src/components/ScrollToTop.tsx
```diff
@@ -0,0 +1,59 @@
...
+      // Adjust 500 to the scroll position you deem appropriate
```
TrevorJoelHarris, line 9: This comment would be a good one to move to the name const you make to store this value (and probably remove reference to 500 in case that value changes in the future

TrevorJoelHarris (changes requested):
Left a few change requests for you!

TrevorJoelHarris:
Talked with @alexneyman-MSFT , says it is no longer needed so closing

PR #2277, opened 2024-04-17
Private MessageChannels: split capability and add dataLayer.getDataLayerPort

kerrynf:
## Description

This PR will rework the private MessageChannels API to properly define sub capabilities and add in a second sub capability for dataLayer, adding getDataLayerPort.

### Main changes in the PR:

1. Rework MessageChannels capability to support separate sub capabilities per channel type
2. Add a second channel type for dataLayer

### Validation performed:

1. Updated the test apps
2. Added and updated the unit tests

### Unit Tests added:

Yes

### End-to-end tests added:

No - E2E will need to be added as fast follow to manage versioning challenges around the refactor.

### Change file added:

Yes

### Related PRs:

[#2140](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2140)

### Next/remaining steps:

- E2E tests will be added as fast follow once the version updates are available.


packages/teams-js/test/private/messageChannels.spec.ts
```diff
@@ -40,27 +41,100 @@ describe('messageChannels', () => {
...
+      await expect(messageChannels.dataLayer.getDataLayerPort()).rejects.toThrowError(
+        new Error(errorLibraryNotInitialized),
+      );
     });
   });
 
   Object.values(FrameContexts).forEach((context) => {
-    describe('Testing isSupported', () => {
-      it('should return true if the capability is supported', async () => {
+    describe('Testing messageChannels isSupported', () => {
```
sumathin-lang, line 72: whats the value in running through each frameContext? the code under test doesnt use this

change/@microsoft-teams-js-f6daee5c-d395-44ed-a4f3-670c3bbc44fe.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added private DataLayer capability",
+  "packageName": "@microsoft/teams-js",
+  "email": "kerrynb@microsoft.com",
+  "dependentChangeType": "patch"
```
sumathin-lang, line 6: i assume it is okay to mark as patch even though it is a breaking change since this is internal API?

sumathin-lang (approved):

change/@microsoft-teams-js-f6daee5c-d395-44ed-a4f3-670c3bbc44fe.json
kerrynf: It isn't a patch, it is minor bump (the top type is the type to look at here). Trevor said that since these APIs are marked beta the breaking changes are not an issue (this would be true even if they weren't private I assume).

`"type": "minor",`

packages/teams-js/test/private/messageChannels.spec.ts
kerrynf: This was a requirement from the teams-js team on my previous PR (I had the same question). Some APIs change based on frame context, so it was recommended to make sure we are testing this just in case some one starts introducing changes in this area.

TrevorJoelHarris: Yes, kerryn is correct. Really, sumathi is too and for this particular code right now there's no difference. Testing for each framecontext will make sure that if someone goes and removes one in the future it will break a test (which is good because removing a frame context would be a breaking change and we don't want to do that accidentally).

Ultimately, I would like to get rid of the framecontext concept entirely and depend entirely on isSupported, allowing the hosts to control what is and isn't supported based on their knowledge of where the app is running, but there are one or two things we'd need to change before we get there

change/@microsoft-teams-js-f6daee5c-d395-44ed-a4f3-670c3bbc44fe.json
TrevorJoelHarris: Correct, if something is marked as beta we reserve the right to change the signature with little or no notice.

TrevorJoelHarris (changes requested):

TrevorJoelHarris:
This looks good to me, will approve once you have a signed off security review

TrevorJoelHarris (approved):

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -198,7 +198,8 @@ export const enum ApiName {
   Menus_SetNavBarMenu = 'menus.setNavBarMenu',
   Menus_SetUpViews = 'menus.setUpViews',
   Menus_ShowActionMenu = 'menus.showActionMenu',
-  MessageChannels_GetTelemetryPort = 'messageChannels.getTelemetryPort',
+  MessageChannels_Telemetry_GetTelemetryPort = 'messageChannels.telemetry.getTelemetryPort',
+  MessageChannels_DataLayer_GetDataLayerPort = 'messageChannels.dataLayer.getDataLayerPort',
```
shrshindeMSFT, line 202: Can we alphabetize this?

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -198,7 +198,8 @@ export const enum ApiName {
   Menus_SetNavBarMenu = 'menus.setNavBarMenu',
   Menus_SetUpViews = 'menus.setUpViews',
   Menus_ShowActionMenu = 'menus.showActionMenu',
-  MessageChannels_GetTelemetryPort = 'messageChannels.getTelemetryPort',
+  MessageChannels_Telemetry_GetTelemetryPort = 'messageChannels.telemetry.getTelemetryPort',
```
shrshindeMSFT, line 201: Also, can you explain how will the telemetry will be handled from previous `messageChannels.getTelemetryPort` to `messageChannels.telemetry.getTelemetryPort`? Can't we use the existing one?

jadahiya-MSFT (approved):

packages/teams-js/src/private/messageChannels.ts
```diff
@@ -14,42 +14,144 @@ import { runtime } from '../public/runtime';
  * Limited to Microsoft-internal use
  */
 export namespace messageChannels {
-  let telemetryPort: MessagePort | undefined;
+  export namespace telemetry {
```
shrshindeMSFT, line 17: Why do we need nested namespace?

PR #2278, opened 2024-04-18
Split iOS E2E Testing pipeline to two

KangxuanYe:
**Summary:**
Separating iOS E2E testing pipeline to be 2 to reduce executing time and avoid hitting time limitation.

**Main Changes:**
Passing test plan name to yaml template to execute E2E test separately in two different pipelines.

**Validation Steps:**
For each of pipeline, you could click 'IOS UI testing' task and at the top of the console, you can read script to figure out which testing plan is executed.

For example, `/usr/bin/xcodebuild -configuration Release -workspace ... -scheme ... -testPlan iosE2ETestPlanA** -sdk iphonesimulator -parallel-testing-enabled YES -parallel-testing-worker-count 2 -destination 'platform=iOS Simulator,name=iPhone 14,OS=16.2' -screenshot-enabled="YES" test`

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -246,7 +246,7 @@ extends:
                   androidAppHostingSdkGitPath: AndroidAppHostingSdk
 
           - job: E2ETestIOS
-            displayName: 'E2E Test - IOS'
+            displayName: 'E2E Test - IOS - Plan A'
             pool:
               name: Azure Pipelines
               image: internal-macos12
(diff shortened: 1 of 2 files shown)


jadahiya-MSFT (approved):

azure-pipelines.yml
```diff
@@ -246,7 +246,7 @@ extends:
                   androidAppHostingSdkGitPath: AndroidAppHostingSdk
 
           - job: E2ETestIOS
-            displayName: 'E2E Test - IOS'
+            displayName: 'E2E Test - IOS - Plan A'
```
jadahiya-MSFT, line 249: I'm not a huge fan of Plan A and Plan B nomenclature, but drawing a blank on alternatives for now.

KangxuanYe: This blank cannot be updated dynamically so I will use plan A/B to indicate which iOS test plan (there is a plan file) is used. Once I figure out how to dynamically do that as web hub sdk does, I can update here to use alphabetical range.

jadahiya-MSFT: Sounds good, thanks

jadahiya-MSFT (dismissed):

AE-MS (approved):
:shipit:

MengyiGong (approved):

aeunms (approved):
🚢

PR #2289, opened 2024-04-24
Turn webStorage capability into a "real" capability

AE-MS:
## Description

The `webStorage` capability isn't "real" in the sense that it provides only hard-coded (and now wrong) answers -- it doesn't ask the host anything. This change updates it so that all function calls (all one of them) are actually plumbed through to the host.

The reason it was originally written in a hard-coded fashion (as far as I can tell) was because the authoring team just wanted it to return a specific value on Teams desktop. Through various iterations it ended up depending on the hard-coded runtime passed from Teams v1. Once Teams v2 started passing its own runtime, this capability "broke."

Rather than continue to depend on hard-coded data or assumptions about what functionality is available where, this updates the code so that the capability acts like a "real" capability and is exposed via the host, allowing the host to indicate how each function should behave.

### Unit Tests added:

Existing unit tests were updated to reflect the current, updated functionality.

### End-to-end tests added:

E2E tests have been added but not in this change. Per the recommendation from @KangxuanYe, the test JSON file is being checked-in blank here. Then, separately, I will check in a change in the host SDK which adds a matching TS file. Then I will come back to this repo and check in the E2E tests I wrote and have already tested locally.

### Change file added:

Yes


apps/teams-test-app/e2e-test-data/webStorage.json
```diff
@@ -0,0 +1 @@
+{}
```
AE-MS, line 1: Per the recommendation from @KangxuanYe , this file is blank for now. The real tests will be added in a separate change, once the corresponding test file is added in the host SDK repo.

packages/teams-js/src/public/authentication.ts
```diff
@@ -341,13 +341,12 @@ export namespace authentication {
     }
   }
 
-
```
AE-MS, line 344: This whitespace change was automatically made by VSCode but I'm checking it in otherwise it will keep making it.

packages/teams-js/test/utils.ts
```diff
@@ -177,6 +177,24 @@ export class Utils {
...
+  /**
+   * This function is used to find a message by the action name provided to the send* functions. Usually the action name is the
+   * name of the function being called..
+   * @param actionName - The action name used in the sent message
+   * @param k - In the case where you expect there are multiple messages sent with the same action name,
+   * use this as a zero-based index to return the kth one. Default is 0 (will return the first match).
+   * @returns {MessageRequest} The found message
+   * @throws {Error} If the message is not found
+   */
+  public findMessageByActionName(actionName: string, k: number = 0): MessageRequest {
```
AE-MS, line 189: >actionName [](http://example.com/codeflow?start=33&length=10)

`actionName` is what the send* functions actually call this parameter.

jadahiya-MSFT (approved):

shrshindeMSFT (approved):
LGTM 🚢

apps/teams-test-app/e2e-test-data/webStorage.json
KangxuanYe: Update: Below is the right json file you should follow when you add a new capability / APIs.

{
  "name": "Your CapabilityName",
  "testCases":[{}]
}

Updated here in case someone gets mislead by my words before.

PR #2301, opened 2024-05-02
Upgrade `pnpm` to v9

shrshindeMSFT:

## Description

Recently there have been an issue where `pnpm-lock.yaml` file is having conflicts by using different `pnpm` versions. To address this issue this repo will be moving to `pnpm v9.0.6` which is the latest version and the change file states that `pnpm v9` has adopted `lockfile v9` format which is more resistance to git conflicts (more about the changelog [here](https://github.com/pnpm/pnpm/releases/tag/v9.0.0)). Also, we will be upgrading `pnpm` version when major bump is available (i.e., v9 -> v10).

### Main changes in the PR:

1. Update pnpm to v9.0.6 in package.json
2. Override jsdom to latest to remove the sub deprecated dependencies.
3. Update the CI pipeline to use pnpm v9.0.6

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v4.1.1
 
       - name: Install pnpm
-        run: npm install -g pnpm
+        run: npm install -g pnpm@9.0.6
 
       - name: Install dependencies
         run: pnpm install
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v4.1.1
 
       - name: Install pnpm
-        run: npm install -g pnpm@7.30.1
+        run: npm install -g pnpm
 
       - name: Install dependencies
         run: pnpm install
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -106,6 +106,7 @@
       "express": "^4.19.2",
       "follow-redirects": "^1.15.6",
       "glob-parent": "^5.1.2",
+      "jsdom": "^24.0.0",
```
shrshindeMSFT, line 109: `jsdom` had two dependencies which were deprecated so need to override it to latest version. Also, this won't affect the package as `jsdom` is only used by `jest-environment-jsdom`.

.github/workflows/prerelease.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v4.1.1
 
       - name: Install pnpm
-        run: npm install -g pnpm@7.30.1
+        run: npm install -g pnpm
```
aeunms, line 22: Does removing the `@<version number>` automatically install the `latest` tag? If not, should we be doing `@latest`?

shrshindeMSFT: Yes, it'll install the latest version if no tag/version number is provided. Here is the official link: https://pnpm.io/installation#using-npm

jekloudaMSFT (approved):

TrevorJoelHarris (approved):

PR #2305, opened 2024-05-03
[DRAFT] Updated DDL cdn deployment and endpoint for shorter cache control

jadahiya-MSFT:
Currently our DDL deployments have a default `cache-control` setting of 7 days to 7300 days. This means that when we make updates to the DDL that need to be seen more immediately we would need to flush the CDN. Moving the `validDomains` artifact into a `*7dttl*` folder designation updates the `cache-control` to one with a 7 day time to live to avoid needing a CDN flush.


tools/yaml-templates/build-test-publish.yml
```diff
@@ -106,6 +106,14 @@ steps:
       flattenFolders: true
     displayName: 'Copy validDomains for CDN'
 
+  - task: CopyFiles@2
```
TrevorJoelHarris, line 109: It would be great to capture why this has the effect stated in a comment in the yaml file. The explanation you gave me makes sense, but I'm worried we will not remember how this works in a couple of months. The functionality described is a side effect of moving the file into a subfolder of the 7dttl folder.

TrevorJoelHarris (approved):

tools/yaml-templates/build-test-publish.yml
```diff
@@ -106,6 +106,15 @@ steps:
       flattenFolders: true
     displayName: 'Copy validDomains for CDN'
 
+  #The`*7dttl*` folder allows us to set everyfile underneath it with a 7 day cache-control instead of the default of 7300 days
```
AE-MS, line 109: every file

tools/yaml-templates/build-test-publish.yml
```diff
@@ -106,6 +106,15 @@ steps:
       flattenFolders: true
     displayName: 'Copy validDomains for CDN'
```
AE-MS, line 107: If we want it to be 7 days, should we remove this line which will still put a copy in the original 7300 days place?

jadahiya-MSFT: We want this for back compat with the older versions of TeamsJS

AE-MS: Oh yes, makes sense. Can you add a comment above this line explaining that? Would not be crazy for a dummy like me to come along later and be like "Why do we have two of these things? DELETE"

TrevorJoelHarris (approved):

PR #2311, opened 2024-05-07
Updated `search` and `otherAppStateChange` e2e tests data in teams-js

Ella-ly:

## Description

The event input data should be sent in the required format. This PR removed unnecessary array wrapping for the event data in e2e tests.

### Main changes in the PR:

1. Updated `search` and `otherAppStateChange` e2e tests data in teams-js.

### Validation performed:

1. e2e tests passed.

### Unit Tests added:

No: test data changes.

### End-to-end tests added:

No: test data changes.

### Change file added:

No: test data changes.

### DIFF
--- a/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
+++ b/apps/teams-test-app/e2e-test-data/otherAppStateChange.json
@@ -23,11 +23,10 @@
       "type": "registerAndRaiseEvent",
       "boxSelector": "#box_otherAppStateChange_registerInstallHandler",
       "eventName": "otherApp.install",
-      "eventData": [
+      "eventData":
         {
           "appIds": ["123", "456"]
-        }
-      ],
+        },
       "expectedTestAppValue": "received"
     },
     {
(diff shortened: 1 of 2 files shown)


AE-MS:
Does the version of the host SDK need to be restricted in these tests? E.g., do tests without the extra wrapping need to only be run against versions >2.17.0?

Ella-ly:
> Does the version of the host SDK need to be restricted in these tests? E.g., do tests without the extra wrapping need to only be run against versions >2.17.0?



> Does the version of the host SDK need to be restricted in these tests? E.g., do tests without the extra wrapping need to only be run against versions >2.17.0?

No, the additional wrapping unintentionally bypassed the old logic in the host SDK. Initially, the data was supposed to be wrapped in an array by host SDK. However, by wrapping the data in tests, it circumvented the old isArray check in the host SDK, avoiding further wrapping, and thus passed in the expected format. By removing the wrapping in tests, for older versions, the isArray check exists and will wrap the data according to the host SDK logic. Additionally, with the logic now corrected, the data should be sent in the required format.

AE-MS (approved):

aeunms (approved):

PR #2328, opened 2024-05-16
[Don't review yet] Access hostname for app's

shrshindeMSFT:

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -70,13 +70,9 @@ interface InitializeResponse {
 }
 
 /**
- * @internal
- * Limited to Microsoft-internal use
+ * set windows for communication
  */
-export function initializeCommunication(
-  validMessageOrigins: string[] | undefined,
-  apiVersionTag: string,
-): Promise<InitializeResponse> {
+function setWindows(validMessageOrigins: string[] | undefined): void {
   // Listen for messages post to our window
   CommunicationPrivate.messageListener = async (evt: DOMMessageEvent): Promise<void> => await processMessage(evt);
 
@@ -102,10 +98,20 @@ export function initializeCommunication(
       extendedWindow.onNativeMessage = handleParentMessage;
     } else {
       // at this point we weren't able to find a parent to talk to, no way initialization will succeed
-      return Promise.reject(new Error('Initialization Failed. No Parent window found.'));
+      throw new Error('Initialization Failed. No Parent window found.');
     }
   }
+}
 
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function initializeCommunication(
+  validMessageOrigins: string[] | undefined,
+  apiVersionTag: string,
+): Promise<InitializeResponse> {
+  setWindows(validMessageOrigins);
   try {
     // Send the initialized message to any origin, because at this point we most likely don't know the origin
     // of the parent window, and this message contains no data that could pose a security risk.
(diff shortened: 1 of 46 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -373,6 +379,33 @@ export function sendNestedAuthRequestToTopWindow(message: string): NestedAppAuth
...
+    {
+      id: CommunicationPrivate.nextMessageId++,
+      timestamp: Date.now(),
+      func: 'getHostName',
+    },
+  ];
+  try {
+    setWindows(undefined);
+    Communication.parentOrigin = '*';
+    return sendMessageToParentAsync(apiVersionTag, 'getHostName', request);
```
TrevorJoelHarris, line 397: Looking at how this function is called in `app.ts`, it looks in some cases the app won't be initialized yet when you hit this line. Can I call like this actually work before the app is initialized? I think it wouldn't but maybe I'm forgetting/missing something. I think if there's no host to send it to the message will just sit forever and never receive a response.

If I assume that this won't work before `initialize` then I think this approach will have some problems. `setWindows` will not throw an error right now if the app is hosted in an iframe that *isn't* the hub-sdk for example.

I wonder if a slightly different approach could work. There are two different configurations TeamsJS can be in if the app that is using it is being hosted:

1) Frameless (app is hosted inside something like a webview)
2) Framed (app is hosted in an iframe).

For case 1, I think you could check if that one is the case relatively easily: if the current window is an `ExtendedWindow` that has `nativeInterface` defined on it, we can assume that the app is probably hosted by a host that is using a webview. Theoretically, someone that is not an official host could be running the app in a webview, but if they are it is unlikely that they have set up the WebView this way. See lines 104-107

For case 2 it is probably trickier. One thing you *might* be able to do is look at the domain that `window.top` is running at and then compare that domain to the list of validOrigins from the CDN. We may want to let the app developer pass in an additional set of "valid" host origins to check for in addition to the CDN list (the same list of valid origins they can pass to initialize).

If either of the cases above are true, we can tell the developer that they are *probably* being hosted by a host that will respond to their initialize call. We still might sometimes get it wrong, but in the worst case app developers will still be doing the same thing they are now: calling initialize and waiting for it to timeout.

What do you think?

shrshindeMSFT:
```
Looking at how this function is called in app.ts, it looks in some cases the app won't be initialized yet when you hit this line. Can I call like this actually work before the app is initialized? I think it wouldn't but maybe I'm forgetting/missing something. I think if there's no host to send it to the message will just sit forever and never receive a response.
```

When the app is not hosted the set windows will throw error `Initialization failed. ...` and we can catch that error and tell the developer that app is not hosted. In `app.ts` we check if `GlobalVars.isInitializeCompleted` then use `sendAndMessageToParentAsync` as this will straight give the host name. For the second part where the app is not initialized `setWindows` sets the windows for communication and if it doesn't find any then it just throws error, if it finds the top window we send the message which will be handled by the hubSDK.

TrevorJoelHarris:
Closing an old draft that doesn't seem like it will be checked in

PR #2337, opened 2024-05-22
Added intAppId parameter to getContext

vikramtha:

## Description

This is adding the internal appID to getContext. This is what hubs use to differentiate the app internally as well as what ends up being used in deeplinking

### Main changes in the PR:

1. Added intAppId to getContext() and the AppInfo interface
2. Added dummy values for testing validation

### Validation performed:

1. Validated on Orange Web

### Unit Tests added:

???

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/change/@microsoft-teams-js-07a56fbc-6d3e-474d-b3d9-6c6fe3a3be50.json
+++ b/change/@microsoft-teams-js-07a56fbc-6d3e-474d-b3d9-6c6fe3a3be50.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added a new parameter to getContext() API called intAppId for Internal App ID",
+  "packageName": "@microsoft/teams-js",
+  "email": "vikramtha@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -354,6 +354,11 @@ export namespace app {
      * ID for the current visible app which is different for across cached sessions. Used for correlating telemetry data.
      */
     appLaunchId?: string;
+
+    /**
+     * Internal app id that is used by Hubs to distinguish between different apps sideloaded or in store
+     */
+    intAppId?: string;
   }
 
   /**
@@ -1007,6 +1012,7 @@ function transformLegacyContextToAppContext(legacyContext: LegacyContext): app.C
         ringId: legacyContext.ringId,
       },
       appLaunchId: legacyContext.appLaunchId,
+      intAppId: legacyContext.intAppId,
     },
     page: {
       id: legacyContext.entityId,
--- a/packages/teams-js/src/public/interfaces.ts
+++ b/packages/teams-js/src/public/interfaces.ts
@@ -757,6 +757,14 @@ export interface Context {
    * They help pre-fill the dialog with necessary information (`dialogParameters`) along with other details.
    */
   dialogParameters?: Record<string, string>;
+
+  /**
+   * @deprecated
+   * As of 2.0.0, please use {link app.Context.intAppId} instead
+   *
+   * Internal app id that is used by Hubs to distinguish between different apps sideloaded or in store
+   */
+  intAppId?: string;
 }
 
 /** Represents the parameters used to share a deep link. */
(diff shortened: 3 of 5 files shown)


packages/teams-js/src/public/app.ts
```diff
@@ -354,6 +354,11 @@ export namespace app {
      * ID for the current visible app which is different for across cached sessions. Used for correlating telemetry data.
      */
     appLaunchId?: string;
+
+    /**
+     * Internal app id that is used by Hubs to distinguish between different apps sideloaded or in store
```
TrevorJoelHarris, line 359:
```suggestion
     * Internal app id that is used by Hosts to distinguish between different apps sideloaded or in store
```

TrevorJoelHarris:
@vikramtha this is failing to build, so it won't get any PR reviews until the build is passing

vikramtha (commented):
Updating

packages/teams-js/src/public/interfaces.ts
```diff
@@ -758,6 +758,14 @@ export interface Context {
...
    */
   dialogParameters?: Record<string, string>;
+
+  /**
+   * @deprecated
+   * As of 2.0.0, please use {@link app.AppInfo.appId | app.Context.app.appId} instead
+   *
+   * App id that is used by Hosts to distinguish between different apps sideloaded or in store
+   */
+  appId: string;
```
Ella-ly, line 768: Since this is a new feature, can I know why we need to add this parameter to a deprecated interface and as a deprecated element?

vikramtha: I believe the Context deprecated interface is still used in the legacyContext transformation so I kept it

Ella-ly (dismissed):

TrevorJoelHarris (dismissed):

jadahiya-MSFT (dismissed):

change/@microsoft-teams-js-07a56fbc-6d3e-474d-b3d9-6c6fe3a3be50.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added a new parameter to `app.getContext` called `appId` which contains the current application's AppId. This value is used in deeplinks and by hosts to identify apps",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Added a new parameter to `app.getContext` called `appId` which contains the current application's AppId. This value is used in deeplinks and hosts to identify apps",
```

TrevorJoelHarris: (I'm just making a small tweak to see if that makes the pipeline retry correctly)

maglims (commented):
I haven't looked into much deeper but first question I have is: doesn't it need changes in hub sdks to populate back appId since it looks it is defined as required?

Are all hub sdk test passing?

packages/teams-js/src/public/interfaces.ts
TrevorJoelHarris: This one should probably also be optional

packages/teams-js/src/public/app.ts
```diff
@@ -354,6 +354,11 @@ export namespace app {
      * ID for the current visible app which is different for across cached sessions. Used for correlating telemetry data.
      */
     appLaunchId?: string;
+
+    /**
+     * App id that is used by Hosts to distinguish between different apps sideloaded or in store
+     */
+    appId: string;
```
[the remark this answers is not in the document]
vikramtha: This will always be populated as long as there is an app def sent back; I rewrote the code in hubSDK so that hubs no longer need to pass the value and instead it is taken from app definition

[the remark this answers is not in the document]
TrevorJoelHarris: What about hosts that aren't yet using the updated host-sdk? I believe in that case this would not be set, right? That's generally why we mark things like this optional, to handle backwards compatibility

TrevorJoelHarris (changes requested):
There are some questions about appId being optional, and I think it should be. If you have questions just send me a message and we'll talk about it.

packages/teams-js/src/public/app.ts
AE-MS, line 361: >string [](http://example.com/codeflow?start=11&length=6)

Can you use the `AppId` type for this? Strong types are better than strings. It should be easy to create this object as an `AppId` in `transformLegacyContextToAppContext` after the data comes in over the wire as a `string`.

jekloudaMSFT:
Resolved by https://github.com/OfficeDev/microsoft-teams-library-js/pull/2694

PR #2343, opened 2024-05-29
Fix the default input for showProfile API

maglims:
According to our official documentation, the available TriggerType for profile api are these: TriggerType = "MouseHover" | "Press" | "AppRequest"
https://learn.microsoft.com/en-us/javascript/api/@microsoft/teams-js/profile.triggertype?view=msteams-client-js-latest

Updating our default input to use the correct one "MouseHover" instead of "MouseClick". This will clear some confusions for devs who are using the default input while doing manual testing. Also, it will not work with "MouseClick" input(throwing an exception back to app) because it doesn't match the enum defined in hub sdk, which is "MouseHover".

## Description

> See above

### Main changes in the PR:

1. Update from "MouseClick" to "MouseHover" for the default input.

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


apps/teams-test-app/e2e-test-data/profile.json
```diff
@@ -23,9 +23,9 @@
           "width": 0,
           "height": 0
         },
-        "triggerType": "MouseClick"
+        "triggerType": "MouseHover"
```
maglims, line 26: This reminds me to check on web hub sdk, where profile is being supported.
 
It also needs a fix there?
https://office.visualstudio.com/ISS/_git/metaos-hub-sdk?path=%2Fpackages%2Fmetaos-hub-sdk%2Fsrc%2FServices%2FMessageHandling%2Fprofile-message-handler.test.ts&_a=contents&version=GBmain

@Kangxuan-Ye  @TrevorJoelHarris could you take a look and comment? thanks

KangxuanYe, line 26: I am so confused how could it work for web if you didn't make any change in Hub SDK side but simply replace MouseClick to MouseHover? It is not an input so I assume it should have a failure here.

maglims: Maybe due to this is the default input for the test app... for example, in iOS, we also don't use this default value for testing. I'd image the default values are mostly used by devs when they are doing the manual testing (so they can just use default and don't have to manually find the input). Does it sound reasonable?

TrevorJoelHarris: No change would be needed in the hub-sdk because the hub-sdk is just taking this entire payload it receives as part of the call and displaying the whole thing it in an alert. In this case, this test is only used in the web, and it's passing up MouseHover as the trigger type, and so the test is expecting that to show up in an alert.

The default value in ProfileAPIs.tsx is only added if a user presses the "Default" button in the test app, it's intended to make manual testing using the test app more straightforward, it can be tricky to figure out what format each call wants added to the text box without it. Making a change there should have no effect on the E2E tests.

TrevorJoelHarris (approved):

TrevorJoelHarris:
Spoke with @maglims and was told this was no longer needed because the test issues were addressed

PR #2355, opened 2024-06-12
Adding *.cloud.microsoft to valid domains list

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Adding `*.cloud.microsoft` to `validDomains`


packages/teams-js/src/artifactsForCDN/validDomains.json
```diff
@@ -37,6 +37,7 @@
     "fa000000125.resources.office.net",
     "fa000000129.resources.office.net",
     "fa000000124.resources.office.net",
-    "fa000000128.resources.office.net"
+    "fa000000128.resources.office.net",
+    "*.cloud.microsoft"
```
maglims, line 41: Just curious, will this cover from line 32-34?

jadahiya-MSFT: The answer is yes, but I think it's important for us to also distinguish the explicit partner requirements if we ever need to remove the wildcard

TrevorJoelHarris:
We should make sure that @sbtron and @AE-MS are both okay with and know this is being checked in before we actually do it.

PR #2361, opened 2024-06-14
Trharris/messageproxy

TrevorJoelHarris:
## Description

Work In Progress: Feel free to read and ask questions, but there is still more work (and tests) needed before this is ready for checkin.

We'd like to disable message proxying from child windows in as many situations as we can. Message proxying is when an application that is using teamsJS receives a post message from a child window (embedded iframe or new window that the app opened) and "repeats" that message up to the host.

Based on research from PM, we *believe* that the only legitimate use of this functionality if when the `authenticate.authentication` function is called from an app running in a web browser. In those cases, TeamsJS opens a new window for the auth to happen in to avoid the browser popup blocker. In that case, we still want TeamsJS to allow messages from the auth window to be repeated up to the host. In all other cases, we want to prevent message proxy-ing by default, only allowing it for apps that truly need that functionality. I'll go through the PR and leave some comments highlighting interesting bits for reviewers to pay attention to.

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -5,7 +5,7 @@
 import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
 import { FrameContexts } from '../public/constants';
 import { SdkError } from '../public/interfaces';
-import { latestRuntimeApiVersion } from '../public/runtime';
+import { latestRuntimeApiVersion, runtime } from '../public/runtime';
 import { version } from '../public/version';
 import { GlobalVars } from './globalVars';
 import { callHandler } from './handlers';
@@ -752,11 +752,15 @@ function isPartialResponse(evt: DOMMessageEvent): boolean {
   return evt.data.isPartialResponse === true;
 }
 
+const handleChildMessageLogger = communicationLogger.extend('handleChildMessage');
+
 /**
  * @internal
  * Limited to Microsoft-internal use
  */
 function handleChildMessage(evt: DOMMessageEvent): void {
+  const logger = handleChildMessageLogger;
+
   if ('id' in evt.data && 'func' in evt.data) {
     // Try to delegate the request to the proper handler, if defined
     const message = deserializeMessageRequest(evt.data as SerializedMessageRequest);
@@ -766,20 +770,25 @@ function handleChildMessage(evt: DOMMessageEvent): void {
       // @ts-ignore
       sendMessageResponseToChild(message.id, message.uuid, Array.isArray(result) ? result : [result]);
     } else {
-      // No handler, proxy to parent
-      sendMessageToParent(
-        getApiVersionTag(ApiVersionNumber.V_2, ApiName.Tasks_StartTask),
-        message.func,
-        message.args,
-        (...args: any[]): void => {
-          if (Communication.childWindow) {
-            const isPartialResponse = args.pop();
-            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-            // @ts-ignore
-            sendMessageResponseToChild(message.id, message.uuid, args, isPartialResponse);
-          }
-        },
-      );
+      if (GlobalVars.allowMessageProxy) {
+        // No handler, proxy to parent
+        sendMessageToParent(
+          getApiVersionTag(ApiVersionNumber.V_2, ApiName.Tasks_StartTask),
+          message.func,
+          message.args,
+          (...args: any[]): void => {
+            if (Communication.childWindow) {
+              const isPartialResponse = args.pop();
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              sendMessageResponseToChild(message.id, message.uuid, args, isPartialResponse);
+              runtime;
+            }
+          },
+        );
+      } else {
+        logger('Message from child not allowed to proxy: %o', message);
+      }
     }
   }
 }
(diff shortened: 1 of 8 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -766,20 +770,25 @@ function handleChildMessage(evt: DOMMessageEvent): void {
...
-        (...args: any[]): void => {
-          if (Communication.childWindow) {
-            const isPartialResponse = args.pop();
-            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-            // @ts-ignore
-            sendMessageResponseToChild(message.id, message.uuid, args, isPartialResponse);
-          }
-        },
-      );
+      if (GlobalVars.allowMessageProxy || GlobalVars.webAuthWindowOpen) {
```
TrevorJoelHarris, line 773: Both of these new GlobalVars start out with a value of false when the app is initialized. `allowMessageProxy` is only true if an undocumented function is called. `webAuthWindowOpen` is only true if `authentication.authenticate` is called from an app running in a browser, and it is changed back to false as soon as that authentication window is closed.

packages/teams-js/src/internal/communication.ts
```diff
@@ -766,20 +770,25 @@ function handleChildMessage(evt: DOMMessageEvent): void {
...
+          getApiVersionTag(ApiVersionNumber.V_2, ApiName.Tasks_StartTask),
+          message.func,
+          message.args,
+          (...args: any[]): void => {
+            if (Communication.childWindow) {
+              const isPartialResponse = args.pop();
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              sendMessageResponseToChild(message.id, message.uuid, args, isPartialResponse);
+              runtime;
```
TrevorJoelHarris, line 785: This section of code takes the `func` and `args` from the message TeamsJS received from its child and sends them in a new post message up to the parent. It also sets up a handler for dealing with the response, passing the response back down to the child (using `sendMessageResponseToChild`) instead of handling the returned data itself.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -177,7 +175,7 @@ export function handleThemeChange(theme: string): void {
     HandlersPrivate.themeChangeHandler(theme);
   }
 
-  if (Communication.childWindow) {
+  if (Communication.childWindow && (GlobalVars.allowMessageProxy || GlobalVars.webAuthWindowOpen)) {
```
TrevorJoelHarris, line 178: This and the load handler below have all been here for a long time, so I decided for now to gate them behind the `allowMessageProxy` logic instead of outright removing them.

packages/teams-js/src/internal/handlers.ts
```diff
@@ -248,17 +246,9 @@ async function handleBeforeUnload(): Promise<void> {
 
   if (HandlersPrivate.beforeSuspendOrTerminateHandler) {
     await HandlersPrivate.beforeSuspendOrTerminateHandler();
-    if (Communication.childWindow) {
```
TrevorJoelHarris, line 251: These sections of code were added specifically for some Sharepoint scenarios that are no longer supported. Because of that I opted just to remove the logic around passing the event to the child entirely. We could alternatively choose to play it safe and gate these behind the `allowMessageProxy` check instead of we wished.

packages/teams-js/src/public/authentication.ts
```diff
@@ -328,6 +328,7 @@ export namespace authentication {
     } finally {
       Communication.childWindow = null;
       Communication.childOrigin = null;
+      GlobalVars.webAuthWindowOpen = false;
```
TrevorJoelHarris, line 331: As far as I can tell, every time the auth window is closed by TeamsJS it is done using this function. Because of that, I am using this as a bottleneck to set `webAuthWindowOpen` back to false.

packages/teams-js/src/public/authentication.ts
```diff
@@ -364,6 +366,8 @@ export namespace authentication {
         : Communication.currentWindow.screenY;
     left += Communication.currentWindow.outerWidth / 2 - width / 2;
     top += Communication.currentWindow.outerHeight / 2 - height / 2;
+
+    GlobalVars.webAuthWindowOpen = true;
```
TrevorJoelHarris, line 370: Immediately before opening an auth window we set `webAuthWindowOpen` to true

packages/teams-js/src/public/pages.ts
```diff
@@ -1,10 +1,8 @@
 import {
-  Communication,
```
TrevorJoelHarris, line 2: All of the changes that were removed in this file were only added for Sharepoint scenarios which are no longer supported. As with the above example, we could choose to gate these behind an `allowMessageProxy` check if we wanted to be safer.

packages/teams-js/src/internal/handlers.ts
TrevorJoelHarris: Thinking about it more over the weekend, I should probably leave these in (and makes sure they're tested). While SharePoint is the reason they were added, it's possible people are using them for auth windows or other use cases. Will get to this "soon"

PR #2364, opened 2024-06-20
Prevent empty string from throwing TypeError while checking message o…

marcinkiewicz:
Validate empty origin

While creating URL from empty string, we would see `Failed to construct 'URL': Invalid URL` while validating the message origin. Add extra check to include this case.

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -564,12 +564,15 @@ async function shouldProcessMessage(messageSource: Window, messageOrigin: string
     messageOrigin === Communication.currentWindow.location.origin
   ) {
     return true;
-  } else {
+  } else if (messageOrigin) {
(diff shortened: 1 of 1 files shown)


TrevorJoelHarris:
Hi @marcinkiewicz, thank you for submitting a PR. We have a contribution guide here: https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md. If you're a Microsoft employee, please make sure to join the correct groups. Your PR also requires a changefile.

Can I ask more about what issue you are running into? The change you've made is in an extremely low-level part of the code that could have security implications.

marcinkiewicz:
Our root error boundary recently started catching the TypeError caused by the creating of URL from either malformed or empty URL (noticed as spike in our monitoring of TypeError issues).

I joined the groups and re-created the PR following the contribution guideline.

PR #2365, opened 2024-06-21
Enable pnpm to automatically install peers and fail if there are conflicts

AE-MS:
## Description

`pnpm` does not automatically install peer dependencies (which is a [recommended](https://nodejs.org/en/blog/npm/peer-dependencies) [behavior](https://javascript-code.dev/articles/206234324) and the default in npm) and this repo doesn't currently enable the setting to fail if conflicting/missing peer dependencies are detected. This change enables automatically installation of peer dependencies and will fail if there are any conflicts. Fortunately, there are none today but this will prevent problems from being accidentally introduced in the future (like in some other repos... <wink wink>)

### Main changes in the PR:

Added `.npmrc` file with appropriate `pnpm` configuration settings to enable automatic peer dependency installation and failure when it doesn't work

### Validation performed:

Neither `pnpm i` nor `pnpm install --resolution-only` fail with these new settings

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+auto-install-peers=true
+strict-peer-dependencies=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 auto-install-peers=true
+# Failing on missing/conflicting peer dependencies is a good thing, don't turn this off even though pnpm suggests it as an option whenever displaying a peer dependency error
 strict-peer-dependencies=true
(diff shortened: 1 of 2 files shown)


jadahiya-MSFT (dismissed):

TrevorJoelHarris (approved):

shrshindeMSFT (approved):
🚢

jadahiya-MSFT (approved):

PR #2372, opened 2024-06-24
Updated most lint packages (and made one whitespace change to clipboard.ts)

AE-MS:
## Description

Multiple lint packages were out of date; this change updates them.

This change also updates one source file (`clipboard.ts`) because VScode automatically formatted it in a preferred way.


packages/teams-js/src/public/clipboard.ts
```diff
@@ -103,7 +103,9 @@ export namespace clipboard {
     if (GlobalVars.isFramelessWindow) {
       return ensureInitialized(runtime) && runtime.supports.clipboard ? true : false;
     } else {
-      return ensureInitialized(runtime) && navigator && navigator.clipboard && runtime.supports.clipboard ? true : false;
+      return ensureInitialized(runtime) && navigator && navigator.clipboard && runtime.supports.clipboard
+        ? true
+        : false;
```
AE-MS, line 108: This was a whitespace change that VSCode did automatically

change/@microsoft-teams-js-8bc1eae5-cc68-4cdb-a5ee-9862ac5e5b8d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated multiple linting packages",
```
jadahiya-MSFT, line 3: While I understand that this is a patch, I wanted to bring up a conversation that I've been thinking about. 

Which is, should items like these that in theory only affects our development of this project be included in the release notes for the next version bump?

jadahiya-MSFT (approved):

change/@microsoft-teams-js-8bc1eae5-cc68-4cdb-a5ee-9862ac5e5b8d.json
AE-MS: FWIW the reason this particular change is tagged as a patch -- and in fact the reason it needed a changelist at all -- is because of that whitespace change in clipboard.ts.

Generally I would not expect package updates that only affect the dev experience to appear in a changelist. The only reason this one did is because a needed automatic change wasn't checked in by previous developers? Or maybe the rules about how to auto-format changed?

Ideally I would have put the change into a completely separate changelist but even I felt that was too much overhead for just whitespace.

jadahiya-MSFT: Would branding this as `"type": "none",` not work? And then adding into the comment why we believe it was branded as `none`. Since `none` gets filtered out when making the overall CHANGELOG during releases.

AE-MS: If that's the recommendation I'm happy to do that. The definition provided by beachball indicates that "none" meant that there was no change to the published packaged, which is not correct for this as even a whitespace change in the published packaged (the non-minimized version) is still a change.

jadahiya-MSFT: I moreso just wanted to bring this up to start a conversation on it. For now, I am happy to keep this as a patch and then do something like add our own guidelines for changefiles as an override to beachball's recommendations.

AE-MS: Per in-person discussion with @TrevorJoelHarris I left it at patch but updated the change description to only focus on the part of the change that affects the published packaged (despite that being the least interesting part of the change 😅)

jadahiya-MSFT: I am happy with that outcome, thanks for making the change

TrevorJoelHarris: I think we should list this as patch because _technically_ it is possible that we have changed the part of the code that consumers use (the clipboard capability). In practice, it's extremely unlikely that this change will affect any customers (especially after minification), but it's always possible that something that looks like an innocuous whitespace change could introduce unexpected behavior. (changing whitespace in a string for example).

The changefile description should always be targeted at what consumers of the library need to know about this change. So for this, something brief like 'Changed whitespace in `clipboard` capability' would be okay with me

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

PR #2375, opened 2024-06-24
Remove packages that don't appear to be used

AE-MS:
## Description

These packages depend on deprecated versions of other packages and none of these appear to be referenced throughout the codebase.


shrshindeMSFT:
Do we know why these packages were added in the first place and where they were used? Also, if you can tell me how did you find unused packages?

sunbrk (approved):

AE-MS:
> Do we know why these packages were added in the first place and where they were used? Also, if you can tell me how did you find unused packages?

Okay, it took me a very long time to track this down across 30 merges/renames and two different repos but I found that `fs-jetpack` was first added with this commit:
https://github.com/OfficeDev/microsoft-teams-library-js/commit/12672d0c75227c69ec61bb8cb16f19324ef46699
...on September 16, 2019. It was used as part of `publish.js` which was a script that the Teams team added at the time to first start publishing their package.

Eventually, we updated how we did publishing as part of the 2.0 release to be more rigorous, official, and locked down. This completely changed the publishing process and moved far away from using that `publish.js` file. The file was eventually deleted on Feb 4, 2022 as part of #907. I can't say for sure which commit first introduced the new way of publishing and essentially deprecated `publish.js`, but it was surely a commit shortly before where `publish.js` was deleted.

It is, of course, possible that some other reference to `fs-jetpack` was added in the meantime, however, and this gets to your question about how I knew they were unused: when I was looking to clean up packages we referenced that were deprecated, one of them was downstream of `fs-jetpack` and rather than just update `fs-jetpack`, I did a text search across the codebase and found that there were no references to it. Moreover, after removing it, everything cleaned, installed, built, and tested just fine.

Relatedly, it's very unlikely that removing either of these packages would break the product code itself, as they are both used for file system manipulation which is not something that teamsjs would be capable of doing.

I'll work on tracking down the `del` package next and respond in a new comment.

AE-MS:
> Do we know why these packages were added in the first place and where they were used? Also, if you can tell me how did you find unused packages?

The `del` package was introduced on October 26, 2016 with [this commit](https://github.com/OfficeDev/microsoft-teams-library-js/commit/20897b7946233c8b06def6d18f01736008e76dcd#diff-25789e3ba4c2adf4a68996260eb693a441b4a834c38b76167a120f0b51b969f7). It was introduced then because it was used by `gulpfile.js`, which was back when the teamsjs library used gulp for its build processes.

It was consequently removed in PR #147 in October of 2018 when the build system was converted over to using webpack. `gulpfile.js` was deleted at that point and as far as I can tell nothing has used the `del` package since then. Similar to the above comment, I was unable to find any references by doing a full text search across the repo.

erinha (approved):
🚀

PR #2385, opened 2024-07-02
Trharris/associated apps

TrevorJoelHarris:

## Description

Proposal for adding a new capability for 1P apps to add associated app tabs to themselves.

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/packages/teams-js/src/private/associatedApps.ts
+++ b/packages/teams-js/src/private/associatedApps.ts
@@ -1,6 +1,6 @@
 import { sendMessageToParentAsync } from '../internal/communication';
 import { ensureInitialized } from '../internal/internalAPIs';
-import { ErrorCode, TabInstance } from '../public';
+import { ErrorCode, SdkError, TabInstance } from '../public';
 import { runtime } from '../public/runtime';
 
 // Open questions
@@ -38,10 +38,10 @@ export namespace associatedApps {
    * TODO: Brief description of what this capability does
    */
   export namespace tab {
-    interface ErrorResponse {
-      errorCode: ErrorCode;
-      message?: string; // TODO: Can remove if you don't have a message to send back to the app developer
-    }
+    // interface ErrorResponse {
+    //   errorCode: ErrorCode;
+    //   message?: string; // TODO: Can remove if you don't have a message to send back to the app developer
+    // }
 
     /**
      * @hidden
@@ -68,14 +68,14 @@ export namespace associatedApps {
 
       validateThreadId(threadId);
 
-      return sendMessageToParentAsync<[boolean, TabInstance | ErrorResponse]>(
+      return sendMessageToParentAsync<[boolean, TabInstance | SdkError]>(
         'apiVersionTag', // TODO: see uses of getApiVersionTag in other files to do this correctly
         'associatedApps.tab.addAndConfigureApp',
         [threadId, appTypes],
-      ).then(([wasSuccessful, response]: [boolean, TabInstance | ErrorResponse]) => {
+      ).then(([wasSuccessful, response]: [boolean, TabInstance | SdkError]) => {
         if (!wasSuccessful) {
           // TODO: Can handle error codes differently here, for example if you don't want "user cancelled" to throw
-          const error = response as ErrorResponse;
+          const error = response as SdkError;
           throw new Error(`Error code: ${error.errorCode}, message: ${error.message ?? 'None'}`);
         }
         return response as TabInstance;
@@ -108,14 +108,14 @@ export namespace associatedApps {
       validateTab(tab);
       validateThreadId(threadId);
 
-      return sendMessageToParentAsync<[boolean, TabInstance | ErrorResponse]>(
+      return sendMessageToParentAsync<[boolean, TabInstance | SdkError]>(
         'apiVersionTag', // TODO: see uses of getApiVersionTag in other files to do this correctly
         'associatedApps.tab.reconfigure',
         [tab, threadId],
-      ).then(([wasSuccessful, response]: [boolean, TabInstance | ErrorResponse]) => {
+      ).then(([wasSuccessful, response]: [boolean, TabInstance | SdkError]) => {
         if (!wasSuccessful) {
           // TODO: Can handle error codes differently here, for example if you don't want "user cancelled" to throw
-          const error = response as ErrorResponse;
+          const error = response as SdkError;
           throw new Error(`Error code: ${error.errorCode}, message: ${error.message ?? 'None'}`);
         }
         return response as TabInstance;
@@ -133,7 +133,7 @@ export namespace associatedApps {
      * @param tab fill in details
      * @param threadId Info about where this comes from, links to external docs if available, etc.
      *
-     * @returns The TabInstance of the newly renamed app
+     * @returns The TabInstance of the newly renamed app tab
      *
      * @throws TODO: Description of errors that can be thrown from this function
      */
@@ -147,14 +147,14 @@ export namespace associatedApps {
       validateTab(tab);
       validateThreadId(threadId);
 
-      return sendMessageToParentAsync<[boolean, TabInstance | ErrorResponse]>(
+      return sendMessageToParentAsync<[boolean, TabInstance | SdkError]>(
         'apiVersionTag', // TODO: see uses of getApiVersionTag in other files to do this correctly
         'associatedApps.tab.rename',
         [tab, threadId],
-      ).then(([wasSuccessful, response]: [boolean, TabInstance | ErrorResponse]) => {
+      ).then(([wasSuccessful, response]: [boolean, TabInstance | SdkError]) => {
         if (!wasSuccessful) {
           // TODO: Can handle error codes differently here, for example if you don't want "user cancelled" to throw
-          const error = response as ErrorResponse;
+          const error = response as SdkError;
           throw new Error(`Error code: ${error.errorCode}, message: ${error.message ?? 'None'}`);
         }
         return response as TabInstance;
@@ -184,14 +184,14 @@ export namespace associatedApps {
       validateTab(tab);
       validateThreadId(threadId);
 
-      return sendMessageToParentAsync<[boolean, TabInstance | ErrorResponse]>(
+      return sendMessageToParentAsync<[boolean, TabInstance | SdkError]>(
         'apiVersionTag', // TODO: see uses of getApiVersionTag in other files to do this correctly
         'associatedApps.tab.remove',
         [tab, threadId],
-      ).then(([wasSuccessful, response]: [boolean, TabInstance | ErrorResponse]) => {
+      ).then(([wasSuccessful, response]: [boolean, SdkError]) => {
         if (!wasSuccessful) {
           // TODO: Can handle error codes differently here, for example if you don't want "user cancelled" to throw
-          const error = response as ErrorResponse;
+          const error = response as SdkError;
           throw new Error(`Error code: ${error.errorCode}, message: ${error.message ?? 'None'}`);
         }
       });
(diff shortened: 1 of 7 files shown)


packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * @hidden
+   * @internal
+   * @beta
+   * Limited to Microsoft-internal use
+   *
+   * TODO: Brief description of what this capability does
+   */
+  export namespace tab {
+    interface ErrorResponse {
```
erinha, line 41: Is there a reason you're not reusing SdkError?

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+     * @internal
+     * @beta
+     * Limited to Microsoft-internal use
+     *
+     * TODO: Add full description of what this function does, ie "Renames the tab associated with an app"
+     *
+     * @param tab fill in details
+     * @param threadId Info about where this comes from, links to external docs if available, etc.
+     *
+     * @returns The TabInstance of the newly renamed app
```
erinha, line 136: Is this renaming a tab or an app? Maybe change this to "The TabInstance of the newly renamed tab" to keep it consistent with the TODO on line 131? (I do recognize that this is largely placeholder text.)

packages/teams-js/src/private/associatedApps.ts
TrevorJoelHarris: Nah, we can use that instead as long as we don't throw it directly, I can update that

packages/teams-js/src/private/associatedApps.ts
erinha: If this is supposed to rename the app, then maybe this function should be renamed to `renameApp`, to be consistent with `addAndConfigureApp`?

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+      }
+
+      validateTab(tab);
+      validateThreadId(threadId);
+
+      return sendMessageToParentAsync<[boolean, TabInstance | ErrorResponse]>(
+        'apiVersionTag', // TODO: see uses of getApiVersionTag in other files to do this correctly
+        'associatedApps.tab.remove',
+        [tab, threadId],
+      ).then(([wasSuccessful, response]: [boolean, TabInstance | ErrorResponse]) => {
```
erinha, line 191: Since this function is returning void, I don't think you expect to receive a `TabInstance` in the success case, right?

packages/teams-js/src/private/associatedApps.ts
TrevorJoelHarris: It's renaming an app...tab? I think? This is just placeholder, I'll make sure to ask them to verify that the info is all correct

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
+import { sendMessageToParentAsync } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { ErrorCode, SdkError, TabInstance } from '../public';
+import { runtime } from '../public/runtime';
+
+// Open questions
+// 1. I've re-used `TabInstance` from the public API, does that contain all of the information you and app developers might need?
```
ht4963-ms, line 7: That's great! We should double check the type since it's been a long time since this was created.

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+// 2. I didn't see any reason to add a `getTabs` function because `pages.tabs.getTabInstances`. Any reason that won't work for you?
```
ht4963-ms, line 8: Does it make sense to deprecate pages.tabs.getTabInstances and use getTabs here instead?

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+// 3. I've added an `AppTypes[]` param to `addAndConfigureApp` to allow for the host to show different app types to the user. Very open to changes.
```
ht4963-ms, line 9: We should probably stick to an enum of sorts. Teams does have the concept of categories when it comes apps. We can look into it and try to map them to this.

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+     * runs through app configuration if required, and then associates the app with the threadId provided. If external docs exist, link to them here"
+     *
+     * @param threadId Info about where this comes from, links to external docs if available, etc.
+     * @param appTypes what type of applications to show the user
+     *
+     * @returns The TabInstance of the newly associated app
+     *
+     * @throws TODO: Description of errors that can be thrown from this function
+     */
+    export function addAndConfigureApp(threadId: string, appTypes: AppTypes[]): Promise<TabInstance> {
```
ht4963-ms, line 62: To be consistent with the other methods, can we stick with just "addAndConfigure"?

packages/teams-js/src/private/associatedApps.ts
TrevorJoelHarris: Probably not because pages.tabs.getTabInstances is public and the the new functions are all private / 1P only

packages/teams-js/src/private/associatedApps.ts
TrevorJoelHarris: Yup, please take a look! It looks like it could work, and since it's whatis returned by `pages.tabs.getTabInstances` there's some nice synergy.

packages/teams-js/src/private/associatedApps.ts
TrevorJoelHarris: That would be great, it's also fine to not put every possible category into this enum to start with. I thought this might be a nice way to make it easy to expand to future scenarios without requiring a bunch of annoying breaking changes on both ends.

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * @hidden
+ * @internal
+ * @beta
+ * Limited to Microsoft-internal use
+ *
+ * TODO: Brief description of what this capability does
+ */
+export namespace associatedApps {
```
debajyoti251088, line 27: I am still a bit iffy with the 'associatedApps' name. I know that we dont want to use extensibilityApps since 'extensibility' is a Teams concept.

But 'platformApps' might be better. Since 'platform' should be generic enough for all hubs where developers want to build their apps on the platform

packages/teams-js/src/private/associatedApps.ts
debajyoti251088: Looked at the object thats returne" [](https://github.com/OfficeDev/microsoft-teams-library-js/blob/931a483ea20f116f6d6f7ca407cea3c03192e1fb/packages/teams-js/src/public/interfaces.ts#L17). It does contain what we need

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
...
+     * runs through app configuration if required, and then associates the app with the threadId provided. If external docs exist, link to them here"
+     *
+     * @param threadId Info about where this comes from, links to external docs if available, etc.
+     * @param appTypes what type of applications to show the user
+     *
+     * @returns The TabInstance of the newly associated app
+     *
+     * @throws TODO: Description of errors that can be thrown from this function
+     */
+    export function addAndConfigure(threadId: string, appTypes: AppTypes[]): Promise<TabInstance> {
```
debajyoti251088, line 62: Should we make it as addRemoveOrConfigure?

The current name covers adding, updating or renaming a tab. But deleting a tab is not really a part of configure. Thoughts?

packages/teams-js/src/private/associatedApps.ts
```diff
@@ -0,0 +1,237 @@
+import { sendMessageToParentAsync } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { ErrorCode, SdkError, TabInstance } from '../public';
+import { runtime } from '../public/runtime';
+
+// Open questions
+// 1. I've re-used `TabInstance` from the public API, does that contain all of the information you and app developers might need?
+// 2. I didn't see any reason to add a `getTabs` function because `pages.tabs.getTabInstances`. Any reason that won't work for you?
+// 3. I've added an `AppTypes[]` param to `addAndConfigureApp` to allow for the host to show different app types to the user. Very open to changes.
+// 4. I've added empty, private `validate` functions for the threadId and TabInstance. Any validation that is possible will help prevent against
```
debajyoti251088, line 10: This might be difficult to do on the TeamsJS layer. On the TMP side we should obviously do it.

TrevorJoelHarris: Yup, I'm not making it a hard requirement but something to please keep in mind: not having validation in TeamsJS for input as the source of a *large* amount of bugs and pain as we tried to expand this to other hosts. In many cases, the TeamsJS documentation and validation was missing because it was baked into the Teams codebase, which made it very difficult in practice for other hosts that wanted to support functions to figure out what they needed to do, what valid input was, etc.

Validating it in the TMP layer is better than nothing obviously, but validation in TeamsJS can help us protect against different definitions of "legal input" on other hosts, which can be a real challenge for making cross platform applications.

Like I said: it's possible there's no way to do validation in TeamsJS (although at the very least I'd like you to check for things like empty string), but the more validation you can do in this layer (or in the host sdk) the better!

packages/teams-js/src/private/associatedApps.ts
TrevorJoelHarris: I am also not wild about the associatedApps name, and happy to get more opinions. I'm also not wild about the "platformApps" name though, because I don't know what "platform apps" are. Aren't Teams apps platform apps? What makes a platform app different than a non-platform app?

packages/teams-js/src/private/associatedApps.ts
richa008: The object that is returned has a lot more fields compared to this. This is missing order, removeUrl, appId. I think we would need to create a new object.

packages/teams-js/src/private/hostEntity.ts
```diff
@@ -0,0 +1,234 @@
...
+     * @param hostEntityId Info about where this value comes from, links to external docs if available, etc. For example:
+     * The id of the host entity that your app wants to associate another app with. In Teams this would be the threadId <link to docs and more explanation>
+     *
+     * @param appTypes what type of applications to show the user
+     *
+     * @returns The TabInstance of the newly associated app
+     *
+     * @throws TODO: Description of errors that can be thrown from this function
+     */
+    export function addAndConfigure(hostEntityId: string, appTypes: AppTypes[]): Promise<TabInstance> {
```
richa008, line 59: Why call it hostEntityId and not threadId here?

packages/teams-js/src/private/hostEntity.ts
```diff
@@ -0,0 +1,234 @@
+import { sendMessageToParentAsync } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { ErrorCode, SdkError, TabInstance } from '../public';
+import { runtime } from '../public/runtime';
+
+// Open questions
+// 1. According to Debo, `TabInstance` from the public API looks like it would work. Helen asked Debo to follow up about getting more recent fields added.
+// 2. I didn't see any reason to add a `getTabs` function because `pages.tabs.getTabInstances`. Any reason that won't work for you?
```
richa008, line 8: getTabInstances may not work for us, I don't think it has updated to return configurable+static tabs.

TrevorJoelHarris:
No longer needed

PR #2402, opened 2024-07-11
Enable errors to be used be callers of getUser

AE-MS:
## Description

In the case of an error returned from `authentication.getUser`, the error message was not being properly unwrapped and so callers couldn't really make heads or tails of it. This change ensures the `SdkError` returned by the host is properly transformed into an `Error` so it can be interpreted by the caller. This was in response to a user report.

Prior to this change, you'd have to know how to dig into multiply nested objects to get the result. Now it's just a standard `Error` with `message`.

I added an E2E test for this (and validated that it ran and passed in these pipelines):
```
      ✓ - Target Web Host SDK v-latest - getUser API Call - Failure (Not full trust)  - iOS Excluded  - >2.25.0
```

I updated the unit tests to validate this behavior. There were already unit tests, but they were validating behavior that never happened in reality because they depended on message responses not having `SdkError` objects, which is not how the host SDK operated.

### DIFF
--- a/packages/teams-js/src/internal/videoEffectsUtils.ts
+++ b/packages/teams-js/src/internal/videoEffectsUtils.ts
@@ -219,7 +219,10 @@ class OneTextureHeader {
   private readonly ONE_TEXTURE_INPUT_ID = 0x6f746931;
   private readonly INVALID_HEADER_ERROR = 'Invalid video frame header';
   private readonly UNSUPPORTED_LAYOUT_ERROR = 'Unsupported texture layout';
-  public constructor(private readonly headerBuffer: ArrayBuffer, private readonly notifyError: (string) => void) {
+  public constructor(
+    private readonly headerBuffer: ArrayBuffer,
+    private readonly notifyError: (string) => void,
+  ) {
     this.headerDataView = new Uint32Array(headerBuffer);
     // headerDataView will contain the following data:
     // 0: oneTextureLayoutId
@@ -322,9 +325,8 @@ class TransformerWithMetadata {
     const timestamp = originalFrame.timestamp;
     if (timestamp !== null) {
       try {
-        const { videoFrame, metadata: { audioInferenceResult } = {} } = await this.extractVideoFrameAndMetadata(
-          originalFrame,
-        );
+        const { videoFrame, metadata: { audioInferenceResult } = {} } =
+          await this.extractVideoFrameAndMetadata(originalFrame);
         const frameProcessedByApp = await this.videoFrameHandler({ videoFrame, audioInferenceResult });
         // the current typescript version(4.6.4) dosn't support webcodecs API fully, we have to do type conversion here.
         const processedFrame = new VideoFrame(frameProcessedByApp as unknown as CanvasImageSource, {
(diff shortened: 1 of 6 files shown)


packages/teams-js/src/internal/videoEffectsUtils.ts
```diff
@@ -219,7 +219,10 @@ class OneTextureHeader {
   private readonly ONE_TEXTURE_INPUT_ID = 0x6f746931;
   private readonly INVALID_HEADER_ERROR = 'Invalid video frame header';
   private readonly UNSUPPORTED_LAYOUT_ERROR = 'Unsupported texture layout';
-  public constructor(private readonly headerBuffer: ArrayBuffer, private readonly notifyError: (string) => void) {
+  public constructor(
+    private readonly headerBuffer: ArrayBuffer,
+    private readonly notifyError: (string) => void,
+  ) {
```
AE-MS, line 225: All the changes in this file were automatically made by prettier. They aren't technically related to my change but since they are innocuous I left them in (and they would keep coming back in future changes until they were made, anyway)

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -1,6 +1,6 @@
 import * as communication from '../../src/internal/communication';
-import * as handlers from '../../src/internal/handlers';
 import { GlobalVars } from '../../src/internal/globalVars';
```
AE-MS, line 2: Again, automatically made by prettier. They aren't technically related to my change but since they are innocuous I left them in (and they would keep coming back in future changes until they were made, anyway)

packages/teams-js/test/public/authentication.spec.ts
```diff
@@ -1,10 +1,9 @@
-// import * as communication from '../../src/internal/communication';
```
AE-MS, line 1: This was just hanging out here.

packages/teams-js/src/public/authentication.ts
```diff
@@ -296,23 +304,24 @@ export namespace authentication {
...
-      .catch((err: Error) => {
+      .catch((err: SdkError) => {
+        const errorMessage = `Error returned, code = ${err.errorCode}, message = ${err.message}`;
         if (userRequest && userRequest.failureCallback) {
-          userRequest.failureCallback(err.message);
+          userRequest.failureCallback(errorMessage);
           return null;
         }
-        throw err;
+        throw new Error(errorMessage);
```
TrevorJoelHarris, line 313: Doesn't this mean we will lose the callstack associated with the original Error object?

TrevorJoelHarris: Oh, I see the original callstack is not actually very helpful because it is basically just from here with no extra data, carry on

TrevorJoelHarris (approved):

packages/teams-js/src/public/authentication.ts
AE-MS: I don't think so because `SdkError` objects don't derive from `Error` and don't have a callstack property or data with them:
```
export interface SdkError {
  /**
  error code
  */
  errorCode: ErrorCode;
  /**
  Optional description for the error. This may contain useful information for web-app developers.
  This string will not be localized and is not for end-user consumption. 
  App should not depend on the string content. The exact value may change. This is only for debugging purposes.
  */
  message?: string;
}
```

I think it would be like if we caught a `string` and then threw it as a new error -- we wouldn't be losing any callstack info there because a `string` doesn't carry any such info.

Does that seem reasonable?

aeunms (approved):

PR #2412, opened 2024-07-17
Add copilot license check

ndangudubiyyam:

## Description

This PR is about how M365Chat application can have access to the information if the user has copilot license or not via runtime info

### Main changes in the PR:

1. Added new API to check if the current user of copilot app has supported m365Chat license or not.
2. Added E2E tests

### Validation performed:

1. Successfully ran the API in Orange

### Unit Tests added:
Yes

### End-to-end tests added:

Yes


packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,22 @@
+import { ensureInitialized } from '../internal/internalAPIs';
+import { runtime } from '../public/runtime';
+
+/**
+ * @beta
```
lakhveerk, line 5: should it also be hidden and internal like other External* apis?

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,22 @@
...
+ * User information required by specific apps
+ */
+export namespace copilot {
+  export namespace license {
```
lakhveerk, line 9: We mark all the tags on all levels: namespace, subnamespace, on top of every funciton etc

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1078,9 +1078,15 @@ export enum DevicePermission {
   Media = 'media',
 }
 
+/** @hidden */
+export interface M365ChatLicenseInformation {
+  hasM365ChatLicense: boolean;
```
erinha, line 1083: I think I left this comment in the doc, but I think having this return an enum offers more flexibility, rather than a boolean.

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * This function is called by M365Chat app.
+     * @returns true if a user had M365Chat license and false otherwise
+     *
+     * @throws Error if {@linkcode app.initialize} has not successfully completed
+     *
+     * @beta
+     */
+    export function isSupported(): boolean {
```
erinha, line 18: If the m365ChatLicenseInfo were changed to return an enum, then I suggest adding a separate function in the `copilot.license` subcapability to return the value of that enum, in addition to `isSupported`.

erinha (changes requested):
Left feedback on what the license API should return.

packages/teams-js/src/public/interfaces.ts
ndangudubiyyam: Oh I thought we will improve upon this design once we have a conformation from the Harmony team. I will follow up with them.

apps/teams-test-app/e2e-test-data/copilot.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "copilot",
+  "platforms": "*",
+  "version": ">=2.22.0",
+  "testCases": [
+    {
+      "title": "copilot hasM365Chat license API Call - Success",
+      "type": "callResponse",
+      "boxSelector": "#box_CheckIfUserOfTheM365ChatAppHasLicense",
+      "expectedTestAppValue": "User does have a license for the M365 Chat app"
+    }
```
erinha, line 11: We should have a test that returns false as well, right?

apps/teams-test-app/e2e-test-data/copilot.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "copilot",
+  "platforms": "*",
+  "version": ">=2.22.0",
```
erinha, line 4: I think it's now 2.25.0, but please confirm.

apps/teams-test-app/e2e-test-data/copilot.json
ndangudubiyyam: We usually don't have failure E2E tests. Also, how can we have both failure and success run at the same time?

packages/teams-js/src/public/interfaces.ts
erinha: IMO we should not have two fields. We should only have the `m365ChatLicenseType` field and app developers should call `copilot.license.getM365ChatLicenseType()` and then they can use that value to take specific actions in their app. 

If we think it's important to have a `copilot.license.hasM365ChatLicense()` function, then we can implement it as something like 
`return runtime.hostVersionsInfo?.m365ChatLicenseInfo?.m365ChatLicenseType !== M365ChatLicenseType.None;`

But I would suspect that the type of license would matter to app developers.

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,52 @@
...
+     * @internal
+     * Limited to Microsoft-internal use
+     * @beta
+     * This function is called by M365Chat app.
+     * @returns the type of the M365Chat license the user associated with the app has.
+     *
+     * @throws Error if {@linkcode app.initialize} has not successfully completed
+     */
+    export function getM365ChatLicenseType(): M365ChatLicenseType {
+      if (ensureInitialized(runtime)) {
```
erinha, line 46: I think this function should call `isSupported()` and if it returns false, it should throw. That matches other API behavior.

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,52 @@
...
+     * @internal
+     * Limited to Microsoft-internal use
+     * @beta
+     * This function is called by M365Chat app.
+     * @returns true if a user had M365Chat license and false otherwise
+     *
+     * @throws Error if {@linkcode app.initialize} has not successfully completed
+     */
+    export function isSupported(): boolean {
+      return (ensureInitialized(runtime) && runtime.hostVersionsInfo?.m365ChatLicenseInfo?.hasM365ChatLicense) ?? false;
```
erinha, line 32: I think this should actually just be:
```suggestion
      return (ensureInitialized(runtime) && runtime.hostVersionsInfo?.m365ChatLicenseInfo) ?? false;
```

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1078,9 +1078,25 @@ export enum DevicePermission {
...
+  m365ChatLicenseType: M365ChatLicenseType;
+  // other fields may be added later
+}
+
+/** @hidden */
+export enum M365ChatLicenseType {
+  WebOnly = 'webOnly',
```
erinha, line 1090: Nit: alphabetize

erinha (changes requested):
Left comments requesting changes to the `M365ChatLicenseInformation` interface.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1078,9 +1078,88 @@ export enum DevicePermission {
...
 
+/**
+ * @hidden
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * @beta
+ */
+export interface AppEligibilityInformation {
```
lakhveerk, line 1089: should these interfaces be moved to copilot.ts file? since this is public/interfaces.ts directory?

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1086,7 +1086,75 @@ export enum DevicePermission {
...
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * @beta
+ */
+export enum UserCohort {
+  None = 'none',
+  M365Copilot = 'm365Copilot',
+  M365CopilotPremium = 'm365CopilotPremium',
+  M365CopilotBasic = 'm365CopilotBasic',
```
erinha: FYI, there's no such thing as a basic license. We'll need to update this.

erinha: Oh, I missed it; you already did! Thanks!

PR #2369, opened 2024-06-21
Avavellozzi/diagnosticAppPoc

ava-vellozzi:
**This PR is fixing issues in my previous PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/2359.
The previous PR was closed due to improper states, so this new PR will address the necessary changes.**

## Description

This pull request is still part of my Proof of Concept (POC) and addresses comments on my first PR linked above.

### Main changes in the PR:

1. Changed .gitignore file back into multiple files
2. Deleted unnecessary comments and big blocks of comments
3. Fixed the "name" in my package.json so it's more readible

### Validation performed:

1. Ran All Unit Tests: Verified that the unit test passes and that the implemented features work as expected.
2. Manual Testing: Manually tested the app to ensure the UI elements are functional and the APIs are working correctly.

### Unit Tests added:

Yes: One unit test was added to validate the initial implementation of the app and authentication APIs (App Initialization Scenario). More will be added later during the implementation of the fully functioning app

### End-to-end tests added:

No: End-to-end tests are currently out of scope for my project.

### Change file added:

Yes

### Related PRs:

https://github.com/OfficeDev/microsoft-teams-library-js/pull/2359

### Next/remaining steps:

- Onboard POC onto Teams Store
- Start full implemention
- Integrate all APIs outlined in Dev Plan (Note: currently the api files have a lot of commented out sections for later use!)
- Implement Hello World Scenario
- Implement Custom Scenario
- Implement new Pre-Set Scenario
- Implement P2 features - log search, new console, settings, download

PR #2415, opened 2024-07-18
Avavellozzi/diagnosticDragDrop

ava-vellozzi:
This is a feature part of my MVP. This is adding onto my previously merged PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/2369

## Description

> The drag and drop functionality of my app is implemented here. 

### Main changes in the PR:

1 Drag and Drop Functionality:

- APIs in API Box: APIs available in the API box can now be dragged and dropped into the custom scenario box.
- Drag and Drop Interaction: Users can select APIs from the API box and drag them to the custom scenario box to build their custom scenarios.

2. Dropdown Menu for Function Selection:

- Function Selection: A dropdown menu is provided to choose specific functions for each API.
- Custom Input: Users can add their own input for each function or use the default input if available.
- Function/Input Display: The selected function and input from the dropdown menu are displayed in the custom scenario box upon dropping.

### Validation performed:

1. Ran All Unit Tests: Verified that the unit test passes and that the implemented features work as expected.
2. Manual Testing: Manually tested the app to ensure the UI elements are functional and the APIs are working correctly.

### Unit Tests added:

Currently, there is just a unit test for the App Initialization scenario. More will be added on later.

No

### End-to-end tests added:

No

### Change file added:
No

### Related PRs:

(https://github.com/OfficeDev/microsoft-teams-library-js/pull/2369)

### Next/remaining steps:

- Multiple apis able to be dragged into custom scenario box
- More apis implemented
- Extra logging features

### Screenshots:

dragDrop


apps/teamsjs-diagnostic-app/public/auth-end.html
```diff
@@ -46,12 +46,20 @@
                   })
                 );
               } else {
+<<<<<<< HEAD
```
noahdarveau-MSFT, line 49: Please resolve this merge conflict

noahdarveau-MSFT (changes requested):

apps/teamsjs-diagnostic-app/package-lock.json
noahdarveau-MSFT, line 1: Is there a reason you deleted the pnpm-lock file from the repo?

noahdarveau-MSFT (changes requested):

apps/teamsjs-diagnostic-app/package-lock.json
ava-vellozzi: Sometimes I have to delete the node_modules and package-lock file and do `pnpm i` again. When I do pnpm i it generates that file and other times it doesn't. Is it important to have it?

apps/teamsjs-diagnostic-app/.gitignore
noahdarveau-MSFT, line 1: Do we need this .gitignore file? Isn't it covered by the .gitignore in the repo root?

apps/teamsjs-diagnostic-app/api/.gitignore
noahdarveau-MSFT, line 1: Same as above

apps/teamsjs-diagnostic-app/src/apis/AppInstallDialogApi.tsx
```diff
@@ -1,8 +1,72 @@
...
+  onDropToScenarioBox: (apiComponent: ApiComponent, func: string, input: string) => void;
+}
+
+const AppInstallDialogAPIs: React.FC<AppInstallDialogAPIsProps> = ({ apiComponent, onDropToScenarioBox }) => {
+  const [selectedFunction, setSelectedFunction] = useState<string>('');
+  const [inputValue, setInputValue] = useState<string>('');
+
+  const handleFunctionChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const selectedFunc = event.target.value;
+    setSelectedFunction(selectedFunc);
```
TrevorJoelHarris, line 16: Something to be aware of: each time you call a `set*` function in React (`setSelectedFunction` and `setInputValue` for example), each function call will re-render the react component. That means that your `handleFunctionChange` function will be causing two renders, one where selectedFunc is set to the new value, and a second one where you have selectedFunc and inputvalue set.

That might not be a problem (it depends on how you've put everything together), but it's worth being aware of. If it is causing bugs (or performance problems) you could put your two pieces of data together into a single interface, something like:

```typescript
const [appInstallProps, setAppInstallProps] = useState<AppInstallProps | null>(null);

interface AppInstallProps {
  selectedFunction: string;
  inputValue: string;
}
```

apps/teamsjs-diagnostic-app/src/apis/AppInstallDialogApi.tsx
```diff
@@ -1,8 +1,72 @@
...
+    } else {
+      setInputValue('');
+    }
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  };
+
+  const [{ isDragging }, drag] = useDrag(() => ({
```
TrevorJoelHarris, line 28: This is cool!

apps/teamsjs-diagnostic-app/src/apis/BarCodeApi.tsx
```diff
@@ -1,8 +1,72 @@
...
+    } else {
+      setInputValue('');
+    }
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  };
+
+  const [{ isDragging }, drag] = useDrag(() => ({
```
TrevorJoelHarris, line 28: Not a requirement, but food for thought: the drag and drop function here looks identical to the one in AppInstallDialogApi.tsx except for one tiny section (the selectedFunction === check). If you add drag and drop functionality to a lot of *Api.tsx files by copy pasting this in, and then you find a bug that needs to be fixed you will have to go update each instance of this hook by hand. It might be helpful to find a way to share this hook with all of your `*Api.tsx` files so you don't have to repeat yourself.

apps/teamsjs-diagnostic-app/src/components/sample/ApiComponents.tsx
```diff
@@ -1,23 +1,39 @@
...
+  {
+    title: 'Bar Code API',
+    name: 'barCode',
+    options: ['checkBarCodeCapability', 'scanBarCode', 'hasBarCodePermission', 'requestBarCodePermission'],
+    defaultInput: '{}',
+    inputType: 'text',
+    onClick: () => console.log('Barcode API called'),
+    },
+  {
+    title: 'Calendar APIs',
```
TrevorJoelHarris, line 30: I don't see a CalendarApi.tsx file, is this entry meant to be checked in?

apps/teamsjs-diagnostic-app/src/apis/AppInstallDialogApi.tsx
ava-vellozzi: Right now, it's not causing any bugs, but I will keep an eye out for this! Thank you!

apps/teamsjs-diagnostic-app/src/components/sample/ApiComponents.tsx
ava-vellozzi: Not yet! That was just there as dummy values, but in the next PR, all the APIs will actually be defined

apps/teamsjs-diagnostic-app/src/components/sample/CustomScenario.tsx
```diff
@@ -1,65 +1,94 @@
...
-              {options.map((option, optionIndex) => (
-                <option key={optionIndex} value={option}>
-                  {option}
-                </option>
-              ))}
-            </select>
-          </div>
-        </div>
+        {api.title === 'App Install Dialog API' ? (
+          <AppInstallDialogAPIs apiComponent={api} onDropToScenarioBox={addToScenario} />
```
TrevorJoelHarris, line 49: This is again a piece of feedback you choose to take or leave. It is a little brittle to have to check the api title field's literal string value here. That string is also specified in ApiComponents.tsx, and if you change it in one place you'd have to change it both or your app will break.

One solution you could use to improve this would be to add something like a factory function to the ApiComponent interface. You already have an "onClick" function, you could add another one that was "function that returns the right React Component when called". If you had something liek this, this section would just turn into:

```typescript
api.renderComponent(addToScenario)
```

TrevorJoelHarris (approved):

sunbrk (approved):

noahdarveau-MSFT (approved):

PR #2420, opened 2024-07-22
Update e2e tests with targetted versions for getUser update

alexneyman-MSFT:
## Description
Updated the getUser test to be in line with the expected changes from the updated getUser function which will enable Micrsoft owned apps the same permissions of full-trust apps. 

### Main changes in the PR:

1. Add hostSdkVersion boundaries
2. Add tests that cover the new cases with a second boolean variable

### Validation performed:

1. tests pass locally in the host SDK

### End-to-end tests added:

Yes

### DIFF
--- a/apps/teams-test-app/e2e-test-data/authentication.json
+++ b/apps/teams-test-app/e2e-test-data/authentication.json
@@ -21,11 +21,62 @@
     {
       "title": "getUser API Call - Failure (Not full trust)",
       "version": ">2.25.0",
+      "hostSdkVersion": {
+        "web": "<3.1.0"
+      },
       "type": "callResponse",
       "platformsExcluded": ["iOS"],
       "boxSelector": "#box_getUser",
       "testUrlParams": [["appDefOverrides", "{\"isFullTrustApp\": false}"]],
       "expectedTestAppValue": "Error: Error: Error returned, code = 500, message = App does not have the required permissions for this operation"
+    },
+    {
+      "title": "getUser API Call - Failure (Not full trust and not Microsoft owned)",
+      "version": ">2.25.0",
+      "hostSdkVersion": {
+        "web": ">3.1.0"
+      },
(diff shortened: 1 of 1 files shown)


erinha (dismissed):

apps/teams-test-app/e2e-test-data/authentication.json
```diff
@@ -21,11 +21,50 @@
...
       "platformsExcluded": ["iOS"],
       "boxSelector": "#box_getUser",
       "testUrlParams": [["appDefOverrides", "{\"isFullTrustApp\": false}"]],
       "expectedTestAppValue": "Error: Error: Error returned, code = 500, message = App does not have the required permissions for this operation"
+    },
+    {
+      "title": "getUser API Call - Failure (Not full trust and not Microsoft owned)",
+      "version": ">2.25.0",
+      "hostSdkVersion": {
+        "web": ">=3.1.0"
```
erinha, line 37: This has to be ">=4.0.0" right? (Applies to all places)

AE-MS: I think since the plan is to patch it back into the 3.x branch it's okay to keep it lower?

AE-MS: Oh wait I'm not sure if you were commenting on the version number or the sign in front it it

I think it needs to be `>3.1.0` ? And then the test case above this should be `<=3.1.0`

AE-MS (changes requested):
:clock1:

apps/teams-test-app/e2e-test-data/authentication.json
alexneyman-MSFT: ah yes my bad, I didn't realize 3.1 was already out, will udate

AE-MS (approved):
:shipit:

erinha (approved):

PR #2429, opened 2024-07-26
Add support for *.cloud.microsoft domains

AE-MS:
## Description

Practically all Microsoft cloud properties are moving the `.cloud.microsoft` domain. This domain is reserved for properties that don't allow 3rd party code to run on it. We already tell developers to allow `*.cloud.microsoft` in their CSPs, this change reflects that the same allowance should be made in the hosts that teamsjs trusts.

### Main changes in the PR:

1. Add *.cloud.microsoft to the allowlist, removing any more specific domains which would be covered by it
2. Update the unit tests to ensure that communication still works from any cloud.microsoft domain

### Validation performed:

1. Unit tests pass
2. Ran a test app using a local build of TJS with these changes and validated that it successfully loads in teams.cloud.microsoft

### Unit Tests added:

No unit tests were added but the existing validation was extended to test new domains in the cloud.microsoft space

### Change file added:

Yes


packages/teams-js/test/private/privateAPIs.spec.ts
```diff
@@ -117,6 +117,10 @@ describe('AppSDK-privateAPIs', () => {
     'https://www.microsoft365.com',
     'https://tasks.office.com',
     'https://www.example.com',
+    'https://teams.cloud.microsoft',
+    'https://outlook.cloud.microsoft',
+    'https://m365.cloud.microsoft',
+    'https://anythingbecauseitsawildcard.cloud.microsoft',
```
jadahiya-MSFT, line 123: I like this, btw this is a duplicate of a previous PR. I will close that one though.

https://github.com/OfficeDev/microsoft-teams-library-js/pull/2355

packages/teams-js/src/artifactsForCDN/validDomains.json
```diff
@@ -29,9 +29,7 @@
     "edgeservices.bing.com",
     "www.bing.com",
     "www.staging-bing-int.com",
-    "teams.cloud.microsoft",
```
jadahiya-MSFT, line 32: Can we make a note somewhere of the specific cloud domains that we want to allowlist as well. Just in case we need to back out this wildcard at some point. It would be useful to document the domains we want too.

AE-MS: Sure, where would you like me to put that note? JSON doesn't support comments. Also note that there are some cloud.microsoft domains that we need to support now but which I can't put in a note in this repo because their existence is not public knowledge.

TrevorJoelHarris: Maybe our internal wiki?

TrevorJoelHarris (approved):

packages/teams-js/src/artifactsForCDN/validDomains.json
jadahiya-MSFT: I was thinking we could document all wildcards in the Wiki, but I'm open to other suggestions. I am also ok that it is the unit test.

jadahiya-MSFT (approved):

PR #2432, opened 2024-07-26
Split the web E2E tests by starting letter

AE-MS:
## Description

This change splits the web E2E tests into groups by starting letter to enable faster overall throughput and more granular reruns, when necessary. The YAML was refactored to maximize code-reuse as part of this.

### Main changes in the PR:

1. Refactor the YAML into a new template, `web-e2e-tests-job.yml` which can run multiple web E2E tests types and groups based on the input parameters.
2. Leveraged this new template to run the same tests as before, but split up by starting letter.

Further details are annotated in the PR where appropriate.

### Validation performed:

1. Ensured that all E2E tests ran and passed by examining output from each job to ensure there were no missed letters
2. Temporarily forced a test to fail to ensure it was still reported in the ADO UI and [still failed](https://office.visualstudio.com/ISS/_build/results?buildId=30229559&view=ms.vss-test-web.build-test-results-tab&runId=1585392436&resultId=100079&paneView=debug) the overall pipeline run


azure-pipelines.yml
```diff
@@ -133,72 +133,28 @@ extends:
...
+              teamsJsReferenceType: 'scriptTag'
+              testPrefixPatternGroups: ['{[0-9],[A-D],[a-d]}', '{[E],[e]}', '{[F-M],[f-m]}', '{[N-Z],[n-z]}']
+
+          - template: tools/yaml-templates/web-e2e-tests-job.yml@self
+            parameters:
+              AppHostingSdk: AppHostingSdk
+              hostingEnvironmentType: 'electron'
+              teamsJsReferenceType: 'npm'
 
           - job: E2ETestCDN
```
AE-MS, line 156: >E2ETestCDN [](http://example.com/codeflow?start=17&length=10)

There is still opportunity to leverage the new YAML template for running these tests as well. However, since I don't understand how these work and aren't sure how to validate them, I didn't want to mess with them quite yet and I didn't want to delay getting the faster tests out to developers while I spent time figuring this out. This will be a follow-up item for me.

azure-pipelines.yml
```diff
@@ -133,72 +133,28 @@ extends:
...
-            displayName: 'E2E Test - CDN (only runs on release builds)'
+            displayName: 'E2E Tests - CDN (only runs on release builds)'
```
AE-MS, line 157: >Tests [](http://example.com/codeflow?start=30&length=5)

I'm nothing if not consistent in my naming.

tools/yaml-templates/web-e2e-tests-job.yml
```diff
@@ -0,0 +1,73 @@
...
+    values:
+      - 'electron'
+      - 'standardWeb'
+  - name: teamsJsReferenceType
+    type: string
+    values:
+      - 'npm'
+      - 'scriptTag'
+  - name: testPrefixPatternGroups
+    type: object
```
AE-MS, line 19: >object [](http://example.com/codeflow?start=10&length=6)

YAML doesn't have arrays as types but objects can be treated like you would expect arrays to be treated, as you can observe by looking at how this value is passed in as well as how it's iterated over below.

tools/yaml-templates/web-e2e-tests-job.yml
```diff
@@ -0,0 +1,73 @@
...
+    default: ['{[A-Z],[a-z],[0-9]}']
+
+jobs:
+  - ${{each testPrefixPattern in parameters.testPrefixPatternGroups}}:
+      - job: E2ETestsWeb_${{parameters.teamsJsReferenceType}}_${{ replace(replace(replace(replace(replace(replace(testPrefixPattern, '[', '_'), ']', '_' ), '{', '_'), '}', '_'), ',', '_'), '-', '_') }}
```
AE-MS, line 24: >replace(replace(replace(replace(replace(replace [](http://example.com/codeflow?start=66&length=47)

I know this looks gross but I don't know of any other way to do it:
- Jobs must have names
- Each job must have a unique name, so the job name must include some data from the parameters that identifies what makes it unique
- Job names can only have alphanumerics and underscores
- This replace function doesn't support RegEx or character groups, so I can't provide a pattern like `[^A-za-z] ` etc.
- Consequently, I have to individually replace each expected "illegal" character with the underscore.
- Job names also can't span multiple lines so this is as readable as I can make it

tools/yaml-templates/web-e2e-tests-job.yml
```diff
@@ -0,0 +1,73 @@
...
+            ${{ else }}:
+              value: 'localScriptTag'
+        steps:
+          - template: build-app-host.yml
+            parameters:
+              appHostGitPath: AppHostingSdk
+
+          - task: CmdLine@2
+            displayName: 'Build Local Test App to Reference via Script Tag'
+            condition: and(succeeded(), eq('${{ parameters.teamsJsReferenceType }}', 'scriptTag'))
```
AE-MS, line 43: >eq('${{ parameters.teamsJsReferenceType }}', 'scriptTag')) [](http://example.com/codeflow?start=40&length=58)

This step is only needed when referencing the local library via the script tag

tools/yaml-templates/web-e2e-tests-job.yml
```diff
@@ -0,0 +1,73 @@
...
+              --useDataFromLocal=true
+              --testPrefixPattern "${{testPrefixPattern}}"
+              --envType=${{ variables.envType }}
+            displayName: 'Run web hosted E2E integration tests (${{testPrefixPattern}})'
+            condition: and(succeeded(), eq('${{ parameters.hostingEnvironmentType }}', 'standardWeb'))
+            workingDirectory: '$(AppHostingSdkProjectDirectory)'
+
+          - bash: |
+              # This sets up the server but runs no tests
+              node tools/cli/runAppsWithE2ETests.js --useDataFromLocal=true --testPrefixPattern "{[]}"
```
AE-MS, line 60: The electron tests require the test host to be running before you run them, but there's no direct script to only setup the test host as needed right now, so I have to use the full "run e2e tests" script but provide it no tests to run (note the `{[]}` prefix pattern) which will cause it to setup the test host which can then be used by playwright below.

As a follow-up item, I will look to refactor the `runAppsWithE2ETests.js` in the other repo so that we can have a script that only sets up the test host that can be used here.

tools/yaml-templates/web-e2e-tests-job.yml
```diff
@@ -0,0 +1,73 @@
...
+              # This runs the Electron tests, which currently cannot be filtered
+              pnpm exec xvfb-maybe playwright test
```
AE-MS, line 62: >pnpm exec xvfb-maybe playwright test [](http://example.com/codeflow?start=14&length=36)

It's a little weird that other tests are run using `runAppsWithE2ETests.js` but electron tests are run using playwright directly. Should that implementation detail of the electron tests also be abstracted into the `runAppsWithE2ETests.js` script?

Probably?

In any case I can't do it here or part of this change because that script lives in the other repo.

tools/yaml-templates/web-e2e-tests-job.yml
erinha: Consider putting a comment above the line indicating what this is doing? I think it's basically replacing all illegal characters with underscore to use as a valid job name.

erinha (approved):
Thanks for making this change! It will be great to reduce the overall runtime of the E2E tests.

I left one suggestion about adding a comment but see nothing that requires functional changes.

TrevorJoelHarris (approved):

PR #2437, opened 2024-07-29
Updated packages with CG vulnerabilities

jadahiya-MSFT:

## Description

### Main changes in the PR:

1. Updated `webpack` package and every package inside of `teamsjs-diagnostics-app`
2. Removed unnecessary `package.lock` files
3. Added `@azure/identity` and `nth-check` to overrides


package.json
```diff
@@ -121,7 +121,10 @@
       "underscore": "1.12.1",
       "url-parse": "^1.5.0",
       "word-wrap": "^1.2.4",
-      "y18n": "^4.0.1"
+      "y18n": "^4.0.1",
+      "nth-check@<2.0.1": ">=2.0.1",
+      "@azure/identity@<4.2.1": ">=4.2.1",
+      "fast-xml-parser@<4.4.1": ">=4.4.1"
```
AE-MS, line 127: Can you add explanation for these to the section below, so we know why they are here and when it would be safe to remove them?

AE-MS, line 127: It looks like this list was previously alphabetized.

AE-MS (changes requested):
:clock1:

package.json
jadahiya-MSFT: This was auto generated. I'm currently looking into seeing if I can just upgrade a single package that has these as a dependency to forgo needing the overrides.

AE-MS (approved):
:shipit:

TrevorJoelHarris (approved):

PR #2448, opened 2024-08-06
Added copilot.eligibility capability

lakhveerk:

## Description

> This PR adds a `copilot.eligibility` capability that provides information about whether the user is eligible to use the copilot app. The information is sent by a host in runtime object. 

### Main changes in the PR:

1. Added copilot.ts file + UTs
2. Edited runtime object to contain AppEligibilityInfo in it.
3. Added test app changes

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

UTs are added.

### End-to-end tests added:

<Yes/No> They will be added in the host sdk

### Change file added:

provided


packages/teams-js/src/public/interfaces.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any*/
 
+import { AppEligibilityInformation } from '../private/interfaces';
```
AE-MS, line 3: Is this creating a dependency cycle? Is the public folder allowed to depend on the private folder? Or does the dependency arrow have to go the other way (private depending on public)?

@TrevorJoelHarris do you have an opinion or knowledge here?

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+  /**
+   * Describes the user’s age group, which can have implications on which product they are able to use.
+   */
+  ageGroup: LegalAgeGroupClassification;
+  /**
+   * Implementation may change to be based on tenant-home region rather than IP.
+   */
+  isCopilotEnabledRegion: boolean;
+  /**
+   * Indicates that the user is eligible for some AAD authenticated Copilot experience.
```
AE-MS, line 303: >AAD [](http://example.com/codeflow?start=50&length=3)

Please change this to "Microsoft Entra ID"

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+   */
+  isCopilotEligible: boolean;
+  /**
+   * Indicates if the tenant admin has opted the user out of Copilot.
+   */
+  isOptedOutByAdmin: boolean;
+  /**
+   * For EDU tenants only. Indicates if the tenant is higher ed or K12.
+   */
+  eduType: EduType;
```
AE-MS, line 313: >EduType [](http://example.com/codeflow?start=11&length=7)

Rather than having a `None` member of `EduType`, it would be clearer if this was declared as:
```
eduType: EduType | null;
```
...and setting the property to `null` when there is no EDU tenant. If you have an `EduType` object instance, the implication is that you can only get one in an applicable EDU tenant.

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+/**
+ * @hidden
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * @beta
+ */
+export enum Cohort {
+  None = 'none',
```
AE-MS, line 325: >None [](http://example.com/codeflow?start=2&length=4)

Please see my comment about the `EduType` elsewhere i this file.

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+/**
+ * @hidden
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * @beta
+ */
+export enum Persona {
+  None = 'none',
```
AE-MS, line 340: >None [](http://example.com/codeflow?start=2&length=4)

Please see my comment about `EduType` elsewhere in this file.

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+ * Limited to Microsoft-internal use
+ *
+ * @beta
+ */
+// https://learn.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#legalagegroupclassification-values
+export enum LegalAgeGroupClassification {
+  /**
+   * Default value, no ageGroup is set for the user.
+   */
+  None = 'none',
```
AE-MS, line 364: >None [](http://example.com/codeflow?start=2&length=4)

Please see my comment about `EduType` elsewhere in this file.

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+  /**
+   * The user is from a country or region that has additional age-related regulations, such as the United States,
+   * United Kingdom, European Union, or South Korea, and the user's age is between a minor and an adult age
+   * (as stipulated based on country or region). Generally, this means that teenagers are considered as notAdult in regulated countries.
+   */
+  NonAdult = 'nonAdult',
+  /**
+   * Value is not found.
+   */
+  Unknown = 'unknown',
```
AE-MS, line 391: Are consumers of this code supposed to treat the "unknown" case different than the "None" case?

packages/teams-js/src/private/interfaces.ts
```diff
@@ -271,3 +271,142 @@ export interface UserJoinedTeamsInformation {
...
+export enum EduType {
+  None = 'none',
+  /**
+   * User is from a tenant labeled as “K12”
+   */
+  K12 = 'k12',
+  /**
+   * User is from a tenant labeled as “HigherEd”
+   */
+  Hed = 'higherEducation',
```
AE-MS, line 411: >Hed [](http://example.com/codeflow?start=2&length=3)

Can we use the full HigherEducation name for this?

packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+  });
+
+  describe('eligibility', () => {
+    it('should throw if called before initialization', () => {
+      utils.uninitializeRuntimeConfig();
+      expect(() => copilot.eligibility.isSupported()).toThrowError(new Error(errorLibraryNotInitialized));
+      expect(() => copilot.eligibility.getEligibilityInfo()).toThrowError(new Error(errorLibraryNotInitialized));
+    });
+    it('should return EligibilityInfo if the app is MChat app', async () => {
```
AE-MS, line 56: >MChat [](http://example.com/codeflow?start=52&length=5)

Since this is only a string it doesn't really matter but here you call it "MChat" and in `copilot.ts` you call it "M365 chat app"

Is there a correct name for it?

AE-MS, line 56: >if the app is MChat app [](http://example.com/codeflow?start=38&length=23)

This isn't really testing that it should return this info if it's the MChat app. It's testing that if the host provided eligibility information, the will be returned from `getEligibilityInfo()`. Nothing MChat-specific about it.

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export namespace copilot {
+  /**
+   * @beta
+   * @hidden
+   * User information required by specific apps
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export namespace eligibility {
```
AE-MS, line 21: >eligibility [](http://example.com/codeflow?start=19&length=11)

Out of curiosity, why are we immediately making this a subcapability?

packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      await utils.initializeWithContext(FrameContexts.content);
+      utils.setRuntimeConfig(copilotRuntimeConfig);
+      expect(copilot.eligibility.isSupported()).toBeTruthy();
+      expect(copilot.eligibility.getEligibilityInfo()).toBe(mockedAppEligibilityInformation);
+    });
+    it('should throw if the value is not set by the host or missing ', async () => {
+      await utils.initializeWithContext(FrameContexts.content);
+      const copilotRuntimeConfigWithOutEligibilityInformation: Runtime = {
```
AE-MS, line 64: >copilotRuntimeConfigWithOutEligibilityInformation [](http://example.com/codeflow?start=12&length=49)

To avoid duplication, consider sharing most of the content between `copilotRuntimeConfigWithOutEligibilityInformation` and `copilotRuntimeConfig` like this:

```
const copilotRuntimeConfig = {
  apiVersion: 4,
  hostVersionsInfo: {
    appEligibilityInformation: mockedAppEligibilityInformation,
  },
  supports: {
    pages: {
      appButton: {},
      tabs: {},
      config: {},
      backStack: {},
      fullTrust: {},
    },
    teamsCore: {},
    logs: {},
  },
};

const copilotRuntimeConfigWithoutEligibilityInformation = {
  ...copilotRuntimeConfig,
  hostVersionsInfo: undefined,
};

```

packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+          logs: {},
+        },
+      };
+      utils.setRuntimeConfig(copilotRuntimeConfigWithOutEligibilityInformation);
+      expect(copilot.eligibility.isSupported()).toBeFalsy();
+      try {
+        copilot.eligibility.getEligibilityInfo();
+      } catch (e) {
+        expect(e).toEqual(errorNotSupportedOnPlatform);
+      }
```
AE-MS, line 84: I don't think there is anything functionally wrong with writing it this way but I believe the canonical Jest way of doing this would be something like this:

```
expect(() => copilot.eligibility.getEligibilityInfo()).toThrowError(errorNotSupportedOnPlatform);
```

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/interfaces.ts
TrevorJoelHarris: To the best of my knowledge this doesn't make a dependency cycle. @lakhveerk is pulling in an interface that isn't in a namespace ( `AppEligibilityInformation and she's only pulling in that interface, and that interface doesn't appear to be pulling in `HostVersionsInfo`, which would the thing that made a cycle.

However, I still wouldn't recommend importing a private interface here and adding it to a public interface. In this repo, things in the private folder are generally meant to be things that are *only* shared with 1P apps (for various reasons) and we do things like suppress their documentation or add properties that only 1P apps should have access to. Because @lakhveerk is adding this interface to the public `HostVersionsInfo` interface, your new private interface with suppressed documentation is being exposed to 3P apps who use `HostVersionsInfo`. That may or may not be a problem today, but if someone in the future adds sensitive 1P information to `AppEligibilityInformation` they could be inadvertently making it accessible to apps that shouldn't have access to it.

Based on what I see here, I assume that `AppEligibilityInformation` is actually okay to expose to 3P apps because we're putting it into `HostVersionsInfo`. If that's the case, `AppEligibilityInformation` should just be made a public interface that people know is okay to expose to all app developers.

packages/teams-js/src/private/copilot.ts
lakhveerk: So, we can keep copilot related APIs into one capability. If we have any other APIs for copilot in future that is not related to `getEligibilityInfo` API, we would want a separate module for that. It would avoid the need to move the existing changes to a subcapability at that time (and will avoid the breaking changes).

Also with a subcapability,  the `isSupported` function will be called on `copilot.eligibility.isSupported` rather than `copilot.isSupported`. 

@erinha

packages/teams-js/src/public/interfaces.ts
lakhveerk: @erinha should I just move the `AppEligibilityInformation` interface into public interface file? 

The interface `AppEligibilityInformation` is used at two places:
1. In `runtime.hostVersionsInfo.appEligibilityInformation` which is public, and 
2. In `copilot.eligibility.getEligibilityInformation` which is private

erinha: I see what Trevor is saying. If it's connected to a public interface, I think we have to include it in the src/public/interface.ts file for documentation purposes. 

It might be worth it to revisit Niharika's original design doc to see if the trade-offs of putting this on the `HostVersionsInfo` interface still make this the right decision.

erinha: Actually, I'm not sure about the documentation part of my comment. The `HostVersionsInfo` interface in this file is marked `/** hidden */` -- does that mean that documentation is not published for it?

packages/teams-js/src/private/copilot.ts
erinha: Based on how Copilot related features evolve, it seemed prudent to not put an API in the top-level capability as we currently don't have any functions that are appropriate for this from a naming perspective. By putting it in a sub-capability, we reserve the top-level capability for future use.

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @hidden
+     * @internal
+     * Limited to Microsoft-internal use
+     * @beta
+     * This function is called by M365Chat app
```
erinha, line 27: I would probably remove this. I don't think it's relevant. Same goes for line 41

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,54 @@
...
+     * @returns boolean to represent whether externalAppCommands capability is supported
```
erinha, line 28: >externalAppCommands [](http://example.com/codeflow?start=45&length=19)

copy/paste error?

packages/teams-js/src/private/copilot.ts
```diff
@@ -0,0 +1,54 @@
...
+      return ensureInitialized(runtime) && !!runtime.hostVersionsInfo?.appEligibilityInformation;
+    }
+
+    /**
+     * @hidden
+     * @internal
+     * Limited to Microsoft-internal use
+     * @beta
+     * This function is called by M365Chat app
+     * @returns the M365Chat eligibility information about the user
```
erinha, line 42: >M365Chat [](http://example.com/codeflow?start=20&length=8)

Nit: suggest replacing with "Copilot"

packages/teams-js/test/private/copilot.spec.ts
erinha: Suggest naming this test:
'should return EligibilityInfo if the host provided eligibility information`

erinha (dismissed):
:clock1:

packages/teams-js/src/public/interfaces.ts
TrevorJoelHarris: I talked with @erinha and @lakhveerk . We agreed that this interface should move to the public folder for now.

Ultimately, there's no way to really solve this problem in a way that is going to make me feel great because the runtime object that this all hangs off of is in the public folder even though it contains data needed and used by both the public and private folders (the `supports` property, for example). The problem here _really_ is that the we don't have the infrastructure to really make the 'private' and 'public' parts of runtime distinct from each other, and I don't think this change is the right time to radically overhaul how that works.

Putting it in the public folder is the best available option right now. It makes sure that the thing we are exporting from public only contains other things defined in the public folder, which will make all of our lives easier if we ever need to change how we deliver those folders to customers or the internal structure of the library.

packages/teams-js/test/private/copilot.spec.ts
lakhveerk: `.toThrowError` works for built in Error object. In our case we are throwing a locally defined SdkError. I am changing it to `expect(copilot.eligibility.getEligibilityInfo()).rejects.toEqual(errorNotSupportedOnPlatform);`. Hope that works
`

packages/teams-js/src/public/interfaces.ts
lakhveerk: Moving the interface to public and closing the comment!

packages/teams-js/test/private/copilot.spec.ts
lakhveerk: `.rejects.toEqual` is for async calls. So that did not work either. I have updated it to 
```
      expect(() => copilot.eligibility.getEligibilityInfo()).toThrowError(
        expect.objectContaining(errorNotSupportedOnPlatform),
      );
```

AE-MS (approved):

jekloudaMSFT (approved):

PR #2450, opened 2024-08-07
Switch all calls of runAppsWithE2ETest.js to the TS version

AE-MS:
## Description

The JS file used to run E2E tests is getting replaced with a TS file. Functionality is currently identical between them, so this change just updates the tests to target the new TS file.

### Validation performed:

Ensured that all E2E tests ran and passed (see the associated pipeline runs)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -122,7 +122,7 @@ extends:
                 condition: succeeded()
                 inputs:
                   targetType: inline
-                  script: 'node tools/cli/runAppsWithE2ETests.js --appUrl=https://localhost:4002 --reportFileName=e2e-tests-report-perf --envType=perf'
+                  script: 'pnpm exec ts-node tools/cli/runAppsWithE2ETests.ts --appUrl=https://localhost:4002 --reportFileName=e2e-tests-report-perf --envType=perf'
                   workingDirectory: '$(AppHostingSdkProjectDirectory)'
 
               - task: PublishTestResults@2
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -122,7 +122,7 @@ extends:
                 condition: succeeded()
                 inputs:
                   targetType: inline
-                  script: 'node tools/cli/runAppsWithE2ETests.js --appUrl=https://localhost:4002 --reportFileName=e2e-tests-report-perf --envType=perf'
+                  script: 'pnpm exec ts-node tools/cli/runAppsWithE2ETests.ts --appUrl=https://localhost:4002 --reportFileName=e2e-tests-report-perf --envType=perf'
```
erinha, line 125: Is this a built-in function of pnpm? Where did ts-node come from?

erinha: Ah I see, it was already present in package.json.

erinha (approved):

KangxuanYe (approved):

azure-pipelines.yml
AE-MS: `pnpm exec` is the pnpm equivalent of `npx`, so that part is built-in to pnpm. `ts-node` is a package used for running TS-based node scripts. Since this script line runs in the other repo where the host package lives, ts-node is something I added to that repo recently for this purpose. I'll share the commit info with you privately.

PR #2454, opened 2024-08-07
Add AppId class to use for storing app ids

AE-MS:
## Description

App Ids are stored and used in various places throughout the code. Today they are frequently stored using strings as that's the most appropriate type to use that exists. That is not ideal as strings will allow many invalid values which are not actually app ids, etc.

This change adds a new `AppId` class that can be used for storing app ids. This will allow future uses of app ids to be type-safe and ensure that all app id data validates appropriately.

Ultimately this change is about all the benefits you get from using the most-appropriate type for the data and using type safe code.

### Main changes in the PR:

1. Add `AppId` class and appropriate validation functions
2. Add unit tests for new code

Note: there are currently no consumers of this class, but as soon as we write new code that needs to store an app id, a ready-made type-safe way will now be available!

### Validation performed:

Ensured that all unit tests for new code passed

### Unit Tests added:

Yes, many.

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/src/internal/appId.ts
```diff
@@ -0,0 +1,30 @@
+import { validateStringAsAppId } from './appIdValidation';
+
+/**
+ * A strongly-typed class used to represent a "valid" app id.
+ *
+ * Valid is a relative term, in this case. Truly valid app ids are UUIDs as documented in the schema:
+ * https://learn.microsoft.com/en-us/microsoftteams/platform/resources/schema/manifest-schema#id
+ * However, there are some older internal/hard-coded apps which violate this schema and use names like
+ * com.microsoft.teamspace.tab.youtube. For compatibility with these legacy apps, we unfortunately cannot
```
TrevorJoelHarris, line 9: Feel free to ignore or otherwise backburner this, but if we had a complete set of what the ids are from telemetry we could probably check for them and then make the AppID validation for *other* appIds a lot more robust

packages/teams-js/src/internal/appIdValidation.ts
```diff
@@ -0,0 +1,45 @@
...
+      `Potential app id (${potentialAppId}) is invalid; its length ${potentialAppId.length} is not within the length limits (${minimumValidAppIdLength}-${maximumValidAppIdLength}).`,
+    );
+  }
+
+  if (doesStringContainNonPrintableCharacters(potentialAppId)) {
+    throw new Error(`Potential app id (${potentialAppId}) is invalid; it contains non-printable characters.`);
+  }
+}
+
+export function doesStringContainScriptTags(str: string): boolean {
```
TrevorJoelHarris, line 28: I feel like we've also had to write code combing URLs for script tags, might be an opportunity to share

TrevorJoelHarris (approved):

packages/teams-js/src/internal/appId.ts
AE-MS: Created a bug to explore future opportunities here.

packages/teams-js/src/internal/appIdValidation.ts
AE-MS: I was unaware of this but I have now found that code. In the process I have also found code that already does app id validation. However, it is not as thorough, does not mention app IDs at all (it just says "id"), and doesn't give specific error messages on failures, just a generic error message.

My plan is to go forward with this code and then make a subsequent change to unify the app validation code using the one that has the most thorough validation and clearest error messages.

packages/teams-js/test/internal/appId.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+// Since there are plenty of tests validating the individual validation functionality, these tests are intentionally not as
+// comprehensive as those. It executes a few basic tests and also validates that the error messages thrown are as expected.
+describe('doesStringContainNonPrintableCharacters', () => {
+  test('should not throw for "valid" app ids', () => {
+    expect(() => new AppId('8e6523aa-97f9-49ad-8614-75cae22f6597')).not.toThrow();
+    expect(() => new AppId('com.microsoft.teamspace.tab.youtube')).not.toThrow();
+  });
+
+  test('should throw error with "script" in message for app id containing script tag', () => {
+    expect(() => new AppId('<script>alert("Hello")</script>')).toThrowError(/script/);
```
erinha, line 13: I don't think I've seen this syntax before or method of validating error messages (I think we usually match the entire error message string).

I like it; is this a recommended future approach and should we document it as guidance for how we want to unit test error messages?

erinha (approved):
🚀

packages/teams-js/test/internal/appId.spec.ts
AE-MS: Sure, I can do that. I actually thing the right longer-term approach here must include a custom error object with an error code, not just a string. But that's a huge feature set all its own.

PR #2461, opened 2024-08-13
Support for keyboard focus navigation 

arsa-MSFT:
## Description

Add support for key board navigation from from teams to hosted app and vice versa

### Main changes in the PR:

1. Add enum for enter focus action item - This is an optional attribute which will be passed to registerFocusEnterHandler API handler callback
2. Add enum for return focus action item - This is an optional attribute which will be passed to returnFocus API 

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/public/interfaces.ts
```diff
@@ -1262,3 +1262,41 @@ export interface ClipboardParams {
...
 }
+
+/**
+ * @hidden
+ * List of enter focus action items
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export enum EnterFocusActionItem {
```
sunbrk, line 1273: Since this is in the interface naming is of importance. Action Item implies a queued lambda or an unprocessed queued action. Is there a different name that would intuitively suggest enumerated choices for Focus stops?

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1262,3 +1262,41 @@ export interface ClipboardParams {
...
+}
+
+/**
+ * Return focus action items
+ */
+export enum ReturnFocusActionItem {
+  /**
+   * Determines the direction to focus in host for previous landmark.
+   */
+  PreviousLandmark,
```
sunbrk, line 1293: Is the name using 'landmark' based on parity with existing code? These appear to be focus stops or focus targets.

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1262,3 +1262,41 @@ export interface ClipboardParams {
...
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export enum EnterFocusActionItem {
+  /**
+   * Determines the direction to focus in app.
+   */
+  PreviousLandmark,
+  NextLandmark,
+  /**
```
sunbrk, line 1279: Maybe the comment could clarify how 'Read' and 'Compose' enumerations are different from the meaning and intent of PreviousLandmark and NextLandmark. And also the semantic meaning of 'Read' or 'Compose' w.r.t content of the app.

packages/teams-js/src/public/interfaces.ts
sunbrk, line 1293: maybe Landmark is implicit and so can the enum value be Previous, Next, ... ?

arsa-MSFT: It used to be true/false for navigateForward or navigateBackward , however with the new enum attribute we are differentiating between these two. Going forward once we have major version released, navigateFoward will be discontinued

packages/teams-js/src/public/interfaces.ts
arsa-MSFT: makes sense, will add more details

packages/teams-js/src/public/interfaces.ts
arsa-MSFT: How about `EnterFocusType` and `ReturnFocusType` .?

change/@microsoft-teams-js-e3bcf90d-72f4-4327-b55b-ac2eb8aa27bd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Removed Beta/Preview tag on `{namespace}` capability. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
```
TrevorJoelHarris, line 3: This comment doesn't seem correct

packages/teams-js/src/public/pages.ts
```diff
@@ -66,12 +74,16 @@ export function tabsNavigateToTabHelper(apiVersionTag: string, tabInstance: TabI
   });
 }
 
-export function returnFocusHelper(apiVersionTag: string, navigateForward?: boolean): void {
+export function returnFocusHelper(
```
TrevorJoelHarris, line 77:
```suggestion
/**
 * @hidden
 */
export function returnFocusHelper(
```

You didn't introduce this but we should probably never have been exporting this and documenting it publicly. Adding a @hidden keyword will prevent docs from getting autogenerated for this function

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1262,3 +1262,41 @@ export interface ClipboardParams {
...
+   */
+  PreviousLandmark,
+  /**
+   * Determines the direction to focus in host for next landmark.
+   */
+  NextLandmark,
+  /**
+   * Determines if the focus should go to the left rail of teams
+   */
+  GoToLeftRail,
```
TrevorJoelHarris, line 1301: Is there a name for `GoToLeftRail` that you could use that doesn't include the word "left"? AppBar maybe?

I mainly ask because it always makes me shudder when we hardcode assumptions about *where* a piece of UI is into an enum like this. When the web design world inevitably decides that these bars have to go on the right or bottom or whatever in 5 years I'd like for this to still be accurate because we used Semantic naming instead of locational naming.

packages/teams-js/src/public/pages.ts
```diff
@@ -66,12 +74,16 @@ export function tabsNavigateToTabHelper(apiVersionTag: string, tabInstance: TabI
...
+  apiVersionTag: string,
+  navigateForward?: boolean,
+  returnFocusActionItem?: ReturnFocusActionItem,
```
TrevorJoelHarris, line 80: We might need to set up a few minutes for me to ask some questions about this PR. I see that you're adding a new optional parameter here for returnFocusActionitem to a public function that is available to all developers, but you're marking the interface itself in `interfaces.ts` as hidden and internal. This seems wrong-ish to me, but I want to ask more questions

packages/teams-js/src/public/pages.ts
```diff
@@ -66,12 +74,16 @@ export function tabsNavigateToTabHelper(apiVersionTag: string, tabInstance: TabI
   });
 }
 
-export function returnFocusHelper(apiVersionTag: string, navigateForward?: boolean): void {
+export function returnFocusHelper(
+  apiVersionTag: string,
+  navigateForward?: boolean,
```
TrevorJoelHarris, line 79: If a developer specifies a boolean for `navigateForward` AND  a `returnFocusActionItem` in the same function call, what happens?

packages/teams-js/src/public/pages.ts
```diff
@@ -178,8 +190,12 @@ export namespace pages {
    * effect and will be a no-op when called.
    * @param navigateForward - Determines the direction to focus in host.
```
TrevorJoelHarris, line 191: New parameters need to be added to the comments so that they get turned into public documentation.

packages/teams-js/src/public/pages.ts
arsa-MSFT: added hidden to restrict it from generating docs

packages/teams-js/src/public/pages.ts
arsa-MSFT: addressed

packages/teams-js/src/public/interfaces.ts
arsa-MSFT: Changed it to `GoToActivityFeed`

packages/teams-js/src/public/interfaces.ts
arsa-MSFT: I would like to keep it as PreviousLandmark and NextLandmark because in future there might be additional values could be populated here which aren't landmark

change/@microsoft-teams-js-e3bcf90d-72f4-4327-b55b-ac2eb8aa27bd.json
arsa-MSFT: addressed

packages/teams-js/src/public/pages.ts
```diff
@@ -177,9 +191,14 @@ export namespace pages {
    * On mobile hosts or hosts where there is no keyboard interaction or UI notion of "focus" this function has no
    * effect and will be a no-op when called.
    * @param navigateForward - Determines the direction to focus in host.
+   * @param returnFocusType - Determines the type of focus to return to in the host.
    */
-  export function returnFocus(navigateForward?: boolean): void {
-    returnFocusHelper(getApiVersionTag(pagesTelemetryVersionNumber, ApiName.Pages_ReturnFocus), navigateForward);
+  export function returnFocus(navigateForward?: boolean, returnFocusType?: ReturnFocusType): void {
```
TrevorJoelHarris, line 196: I spent some time thinking about this, and I think you'll need to modify this in order for it to work (although I don't think the modification will be too tricky). There are two sort of competing "tricky parts" of what you're trying to do here:

1) Because there are two optional parameters in this function and `returnFocusType` is meant to supersede `navigateForward`, the current signature makes it pretty easy for developers to unintentionally pass in invalid/illegal combinations of parameters, for example `returnFocus(false, ReturnFocusType.NextLandmark)` is sort of giving opposite directions to the function.
2) This function can't ever *only* send an enum value because there will be some hosts (older versions of Teams for example) that don't know about the enum and *only* know how to handle the boolean.

Given those restrictions, I think the best path forward would be to make function overloads for returnFocus that take in different parameters, and then figure out how to resolve the different signatures and send up only valid combinations of booleans and enums. I spent a few minutes this morning puttering around and came up with this:

```typescript
// Document this function for developers
export function returnFocus(navigateForward?: boolean): void;

// Document this function for developers
// You could choose to make this a required parameter if you want since it's a new signature, but if you
// would rather keep it as optional and then determine what the default should be that's okay with me
export function returnFocus(returnFocusType?: ReturnFocusType): void;

function returnFocus(arg1?: boolean | ReturnFocusType): void {
  if (arg1 === undefined) {
    // send default boolean and default ReturnFocusType values up to host
    // You will need to document clearly what will happen on the host in this case
    // (for example, if neither is specified will *either* go to NextLandmark if
    // it can, and if not will go forward)
    console.log('Sent default boolean value of true and default ReturnFocusType of NextLandmark')
    return;
  }

  if (typeof arg1 === 'boolean') {
    console.log(`Sent boolean ${arg1} and no ReturnFocusType`);
    return;
  } else {
    let backupBooleanValue: boolean;
    switch (arg1) {
      case (ReturnFocusType.PreviousLandmark):
      case (ReturnFocusType.GoToActivityFeed):
        backupBooleanValue = false;
        break;
      case (ReturnFocusType.NextLandmark):
        backupBooleanValue = true;
    }
    console.log(`Sent ReturnFocusType ${arg1} and backup boolean value ${backupBooleanValue}`);
  }
}
```

Essentially, you would have two exported functions. 1 only accepts booleans and 1 only accepts enums. You then have a third function that is *not* exported where all of the actual implementation is stored. This function figures out whether it has been passed a boolean or enum and then passes the *correct* combination of boolean and enum up to the host. There are 3 cases you would need to handle:

1) No parameter is passed in. In this case you will need to send up logical default values. You could either send up *just* a boolean, or you could send a default boolean and a default enum.

2) A boolean is passed in. You probably just send that boolean to the host.

3) An enum is passed in. In order for this to work on older hosts that don't know about the enum, you will need to map each enum value to it's most logical boolean value so that developers who call that function on older hosts get "basically correct" behavior.

I realize this is a lot to read, but I think the change itself is actually not too bad. If you have questions about any of this, just send me a message and we can chat.

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1262,3 +1262,43 @@ export interface ClipboardParams {
...
+   * Compose - Focus should go to the compose area (such as textbox) of the app.
+   */
+  Read,
+  Compose,
+}
+
+/**
+ * Return focus action items
+ */
+export enum ReturnFocusType {
```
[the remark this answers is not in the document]
arsa-MSFT: addressed

change/@microsoft-teams-js-e3bcf90d-72f4-4327-b55b-ac2eb8aa27bd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add optional enum attribute for registerFocusEnterHandler and returnFocus API",
```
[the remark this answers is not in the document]
arsa-MSFT: addressed

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1262,3 +1262,43 @@ export interface ClipboardParams {
...
 }
+
+/**
+ * @hidden
+ * List of enter focus action items
+ *
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export enum EnterFocusType {
```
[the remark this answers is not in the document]
arsa-MSFT: moved to pages

packages/teams-js/src/public/pages.ts
arsa-MSFT: HI @TrevorJoelHarris, thanks for taking time and review the PR. I tried to incorporate the changes that were called out. Feel free to take another look

apps/teams-test-app/src/components/PagesAPIs.tsx
```diff
@@ -156,6 +156,23 @@ const ReturnFocus = (): React.ReactElement =>
     },
   });
 
+const ReturnFocusWithType = (): React.ReactElement =>
+  ApiWithTextInput<pages.ReturnFocusType>({
+    name: 'returnFocusWithType',
+    title: 'Return Focus With Type',
+    onClick: {
+      validateInput: (input) => {
+        if (!input) {
```
TrevorJoelHarris, line 165: This should validate that input is one of your enum values

apps/teams-test-app/e2e-test-data/pages.json
```diff
@@ -25,6 +25,14 @@
       "expectedAlertValue": "returnFocus called with true",
       "expectedTestAppValue": "Current navigateForward state is true"
     },
+    {
+      "title": "returnFocus With Type API Call - Success",
+      "type": "callResponse",
+      "boxSelector": "#box_returnFocusWithType",
+      "inputValue": 2,
+      "expectedAlertValue": "returnFocus called with false",
```
TrevorJoelHarris, line 33: Is the difference here because you are testing that the "false" fallback is passed up to the host when they don't know about the new enum?

packages/teams-js/src/public/pages.ts
```diff
@@ -172,14 +174,95 @@ export namespace pages {
...
    * On mobile hosts or hosts where there is no keyboard interaction or UI notion of "focus" this function has no
    * effect and will be a no-op when called.
    * @param navigateForward - Determines the direction to focus in host.
    */
-  export function returnFocus(navigateForward?: boolean): void {
-    returnFocusHelper(getApiVersionTag(pagesTelemetryVersionNumber, ApiName.Pages_ReturnFocus), navigateForward);
+  export function returnFocus(navigateForward?: boolean): void;
+
+  /**
+   * Return focus to the host. Will move focus forward, backward or activity feed based on where the application container falls in
```
TrevorJoelHarris, line 233:
```suggestion
   * Return focus to the host. Will attempt to send focus to the appropriate part of the host (as specified by returnFocusType) based on where the application container falls in
```

packages/teams-js/src/public/pages.ts
```diff
@@ -172,14 +174,95 @@ export namespace pages {
...
+    /**
+     * Determines the direction to focus in host for previous landmark.
+     */
+    PreviousLandmark = 0,
+    /**
+     * Determines the direction to focus in host for next landmark.
+     */
+    NextLandmark = 1,
+    /**
+     * Determines if the focus should go to the left rail of teams
```
TrevorJoelHarris, line 217:
```suggestion
     * Determines if the focus should go to the host's activity feed
```

apps/teams-test-app/e2e-test-data/pages.json
```diff
@@ -25,6 +25,14 @@
...
+      "expectedTestAppValue": "Current state is 2"
```
TrevorJoelHarris, line 34: Is the idea that you will come back and add another e2e test testing that your enum value gets sent to the host after you check in the host-sdk part of the change?

apps/teams-test-app/src/components/PagesAPIs.tsx
arsa-MSFT: addressed

packages/teams-js/src/public/pages.ts
arsa-MSFT: addressed

apps/teams-test-app/e2e-test-data/pages.json
arsa-MSFT: well I initially thought to align what hub tests were - but considering there's a new arg in town - I changed the test to reflect that.

apps/teams-test-app/e2e-test-data/pages.json
arsa-MSFT: changed it expect enum. However since it requires hub sdk test change - I'll mostly keep it behind a hubsdk web version

apps/teams-test-app/e2e-test-data/pages.json
```diff
@@ -22,9 +22,25 @@
       "type": "callResponse",
       "boxSelector": "#box_returnFocus",
       "checkboxState": true,
-      "expectedAlertValue": "returnFocus called with true",
+      "version": ">2.27.0",
```
KangxuanYe, line 25: This `version` is Teams-Js library version, I would suggest to change it from '>2.27.0' to  '>= 2.27.0' or you could say '>2.26.0'

apps/teams-test-app/e2e-test-data/pages.json
```diff
@@ -22,9 +22,25 @@
...
+      "hostSdkVersion": {
+        "web": ">=4.2.0"
```
KangxuanYe, line 27:
```suggestion
        "web": ">4.1.0"
```
4.1.0 is current web host sdk version.

apps/teams-test-app/e2e-test-data/pages.json
```diff
@@ -22,9 +22,25 @@
       "type": "callResponse",
```
KangxuanYe, line 22: We can use `version` or/and `hostSdkVersion` to exclude the original test case and then you will need to add two new cases.

TrevorJoelHarris (dismissed):

TrevorJoelHarris (dismissed):

TrevorJoelHarris (approved):

KangxuanYe (approved):

PR #2463, opened 2024-08-14
Update WebStorage to test proper result responses from the host sdk

alexneyman-MSFT:
## Description

This PR updates the WebStorage E2E tests to better handle different host SDK versions. We're adding new test cases for webstorage because we have a new flag to test webstorage behavior.

### Main changes in the PR:

1. Updated the WebStorage E2E test configuration in `apps/teams-test-app/e2e-test-data/webStorage.json`.
2. Modified the `isWebStorageClearedOnUserLogOut` function implementation in `apps/teams-test-app/src/components/WebStorageAPIs.tsx`.

### Validation performed:

1. Ran the updated E2E tests locally to verify that they pass for different host SDK versions.
2. Checked that the tests correctly handle the behavior differences between SDK versions before and after 4.0.2.

### Unit Tests added:

No new unit tests were added as this change primarily affects E2E tests.

### End-to-end tests added:

Yes, the existing E2E tests for WebStorage have been updated and expanded.

### Change file added:

No

### Related PRs:

N/A

### Next/remaining steps:

- [ ] Consider adding similar version-specific tests for other modules if needed.
- [ ] Update documentation to reflect these changes in testing strategy, if applicable.

### Screenshots:

N/A

### DIFF
--- a/apps/teams-test-app/e2e-test-data/webStorage.json
+++ b/apps/teams-test-app/e2e-test-data/webStorage.json
@@ -7,7 +7,7 @@
   },
   "checkIsSupported": {
     "domElementName": "checkWebStorageCapability",
-    "expectedOutput": "webStorage is not supported"
+    "expectedOutput": "webStorage is supported"
   },
   "testCases": [
     {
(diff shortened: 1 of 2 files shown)


apps/teams-test-app/e2e-test-data/webStorage.json
```diff
@@ -7,15 +7,39 @@
...
   "testCases": [
     {
       "title": "isWebStorageClearedOnUserLogOut function Call - Success",
       "type": "callResponse",
       "boxSelector": "#box_isWebStorageClearedOnUserLogOut",
       "expectedAlertValue": "isWebStorageClearedOnUserLogOut called",
-      "expectedTestAppValue": "webStorage is cleared on user log out"
+      "expectedTestAppValue": "webStorage is cleared on user log out",
+      "hostSdkVersion": {
+        "web": "<4.0.2"
```
KangxuanYe, line 20: I think this PR might need to wait for a little bit since Web Host SDK is releasing and I am not sure which version you should exactly to have here.

alexneyman-MSFT: sounds good, will wait

AE-MS:
>   "version": ">2.22.0",

I'm not totally sure I understand how this works but does there need to be another restriction on these cases because for the teamsjs version? The behavior of the test app now changes depending on which version of TJS is being used (e.g., the test app from TJS before your change prints a different message). Is this same JSON file used when the tests are run against TJS 2.26? Or does it use the JSON files that were part of 2.26?

---
Refers to: apps/teams-test-app/e2e-test-data/webStorage.json:4 in 1173e5e. [](commit_id = 1173e5e7647ff1660e6657253e650e4479dd23ea, deletion_comment = False)

AE-MS (approved):
:shipit:

AE-MS:
I approved this but I left an open question that I don't know the answer to and I'm not sure if it should block this getting merged.

PR #2464, opened 2024-08-15
Enable the webStorage capability to function on Teams Mobile even though they run in compatibility mode

AE-MS:
## Description

Teams Mobile uses "compatibility mode" (no host SDK) so they cannot implement the new `webStorage` capability in the same way other hosts would. This change enables the `webStorage.isWebStorageClearedOnUserLogOut` function to return `true` on Teams Mobile, which is what they would return if they implemented the capability. It is also the behavior that this function used to have before it was turned into a capability.

Note that `webStorage.isSupported()` already returns the "correct" value to enable this scenario based on the data in the `mapTeamsVersionToSupportedCapabilities` structure.

### Validation performed:

1. Leveraged test app running on Teams iOS to verify that `webStorage.isWebStorageClearedOnUserLogOut` returns `true` and no longer hangs after this change
2. Ensured unit tests cover this case.

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-  "cSpell.words": ["adal", "teamspace", "uninitialize"],
+  "cSpell.words": ["adal", "ipados", "teamspace", "uninitialize"],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "editor.insertSpaces": true,
--- a/packages/teams-js/src/public/webStorage.ts
+++ b/packages/teams-js/src/public/webStorage.ts
@@ -4,7 +4,7 @@ import { ensureInitialized } from '../internal/internalAPIs';
 import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
 import { app } from './app';
 import { errorNotSupportedOnPlatform, HostClientType, HostName } from './constants';
-import { runtime } from './runtime';
+import { getRuntimeConfigSource, runtime, RuntimeConfigSource } from './runtime';
 
 /**
  * Contains functionality enabling apps to query properties about how the host manages web storage (`Window.LocalStorage`)
@@ -28,6 +28,7 @@ export namespace webStorage {
     }
 
     if (
+      getRuntimeConfigSource() === RuntimeConfigSource.TeamsFallback &&
       (GlobalVars.hostClientType === HostClientType.android || GlobalVars.hostClientType === HostClientType.ios) &&
       (await getHostName()) === HostName.teams
     ) {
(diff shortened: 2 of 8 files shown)


packages/teams-js/src/public/webStorage.ts
```diff
@@ -25,12 +27,34 @@ export namespace webStorage {
       throw errorNotSupportedOnPlatform;
     }
 
+    if (
+      (GlobalVars.hostClientType === HostClientType.android || GlobalVars.hostClientType === HostClientType.ios) &&
```
TrevorJoelHarris, line 31: Should we also return true for HostClientType.ipados?

packages/teams-js/src/public/webStorage.ts
```diff
@@ -25,12 +27,34 @@ export namespace webStorage {
...
+      (await getHostName()) === HostName.teams
+    ) {
+      // On Teams Mobile, they haven't yet implemented this capability. However, for compatibility reasons, we need
+      // to act as if they do. If they did implement it, they would return true, so that's what we do here.
+      // Getting Teams Mobile to implement this is a work-in-progress. Once they do implement it, we can remove this
+      // whole if-block. Until then, we cannot send the message to them because they will not understand it.
+      return true;
+    }
+
     return await sendAndUnwrap(
```
TrevorJoelHarris, line 41: I don't have a good idea here (yet), but it would be great to only return true for these mobile platforms *if* they aren't providing their own implementation. It mightbe possible to look at what happens when we receive a response from the host and return true if we're on a mobile platform AND we receive a specific error? Alternatively, we could build in a sort of "default" implementation of this capability on the host sdks for the mobile platform (although that would be more work).

I don't know that this is necessary, but it would be nice to avoid a bug in the future where mobile hosts implement this and return false, but we are still hardcoding the answer to true.

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "cSpell.words": ["adal", "teamspace", "uninitialize"],
+  "cSpell.words": ["adal", "ipados", "teamspace", "uninitialize"],
```
AE-MS, line 2: >ipados [](http://example.com/codeflow?start=28&length=6)

This is gross but it is the casing that is used in our enum so I guess we have to accept it.

packages/teams-js/src/public/webStorage.ts
AE-MS: Fixed; good catch.

packages/teams-js/src/public/webStorage.ts
AE-MS: Let me know how you feel about the latest iteration of this where I tried to achieve what you were looking for.

AE-MS:

There are no current changes to the `runtime.ts` file.

.vscode/settings.json
TrevorJoelHarris: what do you mean I love ipa dos

TrevorJoelHarris (approved):

alexneyman-MSFT (approved):

PR #2476, opened 2024-08-21
Upgrade package with CG vulnerability

jekloudaMSFT:

## Description

This package was flagged on the Android pipeline. Upgrading to fix the CG alert

### DIFF
--- a/apps/teamsjs-diagnostic-app/package.json
+++ b/apps/teamsjs-diagnostic-app/package.json
@@ -11,7 +11,7 @@
     "@microsoft/teamsfx": "^2.3.2",
     "@microsoft/teamsfx-react": "^3.1.2",
     "@testing-library/dom": "^10.4.0",
-    "axios": "^1.7.3",
+    "axios": "^1.7.4",
     "react": "^18.3.1",
     "react-dnd": "^16.0.1",
     "react-dnd-html5-backend": "^16.0.1",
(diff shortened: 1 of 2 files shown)


jadahiya-MSFT (approved):

apps/teamsjs-diagnostic-app/package.json
```diff
@@ -11,7 +11,7 @@
     "@microsoft/teamsfx": "^2.3.2",
     "@microsoft/teamsfx-react": "^3.1.2",
     "@testing-library/dom": "^10.4.0",
-    "axios": "^1.7.3",
+    "axios": "^1.7.4",
```
AE-MS, line 14: Should an update to `pnpm-lock.yaml` come in this same change?

AE-MS (changes requested):
:clock1:

apps/teamsjs-diagnostic-app/package.json
jekloudaMSFT: Yep, just pushed

jadahiya-MSFT (approved):

pnpm-lock.yaml
```diff
@@ -3632,8 +3632,8 @@ packages:
   axios@0.28.1:
```
AE-MS, line 3632: >axios@0.28.1 [](http://example.com/codeflow?start=2&length=12)

For others who might look at this later, it is okay that we still depend on axios 0.28.1 as the vulnerable versions only start with 1.x. So 0.28.1 is okay (at least as far as this particular vulnerability is concerned.)

AE-MS (approved):
:shipit:

PR #2477, opened 2024-08-21
Move AppId type to the public folder

AE-MS:
## Description

This type is intended to be used by public functions so it should be in the public folder.

I also renamed a misnamed test file (`appValidation.spec.ts` ➡️ `appIdValidation.spec.ts`)


packages/teams-js/test/internal/appIdValidation.spec.ts
jadahiya-MSFT, line 1: Should this also move to the public folder?

AE-MS: Great question! Since the `appIdValidation.ts` file remained in the `internal` folder (since it's not hosting an external type but only hosting internal validation code) the corresponding (now correctly named) `appIdValidation.spec.ts` file remains in the matching `internal` folder.

Did I understand your question correctly?

jadahiya-MSFT: Yup thanks for the clarification

jadahiya-MSFT (approved):

TrevorJoelHarris (approved):

PR #2480, opened 2024-08-22
Add a type-safe option to use when calling pages.navigateToApp

AE-MS:
## Description

Where possible, we should accept data that is as type-safe as possible. Moreover, we should encourage developers to use leverage type-safety to prevent bugs and better understand our API surface. This change enables the `pages.navigateToApp` function to accept a more type-safe input object.

### Main changes in the PR:

1. Create new `AppNavigationParameters` interface which is more type-safe than the previous `NavigateToAppParams` interface:
- It uses the `AppId` type for the `appId` property (instead of `string`)
- It uses the `URL` type for the `webUrl` property (instead of `string`)
2. Mark the old `NavigateToAppParams` interface as deprecated
3. Update `pages.navigateToApp` function to support either of those types as input
4. Add unit tests for both possible input types to `pages.navigateToApp`

### Validation performed:

1. New unit tests and all existing tests pass
2. All E2E tests pass

### Unit Tests added:

Yes

### End-to-end tests added:

No. I intentionally did not add new E2E tests to cover the new input type as I didn't think it would add any value over what the new unit tests already offer. Specifically, the E2E tests will ensure that the overall flow works correctly, and the unit tests will ensure that the right message is emitted from TJS regardless of the input.

### Change file added:

Yes


packages/teams-js/src/public/pages.ts
```diff
@@ -399,10 +401,17 @@ export namespace pages {
         throw errorNotSupportedOnPlatform;
       }
       const apiVersionTag: string = getApiVersionTag(pagesTelemetryVersionNumber, ApiName.Pages_NavigateToApp);
+
       if (runtime.isLegacyTeams) {
-        resolve(sendAndHandleStatusAndReason(apiVersionTag, 'executeDeepLink', createTeamsAppLink(params)));
+        const typeSafeParameters: AppNavigationParameters = !isAppNavigationParametersObject(params)
+          ? convertNavigateToAppParamsToAppNavigationParameters(params)
```
TrevorJoelHarris, line 407: This is nitpicky, but I think if `convertNavigateToAppParamsToAppNavigationParameters` (lol that's a mouthful) was changed to accept both types and then did the `!isAppNavigationParametersObject` check internally, you could write code here that looked like:

```typescript
const typeSafeParameters: AppNavigationParameters = upgradeNavigateToAppParamsIfNeeded(params);
```

TrevorJoelHarris (approved):

packages/teams-js/src/public/pages.ts
AE-MS: I understand your feedback and I do like seeing easy-to-read code. My concern with this approach is that I generally don't prefer functions which "might or might not" do a thing because it makes it difficult for the caller to predict what will happen or see what did happen. If I were going to do it, I'd probably leave the `convertNavigateToAppParamsToAppNavigationParameters` function as it is and then create your suggested function as a wrapper (so that we still had a predictable atomic function).

Given that you said it was nitpicky and that you also said (separately) that I could take or leave, I think I will choose to leave this one. Thank you for your suggestion.

Oh, I would also consider your suggestion further if there were multiple places in the code where I was doing this check but since this seems to be the only one, I'll leave it.

jekloudaMSFT (approved):

PR #2483, opened 2024-08-26
CEA Auth changes

lakhveerk:

## Description
Adding support for `ExternalAppAuthenticationForCEA`. This is a new capability available only for MChat to call in cases the thrid party app is backed with Custom Engine Agent. 
The `ExternalAppAuthenticationForCEA` APIs work similar to ME APIs `AuthenticateAndResendRequest` and `AuthenticateWIthSSOAndResendRequest` that exist in `ExternalAppAuthentication` capability. The few differences are 
1. There are two more APIs, `AuthenticateWithSSO` and `AuthenticateWithOauth`. The purpose of each API can be read in externalAppAuthenticationForCEA.ts file.
2. CEA APIs take coversationId along with other parameters
3. CEA takes AppId object instead of string type

### Main changes in the PR:

1. Added `ExternalAppAuthenticationForCEA` capability to teamsjs
2. Added corresponding test app changes
3. Added e2e test file

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/private/externalAppAuthenticationForCEC.ts
```diff
@@ -0,0 +1,169 @@
+import { sendMessageToParentAsync } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
+import { validateId } from '../internal/utils';
+import { errorNotSupportedOnPlatform, FrameContexts } from '../public/constants';
+import { runtime } from '../public/runtime';
+import {
+  ActionExecuteInvokeRequestType,
```
MengyiGong, line 8: Would be great to have the unit tests for authForCEC APIs.

apps/teams-test-app/src/components/privateApis/ExternalAppAuthenticationAPIs.tsx
```diff
@@ -50,7 +50,7 @@ const AuthenticateAndResendRequest = (): React.ReactElement =>
     defaultInput: JSON.stringify({
       appId: 'b7f8c0a0-6c1d-4a9a-9c0a-2c3f1c0a3b0a',
       authenticateParameters: {
-        url: 'https://www.example.com',
+        url: 'https://localhost:4000',
```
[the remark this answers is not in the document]
lakhveerk: example.com is blocked since the domain does not match. This corrects the default value which previously was throwing error.

apps/teams-test-app/src/components/privateApis/ExternalAppAuthenticationForCEAAPIs.tsx
```diff
@@ -0,0 +1,207 @@
...
+    };
+    originalRequestInfo: externalAppAuthentication.IActionExecuteInvokeRequest;
+  }>({
+    name: 'authenticateAndResendRequestForCEA',
+    title: 'Authenticate And Resend Request',
+    onClick: {
+      validateInput: (input) => {
+        if (!input.appId) {
+          throw new Error('appId is required');
+        }
```
MengyiGong, line 113: Can you also check on the conversationId?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+    ensureInitialized(runtime, FrameContexts.content);
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    validateId(conversationId, new Error('conversation id is not valid.'));
+
+    // Ask the parent window to open an authentication window with the parameters provided by the caller.
+    const [error] = await sendMessageToParentAsync<[externalAppAuthentication.InvokeError]>(
```
MengyiGong, line 84: Would be great to consider using `sendAndUnwrap` so that we can directly get error instead of `[error]`

apps/teams-test-app/src/components/privateApis/ExternalAppAuthenticationForCEAAPIs.tsx
MengyiGong: same for other APIs in this file

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+  ): Promise<void> {
+    ensureInitialized(runtime, FrameContexts.content);
+
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    validateId(conversationId, new Error('conversation id is not valid.'));
+
+    const [error] = await sendMessageToParentAsync<[externalAppAuthentication.InvokeError]>(
```
AE-MS, line 45: >[externalAppAuthentication.InvokeError [](http://example.com/codeflow?start=51&length=38)

It seems unnecessary to have a single value in an array; can you use something like `sendAndUnwrap` instead of `sendMessageToParentAsync` so that you just get back the value you want and not an array with the value in it?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+      getApiVersionTag(
+        externalAppAuthenticationTelemetryVersionNumber,
+        ApiName.ExternalAppAuthenticationForCEA_AuthenticateWithSSO,
+      ),
+      ApiName.ExternalAppAuthenticationForCEA_AuthenticateWithSSO,
+      [appId.toString(), conversationId, authTokenRequest.claims, authTokenRequest.silent],
+    );
+    if (error) {
```
AE-MS, line 53: >error [](http://example.com/codeflow?start=8&length=5)

Based on the type I see being used in the `sendMessageToParentAsync` call, it seems like `undefined` / `null` is not a value return type -- it must be a `externalAppAuthentication.InvokeError` object. If this hub SDK call can actually return `undefined` / `null`, should the return type of the "send" call actually be `externalAppAuthentication.InvokeError | undefined` ?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+      throw errorNotSupportedOnPlatform;
+    }
+
+    validateId(conversationId, new Error('conversation id is not valid.'));
+
+    validateOriginalRequestInfo(originalRequestInfo);
+
+    // Ask the parent window to open an authentication window with the parameters provided by the caller.
+    const [response] = await sendMessageToParentAsync<
+      [externalAppAuthentication.InvokeErrorWrapper | externalAppAuthentication.IActionExecuteResponse]
```
AE-MS, line 135: >InvokeErrorWrapper [](http://example.com/codeflow?start=33&length=18)

Since you didn't introduce this type as part of this PR I don't know if you'll be able to answer this but I am curious: what's the value of having the `InvokeErrorWrapper` type? Could this just return `externalAppAuthentication.InvokeError | externalAppAuthentication.IActionExecuteResponse`?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+        authenticateParameters.isExternal,
+      ],
+    );
+    if (isActionExecuteResponse(response)) {
+      return response;
+    } else if (isInvokeErrorWrapper(response)) {
+      throw response;
+    } else {
+      throw defaultExternalAppError;
+    }
```
AE-MS, line 158: Not the biggest deal but this could also be written like this:
```
else {
    throw isInvokeErrorWrapper(response) ? response : defaultExternalAppError
}
```

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  function isActionExecuteResponse(response: unknown): response is externalAppAuthentication.IActionExecuteResponse {
```
AE-MS, line 161: >isActionExecuteResponse [](http://example.com/codeflow?start=11&length=23)

I usually like to keep type guards defined next to the place where the type itself is defined since they are tightly coupled. Moreover, if anyone needs a `IActionExecuteResponse` type guard in the future, they are less likely to accidentally make a duplicate implementation if the existing one lives right next to the type.

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  /**
+   * @beta
+   * @hidden
+   * Signals to the host to perform authentication using the given authentication parameters and then resend the request to the application backend with the authentication result.
+   * @internal
+   * Limited to Microsoft-internal use
+   * @param appId ID of the application backend to which the request and authentication response should be sent. This must be a UUID
```
AE-MS, line 110: >appId ID [](http://example.com/codeflow?start=12&length=8)

Double ids!

AE-MS, line 110: >application backend [](http://example.com/codeflow?start=28&length=19)

Is this the appid of the bot? I'm not quite sure what "application backend" means here.

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+   * @param conversationId ConversationId To tell the bot what conversation the calls are coming from
+   * @param authenticateParameters Parameters for the authentication pop-up
+   * @param originalRequestInfo Information about the original request that should be resent
+   * @throws InvokeError if the host encounters an error while authenticating or resending the request
+   * @returns A promise that resolves to the IActionExecuteResponse from the application backend and rejects with InvokeError if the host encounters an error while authenticating or resending the request
```
AE-MS, line 115: >InvokeError [](http://example.com/codeflow?start=114&length=11)

Can it also reject with `defaultExternalAppError` ?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: This is the app ID of the third-party app on behalf of which Copilot is making a call

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+    );
+  }
+
+  /**
+   * @beta
+   * @hidden
+   * Signals to the host to perform SSO authentication for the application specified by the app ID and then resend the request to the application backend with the authentication result and originalRequestInfo
+   * @internal
+   * Limited to Microsoft-internal use
+   * @param appId ID of the application backend for which the host should attempt SSO authentication and resend the request and authentication response. This must be a UUID.
```
AE-MS, line 178: >Id ID [](http://example.com/codeflow?start=15&length=5)

Double ids

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  function isInvokeErrorWrapper(err: unknown): err is externalAppAuthentication.InvokeErrorWrapper {
+    if (typeof err !== 'object' || err === null) {
+      return false;
+    }
+
+    const errorWrapper = err as externalAppAuthentication.InvokeErrorWrapper;
+
+    return (
+      errorWrapper?.errorCode === externalAppAuthentication.InvokeErrorCode.INTERNAL_ERROR &&
```
AE-MS, line 259: >? [](http://example.com/codeflow?start=18&length=1)

Is this needed?

AE-MS, line 259: >errorWrapper?.errorCode === externalAppAuthentication.InvokeErrorCode.INTERNAL_ERROR [](http://example.com/codeflow?start=6&length=84)

I don't think it's right to test that `errorWrapper.errorCode` is exactly this value, since it could be some other value in the future? The `errorCode` property can be any `InvokeErrorCode` value. Yes, it is true that `InvokeErrorCode` has only one value right now, but it can have more in the future. If this code only tests for `InvokeErrorCode.INTERNAL_ERROR`, then this function will no longer work when new error codes are added (is the expectation that every new `InvokeErrorCode` gets added to the `enum` and also updated here?

apps/teams-test-app/e2e-test-data/externalAppAuthenticationForCEA.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "ExternalAppAuthenticationForCEA",
+  "version": ">2.29.0",
+  "platform": "Web",
+  "testCases": [
```
AE-MS, line 5: >testCases [](http://example.com/codeflow?start=3&length=9)

Does there have to be a web host SDK version restriction for this, since the support for this capability was only recently added?

AE-MS (changes requested):
:clock1:

apps/teams-test-app/e2e-test-data/externalAppAuthenticationForCEA.json
lakhveerk: I don't think I understand. The platform "Web" will restrict it to web. What else do I need?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: @jekloudaMSFT Jeff, do you have context why we made a wrapper?

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
AE-MS: FWIW that explanation would make for a clearer comment to me, anyway, e.g.,
```
@param appId App ID of the app upon whose behalf Copilot is requesting authentication. This must be a UUID.
```

apps/teams-test-app/e2e-test-data/externalAppAuthenticationForCEA.json
AE-MS: Since we support multiple versions of the host sdk simultaneously, the automated tests are run against those versions when new releases are needed. Because the implementation of this capability does not exist in all of those versions, I think you will need to restrict which host SDK versions this test runs against using a filter like this:
```
  "hostSdkVersion": {
    "web": ">2.16.0"
  },
```

If you search for "hostSdkVersion" you can find some examples in the other JSON files.

Is that clearer?

lakhveerk: that makes sense

lakhveerk: If this was checking against any lower version than when teh support was added (4.3.0), should not it have failed in teh pipelines. I am adding the version to be safe :)

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: Added undefined to the type for non error cases

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: Jess responded that he believes it was just a typing thing. Since I have `isInvokeError`, I don't need it. 
I have updated everywhere from IinvokeErrorWrapper to InvokeError

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: Moved them closer to the interfaces

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: not really. Deleting it. err needs to be an object and if it is not, the first if statement will return false.

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (!isSupported()) {
+      throw errorNotSupportedOnPlatform;
+    }
+
+    validateId(conversationId, new Error('conversation id is not valid.'));
+
+    validateOriginalRequestInfo(originalRequestInfo);
+
+    // Ask the parent window to open an authentication window with the parameters provided by the caller.
+    const [response] = await sendMessageToParentAsync<
```
AE-MS, line 135: >sendMessageToParentAsync [](http://example.com/codeflow?start=29&length=24)

Can this also use `sendAndUnwrap` to avoid needing the seemingly unnecessary array?

AE-MS: (below this, too)

apps/teams-test-app/e2e-test-data/externalAppAuthenticationForCEA.json
AE-MS: Yes, it "should" fail in the pipelines but as of right now the tests are currently only run on the older releases "on demand." This is obviously not supportable because it will lead to people such as yourself making perfectly innocent changes.

There is a bug tracking enabling these older tests in the pipelines that is sitting on Jay's plate but for privacy reasons I cannot provide the bug link to you here.

AE-MS (changes requested):
Thanks for incorporating my feedback; I've left one new comment.

packages/teams-js/src/private/externalAppAuthenticationForCEA.ts
lakhveerk: Fixed!

AE-MS (approved):
:shipit:

MengyiGong (approved):

PR #2489, opened 2024-08-30
Update getMeetingDetailsVerbose Interface

MSJohnnyLee:

## Description

This PR modifies the return interface of the `getMeetingDetailsVerbose` API. Specifically, the `response.details.originalCaller` and `response.details.dialedEntity` properties.

`originalCaller` represents the caller of a call, and `dialedEntity` represents the callee. Those two properties originally were of type `string`, and the return value could've been either an email or phone number in string format.

With this change, two additional properties were added to replace them: `originalCallerInfo` and `dialedEntityInfo`. Those two properties are of a new type `ICallParticipantIdentifiers`. This new type `ICallParticipantIdentifiers` is an object that has 2 explicit fields for `phoneNumber` and `email`.

This new type serves two purposes: 1. The returned value is more explicitly defined 2. The response can return both a phone number and email if that information can be retrieved.

In addition, this change also a new optional property, `callId`.

This PR is a follow up to the original PR that first introduced the `getMeetingDetailsVerbose` API:
https://github.com/OfficeDev/microsoft-teams-library-js/pull/2206

### Main changes in the PR:

1. Create new interface `ICallParticipantIdentifiers`
2. Deprecate `response.details.originalCaller` and `response.details.dialedEntity` 
3. Create two new, optional properties `response.details.originalCallerInfo` and `response.details.dialedEntityInfo` of the new interface `ICallParticipantIdentifiers`
4. Create new optional property `callId`

### Validation performed:

1. Validated in Orange that the new meeting.getMeetingDetailsVerbose API returns the new `response.details.originalCaller` and `response.details.dialedEntity` properties in a supported host
2. Ran automated test cases

### Unit Tests added:

Updated existing test

### End-to-end tests added:

Updated existing test

### Change file added:

Yes, added change file

### Related PRs:

Original getMeetingDetailsVerbose PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/2206


jadahiya-MSFT (dismissed):

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -35,11 +35,11 @@
     },
     {
       "title": "getMeetingDetailsVerbose API Call - Success",
-      "version": ">2.22.0",
```
TrevorJoelHarris, line 38: rather than edit this test, you will want to modify the version to be `> 2.22.0 && <= 2.28.0` and then leave the expected value as is. You can then introduce a new test with version `> 2.28.0` that has your new expected value

packages/teams-js/src/public/meeting.ts
```diff
@@ -94,29 +94,68 @@ export namespace meeting {
...
      * @hidden
      * Phone number of a PSTN callee or email of a VoIP callee
      */
     dialedEntity?: never;
 
+    /**
+     * @hidden
+     * Object representing the entity the caller dialed
+     */
+    dialedEntityInfo?: never;
```
TrevorJoelHarris, line 146: Can you explain why you had a parameter of type never and then added a new one also of type never? I don't think I understand your use of never here. Your PR description says:

"originalCaller represents the caller of a call, and dialedEntity represents the callee. Those two properties originally were of type string, and the return value could've been either an email or phone number in string format.

With this change, those properties have been updated to be of a new type ICallParticipantIdentifiers.
This new type ICallParticipantIdentifiers is an object that has 2 explicit fields for phoneNumber and email."

However, both dialedEntity and dialedEntityInfo are never of type `ICallParticipantIdentifiers`

packages/teams-js/src/public/meeting.ts
```diff
@@ -94,29 +94,68 @@ export namespace meeting {
...
+  /**
+   * @hidden
+   * Hide from docs
+   * Data structure to represent call participant identifiers
+   */
+  interface ICallParticipantIdentifiers {
+    /**
+     * Phone number of a PSTN caller
+     */
+    phoneNumber?: string;
```
TrevorJoelHarris, line 106: Since you made a custom class for EmailAddress (nice!), why did you choose not to make one for phone number? Seems like you could take a similar approach with a regex like:

`/^[\+]?[0-9]{0,3}\W?+[(]?[0-9]{3}[)]?[-\s\.]?[0-9]{3}[-\s\.]?[0-9]{4,6}$/im`

packages/teams-js/src/public/meeting.ts
```diff
@@ -538,6 +577,25 @@ export namespace meeting {
     ScreenShare = 'ScreenShare',
   }
 
+  /**
+   * Represents a validated email.
+   *
+   * @hidden
+   * Hide from docs.
+   */
+  export class EmailAddress {
```
TrevorJoelHarris, line 586: This class should have unit tests ensuring that incorrect email addresses are rejected.

packages/teams-js/src/public/meeting.ts
```diff
@@ -665,7 +723,7 @@ export namespace meeting {
 
     if (
       (response.details?.type == CallType.GroupCall || response.details?.type == CallType.OneOnOneCall) &&
-      !response.details.originalCaller
+      !response.details.originalCallerInfo
```
TrevorJoelHarris, line 726: There should be a unit test for this scenario (I assume there is not because if there was I would see an edited test below)

apps/teams-test-app/e2e-test-data/meeting.json
MSJohnnyLee: Updated test, thanks!

packages/teams-js/src/public/meeting.ts
MSJohnnyLee: Sorry, forgot to add tests as I included additional logic, will add them

packages/teams-js/src/public/meeting.ts
MSJohnnyLee: Will add a test for this scenario, thanks

packages/teams-js/src/public/meeting.ts
MSJohnnyLee: Thanks! I included it upon Aaron's suggestions

The reason I didn't add it for phone numbers is because our service doesn't pre-process phone numbers, we return the number in the format it comes in as. There's no fixed / expected format for phone numbers within our service, and we also don't want to coerce it into a specific format that might work for one customer or scenario but not for others.

packages/teams-js/src/public/meeting.ts
MSJohnnyLee: Actually, it seems like I've included a test in the original PR:
https://github.com/OfficeDev/microsoft-teams-library-js/blob/753866fa9498d97f494ad2342a5a3c414ddbc157/packages/teams-js/test/public/meeting.spec.ts#L684C11-L727C14

The reason why this test is compatible even with this change is because it checks to see if the new property, `originalCaller` / `originalCallerInfo` is missing. The logic is that this new API should return the new property if it is supported, hence if the property isn't returned, the platform isn't supported.

packages/teams-js/src/public/meeting.ts
MSJohnnyLee: Those properties are in the process of being implemented, hence we originally decided to mark them as `never`, to indicate that they should not be used at this point in time. As the implementations are rolled out, we will gradually revert them back to their intended type.

Should we add that as context to the property description?

Regarding `dialedEntity`, it isn't being used now, but I was thinking I'd keep the "schema" consistent with `originalCaller` by adding the corresponding `dialedEntityInfo` property. That way, both the caller info and the callee (dialed entity) info would have the same naming convention.

packages/teams-js/test/public/meeting.spec.ts
```diff
@@ -2260,4 +2266,26 @@ describe('meeting', () => {
       });
     });
   });
+  describe('utility', () => {
```
TrevorJoelHarris, line 2269: Please split these into separate tests (one test for each invalid case, and then a single test for all of the valid ones together is probably okay) This will make it more obvious what is wrong if something breaks.

TrevorJoelHarris (changes requested):
Looks good, just one last request to split EmailAddress and its test into separate files

packages/teams-js/src/public/meeting.ts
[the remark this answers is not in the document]
MSJohnnyLee: Done!

packages/teams-js/test/public/meeting.spec.ts
MSJohnnyLee: Gotcha - I split it into two separate files

packages/teams-js/src/public/emailAddress.ts
```diff
@@ -0,0 +1,17 @@
+import { validateEmailAddress } from '../internal/emailAddressValidation';
+
+/**
+ * Represents a validated email.
+ *
+ * @hidden
+ * Hide from docs.
+ */
```
TrevorJoelHarris, line 8:
```suggestion
/**
 * Represents a validated email.
 */
```

TrevorJoelHarris: I don't think there's any reason we need to hide this from the docs.

TrevorJoelHarris (approved):

apps/teams-test-app/e2e-test-data/meeting.json
```diff
@@ -44,6 +44,14 @@
       "expectedAlertValue": "getMeetingDetails called with shouldGetVerboseDetails: true",
       "expectedTestAppValue": "{\"details\":{\"scheduledStartTime\":\"testStartTime\",\"joinUrl\":\"testJoinUrl\",\"type\":\"oneOnOneCall\",\"originalCaller\":\"testCallerId\",\"dialedEntity\":\"testDnis\",\"trackingId\":\"testTrackingId\"},\"conversation\":{\"id\":\"testConversationId\"},\"organizer\":{\"id\":\"testOrganizerId\",\"tenantId\":\"testTenantId\"}}"
     },
+    {
+      "title": "getMeetingDetailsVerbose API Call - Success",
+      "version": ">2.28.0",
+      "type": "callResponse",
```
KangxuanYe, line 50: I am thinking of that for multi host-sdks testing purpose, I will request for adding `hostSdkVersion` as well for new cases.

KangxuanYe (changes requested):

apps/teams-test-app/e2e-test-data/meeting.json
MSJohnnyLee: Added, thanks

KangxuanYe (approved):

TrevorJoelHarris (approved):

PR #2493, opened 2024-09-03
Fix bug where eslint configuration could not resolve tsconfig in VSCode

juanscr:
This adds a configuration to the workspace vscode settings to make vscode recognize the proper eslint configuration file and the relative working directory for each sub-package in the monorepo.

### Main changes in the PR:

1. Add working directories settings through workspace settings in vscode

## Validation

I opened the files locally and created apparent error lines that eslint should point out and it work flawlessly. As this change is done through the vscode setting, this does not impact any other outside script other than the editor experience.

### Unit Tests added:

Not needed as it related to developer experience instead of a new feature or bug.

### End-to-end tests added:

No

### Change file added:

No

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| image |  image |

Notice the error in the editor disappeared.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,11 +2,11 @@
   "cSpell.words": ["adal", "ipados", "teamspace", "uninitialize", "xvfb"],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "eslint.workingDirectories": [
-    "./packages/teams-js/",
-    "./apps/teams-perf-test-app",
     "./apps/ssr-test-app/",
+    "./apps/teams-perf-test-app",
     "./apps/teams-test-app/",
-    "./apps/typed-dependency-tester/"
+    "./apps/typed-dependency-tester/",
+    "./packages/teams-js/"
   ],
   "editor.formatOnSave": true,
   "editor.insertSpaces": true,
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -1,6 +1,13 @@
 {
   "cSpell.words": ["adal", "ipados", "teamspace", "uninitialize", "xvfb"],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "eslint.workingDirectories": [
+    "./packages/teams-js/",
+    "./apps/teams-perf-test-app",
+    "./apps/ssr-test-app/",
+    "./apps/teams-test-app/",
+    "./apps/typed-dependency-tester/"
```
AE-MS, line 9: Can you alphabetize these? Thanks!

AE-MS:
Thanks for fixing this!

.vscode/settings.json
juanscr: Yes! Give me sec

AE-MS (changes requested):
:clock1:

TrevorJoelHarris (approved):

PR #2499, opened 2024-09-05
Update messaging for measuring message delay between App and Hub

juanscr:
## Description

This pull request has two purposes:
1. Fix the way the message request is timestamped by using a monotonic timer instead of the Date as it could be subject to user clock skew.
2. Add a mechanism for app developers to access Host to App message delay to analyze how much the message travel time impacts the latency of each function call or initialization.

For the second purpose, with this PR, an app developer can register a function that receives the message delays per function call as soon as a response is received from the host so they can log it or store it in their own telemetry. As app developers do not have information about the internal message IDs, the handler gets passed information about the callback that is resolving so they can approximately pinpoint which call had that message delay.

**Note**: it was decided to not expose message IDs as it is a really big architectural change in the codebase and the value of exposing them to app developers would be really small in comparison to the change.

### Main changes in the PR:
1. Change the timestamp on all message requests from `Date.now()` to use `performance.now() + performance.timeOrigin`.
2. Add a function to register a handler that will be called each time a response is received from the host with information about the message delay.

### Validation performed:

1. Ran all the unit tests and passed them successfully.
2. Execute the build and see that all test apps properly build.
3. Create a test app with a registered handler and log the values that were being received.

### Unit Tests added:
Yes

### End-to-end tests added:
No

### Change file added:

Yes

### Data
This is the response message delay for get context:
image
image
image

This is the request message delay for one-way traffic such as change theme: (this one shows more longer timestamps)
image
image
image

This is the request message delay for onw-way traffic `fullScreenChanged`:
image
image
image


juanscr:
@microsoft-github-policy-service agree company="Microsoft"

packages/teams-js/src/internal/utils.ts
```diff
@@ -528,3 +528,17 @@ export function validateUuid(id: string | undefined | null): void {
...
+ */
+const supportsPerformanceTimers = !!performance && 'now' in performance;
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ * @returns current timestamp in milliseconds
+ */
+export function getCurrentTimestamp(): number {
+  return supportsPerformanceTimers ? performance.now() + performance.timeOrigin : new Date().getTime();
```
ndangudubiyyam, line 543: Please add timestamp or date as a return field to support backwards compatibility

packages/teams-js/src/public/app.ts
```diff
@@ -886,6 +898,18 @@ export namespace app {
...
+  /**
+   * Registers a function for handling data of hub to app message delay.
+   *
+   * @remarks
+   * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
+   *
+   * @param handler - The handler to invoke when the metrics are available on each function response.
+   */
+  export function registerHubToAppPerformanceMetricsHandler(handler: HubToAppPerformanceMetricsHandler): void {
+    Handlers.registerHubToAppPerformanceMetricsHandler(handler);
```
maglims, line 910: Do you need "ensureInitialized(runtime)" as well like other handlers?

packages/teams-js/src/internal/interfaces.ts
```diff
@@ -50,3 +50,15 @@ export interface DOMMessageEvent {
   func: string;
   args?: any[];
 }
+
+/**
+ * @hidden
+ * Meant for providing information related to certain callback context.
```
maglims, line 56: Will this be only used in telemetry? If so, might use a name that reflecting it is telemetry related?

CallbackInformation sounds a bit generic.

packages/teams-js/src/internal/communication.ts
```diff
@@ -706,6 +711,19 @@ function handleParentMessage(evt: DOMMessageEvent): void {
     if (callbackId) {
       const callback = CommunicationPrivate.callbacks.get(callbackId);
       logger('Received a response from parent for message %i', callbackId);
+
+      // Send performance metrics information of message delay
```
maglims, line 715: nit: not sure if there is any guidance on how the code is structured in teamsjs.. personally it feels this code can be extracted in a telemetry helper function to keep the communication.ts file less lengthy.

packages/teams-js/src/internal/communication.ts
```diff
@@ -706,6 +711,19 @@ function handleParentMessage(evt: DOMMessageEvent): void {
...
+      const callbackInformation = CommunicationPrivate.callbackInformation.get(callbackId);
+      if (callbackInformation && message.timestamp) {
+        handleHubToAppPerformanceMetrics({
+          actionName: callbackInformation.name,
+          messageDelay: getCurrentTimestamp() - message.timestamp,
```
maglims, line 720: I haven't taken a look at hub sdk side's of code, but just reading codes here, it looks this messageDelay measures between "when app sdk receiving response" and "when hub sdk sends over the response". That sides makes sense. 

What about the other side? i.e, messageDelay between "when app sdk sends the request to hub sdk" and "when hub sdk receives the message request from app sdk"? I do see `messageWasCreatedAt`(the time that app sdk sends the request to hub sdk), do we also need to know "the time when hub sdk receives the message request" to calculate the other side of the latency?

apps/teams-test-app/src/components/AppAPIs.tsx
```diff
@@ -64,6 +64,16 @@ const OpenLink = (): ReactElement =>
     defaultInput: '"https://teams.microsoft.com/l/call/0/0?users=testUser1,testUser2&withVideo=true&source=test"',
   });
 
+const RegisterHubToAppPerformanceMetricsHandler = (): ReactElement =>
+  ApiWithoutInput({
+    name: 'registerHubToAppPerformanceMetricsHandler',
+    title: 'Register Hub to App performance metrics handler',
```
erinha, line 70: Can we use "Host" instead of "Hub" everywhere in this PR please?

packages/teams-js/src/public/app.ts
juanscr: In this instance the initialization is not necessarily a precondition for registering the handler, as it doesn't need to communicate with the host for registering the callback.

packages/teams-js/src/internal/communication.ts
juanscr: I agree with this, I was also worried about clustering a lot the communication layer with small stuff related to telemetry. Will add a utility function instead thanks!

packages/teams-js/src/internal/interfaces.ts
juanscr: Yeah, I agree that is best to separate it to its own telemetry naming convention. Will correct it.

packages/teams-js/src/internal/utils.ts
sumathin-lang, line 543: on what basis is this feature flag set?
Since you are using a completely new field in the messageObject for monotonic timestamp, why is this flag needed?

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -50,6 +56,60 @@ export const enum ApiVersionNumber {
...
+   */
+  public static storeCallbackInformation(messageUUID: MessageUUID, callbackInformation: CallbackInformation): void {
+    HostToAppMessageDelayTelemetry.callbackInformation.set(messageUUID, callbackInformation);
+  }
+
+  /**
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  public static clearMessages(): void {
```
sumathin-lang, line 82: when is clear invoked?

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -50,6 +56,60 @@ export const enum ApiVersionNumber {
...
+   * @param message The response from the host.
+   * @param logger A logger for logging any possible error.
+   */
+  public static telemetryHostToAppPerformanceMetrics(
+    callbackID: MessageUUID,
+    message: MessageResponse,
+    logger: debug.Debugger,
+  ): void {
+    const callbackInformation = HostToAppMessageDelayTelemetry.callbackInformation.get(callbackID);
+    if (callbackInformation && message.timestamp) {
```
sumathin-lang, line 101: this addresses two way request/response API.
What about one-way API like theme changed notification from Host to App. How would that be addressed?

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -50,6 +56,60 @@ export const enum ApiVersionNumber {
...
+      handleHostToAppPerformanceMetrics({
+        actionName: callbackInformation.name,
+        messageDelay: getCurrentTimestamp() - message.timestamp,
```
sumathin-lang, line 104: from you local testing, can you please add some data points for values recorded both for two-way and one-way messages

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -46,6 +48,7 @@ export interface SerializedMessageResponse {
   uuidAsString?: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   args?: any[];
+  timestamp?: number;
```
sumathin-lang, line 51: why wouldn't this have monotonicTimestamp?

packages/teams-js/src/internal/messageObjects.ts
```diff
@@ -58,6 +61,7 @@ export interface MessageResponse {
   uuid?: MessageUUID;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   args?: any[];
+  timestamp?: number;
```
sumathin-lang, line 64: same as above

packages/teams-js/test/utils.ts
```diff
@@ -39,11 +39,14 @@ export class Utils {
...
   public topWindow: Window;
 
+  public respondWithTimestamp: boolean;
+
   private onMessageSent: null | ((messageRequest: MessageRequest) => void) = null;
 
   public constructor() {
     this.messages = [];
     this.childMessages = [];
+    this.respondWithTimestamp = false;
```
sumathin-lang, line 49: who sets this to true?

sumathin-lang (changes requested):

packages/teams-js/src/internal/telemetry.ts
juanscr: I agree! We can do this after the registered handler invocation. Maybe not clear all the messages but at least the one that was already handled by the function.

packages/teams-js/src/internal/utils.ts
juanscr: This flag is needed in case the app is running in an environment that `performance` timers are not yet available. Like old versions of chrome, internet explorer and stuff like that... Mostly it shouldn't present any issues but is just a safeguard for edge cases.

packages/teams-js/src/internal/messageObjects.ts
juanscr: This is only for the response that the Hosts sends to Teams JS. For requests, as Teams JS was already using the `timestamp` field we needed to create a new one for backwards compatibiltiy. For responses, we can use `timestamp` instead as it is a completely new field...

If you find it more homogeneous, we can still call it `monotonicTimestamp` for the response to comply with the field name in the request. But it wouldn't be necessary.

packages/teams-js/test/utils.ts
juanscr: I set is as default by `false` to not add any possible issue with current tests that already ran. On the tests I made for testing the handler I set it to `true` to see if the handler gets called.

packages/teams-js/src/internal/telemetry.ts
juanscr: Yes, you're right, I actually didn't know this feature existed until you pointed it out. Thank you!

After digging around in the codebase, I created a commit to handle the one way API cases: https://github.com/OfficeDev/microsoft-teams-library-js/pull/2499/commits/7b23b3f11f146fb9d8fca94b89a443e44f3dd734

packages/teams-js/src/internal/messageObjects.ts
sumathin-lang: yes, pls use the same name to minimize confusion

ndangudubiyyam (commented):
Looks good to me

packages/teams-js/src/internal/handlers.ts
```diff
@@ -182,6 +183,27 @@ export function handleThemeChange(theme: string): void {
   }
 }
 
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function registerHostToAppPerformanceMetricsHandler(
```
ndangudubiyyam, line 190: nit: Do you think we can give the app devs the option to unregister?

ndangudubiyyam (approved):

packages/teams-js/src/internal/handlers.ts
juanscr: Good question. I wouldn't think there is a need for unregistering for now. Normally apps want to recollect the data or none at all, which they wouldn't register the handler in the first place.

packages/teams-js/src/internal/communication.ts
```diff
@@ -9,6 +9,7 @@ import { latestRuntimeApiVersion } from '../public/runtime';
 import { version } from '../public/version';
 import { GlobalVars } from './globalVars';
 import { callHandler } from './handlers';
+import HostToAppMessageDelayTelemetry from './hostToAppTelemetry';
```
maglims, line 12: Noticing that this is defined as default export and it looks there are not so many existing default export.. any reasoning behind it?

packages/teams-js/src/internal/hostToAppTelemetry.ts
```diff
@@ -0,0 +1,90 @@
...
+    const callbackInformation = HostToAppMessageDelayTelemetry.callbackInformation.get(callbackID);
+    if (!callbackInformation || !message.monotonicTimestamp) {
+      logger(
+        'Unable to send performance metrics for callback %s with arguments %o',
+        callbackID.toString(),
+        message.args,
+      );
+      return;
+    }
+    handleHostToAppPerformanceMetrics({
```
maglims, line 83: Is this all the properties that we will ask apps to  log? Not sure if I am mssing something but I don't see we are sending any ids(messageId or correlationId)? Don't they need that to be able to corelate the data?

packages/teams-js/src/internal/hostToAppTelemetry.ts
```diff
@@ -0,0 +1,90 @@
...
+      actionName: callbackInformation.name,
+      messageDelay: getCurrentTimestamp() - message.monotonicTimestamp,
+      messageWasCreatedAt: callbackInformation.calledAt,
```
maglims, line 86: noticing that here we are asssigning messageWasCreatedAt with callbackInformation.calledAt, while in handleOneWayPerformanceMetrics, we are messageWasCreatedAt = message.monotonicTimestamp;

What is the diff between these two methods: handlePerformanceMetrics and handleOneWayPerformanceMetrics?

packages/teams-js/src/internal/hostToAppTelemetry.ts
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public static handleOneWayPerformanceMetrics(message: MessageRequest, logger: Debugger): void {
+    const timestamp = message.monotonicTimestamp;
+    if (!timestamp) {
+      logger('Unable to send performance metrics for event %s', message.func);
+      return;
+    }
+    handleHostToAppPerformanceMetrics({
+      actionName: message.func,
+      messageDelay: getCurrentTimestamp() - timestamp,
```
maglims, line 59: is there anyway we can quickly check on the data? does console.log print work?

packages/teams-js/src/internal/communication.ts
juanscr: Not any particular reason. This module will only use this class and I don't think we will export additional things. So that's why I defaulted to use the default export

packages/teams-js/src/internal/hostToAppTelemetry.ts
juanscr: Yes! So, in a normal callback, the original request was created when teams-js sent the message through post message. On the other hand, in one-way functions the request was created when the message was created to be passed.

I do agree the name is confusing, maybe is way more appropriate `requestStartedAt`, do you agree?

packages/teams-js/src/internal/hostToAppTelemetry.ts
juanscr: Just added data for one-way and normal callbacks so you can see 😄

packages/teams-js/src/internal/hostToAppTelemetry.ts
juanscr: We are currently not sending messageIds as they do not necessarily have access to them prior to this, so it would be useless information for the app. This handler is mostly designed for apps to also have valuable information and exposing the internal message ids we use is unnecessary for now.

packages/teams-js/src/internal/communication.ts
```diff
@@ -807,6 +812,7 @@ function handleIncomingMessageFromParent(evt: DOMMessageEvent): void {
     if (callbackId) {
       const callback = CommunicationPrivate.callbacks.get(callbackId);
       logger('Received a response from parent for message %s', callbackId.toString());
+      HostToAppMessageDelayTelemetry.handlePerformanceMetrics(callbackId, message, logger);
```
maglims, line 815: From the screenshot you have attached, one observation is that the themeChange messageDelay(from 27-39s) is much longer than getContext(in the range of 2.3s-9s). Not sure if it reflects the reality that themeChange MessageDelay is slower than getContext, or it includes some random variation which is common issue with manual testing

Also, If we look at the handlePerformanceMetrics and handleOneWayPerformanceMetrics method, both of them use getCurrentTimestamp as the ending timestamp. Do you think it makes sense to record a common ending time in handleIncomingMessageFromParent(for example, after line 805), then pass that endtime to handlePerformanceMetrics and handleOneWayPerformanceMetrics to calculate the MessageDelay later?

maglims:
Just FYI, I think @TrevorJoelHarris  has checked in a code refactoring how communication works. You might have to rebase on that and make changes accordingly. @juanscr

packages/teams-js/src/internal/communication.ts
juanscr: The timestamp is in milliseconds hehe... So, the difference is not as big as it seems. It could be a lot of reasons of why this occurs, but it would be good to see the data in a production environment to be totally sure.

I find it reasonable to use the same ending time, the difference tho would be of some small milliseconds at most.

juanscr:
> Just FYI, I think @TrevorJoelHarris has checked in a code refactoring how communication works. You might have to rebase on that and make changes accordingly. @juanscr

Yeah, I thought there was going to be an issue with Trevor's changes. But everything is still compliant with this PR, the sync is made tho. Later down the line when the `sendMessageToParentAsync` function gets removed some changes will need to be done but for now everything keeps working as usual 😄

packages/teams-js/src/internal/communication.ts
juanscr: Pushed in latest commit 😄

packages/teams-js/src/internal/hostToAppTelemetry.ts
juanscr: Fixed in commit https://github.com/OfficeDev/microsoft-teams-library-js/pull/2499/commits/54820b8e0f5a017f0a54ef103594cf4a47ac4639

ndangudubiyyam (approved):

maglims (approved):

ndangudubiyyam (dismissed):

maglims (dismissed):

packages/teams-js/src/internal/communication.ts
```diff
@@ -1090,6 +1098,7 @@ function createMessageRequest(
     uuid: messageUuid,
     func: func,
     timestamp: Date.now(),
+    monotonicTimestamp: getCurrentTimestamp(),
```
sumathin-lang, line 1101: as i mentioned in the other comment, this should not take on "date.now" value

sumathin-lang (changes requested):

packages/teams-js/src/internal/communication.ts
juanscr: Good catch. Just pushed the change to never mix up both 😄

packages/teams-js/src/internal/hostToAppTelemetry.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @internal
+   * Limited to Microsoft-internal use
+   *
+   * Executes telemetry actions related to host to app performance metrics.
+   * @param callbackId The message id for the request.
+   * @param message The response from the host.
+   * @param logger The logger in case an error occurs.
+   * @param endTime The ending time for calculating the elapsed time
+   */
+  public static handlePerformanceMetrics(
```
sumathin-lang, line 74: @juanscr what is this measuring again - is this measuring E2E time for a request like "getContext" or just the time for getContext response to travel from Host to App. I expect it to be the latter

sumathin-lang (changes requested):

packages/teams-js/src/internal/communication.ts
sumathin-lang, line 1101: can you please confirm one-way messages from App to Hub also go through this function?

sumathin-lang (changes requested):

packages/teams-js/src/internal/communication.ts
juanscr: Yes, I can confirm. Normally one-way app to host functions use `sendMessageToParent` which uses `sendMessageToParentHelper` which uses `createMessageRequest`. Here is an example with `notifySuccess`:
image

packages/teams-js/src/internal/communication.ts
```diff
@@ -518,9 +520,12 @@ function sendMessageToParentHelper(
   args: any[] | undefined,
 ): MessageRequestWithRequiredProperties {
   const logger = sendMessageToParentHelperLogger;
-
   const targetWindow = Communication.parentWindow;
   const request = createMessageRequest(apiVersionTag, actionName, args);
+  HostToAppMessageDelayTelemetry.storeCallbackInformation(request.uuid, {
+    name: actionName,
+    calledAt: request.timestamp,
```
sumathin-lang, line 527: @juanscr isnt this supposed to be monotonicTimestamp

sumathin-lang (changes requested):

ndangudubiyyam (approved):

packages/teams-js/src/internal/communication.ts
juanscr: Hey @sumathin-lang, not necessarily. This information is merely contextual around when the callback happened and isn't used for any calculation. I selected the date instead of the `monotonicTimestamp` as it should be more proper for a global timestamp as it does not possess issues as not ticking while sleeping and so forth (https://github.com/w3c/hr-time/issues/115#issuecomment-1172985601). I can change it to `monotonicTimestamp` if you want but it's a minute detail.

sumathin-lang: @juanscr I would suggest using the same "units" for messageDelay and calledAt. Even if the App wants to use this "calledat" to measure some kind of delta,  perf.now is a better choice compared to date,now. But i will let the teams-js team make this call

sumathin-lang (approved):

PR #2502, opened 2024-09-06
Add TJS version in default log

KangxuanYe:

## Description

Adding TJS version into default log so that developers will know what version of TJS is using clearly from console.

### DIFF
--- a/packages/teams-js/src/public/app.ts
+++ b/packages/teams-js/src/public/app.ts
@@ -747,8 +747,7 @@ export namespace app {
     // This will always be the current script because browsers load and execute scripts in order.
     // Whenever a script is executing for the first time it will be the last script in this array.
     const currentScriptSrc = scripts && scripts[scripts.length - 1] && scripts[scripts.length - 1].src;
-    const scriptUsageWarning =
-      'Today, teamsjs can only be used from a single script or you may see undefined behavior. This log line is used to help detect cases where teamsjs is loaded multiple times -- it is always written. The presence of the log itself does not indicate a multi-load situation, but multiples of these log lines will. If you would like to use teamjs from more than one script at the same time, please open an issue at https://github.com/OfficeDev/microsoft-teams-library-js/issues';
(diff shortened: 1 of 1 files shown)


AE-MS:
Thank you for proactively making this change! It will be very helpful!

packages/teams-js/src/public/app.ts
```diff
@@ -747,8 +747,7 @@ export namespace app {
...
     const currentScriptSrc = scripts && scripts[scripts.length - 1] && scripts[scripts.length - 1].src;
-    const scriptUsageWarning =
-      'Today, teamsjs can only be used from a single script or you may see undefined behavior. This log line is used to help detect cases where teamsjs is loaded multiple times -- it is always written. The presence of the log itself does not indicate a multi-load situation, but multiples of these log lines will. If you would like to use teamjs from more than one script at the same time, please open an issue at https://github.com/OfficeDev/microsoft-teams-library-js/issues';
+    const scriptUsageWarning = `Today, teamsjs version: ${version} can only be used from a single script or you may see undefined behavior. This log line is used to help detect cases where teamsjs is loaded multiple times -- it is always written. The presence of the log itself does not indicate a multi-load situation, but multiples of these log lines will. If you would like to use teamjs from more than one script at the same time, please open an issue at https://github.com/OfficeDev/microsoft-teams-library-js/issues`;
```
AE-MS, line 750: >version: ${version}  [](http://example.com/codeflow?start=47&length=20)

Rather than putting this in the script usage warning string, can we just log this version as a separate log line? It can even be written from in this function.

I just worry that if we put it in the warning string, then when we remove the warning string someday (we will eventually maybe support multiple instances) the string will be removed and the version log will go with it.

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/app.ts
KangxuanYe: @AE-MS, I started another thread of conversation with Saurabh and Trevor. Saurabh has more concern from PM side and I will keep discussing with them about what to do. So far, we summarized the problem as difficulty of discoverability and had some of ideas but not sure which path we will go to.

PR #2503, opened 2024-09-06
Add button to test app to reload itself

AE-MS:
## Description

There have recently been some bugs exposed when apps reload themselves internally (as opposed to the host reloading the whole app container). This change adds a button to the test app which will trigger an internal reload so that tests can be written which trigger this behavior.

### Validation performed:

- Manually verified that when I click the button it reloads
- Verified that reloading in this fashion on older host SDKs exposes some of the recently fixed bugs

### DIFF
--- a/apps/teams-test-app/src/pages/TestApp.tsx
+++ b/apps/teams-test-app/src/pages/TestApp.tsx
@@ -71,6 +71,7 @@ export const TestApp: React.FC = () => {
 
   return (
     <>
+      <button onClick={() => window.location.reload()}>Reload This App</button>
       <div className="App-container">
         <AppAPIs />
         <AppInitializationAPIs />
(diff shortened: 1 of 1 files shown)


TrevorJoelHarris (dismissed):

apps/teams-test-app/src/pages/TestApp.tsx
```diff
@@ -71,6 +71,7 @@ export const TestApp: React.FC = () => {
 
   return (
     <>
+      <button onClick={() => window.location.reload()}>Reload This App</button>
```
TrevorJoelHarris, line 74: If you want to call this from tests it should have an id so it can be retrieved from the DOM easily

AE-MS: Good to know; will make that change now.

TrevorJoelHarris (approved):

noahdarveau-MSFT (approved):

PR #2513, opened 2024-09-16
noahdarveau/Rollup changes for treeshaking

noahdarveau-MSFT:

## Description

### Main changes in the PR:

1. Added Rollup to the repository which will be used to build a treeshakable version of Teams-JS once the treeshaking code changes go in.
2. Changed the pkg.json file to have the `main` field point to the minified cdn bundle and added the `module` field to point to the esm bundle.
3. Updated the Build-Test-Publish pipeline so the CDN package does not included the Rollup-built code.
4. Added a basic tree shaking test app that was used to verify tree shaking is working in teams-js. Note that there are currently no automated tests to check if tree shaking is working and verification must be done manually by examining the bundled package built with the test app.

### Validation performed:

1. All unit and e2e tests continue to pass.


tools/yaml-templates/build-test-publish.yml
```diff
@@ -118,10 +118,10 @@ steps:
 
   - task: CopyFiles@2
     inputs:
-      Contents: |
-        packages\teams-js\dist\**\?(*.js|*.ts|*.map)
+      sourceFolder: 'packages\teams-js\dist'
```
jadahiya-MSFT, line 121: would this include the packages in `umd`?

noahdarveau-MSFT: Yes, it will include both the umd and esm packages. We still need to publish both the umd package to npm to backwards support any apps that are not configured to use es modules

packages/teams-js/rollup.config.mjs
```diff
@@ -0,0 +1,58 @@
...
+import terser from '@rollup/plugin-terser';
+import typescript from '@rollup/plugin-typescript';
+import nodePolyfills from 'rollup-plugin-polyfill-node';
+
+import version from './package.json' assert { type: 'json' };
+
+export default {
+  input: './src/index.ts',
+  output: {
+    dir: 'dist/esm',
```
jadahiya-MSFT, line 16: How is `esm` being used when it comes to publishing?

jadahiya-MSFT: Your above answer, answers this

noahdarveau-MSFT: Can you please elaborate a bit more? I'm not sure I fully understand what you're asking

tools/yaml-templates/build-test-publish.yml
```diff
@@ -118,10 +118,10 @@ steps:
...
+      contents: '**\?(*.js|*.ts|*.map)'
       TargetFolder: '$(Build.ArtifactStagingDirectory)\NPMFeed\dist'
-      flattenFolders: true
+      flattenFolders: false
```
jadahiya-MSFT, line 124: I assume this is to for backcompat in case a developer doesn't have ESModules configured

packages/teams-js/rollup.config.mjs
jadahiya-MSFT: Oh I was just curious how the `esm` bits were being pulled in and published and your answer about publishing both `esm` and `umd` when publishing to the NPM Feed answered that for me

tools/yaml-templates/build-test-publish.yml
noahdarveau-MSFT: Yes, that is one reason, but mostly it is because the rollup-built package keeps all of the files separate, so if we flatten all of the files to one folder level, our published artifact would be a huge mess of every file in teams-js condensed to a single folder level. Keeping the folder structure helps keep the artifact neat.

jadahiya-MSFT: Got it, thanks for the explanation

apps/tree-shaking-test-app/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import { barCode } from 'testlibraryfortreeshaking';
+//import { PrintTestMessage } from 'testlibraryfortreeshaking/app';
+import { geoLocation } from '@microsoft/teams-js';
+//import { app } from '@microsoft/teams-js';
```
jadahiya-MSFT, line 4: Do we want to keep these commented out? If so can we have an added comment explaining why

noahdarveau-MSFT: Oh, yeah, I forgot about that file. The whole tree-shaking test app isn't actually doing anything. I was just using it for testing, and wanted to check it in for use later. I'll go ahead and remove the comments so it looks nicer.

jadahiya-MSFT (approved):

packages/teams-js/src/private/index.ts
```diff
@@ -18,7 +18,7 @@ export {
   registerUserSettingsChangeHandler,
   openFilePreview,
 } from './privateAPIs';
-export { conversations } from './conversations';
+export { conversations, OpenConversationRequest, ConversationResponse } from './conversations';
```
TrevorJoelHarris, line 21: What is this change needed for?

TrevorJoelHarris (commented):
Had one question about a change to the `private\index.ts` file

packages/teams-js/src/private/index.ts
noahdarveau-MSFT: While developing, I was running into a bug where OpenConversationRequest and ConversationResponse was not able to be imported by the teams test app, so I had to export it directly to resolve the issue. I just tried removing the change and it seems to be working ok, so I've reverted the change.

jadahiya-MSFT (approved):

TrevorJoelHarris (approved):

PR #2514, opened 2024-09-17
Improve message logging from communications module

AE-MS:
## Description

From having read a lot of logs recently, the messages emitted from the communications module were often confusing or hard to follow. This change fixes the issues seen:
- Message ids logged were originally confusing and inconsistent. This was because the code was only logging some of the ids some of the time (e.g. only the uuids and not the legacy ids). Moreover, when it was logging the UUIDs, it was using the `%i` format specifier, which I _think_ tells it to format it as an integer (although [the official debug docs](https://www.npmjs.com/package/debug) don't list `%i` as an officially supported formatter). This change logs both IDs all the time and it now correctly logs the UUIDs as strings.
- This change now explicitly logs information about when the library is relaying messages to and from child frames
- I renamed some of the functions to be clearer about their purpose (e.g. whether they were handling incoming messages from a place or outgoing message to a place)

There is no functionality change in this, it's all renames and logging (which I guess is a sort of functionality change, but you know what I mean).

### Validation performed:

Read a lot of logs as I was implementing and testing this, including nested app scenarios, to ensure that they all looked right and could be followed.

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -767,7 +767,7 @@ function handleIncomingMessageFromParent(evt: DOMMessageEvent): void {
   } else if ('func' in evt.data && typeof evt.data.func === 'string') {
     // Delegate the request to the proper handler
     const message = evt.data as MessageRequest;
-    logger('Received a message from parent, id: %s, action: "%s"', getMessageIdsAsLogString(message), message.func);
+    logger('Received a message from parent %s, action: "%s"', getMessageIdsAsLogString(message), message.func);
     callHandler(message.func, message.args);
   } else {
     logger('Received an unknown message: %O', evt);
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -87,8 +87,7 @@ export function initializeCommunication(
   apiVersionTag: string,
 ): Promise<InitializeResponse> {
   // Listen for messages post to our window
-  CommunicationPrivate.messageListener = async (evt: DOMMessageEvent): Promise<void> =>
-    await processIncomingMessage(evt);
+  CommunicationPrivate.messageListener = (evt: DOMMessageEvent): Promise<void> => processIncomingMessage(evt);
 
   // If we are in an iframe, our parent window is the one hosting us (i.e., window.parent); otherwise,
   // it's the window that opened us (i.e., window.opener)
@@ -1066,6 +1065,7 @@ function createMessageEvent(func: string, args?: any[]): MessageRequest {
 function getMessageIdsAsLogString(
   message:
     | SerializedMessageRequest
+    | SerializedMessageResponse
     | MessageRequestWithRequiredProperties
     | MessageRequest
     | MessageResponse
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -369,7 +369,7 @@ export function sendNestedAuthRequestToTopWindow(message: string): NestedAppAuth
   const targetWindow = Communication.topWindow;
   const request = createNestedAppAuthRequest(message);
 
-  logger('Message %s (legacy id: %i) information: %o', request.uuid.toString(), request.id, {
+  logger('Message %s information: %o', getMessageIdsAsLogString(request), {
     actionName: request.func,
   });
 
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/internal/communication.ts
```diff
@@ -368,7 +369,9 @@ export function sendNestedAuthRequestToTopWindow(message: string): NestedAppAuth
   const targetWindow = Communication.topWindow;
   const request = createNestedAppAuthRequest(message);
 
-  logger('Message %i information: %o', request.uuid, { actionName: request.func });
+  logger('Message %s (legacy id: %i) information: %o', request.uuid.toString(), request.id, {
```
TrevorJoelHarris, line 372: How would you feel about some sort of helper function that takes in request and spits out `%s (legacy id: %i)`? That would be nice both from a reduce typing point of view but also from a "what if we change what we think of as the request's id in the future" point of view

TrevorJoelHarris (changes requested):
One request for a helper function that prints message identifiers

packages/teams-js/src/internal/communication.ts
AE-MS: Done

TrevorJoelHarris (dismissed):

packages/teams-js/src/internal/communication.ts
```diff
@@ -87,7 +87,8 @@ export function initializeCommunication(
   apiVersionTag: string,
 ): Promise<InitializeResponse> {
   // Listen for messages post to our window
-  CommunicationPrivate.messageListener = async (evt: DOMMessageEvent): Promise<void> => await processMessage(evt);
+  CommunicationPrivate.messageListener = async (evt: DOMMessageEvent): Promise<void> =>
+    await processIncomingMessage(evt);
```
erinha, line 91: Should this be in curly braces now that it's on a new line?

packages/teams-js/src/internal/communication.ts
```diff
@@ -368,7 +369,9 @@ export function sendNestedAuthRequestToTopWindow(message: string): NestedAppAuth
   const targetWindow = Communication.topWindow;
   const request = createNestedAppAuthRequest(message);
 
-  logger('Message %i information: %o', request.uuid, { actionName: request.func });
+  logger('Message %s information: %o', getMessageIdsAsLogString(request), {
```
erinha, line 372: nit: `getMessageIdAsLogString` ? (Id singular vs plural)

packages/teams-js/src/internal/communication.ts
```diff
@@ -389,7 +392,11 @@ function sendRequestToTargetWindowHelper(
 
   if (GlobalVars.isFramelessWindow) {
     if (Communication.currentWindow && Communication.currentWindow.nativeInterface) {
-      logger(`Sending message %i to ${targetWindowName} via framelessPostMessage interface`, request.uuidAsString);
+      logger(
+        'Sending message %s to %s via framelessPostMessage interface',
+        getMessageIdsAsLogString(request),
+        targetWindowName,
```
erinha, line 398: Curious if you had a reason for passing input as arguments vs using an interpolated string?

packages/teams-js/src/internal/communication.ts
```diff
@@ -1020,3 +1062,20 @@ function createMessageEvent(func: string, args?: any[]): MessageRequest {
     args: args || [],
   };
 }
+
+function getMessageIdsAsLogString(
+  message:
```
erinha, line 1067: Intentional that it does not include `SerializedMessageResponse` as one of the possible types?

erinha (changes requested):
Left a couple questions

packages/teams-js/src/internal/communication.ts
AE-MS: 1. Consistency
2. I don't know if debug works this way but at least _some_ logger packages don't evaluate the arguments if they know that the log statement is being filtered, for performance. E.g., in this example, if `logger` knew that logging was disabled, it wouldn't call the `getMessageIdsAsLogString` function to get that value when it knew it didn't need it. Come to think of it now though, that would really be up to typescript where it evaluates all function arguments before actually calling the `logger` function, so this reason is completely irrelevant in TypeScript/JavaScript. So it's really just #1 then.

packages/teams-js/src/internal/communication.ts
AE-MS: It's sort of intentional in that no one is calling this function with an object of that type but it's easy enough to add it, so I did.

packages/teams-js/src/internal/communication.ts
AE-MS: This was an automatic change that was done by prettier after the mechanical refactor that was the rename of `processMessage` but I can add the curlies now.

AE-MS: After much discussion I actually removed the `async` and `await` since they are not necessary. `processMessage` returns a `Promise<void>` and so now it just keeps returning that up through the caller, enabling the caller to `await` it at their leisure, rather than enforcing the `await` to happen right here in this function.

erinha (approved):
:shipit:

TrevorJoelHarris (approved):

PR #2519, opened 2024-09-23
Unify some of the appid code, improve script tag detection and add validation

AE-MS:
## Description

The codebase currently has some duplication in a few areas:
- Multiple different ways of accepting, storing, and validation of app ids
- Multiple (different) ways of identifying the presence of script tags

### Main changes in the PR:

1. Unify all App Id logic around the `AppId` type -- note: the input parameters for appids unfortunately cannot be changed from `string` to `AppId` at this time since that would be a API-breaking change. 
2. Remove incomplete implementation of HTML entity decoding (since it was missing a lot of other entities) and remove need for such functionality by just looking specifically for encoded script tags (script tag searching was the only function using the entity decoding helper)
3. All related tests that were using a regex to test for a specific substring were updated to be case-insensitive, since they are intentionally not looking to exact match the error message (the exact error message is not part of the API contract)
4. Explicitly added unit tests for validate the `hasScriptTags` function
5. Updated unit tests for functions that were doing app id validation to no longer test on the exact error message (it's not part of the API contract) and only test for the presence of a descriptive word (in a case-insensitive way). Relatedly: updated these tests to more directly use Jest-provided affordances for testing async functions that throw errors.
6. Updated unit test usage of domains not controlled by Microsoft to use Microsoft-owned domains

### Validation performed:

1. All existing and new unit tests were run and passed
2. Private helper functions that previously didn't have specific unit tests now do

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

### Next/remaining steps:
There are still a few places that use the old `validateId` function but that I did not change to use `AppId`. This is because the data being validated were not actually an app ids -- they are fields like "title id" and "oauthConfigId". It's not clear why these other ID types were using the app id validation code so I can't tell whether to unify them now. I have questions out to the owners about this and will follow up in a subsequent change.

### DIFF
--- a/change/@microsoft-teams-js-81b620f2-8af1-4450-89b0-20600bac843e.json
+++ b/change/@microsoft-teams-js-81b620f2-8af1-4450-89b0-20600bac843e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Unified app id validation, removed incomplete HTML element transformation, and refreshed how script tags are identified in strings",
+  "packageName": "@microsoft/teams-js",
+  "email": "36546967+AE-MS@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-81b620f2-8af1-4450-89b0-20600bac843e.json
+++ b/change/@microsoft-teams-js-81b620f2-8af1-4450-89b0-20600bac843e.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Unified app id validation, removed incomplete HTML element transformation, and refreshed how script tags are identified in strings",
+  "comment": "Updated internal app id validation",
   "packageName": "@microsoft/teams-js",
   "email": "36546967+AE-MS@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/packages/teams-js/test/internal/appIdValidation.spec.ts
+++ b/packages/teams-js/test/internal/appIdValidation.spec.ts
@@ -1,6 +1,5 @@
 import {
   doesStringContainNonPrintableCharacters,
-  doesStringContainScriptTags,
   isStringWithinAppIdLengthLimits,
   maximumValidAppIdLength,
   minimumValidAppIdLength,
--- a/packages/teams-js/test/internal/appIdValidation.spec.ts
+++ b/packages/teams-js/test/internal/appIdValidation.spec.ts
@@ -58,37 +58,6 @@ describe('isStringWithinAppIdLengthLimits', () => {
   });
 });
 
-describe('doesStringContainScriptTags', () => {
-  test('should return true for strings containing script tags', () => {
-    expect(doesStringContainScriptTags('<script>alert("Hello")</script>')).toBe(true);
-    expect(doesStringContainScriptTags('<script src="example.js"></script>')).toBe(true);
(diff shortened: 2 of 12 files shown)


packages/teams-js/test/internal/appIdValidation.spec.ts
```diff
@@ -58,37 +57,6 @@ describe('isStringWithinAppIdLengthLimits', () => {
...
-
-  test('should return false for empty string', () => {
-    expect(doesStringContainScriptTags('')).toBe(false);
-  });
-
-  test('should return true for strings with multiple script tags', () => {
-    expect(doesStringContainScriptTags('<script>alert("Hello")</script><script>console.log("test")</script>')).toBe(
-      true,
-    );
-  });
```
AE-MS, line 89: These tests have been removed in favor of the tests for the other script tag identification function

change/@microsoft-teams-js-81b620f2-8af1-4450-89b0-20600bac843e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Unified app id validation, removed incomplete HTML element transformation, and refreshed how script tags are identified in strings",
```
TrevorJoelHarris, line 3: I go back and forth on patch notes like this, because for most external devs they will read a description like this and have no idea what it means for them (which is, I assume, almost nothing).

Is there a punchy sentence that describes which parts of this change an external dev would care about? Maybe something like "Updated internal app id validation"

AE-MS: I'm not actually sure any external dev will care about this change at all since it's purely internal infrastructural updates. They shouldn't observe any API or behavior difference at all. I guess the library will be a little smaller now (less duplicated code) and it will be harder to hack, theoretically. Not sure if that makes for good patch notes...

AE-MS: I can, of course, change it to "updated internal app id validation" but I'm not sure an external dev would care much about that either. :)

TrevorJoelHarris (approved):

packages/teams-js/test/private/externalAppAuthentication.spec.ts
```diff
@@ -886,11 +880,11 @@ describe('externalAppAuthentication', () => {
           await utils.initializeWithContext(frameContext);
           utils.setRuntimeConfig({ apiVersion: 2, supports: { externalAppAuthentication: {} } });
           // eslint-disable-next-line @microsoft/sdl/no-insecure-url
-          const invalidsingInUrl = new URL('http://pishingsite.com');
+          const invalidStingInUrl = new URL('http://adatum.com');
```
erinha, line 883: >invalidStingInUrl [](http://example.com/codeflow?start=16&length=17)

invalidStringInUrl

packages/teams-js/test/internal/utils.spec.ts
```diff
@@ -367,6 +368,105 @@ describe('utils', () => {
     });
```
erinha, line 368: >}); [](http://example.com/codeflow?start=4&length=3)

Unless I'm missing something I think we don't have a unit test that validates a valid appId can be constructed successfully.

PR #2524, opened 2024-09-23
Added the teamsjs instance identifier and timestamp to every log line

AE-MS:
## Description

Added the teamsjs instance identifier and timestamp to every log line. This will help us debug issues where multiple teamsjs instances are being used (which is an unsupported scenario).

### Validation performed:

Ensured everything looked right in the logs:
image
...even in the multi-instance scenario:
image

### Change file added:

Yes

### DIFF
--- a/packages/teams-js/src/internal/telemetry.ts
+++ b/packages/teams-js/src/internal/telemetry.ts
@@ -8,7 +8,7 @@ export const teamsJsInstanceIdentifier = new UUID();
 // Every log statement will get prepended with the teamsJsInstanceIdentifier and a timestamp
 const origFormatArgs = registerLogger.formatArgs;
 registerLogger.formatArgs = function (args) {
-  args[0] = `${teamsJsInstanceIdentifier.toString()} (${new Date().toISOString()}): ${args[0]}`;
+  args[0] = `(${new Date().toISOString()}): ${args[0]} [${teamsJsInstanceIdentifier.toString()}]`;
   origFormatArgs.call(this, args);
 };
 
--- a/packages/teams-js/src/internal/telemetry.ts
+++ b/packages/teams-js/src/internal/telemetry.ts
@@ -1,5 +1,14 @@
 import { debug as registerLogger, Debugger } from 'debug';
 
+import { UUID } from './uuidObject';
+
+export const teamsJsInstanceIdentifier = new UUID();
+const origFormatArgs = registerLogger.formatArgs;
+registerLogger.formatArgs = function (args) {
+  args[0] = `${teamsJsInstanceIdentifier.toString()} (${new Date().toISOString()}): ${args[0]}`;
+  origFormatArgs.call(this, args);
+};
+
 const topLevelLogger = registerLogger('teamsJs');
 
 /**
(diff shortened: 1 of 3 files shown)


packages/teams-js/src/internal/telemetry.ts
```diff
@@ -1,5 +1,17 @@
...
 
+import { UUID } from './uuidObject';
+
+// Each teamsjs instance gets a unique identifier that will be prepended to every log statement
+export const teamsJsInstanceIdentifier = new UUID();
+
+// Every log statement will get prepended with the teamsJsInstanceIdentifier and a timestamp
+const origFormatArgs = registerLogger.formatArgs;
+registerLogger.formatArgs = function (args) {
+  args[0] = `${teamsJsInstanceIdentifier.toString()} (${new Date().toISOString()}): ${args[0]}`;
```
TrevorJoelHarris, line 11:
```suggestion
  args[0] = `(${new Date().toISOString()}): ${args[0]} (${teamsJsInstanceIdentifier.toString()})`;
```

This is really just a stylistic thing, but given how long uuids are I'd personally prefer them at the end to improve readability. Fine to leave as-is if you like it better, just my two cents.

TrevorJoelHarris (approved):

TrevorJoelHarris (dismissed):

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -1,5 +1,17 @@
 import { debug as registerLogger, Debugger } from 'debug';
 
+import { UUID } from './uuidObject';
+
+// Each teamsjs instance gets a unique identifier that will be prepended to every log statement
+export const teamsJsInstanceIdentifier = new UUID();
+
+// Every log statement will get prepended with the teamsJsInstanceIdentifier and a timestamp
+const origFormatArgs = registerLogger.formatArgs;
```
erinha, line 9:
```suggestion
const originalFormatArgs = registerLogger.formatArgs;
```
I wasn't sure if it was "origin" or "original" (though one could argue that the code makes it obvious). Suggestion for clarity

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -1,5 +1,17 @@
...
+registerLogger.formatArgs = function (args) {
+  args[0] = `(${new Date().toISOString()}): ${args[0]} [${teamsJsInstanceIdentifier.toString()}]`;
```
erinha, line 11: How often is this called? Just the first time the library initializes, or every time we log? Wondering if we should be concerned about the perf of `new Date()`. For some reason I have a memory that it is not performant? But I could definitely be misremembering.

erinha: I did some light searching online and couldn't find anything to confirm my memory. Maybe it's fine?

erinha (dismissed):
Left suggestions for potential added clarity and asked about perf. Overall, I love it!

packages/teams-js/src/internal/telemetry.ts
AE-MS: Per in-person discussion, logs are not written frequently enough for us to worry about performance right now, but it will be something we keep an eye on.

TrevorJoelHarris (approved):

erinha (approved):
🚀

PR #2526, opened 2024-09-25
Pipeline Fix

noahdarveau-MSFT:
After merging in the Rollup changes, it istreating all .js files in the packages/teams-js directory as esm files. The prepBetaRelease.js is written in commonjs, so it needs to be renamed to prepBetaRelease.cjs to be treated as such and unblock the pipeline since whenever it runs it will try to treat the file as an esm file and wil fail.

This did not break on the PR because the prepBetaRelease step in the pipeline has a condition so that it only runs for pipelines running on the main branch, not PR branches. Since the step was skipped, the file was not run and therefore did not error out.


jadahiya-MSFT (approved):

AE-MS (dismissed):

AE-MS:
Can you add details on how the pipelines didn't break on your PR but only broke after it was merged?

TrevorJoelHarris (approved):

jadahiya-MSFT (approved):

Ella-ly (approved):

PR #2529, opened 2024-09-26
[DRAFT/DO NOT REVIEW] Kx/debugging discoverability poc

KangxuanYe:

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/public/logger.ts
```diff
@@ -0,0 +1,21 @@
+import { GlobalVars } from '../internal/globalVars';
+
+export namespace logger {
+  /**
+   * todo: doc
+   * @returns bool
+   */
+  export function turnOnConsoleLog(): boolean {
```
TrevorJoelHarris, line 8: probably no reason you need to return these booleans

packages/teams-js/src/internal/telemetry.ts
```diff
@@ -12,7 +13,59 @@ registerLogger.formatArgs = function (args) {
   originalFormatArgsFunction.call(this, args);
 };
 
-const topLevelLogger = registerLogger('teamsJs');
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const createDebuggerFunction = (namespace: string): Debugger => {
```
TrevorJoelHarris, line 17: This is cool and it will probably need a good dose of explanatory comments

packages/teams-js/src/public/logger.ts
KangxuanYe: Updated. Change `boolean` to `void` now.

AE-MS:
This looks neat -- can you say more about the use-case for it? Is it something we expect customers might find easier to enable than the current set of logging steps? Or is it something you expect to be able to use when debugging issues? Etc.?

KangxuanYe:
> This looks neat -- can you say more about the use-case for it? Is it something we expect customers might find easier to enable than the current set of logging steps? Or is it something you expect to be able to use when debugging issues? Etc.?

Based on our discussion with Saurabh, I think the use-case would be customers find the bug/error/failure in their code from Teams-Js library and they wish to know what happened. Then they can read the documentation of `Logger` capability and by calling APIs in their codebase, they can get all/partial of client debugging loggings depending on where they turn on the client debugging loggings. In this scenario, I think customers will have easier steps for debugging and have more control on which part they would like to focus on. They can use this method for debugging but for my work, I would still request them to follow the steps we have right now to make sure they send ALL of client loggings to me from the beginning of initialization.

packages/teams-js/src/internal/telemetry.ts
KangxuanYe: Could you please take a look on comments? Basically I just explained the code in detail.

KangxuanYe:
Need to turn it back to draft again because after rebasing on main, the console log complains that it cannot find timer, which is used by `console.timeLog`.

PR #2535, opened 2024-10-01
Revert the changes of moving common data models to interface file for externalAppCardAction(ForCEA) files

MengyiGong:

## Description

> Revert the changes of moving common data models to interface file for externalAppCardAction(ForCEA) APIs to avoid the breaking changes.

### DIFF
--- a/change/@microsoft-teams-js-4fe5ca49-b46c-4a6b-b580-bd6035df67e6.json
+++ b/change/@microsoft-teams-js-4fe5ca49-b46c-4a6b-b580-bd6035df67e6.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Keep the common data models in the `externalAppCardAction` API file to share with the `externalAppCardActionForCEA` API.",
+  "packageName": "@microsoft/teams-js",
+  "email": "maggiegong@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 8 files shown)


AE-MS:
If you are reverting the change that was meant to avoid breaking changes, does that mean that we're now intentionally making a breaking change? Why did we want to avoid it before but now we are okay with it?

change/@microsoft-teams-js-4fe5ca49-b46c-4a6b-b580-bd6035df67e6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Keep the common data models in the `externalAppCardAction` API file to share with the `externalAppCardActionForCEA` API.",
```
AE-MS, line 3: >Keep the common data models in the `externalAppCardAction` API file to share with the `externalAppCardActionForCEA` API. [](http://example.com/codeflow?start=14&length=120)

Consider:
"Unified common data models for external card actions into `externalAppCardActions` namespace"

AE-MS (changes requested):
:clock1:

MengyiGong:
> If you are reverting the change that was meant to avoid breaking changes, does that mean that we're now intentionally making a breaking change? Why did we want to avoid it before but now we are okay with it?
From the changes in the last PR, we did not realize we are making a "breaking change" (all the data models and functions are marked with `@beta`, so it is not breaking). It was not intentional change.
Confirmed with @lakhveerk and @erinha please correct me if I'm wrong:
We believe MChat have not consumed the beta version from last week. We are fixing it before they consume our new version.

change/@microsoft-teams-js-4fe5ca49-b46c-4a6b-b580-bd6035df67e6.json
MengyiGong: Updated with your suggestion. Thanks!

lakhveerk (approved):

AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

PR #2544, opened 2024-10-04
added a listener for testing host to test manual initialization on test app

KangxuanYe:

## Description

This is a change on test app that it will not go through fixed initialization flow we hard-coded here if one of url parameter is `testInit`. Instead, it will have a listener listening to our testing host, waiting for message to call corresponding APIs to complete the initialization flow.

### Main changes in the PR:

1. Add a message listener for testing host to call initialization relevant APIs
2. Set a url parameter so that hard-coded initialization flow will not be handled.


apps/teams-test-app/src/App.tsx
```diff
@@ -16,29 +16,61 @@ const urlParams = new URLSearchParams(window.location.search);
...
-    app.notifyAppLoaded();
-    app.notifySuccess();
+window.addEventListener('message', handleMessageFromMockedHost);
+
+function handleMessageFromMockedHost(msg: MessageEvent): void {
+  if (!msg || !msg.data) {
+    console.log('Unrecognized message format received by app, message being ignored. Message: %o', msg);
+    return;
+  }
+  // Ensure the message is from a trusted origin
```
TrevorJoelHarris, line 54: How would you feel about writing a log statement for the cases where we are handling the message? It might be useful to be able to tell from the console logs that `app.initialize` was called because of this handler as opposed to some other reason.

Maybe the simplest way to do it would be something like 
```suggestion
console.log(`Received message from test host: ${msg.data}`);

  // Ensure the message is from a trusted origin
```

apps/teams-test-app/src/App.tsx
```diff
@@ -16,29 +16,61 @@ const urlParams = new URLSearchParams(window.location.search);
...
-) {
-  console.info('Not calling appInitialization because part of App Initialization Test run');
-} else {
-  if (isTestBackCompat()) {
-    appInitialization.notifyAppLoaded();
-    appInitialization.notifySuccess();
-  } else {
-    app.notifyAppLoaded();
-    app.notifySuccess();
+window.addEventListener('message', handleMessageFromMockedHost);
```
TrevorJoelHarris, line 47: Should this listener only be set up when testInit is present on the url paremeters and set to true?

apps/teams-test-app/src/App.tsx
```diff
@@ -16,29 +16,61 @@ const urlParams = new URLSearchParams(window.location.search);
...
+      break;
+    case 'app.notifyExpectedFailure':
+      app.notifyExpectedFailure({ reason: app.ExpectedFailureReason.Other, message: 'Failed on test app on purpose' });
+      break;
+    case 'app.notifyAppLoaded':
+      app.notifyAppLoaded();
+      break;
+    // Add more cases for other API calls as needed
+    default:
+      console.log('Unknown API call or response:', msg);
```
TrevorJoelHarris, line 73: Should this be console.warn instead of console.log?

apps/teams-test-app/src/App.tsx
```diff
@@ -16,29 +16,61 @@ const urlParams = new URLSearchParams(window.location.search);
...
-    appInitialization.notifyAppLoaded();
-    appInitialization.notifySuccess();
-  } else {
-    app.notifyAppLoaded();
-    app.notifySuccess();
+window.addEventListener('message', handleMessageFromMockedHost);
+
+function handleMessageFromMockedHost(msg: MessageEvent): void {
+  if (!msg || !msg.data) {
+    console.log('Unrecognized message format received by app, message being ignored. Message: %o', msg);
```
TrevorJoelHarris, line 51: Should this be console.warn instead of console.log?

TrevorJoelHarris (changes requested):
Left some thoughts about when to set up the listener

apps/teams-test-app/src/App.tsx
```diff
@@ -16,29 +16,61 @@ const urlParams = new URLSearchParams(window.location.search);
 const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? urlParams.get('origins') : '';
 const validMessageOrigins: string[] | undefined = getOriginsParam ? getOriginsParam.split(',') : undefined;
 
-// This is added for custom initialization when app can be initialized based upon a trigger/click.
-if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
-  if (isTestBackCompat()) {
-    initialize(undefined, validMessageOrigins);
+if (!urlParams.has('testInit') || !urlParams.get('testInit')) {
```
AE-MS, line 19: >!urlParams.has('testInit') || !urlParams.get('testInit') [](http://example.com/codeflow?start=4&length=56)

Isn't checking `urlParams.get('testInit') === null` sufficient?

AE-MS, line 19: >testInit [](http://example.com/codeflow?start=20&length=8)

Is it possible to name this parameter in a way that is more description of what it does? I.e. "testInit" doesn't exactly say what behavior is being triggered. Perhaps the parameter could be named something like "doNotCallInitializeFunction" ? I'm not actually sure what this parameter is doing... except telling it to ignore the customInit parameter? Could we just not send the customInit parameter in situations where you don't want the behavior it triggers?

apps/teams-test-app/src/App.tsx
KangxuanYe: Yes, I will put it into the situation when init is skipped.

apps/teams-test-app/src/App.tsx
KangxuanYe: Logically it seems like it doesn't work if there is a url link looks as xxxxxx?testInit=false. And `urlParams.get('testInit') === null` has the same behavior if `testInit=true` or `testInit=false`

KangxuanYe: There is a possibility I will remove this name because `customInit` and `appInitializationTest` url parameters do the same thing as it does.

apps/teams-test-app/src/App.tsx
```diff
@@ -16,29 +16,61 @@ const urlParams = new URLSearchParams(window.location.search);
...
+  // This is added for custom initialization when app can be initialized based upon a trigger/click.
+  if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
```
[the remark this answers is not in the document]
KangxuanYe: Synced offline. Basically, `customInit` and `appInitializationTest` are all used to skip the hardcode initialization flow.

apps/teams-test-app/src/App.tsx
```diff
@@ -42,6 +43,35 @@ if (
   }
 }
 
+function handleMessageFromMockedHost(msg: MessageEvent): void {
+  if (!msg || !msg.data) {
```
AE-MS, line 47: >!msg [](http://example.com/codeflow?start=6&length=4)

Is it necessary to perform this check? When would the event object ever be `undefined` or `null`? The prototype for `addEventListener` doesn't indicate that the event should ever be `undefined`.

apps/teams-test-app/src/App.tsx
```diff
@@ -42,6 +43,35 @@ if (
...
+    console.warn('Unrecognized message format received by app, message being ignored. Message: %o', msg);
+    return;
+  }
+  console.log(`Received message from test host: ${msg.data}`);
+  // Ensure the message is from a trusted origin
```
AE-MS, line 52: Is this comment accurate? I'm not sure I see how the code below this is ensuring the message is coming from a trusted origin.

AE-MS (changes requested):
:clock1:

apps/teams-test-app/src/App.tsx
TrevorJoelHarris: Maybe something like

```suggestion
  // Handle messages that are correctly formatted and for func values we recognize
```

apps/teams-test-app/src/App.tsx
KangxuanYe: My initial thought is that, If I don't add this check, it is to say I need to make sure, or we have an implicit agreement, that our testing host WILL NOT send empty message. It would not happen right now but we cannot make sure we don't want to do it in the future. I don't want this function to rely on what testing host does/sends so I added this check.

But now I think it would be better if it fails directly if people trying to pass in empty msg instead of filter it out by if-else condition check so that developer might not get aware of himself sending empty msg. Will remove it.

apps/teams-test-app/src/App.tsx
KangxuanYe: I think that comment is for user to make sure the message is from a trusted origin rather than saying the functionality of that piece of code lol.

Thank you @TrevorJoelHarris , I have taken your suggestion.

AE-MS (approved):
:shipit:

TrevorJoelHarris (approved):

PR #2551, opened 2024-10-09
New "correct" way to communicate with the host

TrevorJoelHarris:

## Description

It was getting increasingly confusing in this codebase to send messages to the host and receive responses. We have 6 or 7 different functions that did some variation of that job (sendAndUnwrap! sendAndHandleStatusAndReason!, sendMessageToParentAsync! etc!), but they all behaved a little differently which was causing a lot of confusion when someone was trying to do things the "right" way. Furthermore, none of those functions did anything to encourage callers to serialize their arguments before passing them to the host, deserialize return values from the host, validate that responses from the host were correctly shaped, or handle errors in a uniform way.

No longer!

There are now _only two_ functions that should be used to communicate with the host:

`communication.callFunctionInHost` and `communication.callFunctionInHostAndHandleResponse`.

`communication.callFunctionInHost` should be used when you don't need to receive any data back from the host after calling your function. This function expects to receive a response from the host that is either an error or undefined. If the function receives an error from the host it throws it as an uppercase-E-Error, and if it receives anything else that isn't undefined it also throws an Error. The return type of this function is `Promise<void>`.

`communication.callFunctionInHostAndHandleResponse` should be used when you **do** need to receive data back from the host after calling your function. This function expects to receive a response from the host that is either an error or the return type specified by the first generic type. If the function receives an error from the host it throws it as an uppercase-E-Error, and if it receives anything else that isn't expected return type with the correct values it also throws an Error. The return type of this function is `Promise<DeserializedExpectedReturnType>` (generic type 2).

This function takes in a `ResponseHandler` which is in charge of validating that the non-error response looks valid and also in charge of deserializing the response value into whatever is supposed to be handed back to the caller (if necessary).

**BOTH** new functions also put some additional guard rails in place around passing args up to the host. The arg array can now consist of:

A simple type: `type SimpleType = string | number | boolean | null | undefined | SimpleType[];`

OR

An object that implements `ISerializable`, which means that it has a `serialize` function on it that returns `unknown`.

If you try to pass an arg that is an object that does not implement `ISerializable` you will get a transpilation error. This should make it easier for people to figure how what they actually need to serialize before passing it to the host.

Finally, I have marked all of the old functions for host communication in communication.ts as `@deprecated`. You can still call them, but if you do I will come be sad in your general vicinity. Please use the new ones only.

### Main changes in the PR:

1. Changed what functions can be used to communicate with the host
2. Made hostEntity.ts use these new functions
3. Made externalAppAuthenticationForCEA.ts use these functions

### Validation performed:

1. Ran all unit tests

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes


packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +259,43 @@ export function sendMessageToParentAsync<T>(
   });
 }
 
+export async function sendMessage<ReceivedFromHost, DeserializedFromHost>(
+  apiVersionTag: string,
+  actionName: string,
+  responseHandler: ResponseHandler<ReceivedFromHost, DeserializedFromHost>,
+  args: ArgsForHost | undefined = undefined,
```
AE-MS, line 266: I wonder if this could be simplified in a way that doesn't need the custom type (I know, look at me, the custom type advocate saying maybe we don't need a custom type). Ignoring all the other goodness you have here and just focusing on the input arguments, what do you think about something like this?

```
interface ISerializable {
    serialize(): unknown;
}

function isSerializable(arg: unknown): arg is ISerializable {
    return (arg as ISerializable).serialize !== undefined && typeof (arg as ISerializable).serialize === 'function';
}

function sendMessage(actionName: string, args: unknown[]) {
    const outgoingArgs: unknown[] = [];
    for (const arg of args) {
        outgoingArgs.push(isSerializable(arg) ? arg.serialize() : arg);
    }
}
```

Here are a few thoughts I have about it:
- Doesn't require the custom type so theoretically the usage should be simpler
- If a particular action doesn't require any arguments, the caller can just pass [] for the `args` parameter
- Uses what I believe to be more common naming
- Doesn't use `any` so nothing has to be suppressed
- Doesn't need a list of all simple types -- just interrogates each type and says "do you need to be serialized? If so, let me do that. Otherwise, I'll just pass through."

One thing that does seem different but I can't tell if it's good or bad. In `ArgsForHost` you don't list `object` as a simple type, which I think implies that if you want to pass an `object` as an argument, it must implement `ISerializable`. Is that true? If so, is that a necessary restriction? E.g. there are many objects that don't actually need to be serialized because they are just key value pairs of POD (plain old data), e.g.,
```
{
    name: string,
    age: number,
    isStudent: boolean
}
```

To be clear, I'm not saying to get rid of all the other stuff you have (respond handlers, error checker, etc.) -- I haven't looked at those yet, I was just thinking so far on how argument input works.

TrevorJoelHarris: It's an interesting idea. I omitted object from that list very intentionally because I want people who are passing more complicated objects to have to _think_ about whether or not their object requires serialization. I like a lot of the suggestions you have above, but I worry a little bit that if we go that direction fully we'll still be making it pretty easy for people to be passing in things that *should* be serialized without realizing that they *need to be* serialized.

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+ * @param functionName The function name to call in the host.
+ * @param responseHandler When the host responds, this handler will validate and deserialize the response.
+ * @param args A collection of data to pass to the host. This data must be an array of either simple types or objects that implement {@link ISerializable}.
+ * @param errorChecker This optional property can be used to override the default ErrorChecking this function uses to decide whether to throw the host response as a new Error. Specify this is your function needs to do any logic verifying that the object received is an error that goes beyond the logic found in {@link isSdkError}.
+ *
+ * @returns The response received from the host after deserialization.
+ *
+ * @throws An Error containing the SdkError information ({@link SdkError.errorCode} and {@link SdkError.message}) if the host returns an SdkError, or an Error if the response from the host is an unexpected format.
+ */
+export async function callFunctionInHostAndHandleResponse<ReceivedFromHost, DeserializedFromHost>(
```
jadahiya-MSFT, line 301: Do you think that calling this `callFunctionInHost` might be confusing if the expectation is for this to `sendFunctionToHost`?

change/@microsoft-teams-js-19c8476a-7a54-4481-b8cb-811fcd3bab65.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
jadahiya-MSFT, line 2: because we are marking thecurrent `send` functions as deprecated I believe this should be a `minor` rather than a `patch`

change/@microsoft-teams-js-19c8476a-7a54-4481-b8cb-811fcd3bab65.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Created unified way for capability functions to communicate with hosts and deprecated old functions used for this purpose. This change is internal to the library and should not affect any callers.",
```
jadahiya-MSFT, line 3: Can we add all the functions that are now deprecated with reference to their replacements in this comment for anyone that looks through the changelog of the release this change will go out with?

packages/teams-js/src/public/serializable.interface.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Interface for objects that can be serialized and passed to the host
+ */
+export interface ISerializable {
```
jadahiya-MSFT, line 4: This is a great change! Thank you for adding this!

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: I made some changes to move further in this direction while still maintaining the "any objects need to implement ISerializable or you get typescript errors". See what you think!

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: I think there's definitely going to be some initial confusion because I'm changing what this type of function is called; moving from `sendMessage*` to a `callFunction*` convention. My _hope_ is that with this new name change 99% of the team will never have to think about what's actually going on under the covers within the communication layer and they can instead think about what they are doing semantically.

_Technically_ what these functions are doing are sending a bunch of data to the host sdk using one of a few different transport mechanisms, (postMessage, callbacks).

_Semantically_ though, these functions are really calling a function on the host that may or may not return data.

I'm not attached to these specific names, but I *do* like Aaron's suggestion of changing the names to be things that are more semantically meaningful. I'm open to other names along those lines if you have ideas!

change/@microsoft-teams-js-19c8476a-7a54-4481-b8cb-811fcd3bab65.json
TrevorJoelHarris: I marked this as patch because these functions are internal to the library and shouldn't affect library consumers at all. Although, come to think of it this may need to be upgraded to minor if we keep the changes to the authCEA capability, so I'll leave this open until we decide on that

change/@microsoft-teams-js-19c8476a-7a54-4481-b8cb-811fcd3bab65.json
TrevorJoelHarris: This change shouldn't impact external callers, but until we decide if we're keeping the authenticationCEA part of the change I will leave this as a reminder to update the message about those changes if we keep them.

jadahiya-MSFT (approved):

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: A bit of an aside, do you think making these changes would be better placed in a separate PR instead and then to either keep the `ISerializable` interface as is or to remove it entirely from this PR?

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+  apiVersionTag: string,
```
AE-MS, line 302: I guess this is sort of nitpicky but it seems kind of weird that the first parameter to this function is an api version tag instead of, you know, the name of the function you want to call.

AE-MS: It seems like a sensible order would be: functionName, args, responseHandler, apiVersionTag, errorChecker, based on the way people might think about "calling" a function.

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
   });
 }
 
+type SimpleType = string | number | boolean | null | undefined | SimpleType[];
+
+function isSerializable(arg: unknown): arg is ISerializable {
```
AE-MS, line 269: >isSerializable [](http://example.com/codeflow?start=9&length=14)

I think this would more appropriately live next to the definition of `ISerializable` since its conceptually part of the type definition.

packages/teams-js/src/public/serializable.interface.ts
```diff
@@ -0,0 +1,9 @@
...
+  /**
+   * @returns A serializable representation of the object, used for passing objects to the host.
+   */
+  serialize(): unknown;
```
AE-MS, line 8: >unknown [](http://example.com/codeflow?start=15&length=7)

Should the return type of this be `string`? Or at least `SimpleType`

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+ * @param errorChecker This optional property can be used to override the default ErrorChecking this function uses to decide whether to throw the host response as a new Error. Specify this is your function needs to do any logic verifying that the object received is an error that goes beyond the logic found in {@link isSdkError}.
+ *
+ * @returns The response received from the host after deserialization.
+ *
+ * @throws An Error containing the SdkError information ({@link SdkError.errorCode} and {@link SdkError.message}) if the host returns an SdkError, or an Error if the response from the host is an unexpected format.
+ */
+export async function callFunctionInHostAndHandleResponse<ReceivedFromHost, DeserializedFromHost>(
+  apiVersionTag: string,
+  functionName: string,
+  responseHandler: ResponseHandler<ReceivedFromHost, DeserializedFromHost>,
```
AE-MS, line 304: >ReceivedFromHost [](http://example.com/codeflow?start=35&length=16)

Consider `SerializedReturnValueFromHostFunction` -- I know, it's a bit long, but it's also very clear

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+  args: (SimpleType | ISerializable)[],
+  errorChecker?: (response: unknown) => response is SdkError,
```
AE-MS, line 306: >errorChecker [](http://example.com/codeflow?start=2&length=12)

What's an example use case for this? Something about this strikes me as a little funny. It seems like this code should always be able to determine itself if the response object is an `SdkError` (e.g., we already have the `isSdkError` function).

I'm looking at the example in the `authenticateWithOauth` where it provides a custom checker that checks if the object is an `InvokeError` -- and then it _essentially_ lies by saying "I just identified this as an `InvokeError` but yeah, it's totally an `SdkError`"

I get that those two types are strikingly similar but they aren't actually the same. 

Let me think about why this seems funky to me.

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+): Promise<DeserializedFromHost> {
+  const argsSafeToTransfer = serializeItemArray(args);
```
AE-MS, line 308: >argsSafeToTransfer [](http://example.com/codeflow?start=8&length=18)

Internal variable so it doesn't matter that much but consider `serializedArguments` or `serializedArgumentArray`

packages/teams-js/src/internal/communication.ts
AE-MS: It seems like what you actually want this function to be is something like "If you aren't using `SdkError`, then this function's job is to return an error object containing an error code and a message field that is a transformation of your custom error type" -- which isn't the same thing as saying that this custom object IS an `SdkError`, because they aren't. At least in the case of `InvokeError` the interpretation of the `errorCode` field should be different.

Maybe you should create a new interface like `ICommonError` and this parameter should be `transformErrorToCommonError?: (response: unknown) => ICommonError` and then your code below would look like this:

```
  if ((transformErrorToCommonError&& transformErrorToCommonError(response)) || (!transformErrorToCommonError&& isSdkError(response))) {
    throw new Error(`${response.errorCode}, message: ${response.message ?? 'None'}`);
```

I know that in practice this may seem a bit silly because all the property names are the same but treating two different types as the same seems like it will eventually lead to trouble.

packages/teams-js/src/internal/responseHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+   * @returns True if the response is valid, false otherwise
+   */
+  public abstract validate(response: ReceivedFromHost): boolean;
+
+  /**
+   * This function converts the response from the host into a different format
+   * before returning it to the caller (if needed).
+   * @param response
+   */
+  public abstract deserialize(response: ReceivedFromHost): DeserializedFromHost;
```
AE-MS, line 22: I'm curious about the need to make these two different steps. Does `validate` returning `true` mean that `deserialize` is guaranteed to succeed? Said differently, does part of what `validate` does is ensure that the response data can be deserialized? Or can there still be potential problems deserializing even if the response is "valid?"

I'm thinking that you could combine these an only have a `deserialize` method, and part of that method also validates the response and fails if its invalid.

Generally I am for small-scoped single-purpose functions, not large combo functions. However, in this particular case, it feels like the exact value of splitting is unclear because both functions could error-out (possibly for different reasons?) and it's not clear how the caller should behave differently in an error for one vs the other. E.g., is a "valid" but "not deserializable" response authentically different such that we want to detect each condition separately? Or should we just have deserialize and the job of that function is to validate the data as it is deserializing?

packages/teams-js/src/internal/responseHandler.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * This class is used for validating and deserializing the response from the host.
+ *
+ * @typeParam ReceivedFromHost The type of the response received from the host
+ * @typeParam DeserializedFromHost The type of the response after deserialization
+ */
+export abstract class ResponseHandler<ReceivedFromHost, DeserializedFromHost> {
```
AE-MS, line 7: >ReceivedFromHost, DeserializedFromHost [](http://example.com/codeflow?start=38&length=38)

I left this in the communication.ts file but I now realize I should probably have put it here -- but I don't want to delete my original comment since Jay responded to it.

Consider naming these types something like `SerializedReturnValueFromHost` and `DeserializedReturnValueFromHost`

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: I'm not sure what you mean Jay, can you say more? The ISerializable interface is new in this change

packages/teams-js/src/internal/communication.ts
AE-MS, line 302: Should people going forward ever really need to provide this? Could we just use the function name string as the apiVersionTag for all new capabilities/functions? And then there's a legacy version of this that takes an apiVersionTag but only for those old back-compat functions that have multiple versions (e.g., startTask vs dialog.url.open)

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -133,6 +135,18 @@ export namespace externalAppAuthentication {
...
 
+  /**
+   * @hidden
+   * @internal
+   * Limited to Microsoft-internal use
+   */
+  export class SerializableActionExecuteInvokeRequest implements ISerializable {
+    public constructor(private invokeRequest: externalAppAuthentication.IActionExecuteInvokeRequest) {}
+    public serialize(): object | string {
+      return this.invokeRequest; // I wonder if the Record actually needs to be serialized. Might be a real bug?
```
lakhveerk, line 146: Record is consisting of only primitive types that are serialized internally in post message. This means we can pass the data without serializing them. I will put a PR out that will make sure the Record actually contains only primitive types. 

Automatic serialization would not happen on `symbol` type that is a primitive type. I am talking to 3pcxp to make sure that won't be sent as part of Record.

packages/teams-js/src/public/serializable.interface.ts
TrevorJoelHarris: We should talk about this. I think we could definitely require string here, but there's no problem with passing lowercase-o `object`s through postMessage as long as those objects only contain primitives or objects that only contain primitives.

If we require string here that will mean that anyone who wants to serialize an object will have to turn it into a string representation. For simple objects like AppId that is very straightforward, but for objects with a lot of properties it would be less straightforward.

What I have here was my attempt at a compromise between "forcing correctness" and "developer convenenience", but I can be swayed if you feel strongly.

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+    functionName,
+    argsSafeToTransfer,
+  );
+
+  if ((errorChecker && errorChecker(response)) || (!errorChecker && isSdkError(response))) {
+    throw new Error(`${response.errorCode}, message: ${response.message ?? 'None'}`);
+  } else if (!responseHandler.validate(response as ReceivedFromHost)) {
+    throw new Error(`${ErrorCode.INTERNAL_ERROR}, message: Invalid response from host`);
+  } else {
+    return responseHandler.deserialize(response as ReceivedFromHost);
```
AE-MS, line 320: >deserialize [](http://example.com/codeflow?start=27&length=11)

I have been thinking about this and it seems like there's some semantic asymmetry in how teamjs sends and receives data for this communication.

Is the intended design that the hub SDK can send a single data value to teamjs, or that it can send multiple data values?

The current design suggests that teamsjs is allowed to send multiple data values to the hub SDK (`args` is an array where each element can individually be a `SimpleType` or a type that serializes) but that the hub SDK can only send a single value to teamsjs, and that single value is deserialized as a whole. If you are in the hub SDK and want to send multiple values to teamsjs, you have to either send them as an object or send them in an array (which I guess is then wrapped in its own array?).

Such a design isn't necessarily too weird, if you think about how you would just regularly call a function, functions can have multiple input parameters and can only have a single return value. If you want to make a function return multiple values, you have to make the return type be either an object or an array or a tuple.

I wanted to make sure I understood the intended design, as to be observed by capability developers. Should they think about the hub SDK being able to send back multiple values, or only a single value (and if they want multiples they have to wrap them in an object or array)?

packages/teams-js/src/internal/communication.ts
jadahiya-MSFT: Aah I was looking over the code and while it is tied together, I think there could be one PR that was just focused on the introduction of the `ISerializable` interface and all the code that touched and another would be the deprecation of the `send` functions and addition of the `callFunctionInHost` ones. The reason I think it would be a good call is that the majority of the feedback on this PR appears to be about the `ISerializable` interface. So for now you can change this block of code to something like the following and then we can follow this PR up with another one that says introducing a new `ISerializable` object.

```javascript
export async function callFunctionInHost(
  apiVersionTag: string,
  functionName: string,
  args: any[],
  errorChecker?: (response: unknown) => response is SdkError,
): Promise<void> {
 //The following probably will need to be changed, but serializing the args is still important
  const argsSafeToTransfer = serializeItemArray(args);
  const [response] = await sendMessageToParentAsync<[SdkError]>(apiVersionTag, functionName, argsSafeToTransfer);

  if ((errorChecker && errorChecker(response)) || (!errorChecker && isSdkError(response))) {
    throw new Error(`${response.errorCode}, message: ${response.message ?? 'None'}`);
  } else if (response !== undefined) {
    // If we receive a response from the host that is not a recognized error type it is an invalid response
    throw new Error(`${ErrorCode.INTERNAL_ERROR}, message: Invalid response from host`);
  }
}
```

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: How about `SerializedReturnValueFromHost`, it removes the `Function` part like jay said and is more clear that what I had originally. I don't have a problem with long type names because autocomplete exists, but the `Function` suffix seems unnecessary to me given the context this is used in

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+    } else {
+      return item;
+    }
+  });
+}
+
+/**
+ * Call a function in the host and receive a response. If the host returns an {@link SdkError} instead of a normal response, this function will throw a new Error containing the SdkError's information
+ *
+ * @param apiVersionTag A unique tag used to identify the API version for telemetry purposes.
```
TrevorJoelHarris, line 291: Update documentation about this

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: Totally valid, and thinking about how to improve this led to a design I like a lot better. I changed this line to instead read:

`isResponseAReportableError?: (response: unknown) => response is { errorCode: number | string; message?: string },`

Essentially, "Is the thing you got back something that has an errorCode and a message on it in a format that you accept as valid?"

One of the really nice things about doing it this way is that people can pass their functions like `isInvokeError` in to this directly now instead of wrapping it in a awkward wrapper function that lies about it being an SdkError. Handshakes all around!

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: I think for now there's no good way to simplify this. We have too many cases where we actually do need the developer to pass in a unique value here so that our telemetry tells us which functions are being used (registerHandler for example). I will update the comment for this though to make it clearer what this actually needs to be set to (and why).

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: You're correct that there's asymmetry there, and it's by design for the reasons you're guessing. In order to make these functions as semantically simple as I can (because they are inherently hard to understand) I wanted to make them mimic how developers think of calling "normal" functions: pass the function a list of arguments, receive a single return value. That return value could be an object, a simple type, or an array of things, but whatever it is it's _just one thing_.

Can someone do something fancy and return a list of various return values from the host? Sure. Is it more complicated to understand? Yes. Does it actually help with anything? No!

packages/teams-js/src/public/serializable.interface.ts
TrevorJoelHarris: After some offline discussion, we decided to change the type to `string | object` to be as restrictive as possible. If someone really needs to turn an object into a boolean (?) in the future we can talk about it.

packages/teams-js/src/internal/responseHandler.ts
TrevorJoelHarris: Here was my thinking, but open to discussing more:

Separating these into two steps helps encourage consistency for the callers of our sdk functions. If we receive a payload from the host that is invalid, we will now return a consistent error to the function:

```typescript
 } else if (!responseHandler.validate(response as SerializedReturnValueFromHost)) {
    throw new Error(`${ErrorCode.INTERNAL_ERROR}, message: Invalid response from host`);
  } else {
```

My thinking was that we should *never* run into something that was unserializable if the caller marked it as valid: if a payload we receive from the host is "valid" one of the things that means is that we know we can successfully deserialize it into something meaningful and correct.

If we were to merge these steps into a single `validateAndDeserialize` step, we would need to depend on each developer that uses these functions to throw consistent errors when they receive invalid data. It would work, but it felt messier to me.

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: Talked with Jay some about this. We agreed that in practice it would have made sense to split this up into two PRs: one that just introduced the ISerializable interface and then had certain objects implement it (so reviewers could focus on that item) and then a second one that introduced the new communication functions and made use of them. Unfortunately, because this work started as an FHL project I did not do a great job of planning and sequencing the work into multiple checkins. Because people have already reviewed and read this, splitting it into multiple checkins would not be of much practical use at this point.

Still, good idea and that will teach me to cut corners even during FHL projects :D

packages/teams-js/src/internal/responseHandler.ts
TrevorJoelHarris: Talked with @AE-MS offline. We agreed that there probably is a way to merge these into one function call, something like `validateAndDeserialize` and then have that function return `string | object | 'InvalidResponse'` which the communication layer could then use to determine that an invalid response was received from the host so the communication code could then throw a consistent error for this scenario.

Ultimately, Aaron liked that approach better (but wasn't strongly opposed to keeping it as-is) and I liked the current approach better (but was willing to change if someone preferred the other approach by a lot).

To break the tie, I've decided to just keep this as-is and Aaron has agreed he's okay with that.

change/@microsoft-teams-js-f9ac3559-b7e4-4cdd-bff9-54aecd663d48.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Made `externalAppAuthenticationForCEA.ts` throw `Error` objects instead of `SdkErrors`",
+  "packageName": "@microsoft/teams-js",
+  "email": "trharris@microsoft.com",
+  "dependentChangeType": "patch"
```
jadahiya-MSFT, line 6: Should this part still be patch?

jadahiya-MSFT (approved):

change/@microsoft-teams-js-f9ac3559-b7e4-4cdd-bff9-54aecd663d48.json
TrevorJoelHarris: I believe this is correct, yes, it's just saying what a package that depends on this should bump it's version number by when it consumes a new version.

The funny (?) part is that I think this isn't really necessary because patch should be the default here? Never hurts to be specific I guess.

jadahiya-MSFT (approved):

packages/teams-js/src/internal/communication.ts
```diff
@@ -257,6 +264,90 @@ export function sendMessageToParentAsync<T>(
...
+ *
+ * @param functionName The function name to call in the host.
+ * @param args A collection of data to pass to the host. This data must be an array of either simple types or objects that implement {@link ISerializable}.
+ * @param responseHandler When the host responds, this handler will validate and deserialize the response.
+ * @param apiVersionTag A unique tag used to identify the API version for telemetry purposes. This should be set using {@link getApiVersionTag}, which should be passed a unique string identifying the function being called by the app developer as well as a version number that is incremented whenever meaningful changes are made to that function.
+ * @param isResponseAReportableError This optional property can be used to override the default ErrorChecking this function uses to decide whether to throw the host response as a new Error. Specify this is your function needs to do any logic verifying that the object received is an error that goes beyond the logic found in {@link isSdkError}.
```
AE-MS, line 286: >is [](http://example.com/codeflow?start=203&length=2)

if

AE-MS, line 286: >logic found in {@link isSdkError [](http://example.com/codeflow?start=310&length=32)

I'm not... actually sure if it matters to update this but this function isn't using `isSdkError` anymore. Writing this documentation to be 100% accurate may not be necessary, though.

packages/teams-js/src/internal/responseHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+   * before returning it to the caller (if needed).
+   * @param response
+   */
+  public abstract deserialize(response: SerializedReturnValueFromHost): DeserializedReturnValueFromHost;
+}
+
+/**
+ * This class is used for validating and deserializing boolean responses from the host.
+ */
+export class BooleanResponseHandler extends ResponseHandler<boolean, boolean> {
```
AE-MS, line 28: >BooleanResponseHandler [](http://example.com/codeflow?start=13&length=22)

Could you actually write this for all `SimpleType`s the same way?

```
export class SimpleResponseHandler<T extends SimpleType> extends ResponseHandler<T, T> {
  public validate(_response: T): boolean {
    return true;
  }

  public deserialize(response: T): T {
    return response;
  }
}
```

I guess maybe `string` types might have a custom validator but the others seem like maybe not.

packages/teams-js/src/internal/communication.ts
TrevorJoelHarris: I think everything I wrote in here is accurate: the host returns SdkError objects and we turn them into Errors that contain the SdkError information. Let me know if anything in there is confusing or you want more detail

packages/teams-js/src/internal/responseHandler.ts
TrevorJoelHarris: Yup, good call I hadn't thought of the `extends` thing that's neat

jadahiya-MSFT (dismissed):

change/@microsoft-teams-js-f9ac3559-b7e4-4cdd-bff9-54aecd663d48.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Made `externalAppAuthenticationForCEA.ts` throw `Error` objects instead of `SdkErrors`",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Made `externalAppAuthenticationForCEA.ts` throw `Error` objects instead of `SdkErrors`.",
```

AE-MS (approved):
:shipit:

AE-MS:
This is really great to see; I love that we have "a way" to do this communication now.

jadahiya-MSFT (approved):

PR #2554, opened 2024-10-10
Update `copilot.eligibility.getEligibilityInfo` to retrieve info from host

erinha:
## Description

Update `copilot.eligibility.getEligibilityInfo` to retrieve the eligibility info from the host if it was not provided in the `initialize` response. 

### Main changes in the PR:

1. `copilot.eligibility.getEligibilityInfo` is now an async function.
2. The function first checks if the eligibility info has been provided in the `initialize` response and returns that if so. If not, it makes a request to the host (parent window) to retrieve it.
3. If the `copilot.eligibility` capability is not supported, `getEligibilityInfo` now throws an `Error` instead of a string.

### Validation performed:

1. Added new unit tests and updated existing ones.
2. Updated the Test App to allow for the async call and new response info.
3. Manually validated with the corresponding host SDK change that the functionality works as expected.

### Unit Tests added:

Yes

### End-to-end tests added:

Yes

### Change file added:

Yes


change/@microsoft-teams-js-286cc515-1cef-4d54-9af6-d869e00d96c9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
lakhveerk, line 2: I think changing the function signature would be a breaking change for MChat.

packages/teams-js/src/private/copilot.ts
```diff
@@ -41,12 +48,53 @@ export namespace copilot {
...
+
+      // Return the eligibility information if it is already available
+      if (runtime.hostVersionsInfo?.appEligibilityInformation) {
+        return Promise.resolve(runtime.hostVersionsInfo!.appEligibilityInformation);
+      }
+
+      // Send message to host SDK to get eligibility information
+      return sendMessageToParentAsync<AppEligibilityInformation | SdkError>(
+        getApiVersionTag(copilotTelemetryVersionNumber, ApiName.Copilot_Eligibility_GetEligibilityInfo),
+        'copilot.eligibility.getEligibilityInfo',
```
lakhveerk, line 65: Can this be `ApiName.Copilot_Eligibility_GetEligibilityInfo` instead of a string

packages/teams-js/src/private/copilot.ts
```diff
@@ -41,12 +48,53 @@ export namespace copilot {
...
+      ).then((response: AppEligibilityInformation | SdkError) => {
+        if (isSdkError(response)) {
```
lakhveerk, line 67: In case, some error comes that is not of type sdkError, this function cast it as `AppEligibilityInformation` and send to the app. I think if we have a type-guard for the response as well that would be good

packages/teams-js/src/private/copilot.ts
```diff
@@ -41,12 +48,53 @@ export namespace copilot {
      *
      * @throws Error if {@linkcode app.initialize} has not successfully completed
      */
-    export function getEligibilityInfo(): AppEligibilityInformation {
+    export function getEligibilityInfo(): Promise<AppEligibilityInformation> {
```
lakhveerk, line 51: don't we need `export async getEligibilityInfo`

packages/teams-js/src/private/copilot.ts
```diff
@@ -41,12 +48,53 @@ export namespace copilot {
...
       ensureInitialized(runtime);
       if (!isSupported()) {
         throw errorNotSupportedOnPlatform;
       }
-      return runtime.hostVersionsInfo!.appEligibilityInformation!;
+
+      // Return the eligibility information if it is already available
+      if (runtime.hostVersionsInfo?.appEligibilityInformation) {
+        return Promise.resolve(runtime.hostVersionsInfo!.appEligibilityInformation);
```
lakhveerk, line 59: If we write async in the function header, the returned values will automatically be wrapped in promise. I think it can be simplified to  `return runtime.hostVersionsInfo!.appEligibilityInformation!`

change/@microsoft-teams-js-286cc515-1cef-4d54-9af6-d869e00d96c9.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Updated `copilot.eligibility.getEligibilityInfo` to get the eligibility info from the host if it is not already available.",
```
lakhveerk, line 3: API consumers might be more interested in the changing function definition from sync to async

change/@microsoft-teams-js-286cc515-1cef-4d54-9af6-d869e00d96c9.json
erinha: Because it's marked beta, we have some leeway. We will work with them to ensure they're expecting this.

lakhveerk: oh, since the API was marked @beta, it would still be a minor change :)

lakhveerk: haha it looks like we commented at the same time

packages/teams-js/src/private/copilot.ts
```diff
@@ -29,7 +32,11 @@ export namespace copilot {
      * @throws Error if {@linkcode app.initialize} has not successfully completed
      */
     export function isSupported(): boolean {
-      return ensureInitialized(runtime) && !!runtime.hostVersionsInfo?.appEligibilityInformation;
+      return (
+        ensureInitialized(runtime) &&
+        (!!runtime.hostVersionsInfo?.appEligibilityInformation ||
+          !!(runtime.supports.copilot && runtime.supports.copilot.eligibility))
```
lakhveerk, line 38: can this be simplified to `!!runtime.supports.copilot?.eligibility`

change/@microsoft-teams-js-286cc515-1cef-4d54-9af6-d869e00d96c9.json
erinha, line 3:
```suggestion
  "comment": "Updated `copilot.eligibility.getEligibilityInfo` to be async and get the eligibility info from the host if it is not already available.",
```

azure-pipelines.yml
```diff
@@ -20,7 +20,7 @@ resources:
     - repository: AppHostingSdk
       type: git
       name: ISS/metaos-hub-sdk
-      ref: refs/heads/main
+      ref: refs/heads/erinha/elig-message-handler
```
erinha, line 23: Do not merge with this change.

packages/teams-js/src/private/copilot.ts
```diff
@@ -41,12 +48,49 @@ export namespace copilot {
      *
      * @throws Error if {@linkcode app.initialize} has not successfully completed
      */
-    export function getEligibilityInfo(): AppEligibilityInformation {
+    export async function getEligibilityInfo(): Promise<AppEligibilityInformation> {
       ensureInitialized(runtime);
       if (!isSupported()) {
-        throw errorNotSupportedOnPlatform;
+        throw new Error(`Error code: ${errorNotSupportedOnPlatform.errorCode}, message: Not supported on platform`);
```
jekloudaMSFT, line 54: Why has this error changed?

erinha: I am trying to follow best practices and throwing an Error means that the app developer will get the same information but also a call stack, which is more helpful when debugging.

packages/teams-js/src/private/copilot.ts
erinha: I agree that hard coding is not awesome.

I don't think we should use `ApiName.Copilot_Eligibility_GetEligibilityInfo` here because that value is intended for telemetry use and could theoretically change without notice, causing the communication between teams-js and the host to break. 

Probably there should be an effort to use constants for the names of the functions that are passed to postMessage calls, but I'm not going to start that effort as part of this PR.

azure-pipelines.yml
erinha: Reverted

packages/teams-js/src/private/copilot.ts
```diff
@@ -41,12 +48,49 @@ export namespace copilot {
...
+        );
+      }
+      // validate response
+      if (!isEligibilityInfoValid(response)) {
+        throw new Error('Error deserializing eligibility information');
+      }
+      return response;
+    }
+
+    function isEligibilityInfoValid(eligibilityInfo: AppEligibilityInformation): boolean {
```
lakhveerk, line 82: Even though it was always expected from hubs to not send undefined for eligibilityInfo properties, but we were not enforcing it in code in case they send undefined. Do you think we should inform them about this change?

lakhveerk (dismissed):

packages/teams-js/src/private/copilot.ts
erinha: Sure, I can do that. It's also relatively easy to adjust this validation (vs rolling out host side changes) if we need to change it in the future.

packages/teams-js/src/private/copilot.ts
AE-MS: I have been encouraging people in PRs to use the constant in both places since the string values are identical and every string you add makes the teamsjs package size bigger, so we should reuse where possible.

I understand your concern about the two values serving a different purpose but I think in practice that isn't likely to happen -- we'd have to change everything about the telemetry pipeline, reports, etc. if this string were to ever change. Minimizing package size increase outweighs the "different purpose" concern IMO.

jekloudaMSFT (dismissed):

jekloudaMSFT (approved):

packages/teams-js/src/private/copilot.ts
erinha: Discussed offline. I'm making the change here and opening a work item to track improving the code to prevent misuse of the telemetry constant vs a function name constant.

lakhveerk (approved):

PR #2556, opened 2024-10-14
Added a url param to skip initialization when running test app

KangxuanYe:

## Description

Added a `skipInit` URL parameter in test app to make it (will happen later) as the only URL parameter to skip initialization for test app. `customInit` and `appInitializationTest` URL parameters will be removed.

### Change file added: No

### DIFF
--- a/apps/teams-test-app/src/App.tsx
+++ b/apps/teams-test-app/src/App.tsx
@@ -17,7 +17,12 @@ const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? u
 const validMessageOrigins: string[] | undefined = getOriginsParam ? getOriginsParam.split(',') : undefined;
 
 // This is added for custom initialization when app can be initialized based upon a trigger/click.
-if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
+if (
+  !urlParams.has('customInit') ||
+  !urlParams.get('customInit') ||
+  !urlParams.has('skipInit') ||
+  !urlParams.get('skipInit')
+) {
   if (isTestBackCompat()) {
     initialize(undefined, validMessageOrigins);
   } else {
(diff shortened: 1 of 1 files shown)


apps/teams-test-app/src/App.tsx
```diff
@@ -17,7 +17,12 @@ const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? u
 const validMessageOrigins: string[] | undefined = getOriginsParam ? getOriginsParam.split(',') : undefined;
 
 // This is added for custom initialization when app can be initialized based upon a trigger/click.
-if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
+if (
+  !urlParams.has('customInit') ||
```
MengyiGong, line 21: nit: It would be better to extract this into a separate function, perhaps named checkUrlParam(urlParam: string) for all the url param checks . This function could also be reused on Line 36, 37 and 38 of this file. The benefit is that it would make the code cleaner and more maintainable.

KangxuanYe: It is unnecessary because most of them would be removed and after remove, the function you suggested will only be written once here for `skipInit`.

KangxuanYe: Also suppose you have a function checking as `!A || !B`, `A && B` is not equivalent to say `not !A || !B` because `A && !B` is a sufficient scenario for `not !A || !B`

apps/teams-test-app/src/App.tsx
```diff
@@ -17,7 +17,12 @@ const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? u
...
+  !urlParams.get('customInit') ||
+  !urlParams.has('skipInit') ||
+  !urlParams.get('skipInit')
```
AE-MS, line 24: Wouldn't checking `urlParams.get('paramName') === null` be sufficient? What cases does the `.has` call also cover?

AE-MS:
Can you explain more about the difference between skipInit and customInit parameters and how they relate?

PR #2563, opened 2024-10-21
noahdarveau/app helpers refactor

noahdarveau-MSFT:
**This PR is part of the larger Treeshaking effort with more PR's to come**

The app.ts file has a large number of helper functions that are not a part of the app namespace and are also not exported through the index file to be used by the package consumer, they are only exported to be used by other parts of teams-js. As such I have moved all of the helper functions that are not a part of the app namespace to their own file, appHelpers.ts, located in the internal folder. This helps keep teams-js's structure more consistent. Also, in a future PR as part of the treeshaking work, I will need to remove the app namespace and update the index file with `export  * as app`. Since the helper functions are marked as exports so they can be used by the rest of the library, that change would cause them to be exported through the index file to the library consumer. As such they need to be moved to their own separate file anyways. 

I also updated a reference in the private index folder to export an interface that was missing and should be exported. This was a bug not caught by our docs generation as it was covered up by the bundler. Since we've moved to a new bundler that will not cover for the bug, this change will need to be added.

### DIFF
--- a/change/@microsoft-teams-js-0dfea369-ca36-49f5-87e9-4f2825d7c380.json
+++ b/change/@microsoft-teams-js-0dfea369-ca36-49f5-87e9-4f2825d7c380.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Refactored helper functions from app file to their own file",
+  "packageName": "@microsoft/teams-js",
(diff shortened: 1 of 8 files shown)


AE-MS (approved):
:shipit:

maglims:
Great idea on this change!

TrevorJoelHarris (approved):

PR #2565, opened 2024-10-21
noahdarveau/dts plugin and some treeshaking changes

noahdarveau-MSFT:
**This PR is part of the larger Treeshaking effort with more PR's to come**

This PR has a few major components to it. Unfortunately, they all need to be checked in at the same time, so this PR cannot be made any smaller, or broken into multiple PR's.

The Changes:
--------------------
- The main change in this PR is the addition of the `dts` plugin to the rollup config. This plugin is being added to generate a bundled typings file (MicrosoftTeams.d.ts) for the library. This was previously being done by the webpack plugin, however for some reason (that I have not been able to determine the cause of, likely due to the inner implementation of webpack), the webpack typings bundle was not working properly with the treeshakable code (which will be coming soon in subsequent PR's). I was previously using just the unbundled tsc generated typings files however, that caused an issue with the typed-dependency-tester. As such, a rollup typings bundler is needed to be added. There were a few other issues that needed to be resolved in order for the rollup dts plugin to work properly, with the changes outlined below: 
- There are a few instances throughout the codebase where we use the syntax `export import foo = namespace.bar`. This syntax is outdated with esm modules and so the dts() plugin would error when encountering it. In order to keep the same functionality, I had to replace the syntax with named imports and exports. This cannot be done from within a namespace, so I had to go ahead and remove the namespaces for files that are referenced in these problematic lines (App, AppInitialization, and Settings). 
- Since the app namespace was removed, all files that import the app namespace directly from the app file and not from the index file needed to have their references updated.
- Since the rollup dts plugin is no generating the typings file, we can remove the webpack plugin that was previously being used to generate the bundled typings file.
- There are also a few various whitespace changes auto-generated by prettier when I saved the files

### DIFF
--- a/change/@microsoft-teams-js-b23ef57f-9184-4659-93c8-48d862788dd3.json
+++ b/change/@microsoft-teams-js-b23ef57f-9184-4659-93c8-48d862788dd3.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Made `app`, `appInitialization`, and `settings` treeshakable as well as added dts rollup plugin",
+  "comment": "Made the `app`, `appInitialization`, and `settings` files treeshakable",
   "packageName": "@microsoft/teams-js",
   "email": "109628470+noahdarveau-MSFT@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-b23ef57f-9184-4659-93c8-48d862788dd3.json
+++ b/change/@microsoft-teams-js-b23ef57f-9184-4659-93c8-48d862788dd3.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Made `app`, `appInitialization`, and `settings` treeshakable as well as added dts rollup plugin",
+  "packageName": "@microsoft/teams-js",
+  "email": "109628470+noahdarveau-MSFT@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/rollup.config.mjs
+++ b/packages/teams-js/rollup.config.mjs
@@ -6,53 +6,65 @@ import { nodeResolve } from '@rollup/plugin-node-resolve';
 import replace from '@rollup/plugin-replace';
 import terser from '@rollup/plugin-terser';
 import typescript from '@rollup/plugin-typescript';
+import dts from 'rollup-plugin-polyfill-node';
 import nodePolyfills from 'rollup-plugin-polyfill-node';
 
 import version from './package.json' assert { type: 'json' };
 
-export default {
-  input: './src/index.ts',
-  output: {
-    dir: 'dist/esm',
-    name: '@microsoft/teams-js',
-    format: 'es',
-    preserveModules: true,
-    entryFileNames: '[name].js',
-    sourcemap: false,
-    plugins: [terser()],
-    globals: {
-      buffer: 'Buffer',
-      tty: 'tty',
-      util: 'util',
-      os: 'os',
+export default [
+  {
+    input: './src/index.ts',
+    output: {
+      dir: 'dist/esm',
+      name: '@microsoft/teams-js',
+      format: 'es',
+      preserveModules: true,
+      entryFileNames: '[name].js',
+      sourcemap: false,
+      plugins: [terser()],
+      globals: {
+        buffer: 'Buffer',
+        tty: 'tty',
+        util: 'util',
+        os: 'os',
+      },
+    },
+    preserveEntrySignatures: 'strict',
+    plugins: [
+      nodeResolve({
+        extensions: ['.js', '.ts', '.d.ts', '.json'],
+      }),
+      replace({
+        preventAssignment: true,
+        'process.env.NODE_ENV': JSON.stringify('production'),
+        PACKAGE_VERSION: JSON.stringify(version.version),
+      }),
+      typescript(),
+      json(),
+      commonjs(),
+      nodePolyfills(),
+    ],
+    treeshake: {
+      moduleSideEffects: [
+        'src/internal/communication.ts',
+        'src/internal/nestedAppAuthUtils.ts',
+        'src/internal/utils.ts',
+        'src/internal/videoEffectsUtils.ts',
+        'src/private/constants.ts',
+        'src/private/interfaces.ts',
+        'src/public/constants.ts',
+        'src/public/handlers.ts',
+        'src/public/interfaces.ts',
+      ],
     },
   },
-  preserveEntrySignatures: 'strict',
-  plugins: [
-    nodeResolve({
-      extensions: ['.js', '.ts', '.d.ts', '.json'],
-    }),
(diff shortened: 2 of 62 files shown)


packages/teams-js/rollup.config.mjs
```diff
@@ -6,53 +6,65 @@ import { nodeResolve } from '@rollup/plugin-node-resolve';
...
+        'process.env.NODE_ENV': JSON.stringify('production'),
+        PACKAGE_VERSION: JSON.stringify(version.version),
+      }),
+      typescript(),
+      json(),
+      commonjs(),
+      nodePolyfills(),
     ],
+    treeshake: {
+      moduleSideEffects: [
```
jekloudaMSFT, line 48: Are these the files that can't get tree shaken out?

noahdarveau-MSFT: Yes, the bundler should mark those files as containing side-effects, so they should be ignored by the treeshaker. Most of them are references to the `window` object. While allowing them to be treeshaken won't necessarily cause an issue, it is possible that they might. Fortunately, most of these files are frequently used, so there aren't many situations where they would be treeshaken out anyways.

noahdarveau-MSFT: Those files are also identified in the package.json, which should instruct the bundler of anyone consuming the library to not treeshake them out either.

packages/teams-js/src/public/appInitialization.ts
```diff
@@ -15,86 +15,84 @@ import { app } from './app';
...
-   * @deprecated
-   * As of TeamsJS v2.0.0, please use {@link app.IExpectedFailureRequest} instead.
-   */
-  export import IExpectedFailureRequest = app.IExpectedFailureRequest;
+/**
+ * @deprecated
+ * As of TeamsJS v2.0.0, please use {@link app.Messages} instead.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export { Messages };
```
TrevorJoelHarris, line 23: I am so happy we are changing this syntax

packages/teams-js/test/internal/communication.spec.ts
```diff
@@ -6,7 +6,7 @@ import { NestedAppAuthMessageEventNames, NestedAppAuthRequest } from '../../src/
 import { ApiName, ApiVersionNumber, getApiVersionTag } from '../../src/internal/telemetry';
 import { UUID } from '../../src/internal/uuidObject';
 import { FrameContexts } from '../../src/public';
-import { app } from '../../src/public/app';
+import * as app from '../../src/public/app';
```
TrevorJoelHarris, line 9: Just testing my knowledge: there's no reason we *shouldn't* `import * from bla` instead of importing specific types, correct? I think tree shaking will be "smart" enough to still remove anything you aren't literally referencing in the file.

packages/teams-js/test/public/appInitialization.spec.ts
```diff
@@ -1,6 +1,6 @@
 import * as AppHelpersFile from '../../src/internal/appHelpers';
 import * as AppFile from '../../src/public/app';
-import { appInitialization } from '../../src/public/appInitialization';
+import * as appInitialization from '../../src/public/appInitialization';
```
TrevorJoelHarris, line 3: could this instead be

`import * as app from '../../src/public/appInitialization'`;

And then you could delete line 13? Not a big deal, but seems like a little bit of indirection that we don't need anymore with your change

TrevorJoelHarris (changes requested):
Extremely small change asked for and a couple of questions to test my knowledge about best practices with tree shaking

packages/teams-js/test/public/appInitialization.spec.ts
noahdarveau-MSFT: That wouldn't work with just a one-line code change since appInitialization doesn't have the `_uninitialize` and `_initialize` funtions. I'd have to refactor that as well, which feels out of scope for this work. Did you possibly mean to refer to line 2's import from app.ts?

packages/teams-js/test/internal/communication.spec.ts
noahdarveau-MSFT: Typically, it's best practice to name all of the imports you would want to use (provided they are named exports). A bundler with well implemented treeshaking _**should**_ be smart enough to shake out all unused functions making this syntax fine, but in theory there is a small possibility that they don't get treeshaken out since it is not something we can directly control. In this scenario it is perfectly fine since it's just a test file, and we're not running it through a bundler anyways nor are we particularly concerned with storage complexity. Also, by using this import syntax we can still treat all of the imports as `app.whatever`, so I don't have to update all of the function call references.

packages/teams-js/src/public/appInitialization.ts
noahdarveau-MSFT: same

packages/teams-js/test/public/appInitialization.spec.ts
TrevorJoelHarris: Yes sorry, highlighted the wrong line apparently

packages/teams-js/test/internal/communication.spec.ts
TrevorJoelHarris: Okay, makes sense! Just wanted to verify this was a pattern we were okay with people copying (and I think we are)

TrevorJoelHarris (dismissed):

change/@microsoft-teams-js-b23ef57f-9184-4659-93c8-48d862788dd3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Made `app`, `appInitialization`, and `settings` treeshakable as well as added dts rollup plugin",
```
jadahiya-MSFT, line 3: small nit and feel free to ignore it if you disagree. 
```suggestion
  "comment": "Made the `app`, `appInitialization`, and `settings` files treeshakable and added dts rollup plugin",
```

Additionally, does the dts rollup plugin only impact internal teams? If so we can probably just remove the dts part from this.

noahdarveau-MSFT: Updated!

packages/teams-js/webpack.config.cjs
```diff
@@ -73,14 +73,6 @@ module.exports = {
       PACKAGE_VERSION: JSON.stringify(packageVersion),
     }),
 
-    new DtsBundleWebpack({
```
jadahiya-MSFT, line 76: Does this mean the `dts` related package we consume in our root `package.json` can be removed?

noahdarveau-MSFT: Yes, good catch. I forgot to fully remove it. Fixed!

jadahiya-MSFT (approved):

TrevorJoelHarris (approved):

PR #2566, opened 2024-10-22
Add support for Node 22 without modifying minimum version

juanscr:
## Description

With this change, a deprecated experimental syntax using `assert` is changed for directly reading the file instead. With this change, all node versions including old and new ones are supported for building the project without any major change in minimum supported version. Moreover, this does not change the current behavior of the rollup file as it uses a more direct approach for reading the file.

### Main changes in the PR:

1. Removed the `assert` experimental and deprecated syntax.
2. For changing it, added a supported file read syntax supported on all node versions.

### Validation performed:

1. Build the project manually.

### Unit Tests added:

No. This is a backwards compatible change with the rollup build file.

### End-to-end tests added:

No.

### Change file added:

Yes.


change/@microsoft-teams-js-f6526430-887f-46bc-bfb4-8c9750214364.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
TrevorJoelHarris, line 2: I'd actually mark this change as type `none` instead of patch I think. Should have no noticeable effect on library consumers, only matters if you are building the library yourself.

TrevorJoelHarris (changes requested):
Small question about change type but looks good to me otherwise

change/@microsoft-teams-js-f6526430-887f-46bc-bfb4-8c9750214364.json
juanscr: Sure, I agree. Let me correct the change file

noahdarveau-MSFT (dismissed):

noahdarveau-MSFT (approved):

TrevorJoelHarris (approved):

TrevorJoelHarris (approved):

noahdarveau-MSFT (approved):

PR #2570, opened 2024-10-23
Added JSON.Stringfy function to convert potential json format msg

KangxuanYe:
## Description

Though all of API calls from testing host will be string format, but listener in test app will listen to all of messages, including response from Web Host SDK. Some of responses are in JSON format. It does no harm for testing, but in case people want to know what it is, I added one line to stringfy those kind of data.


AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

PR #2572, opened 2024-10-24
Validate the unknown object is either plain object of primitive

lakhveerk:

## Description

In `externalAppAuthnetication` and `externalAppAutneticationForCEA`, some APIs accept `IActionExecuteInvokeRequest` as input. This interface has a `data` property that can be of type `string | Record<string, unknown>`. The unknown type in Record is unknown. Currently, we post the value as it is to host sdk. However, it could have contained values that are not serializable. 
The copilot team that consumes the API confirms that the type would be either primitive or plain object. This PR validate the data and throws if that is not the case.  

### Main changes in the PR:

1. Added validation function `isPrimitiveOrPlainObject` in `utils.ts` file.
2. Added validation for `IActionExecuteInvokeRequest.data` element in `ExternalAppAuthentication` and `ExternalAppAuthenticationForCEA` capabilities.

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/internal/utils.ts
```diff
@@ -504,3 +504,46 @@ export function validateUuid(id: string | undefined | null): void {
     throw new Error('id must be a valid UUID');
   }
 }
+
+/**
+ * @beta
```
TrevorJoelHarris, line 509: Probably no need for a @beta tag here since this isn't exported outside of the library

packages/teams-js/src/internal/utils.ts
```diff
@@ -504,3 +504,46 @@ export function validateUuid(id: string | undefined | null): void {
...
+    typeof value === 'object' &&
+    Object.prototype.toString.call(value) === '[object Object]' &&
+    (Object.getPrototypeOf(value) === Object.prototype || Object.getPrototypeOf(value) === null);
+
+  if (!isPlainObject) {
+    return false;
+  }
+
+  // Check all properties of the object recursively
+  return Object.keys(value).every((key) => isPrimitiveOrPlainObject(value[key]));
```
TrevorJoelHarris, line 548: I couldn't tell by looking at your unit tests: do you have any unit tests that hit line 548 and actually recurse through? That would be I guess an object with no functions at the top level but a function inside of one of its members.

TrevorJoelHarris (changes requested):
Small questions about testing

packages/teams-js/src/internal/utils.ts
lakhveerk: removed!

packages/teams-js/src/internal/utils.ts
```diff
@@ -518,3 +518,49 @@ const supportsPerformanceTimers = !!performance && 'now' in performance;
...
+ * @internal
+ * Limited to Microsoft-internal use
+ *
+ * Function to check whether the data is a primitive type or a plain object.
+ * @param value The value to check
+ * @returns true if the value is a primitive type or a plain object, false otherwise
+ *
+ */
+export function isPrimitiveOrPlainObject(value: unknown, depth: number = 0): boolean {
+  if (depth > 1000) {
```
lakhveerk, line 533: I added this change as well since you last reviewed @TrevorJoelHarris

packages/teams-js/src/internal/utils.ts
```diff
@@ -518,3 +518,49 @@ const supportsPerformanceTimers = !!performance && 'now' in performance;
...
+    return false; // Limit recursion depth
+  }
+
+  // Check if the value is a primitive type or null
+  if (
```
MengyiGong, line 538: nit: what about this approach:
 ```function isPrimitiveOrNull(value) {
  return value === null || (typeof value !== 'object' && typeof value !== 'function');
}```
 it covers all primitive types without needing to list each one individually.

packages/teams-js/src/internal/utils.ts
lakhveerk: Because Anand from 3pcxp said, he does not see it going any further than that

packages/teams-js/src/private/externalAppAuthentication.ts
```diff
@@ -373,28 +373,55 @@ export type InvokeErrorWrapper = InvokeError & { responseType: undefined };
...
+}
+
+/**
+ * @beta
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+function validateQueryMessageExtensionRequest(originalRequestInfo: IQueryMessageExtensionRequest): void {
+  if (originalRequestInfo.commandId.length > 64) {
```
MengyiGong, line 413: I might not know much of the background context, and I feel confused and curious about where these magic numbers (like 64, 5, 512) come from.

packages/teams-js/src/internal/utils.ts
MengyiGong: nit: It would be great if you could add this context as a comment for this function to help the next developer understand the logic.

packages/teams-js/src/internal/utils.ts
lakhveerk: The function needs to allow the nested primitives in an object as well. This check will incorrectly return false if the value was a plain object or an array of primitives.

packages/teams-js/src/private/externalAppAuthentication.ts
lakhveerk: They all came from 3pcxp team. In this PR, I am just dividing the already existing function into smaller functions. These numbers were added when the support was added for MEs in last December

packages/teams-js/test/internal/utils.spec.ts
```diff
@@ -548,4 +549,46 @@ describe('utils', () => {
       });
     });
   });
+  describe('isPrimitiveOrPlainObject', () => {
+    it('should return true for undefined or null', () => {
```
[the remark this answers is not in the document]
lakhveerk: Just added a test case, let me know if you think that is sufficient

packages/teams-js/src/internal/utils.ts
lakhveerk: I added a comment saying `Recursion is limited to a maximum depth of 1000 to prevent excessive nesting and potential stack overflow.` Does that work?

MengyiGong (approved):

TrevorJoelHarris (approved):

PR #2576, opened 2024-10-25
noahdarveau/dialog treeshaking changes

noahdarveau-MSFT:
**This PR is part of the larger Treeshaking effort, with more PRs to come in the future**

This PR refactors the `dialog` capability as well as all of its sub-capabilities to make them treeshakable. Since `dialog` has many sub-capabilities, some with their own set of sub-capabilities, it's getting its own PR since there's quite a few changes that had to be made. Those changes being:
- Moved all helper functions and `dialogTelemetryVersionNumber` constant to their own file, `dialogHelpers.ts` in the internal folder.
- Moved each sub-capability to their own files and reorganized the file structure to help keep the repo organized.
- Removed the namespaces from each capability file to enable treeshaking

The new folder structure for the `dialog` capability and sub-capabilites is as follows:
```
public
  dialog
    adaptiveCard
      adaptiveCard.ts
      bot.ts
    dialog.ts
    update.ts
    url
      bot.ts
      parentCommunication.ts
      url.ts 
```

One question I have is do you think having the parent capability match the folder name that they are in, or would it make more sense to name them index.ts? For example, would you prefer having the folder `adaptiveCard` contain `adaptiveCard.ts` and `bot.ts`. Or would you prefer the folder `adaptiveCard` contain `index.ts` and `bot.ts`?

### DIFF
--- a/packages/teams-js/src/public/dialog/adaptiveCard/adaptiveCard.ts
+++ b/packages/teams-js/src/public/dialog/adaptiveCard/adaptiveCard.ts
@@ -8,7 +8,6 @@ import { errorNotSupportedOnPlatform, FrameContexts } from '../../constants';
 import { AdaptiveCardDialogInfo, DialogInfo } from '../../interfaces';
 import { runtime } from '../../runtime';
 import { DialogSubmitHandler } from '../dialog';
-import * as url from '../url/url';
 import * as bot from './bot';
 
 /**
@@ -22,7 +21,7 @@ import * as bot from './bot';
  * This function cannot be called from inside of a dialog
  *
  * @param adaptiveCardDialogInfo - An object containing the parameters of the dialog module {@link AdaptiveCardDialogInfo}.
- * @param submitHandler - Handler that triggers when a dialog calls the {@linkcode url.submit} function or when the user closes the dialog.
+ * @param submitHandler - Handler that triggers when a dialog fires an [Action.Submit](https://adaptivecards.io/explorer/Action.Submit.html) or when the user closes the dialog.
  *
  * @beta
  */
--- a/packages/teams-js/src/public/dialog/adaptiveCard/adaptiveCard.ts
+++ b/packages/teams-js/src/public/dialog/adaptiveCard/adaptiveCard.ts
@@ -1,18 +1,65 @@
-import { sendMessageToParent } from '../../internal/communication';
-import { botUrlOpenHelper, updateResizeHelper, urlOpenHelper, urlSubmitHelper } from '../../internal/dialogHelpers';
-import { GlobalVars } from '../../internal/globalVars';
-import { registerHandler, removeHandler } from '../../internal/handlers';
-import { ensureInitialized } from '../../internal/internalAPIs';
-import { ApiName, ApiVersionNumber, getApiVersionTag } from '../../internal/telemetry';
-import { isHostAdaptiveCardSchemaVersionUnsupported } from '../../internal/utils';
-import { DialogDimension, errorNotSupportedOnPlatform, FrameContexts } from '../constants';
-import {
-  AdaptiveCardDialogInfo,
-  BotAdaptiveCardDialogInfo,
-  BotUrlDialogInfo,
-  DialogInfo,
-  DialogSize,
-  M365ContentAction,
-  UrlDialogInfo,
-} from '../interfaces';
-import { runtime } from '../runtime';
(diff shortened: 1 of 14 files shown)


TrevorJoelHarris:
> One question I have is do you think having the parent capability match the folder name that they are in, or would it make more sense to name them index.ts? For example, would you prefer having the folder adaptiveCard contain adaptiveCard.ts and bot.ts. Or would you prefer the folder adaptiveCard contain index.ts and bot.ts?

Just personal opinion, but I like making the folder name match the parent capability name.

packages/teams-js/src/public/dialog/adaptiveCard/adaptiveCard.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @beta
+ */
+/**
+ * Allows app to open an adaptive card based dialog.
+ *
+ * @remarks
+ * This function cannot be called from inside of a dialog
+ *
+ * @param adaptiveCardDialogInfo - An object containing the parameters of the dialog module {@link AdaptiveCardDialogInfo}.
+ * @param submitHandler - Handler that triggers when a dialog calls the {@linkcode url.submit} function or when the user closes the dialog.
```
TrevorJoelHarris, line 25: I think this comment is actually wrong, and that fixing it will let you remove the reference to the url file.

For adaptive card dialogs, this handler is not triggered by calls to the submit function because that function can only be called from url dialogs which are essentially Teams apps running inside of a dialog shell. Instead, for adaptive card dialogs the equivalent is when the adaptive card fires an [Action.Submit](https://adaptivecards.io/explorer/Action.Submit.html). If we update this comment to be something like this:

```suggestion
 * @param submitHandler - Handler that triggers when a dialog fires an [Action.Submit](https://adaptivecards.io/explorer/Action.Submit.html) or when the user closes the dialog.
```

This will also make sure we don't unintentionally use any url-only logic in the adaptive card functionality

packages/teams-js/src/public/dialog/adaptiveCard/bot.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+/**
+ * Allows an app to open an adaptive card-based dialog module using bot.
+ *
+ * @param botAdaptiveCardDialogInfo - An object containing the parameters of the dialog module including completionBotId.
+ * @param submitHandler - Handler that triggers when the dialog has been submitted or closed.
+ *
+ * @beta
+ */
+export function open(botAdaptiveCardDialogInfo: BotAdaptiveCardDialogInfo, submitHandler?: DialogSubmitHandler): void {
```
TrevorJoelHarris, line 27: Why did bot not end up with its own folder, given that it is a subcapability of adaptiveCard? I don't feel strongly about it, except that I want to make sure I understand why we're making this decision so we can give clear guidance to the rest of the engineering team about when and why to make new subfolders

TrevorJoelHarris (changes requested):
Just a couple small questions/comments. Let me know when it's ready for a second look

packages/teams-js/src/public/dialog/adaptiveCard/bot.ts
noahdarveau-MSFT: I didn't give `bot` its own folder since it doesn't have any sub-capabilites of its own, so the only thing that would be in the folder would be the `bot.ts` file itself. The same goes for `url.bot` and `url.parentCommunication`. If you feel like it would make more sense to have all of the bottom level sub-capabilites  be in their own folder, I'd be happy to add them, though it felt a bit extraneous to me, since following that schema, each capability/sub-capability would get its own folder, which feels unnecessary. Let me know your thoughts.

packages/teams-js/src/public/dialog/adaptiveCard/adaptiveCard.ts
noahdarveau-MSFT: Fixed!

packages/teams-js/src/public/dialog/adaptiveCard/bot.ts
TrevorJoelHarris: I think that's fine, and seems like clear guidance we could give to people working on this repo. If a capability has subcapabilities, put it into a folder.

TrevorJoelHarris (approved):

jekloudaMSFT (approved):

AE-MS (approved):
:shipit:

PR #2589, opened 2024-10-30
Add promise response to notify success.

juanscr:
## Description

The notify success function, previously, was a fire and forget function which hinder the possibility of the app knowing when it finished in the host. This change modifies this and adds a response to the function that indicates if the function finished.

Additionally, as we can't assure the host has already implemented the answer back mechanism for this function a runtime property was added that will determine if the host supports it or not. In this manner, the `notifySuccess` function now needs the runtime to be populated before even passing it to the communication layer.

### Main changes in the PR:

1. Changed the notify success function to wait for a response from the host.
2. Added a new runtime attribute to specify if the host supports or not a response from the `notifySuccess` function. As such, I need for a complete initialized promise to finish before even passing the message to the communication functions.

### Validation performed:

1. Ran the unit tests
2. Ran the test app to see if it worked properly.

### Unit Tests added:

Yes

### End-to-end tests added:

No

### Change file added:

Yes

## Screenshots
Test app on host that do not support answering back to the `notifySuccess` callback:
image

Test app on host that supports answering back to the `notifySuccess` callback:
image


juanscr:

This should not need a version change in the runtime as a previous version can still be transformed to the newest one by setting the new field to false. The `notifySuccess` function will fallback to behave as before without any issue. Therefore, the changes are backwards compatible and supports old and new behavior without any change in the runtime config.

packages/teams-js/src/public/app/app.ts
```diff
@@ -672,9 +672,8 @@ export function notifyAppLoaded(): void {
 /**
  * Notifies the frame that app initialization is successful and is ready for user interaction.
  */
-export function notifySuccess(): void {
-  ensureInitializeCalled();
```
maglims, line 676: Maybe it is already discussed that i have missed: are we allowing parallel appinit/notifySuccess sending in this PR?  Will this change be adopted by all apps, not only bizchat?

packages/teams-js/src/internal/appHelpers.ts
```diff
@@ -33,6 +40,16 @@ const initializationTimeoutInMs = 5000;
 
 const appLogger = getLogger('app');
 
+/**
+ * The response of the notify success callback.
+ */
+export interface NotifySuccessResponse {
+  /**
+   * It shows if the callback resolved successfully in the host. If the host does not support answering back to the callback, the result is unknown.
```
maglims, line 48: I think Niharika mentioned that the notifySuccess is only to track up to the hub sdk receiving it? Might be good to check with her

packages/teams-js/src/internal/appHelpers.ts
```diff
@@ -70,8 +87,37 @@ export function notifyFailureHelper(apiVersiontag: string, appInitializationFail
   ]);
 }
 
-export function notifySuccessHelper(apiVersionTag: string): void {
-  sendMessageToParent(apiVersionTag, app.Messages.Success, [version]);
+function supportsNotifySuccessResponse(): boolean {
```
maglims, line 90: Am I reading it right that we are treating  notifySuccess a capability? Is there a specific reason why?

packages/teams-js/src/public/app/app.ts
juanscr: Yes, well previously it allowed parallel initialize and notify success call which this PR keeps supporting in some sense. The behavior is as follows:

If some calls `notifySuccess` there could be three scenarios:
1. Initialize already finished which the `notifySuccess` function will immediately send the request.
2. Initialize hasn't finished yet, therefore the `notifySuccess` message will be sent as soon as the `initialize` callback resolves so I can read the runtime.
3. Initialize hasn't even been called which passes an error.

The unit tests cover all three cases :)

packages/teams-js/src/internal/appHelpers.ts
juanscr: Of what I understood the idea was to resolve the promise when the Hub finishes processing the message, that is rewriting the function from a fire-and-forget to a request-and-wait-for-response. If this assumption is wrong, please correct me Niharika

packages/teams-js/src/internal/appHelpers.ts
juanscr: The problem is that the previous behavior for `notifySuccess` was a fire-and-forget callback, therefore the host didn't send any response back to teams-js after the callback finished processing. As such, if I did the change without checking if the host answers back, I will wait indefinitely for a response to the callback as the host will never send one. In this manner, there are some newer hosts that will get a response from the callback and some older ones which will not, and I need to make the function behave accordingly.

In discussions with Trevor, we decided the best way to do it is following the `runtime` object in teams-js. Normally this is used for capabilities but on this case, it makes sense to use it, as it changes the behavior depending on the host is running in.

packages/teams-js/src/public/appInitialization.ts
```diff
@@ -58,12 +65,12 @@ export function notifyAppLoaded(): void {
...
 /**
  * @deprecated
- * As of TeamsJS v2.0.0, please use {@link app.notifySuccess app.notifySuccess(): void} instead.
+ * As of TeamsJS v2.0.0, please use {@link app.notifySuccess app.notifySuccess(): Promise<NotifySuccessResponse>} instead.
  *
  * Notifies the frame that app initialization is successful and is ready for user interaction.
  */
-export function notifySuccess(): void {
-  notifySuccessHelper(
+export function notifySuccess(): Promise<NotifySuccessResponse> {
```
TrevorJoelHarris, line 72: we shouldn't change the signature on this function because it's an old back compat function.

packages/teams-js/src/public/runtime.ts
```diff
@@ -223,6 +223,9 @@ interface IRuntimeV4 extends IBaseRuntime {
   readonly isNAAChannelRecommended?: boolean;
   readonly isLegacyTeams?: boolean;
   readonly supports: {
+    readonly app?: {
+      readonly notifySuccessResponse: boolean;
```
TrevorJoelHarris, line 227: I don't think there's a reason we can't use a boolean instead of {} here except for convention, but let's use {} because that's what we do for other capabilities/subcapabilities

TrevorJoelHarris: (really, it's because we want to be able to build hierarchies of nested capabilities if needed)

packages/teams-js/src/internal/appHelpers.ts
```diff
@@ -70,8 +87,37 @@ export function notifyFailureHelper(apiVersiontag: string, appInitializationFail
...
+  return ensureInitialized(runtime) && !!runtime.supports.app?.notifySuccessResponse;
+}
+
+export async function notifySuccessHelper(apiVersionTag: string): Promise<NotifySuccessResponse> {
```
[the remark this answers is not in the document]
TrevorJoelHarris: It would be good for this comment to have a function explaining what it is doing conceptually, something about how it lets a dev call notifySuccess even while initialize is pending receiving a response, even though the runtime object is required. I want to make sure that people understand the cleverness in here and don't accidentally mess it up in the future

apps/teams-test-app/src/components/AppInitialization.tsx
```diff
@@ -27,12 +27,10 @@ const NotifySuccess = (): React.ReactElement =>
     title: 'appInitialization.success',
     onClick: {
       withPromise: async () => {
-        app.notifySuccess();
-        return 'called';
+        return JSON.stringify(await app.notifySuccess());
       },
       withCallback: (setResult) => {
-        appInitialization.notifySuccess();
```
TrevorJoelHarris, line 34: Do we need to add or update notifySuccess e2e tests?

TrevorJoelHarris (changes requested):

packages/teams-js/src/public/appInitialization.ts
juanscr: Ohh got it, that makes sense. Will revert the change.

packages/teams-js/src/public/runtime.ts
juanscr: Sure, it makes sense!

packages/teams-js/src/internal/appHelpers.ts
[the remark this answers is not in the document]
juanscr: I'll add the comment 😄

packages/teams-js/src/internal/appHelpers.ts
```diff
@@ -70,8 +87,37 @@ export function notifyFailureHelper(apiVersiontag: string, appInitializationFail
...
+  // finishes to have the full runtime object instantiated.
+  if (!GlobalVars.initializePromise) {
+    throw new Error(errorLibraryNotInitialized);
+  }
+  return GlobalVars.initializePromise.then(() => callNotifySuccessInHost(apiVersionTag));
+}
+
+export async function callNotifySuccessInHost(apiVersionTag: string): Promise<NotifySuccessResponse> {
+  if (!supportsNotifySuccessResponse()) {
+    callFunctionInHost(app.Messages.Success, [version], apiVersionTag);
```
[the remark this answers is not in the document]
juanscr: That makes sense, reading into the implementation I can see that function is waiting for the response. I think is a simple change reverting to the deprecated function. 

Still, there may be some needs in the future to add fire-and-forget type capabilities and it would be great to support that behavior in the "correct" way. Maybe we should add a backlog task for creating another function to do so? Or add a flag in the `callFunctionInHost` to be able to create fire-and-forget capabilities? I'll push the change, but I could include something like that if you want 😄

apps/teams-test-app/src/components/AppInitialization.tsx
juanscr: I do think we need to do it as the test app of course changed to display these new values. When I noticed the E2E tasks didn't failed, I stop thinking about it as maybe the current E2E tests do not test the specific content of the `notifySuccess` text box but only that it doesn't error out. I'll check the E2E tests tomorrow and let you know lol

packages/teams-js/src/internal/appHelpers.ts
ndangudubiyyam: @maglims Biz Chat requested that need to know if app.notifySuccess has been successfully resolved to help plugin holes in the telemetry. They are trying to get 99.99% success rate for P95 telemetry and because of the missing telemetry from hub/hubsdk, they want to rely on their own telemetry to fund the effort. This might help in moving the 'Other' percentage to 'Success' bucket once the biz chat gets a response back from hubsdk that it indeed received and processed notifysuccess.

packages/teams-js/src/internal/appHelpers.ts
```diff
@@ -70,8 +86,43 @@ export function notifyFailureHelper(apiVersiontag: string, appInitializationFail
...
+  }
+
+  // If initialize hasn't been called yet, throw an error to the dev as the app hasn't initialized yet
+  if (!GlobalVars.initializePromise) {
+    throw new Error(errorLibraryNotInitialized);
+  }
+
+  // If initialize is still waiting for response, dispatch the call after initialize
+  // finishes to have the full runtime object instantiated.
+  return GlobalVars.initializePromise.then(() => callNotifySuccessInHost(apiVersionTag));
```
ndangudubiyyam, line 106: Don't we want to handle if the initializePromise is rejected?

packages/teams-js/src/internal/appHelpers.ts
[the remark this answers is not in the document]
TrevorJoelHarris: I would like to wait to add a "correct" way to fire and forget from TeamsJS until we think of a legitimate reason to need one. So far, I haven't been able to think of a legitimate reason to ever call a function in the host that can't *at least* potentially receive errors from the host if something goes wrong (which is what `callFunctionInHost` is for).

packages/teams-js/src/internal/appHelpers.ts
TrevorJoelHarris: Good question! If you look at `communication.initializeCommunication`, if we never receive a response to our initialize request we never change the value of `Communication.parentOrigin` from null to the host's actual origin, which will prevent the notifySuccess message from being sent to the host at all (which you can see in `communication.sendRequestToTargetWindowHelper`)

apps/teams-test-app/src/components/AppInitialization.tsx
TrevorJoelHarris: What was the answer to this?

juanscr: So, for now I decided not to add nor update the current E2E tests. The reasoning behind this is that, although the explicit response of the function changed, the expected client behavior for using this function is exactly the same which is stopping the loading indicator for the app. As such, the unit tests are properly modified and enhanced to consider when should or not should the function receive a response. But as E2E tests are more of a "black box" testing to see if the behavior is preserved, I wouldn't consider modifying the existing ones 😄

TrevorJoelHarris (approved):

ndangudubiyyam (approved):

PR #2592, opened 2024-10-31
noahdarveau/treeshaking private capabilities

noahdarveau-MSFT:
**This PR is part of the larger Treeshaking effort, with more PRs to come in the future**

This PR updates the following private capabilities to make them treeshakable. They are all fairly simple changes that only required removing the namespaces and updating the references accordingly.

- appEntity
- conversations
- copilot
- externalAppAuthentication
- externalAppAuthenticationForCEA
- externalAppCardActions
- externalAppCardActionsForCEA
- externalAppCommands
- files
- logs
- meetingRoom
- notifications
- otherAppStateChange
- remoteCamera


packages/teams-js/src/private/index.ts
```diff
@@ -18,20 +18,23 @@ export {
   registerUserSettingsChangeHandler,
   openFilePreview,
 } from './privateAPIs';
-export { conversations, OpenConversationRequest } from './conversations';
+export * as conversations from './conversations';
+//It is necessary to export OpenConversationRequest from conversations.ts individually as well
```
jekloudaMSFT, line 22: What's special about OpenConversationRequest that we have to export it individually?

noahdarveau-MSFT: Prior to this change, it was listed as a named export, so anyone could have been referencing it as `import { OpenConversationRequest} from '@microsoft/teams-js'`. If we only had the `export * as conversations` line exporting, they would have to reference it as `conversations.OpenConversationRequest`, which would be a breaking change.

jekloudaMSFT (approved):

AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

PR #2598, opened 2024-11-05
noahdarveau/buffer removal

noahdarveau-MSFT:
This PR replaces use of the `Buffer` package with [uint8array-extras](https://www.npmjs.com/package/uint8array-extras). This change allows for the removal of the `buffer` polyfill, which was previously required as `Buffer` was not supported in browsers. Since `uint8array-extras` is browser-compatible we do not need to include a polyfill for it. The `Buffer` polyfill is rather large, ~22KB when minified, and cannot reap the benefits of treeshaking since polyfills are always included in the global scope. We are only using the `Buffer` package in the `clipboard` capability, so even if a consumer was treeshaking out `clipboard`, they were still getting the `Buffer` polyfill included in their bundle, even though it was not doing anything. With this change, anyone not using the `clipboard` capability should see a flat ~22KB package size reduction.

To include this package, despite it being and NPM package, it was having issues working with Jest for our tests. I'm still not entirely sure what exactly the issue was, but copying the package locally and running it through our typescript transpiler fixed the issue. There was another issue with `TextDecoder` not being supported in jsdom, even though it is supported in both Node and browsers, so I included a jest-setup file that polyfills `TextDecoder` with the Node implementation.


jadahiya-MSFT (dismissed):

package.json
```diff
@@ -146,7 +146,6 @@
     }
   },
   "dependencies": {
-    "skeleton-buffer": "file:./skeleton-buffer",
```
TrevorJoelHarris, line 149: Should we also delete the skeleton buffer folder and files as part of this change?

packages/teams-js/jest-setup.cjs
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable strict-null-checks/all */
```
TrevorJoelHarris, line 1: The comment you put in the description about why this was needed was helpful, could you add a comment to the file so that if people look at this they know why it was added?

packages/teams-js/src/internal/uint8array-extras/README.md
```diff
@@ -0,0 +1 @@
+This package is a stripped out version of the [uint8arrays-extras](https://www.npmjs.com/package/uint8array-extras) package, containing only the functionality used by teams-js. Itis being directly included in the library instead of an NPM package due to a Jest issue preventing its usage, which is solved by transpiling the file ourselves. This package is used to replace the `Buffer` package as this is both treeshakable and browser-compatible, so we no longer need to polyfill `Buffer`, which is quite size intensive. This provides a much more space-efficient method to providing the needed functionality.
```
TrevorJoelHarris, line 1:
```suggestion
This package is a stripped out version of the [uint8arrays-extras](https://www.npmjs.com/package/uint8array-extras) package, containing only the functionality used by teams-js. It is being directly included in the library instead of an NPM package due to a Jest issue preventing its usage, which is solved by transpiling the file ourselves. This package is used to replace the `Buffer` package as this is both treeshakable and browser-compatible, so we no longer need to polyfill `Buffer`, which is quite size intensive. This provides a much more space-efficient method to providing the needed functionality.
```

TrevorJoelHarris (changes requested):
Couple of small questions and I want to follow up with jay about the transpiled uint8array-extras folder

jadahiya-MSFT (approved):

TrevorJoelHarris (approved):

PR #2620, opened 2024-11-14
Skip failure on task to install XCTestHTMLReport and only publish result to artifact when the task succeeded

KangxuanYe:
## Description

As title says, this PR is to skip failure on task to install XCTestHTMLReport and only publish result to artifact when the task succeeded, considering we encounter the error as below every day. Pipeline will be considered to be successful even if XCTestHTMLReport fails to be installed since we have a `continueOnError` tag for that.

```dyld[7806]: terminating because inserted dylib '/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' could not be loaded: tried: '/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' (fat file, but missing compatible architecture (have 'x86_64,arm64', need 'arm64e')), 
'/System/Volumes/Preboot/Cryptexes/OS/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' (no such file), '/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' (fat file, but missing compatible architecture (have 'x86_64,arm64', need 'arm64e'))
dyld[7806]: tried: '/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' (fat file, but missing compatible architecture (have 'x86_64,arm64', need 'arm64e')), 
'/System/Volumes/Preboot/Cryptexes/OS/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' (no such file), '/Users/runner/work/_temp/codeql3000/distribution/codeql/tools/osx64/libtrace.dylib' (fat file, but missing compatible architecture (have 'x86_64,arm64', need 'arm64e'))
/Users/runner/work/_temp/939bbdf4-66ff-4572-8b7e-3fdb96cc1197.sh: line 1:  7806 Abort trap: 6           brew install xctesthtmlreport
```

## Validation

Hopefully it works when next time we encounter the same issue.

### DIFF
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -14,11 +14,13 @@ steps:
 
   - task: Bash@3
     displayName: 'Install XCTestHtmlReport for publishing result'
+    id: installXCTestHtmlReport
     inputs:
       targetType: inline
       script: |
         brew install xctesthtmlreport
       workingDirectory: '$(Agent.BuildDirectory)/iOSHost'
+      continue-on-error: true
       condition: always()
 
   - task: NodeTool@0
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -166,5 +166,5 @@ steps:
     inputs:
       path: '$(agent.buildDirectory)/Logs/$(agent.JobName)/$(agent.JobName)_DebugLogs.zip'
       artifact: iOSDebugLogs - ${{ parameters.testPlan }} - Attempt $(System.JobAttempt)
-    condition: eq(variables['task.iOS_E2E_Test_Task.status'], 'failed')
+    condition: and(eq(variables['task.iOS_E2E_Test_Task.status'], 'failed'), eq(variables['xctestHtmlReportInstalled'], 'Yes'))
     continueOnError: true
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -14,7 +14,6 @@ steps:
 
   - task: Bash@3
     displayName: 'Install XCTestHtmlReport for publishing result'
-    id: installXCTestHtmlReport
     inputs:
       targetType: inline
       script: |
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -144,6 +144,8 @@ steps:
       mergeTestResults: true
     condition: succeededOrFailed()
 
+    # The condition of this task is to say, only when xctestHtmlReportInstall task successfully be executed,
+    # this task will then be executed to publish results to the artifact.
   - bash: |
       output="$(agent.buildDirectory)/Logs/$(agent.JobName)"
       rm -rf "${output}" > /dev/null
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -19,8 +19,10 @@ steps:
       targetType: inline
       script: |
         brew install xctesthtmlreport
+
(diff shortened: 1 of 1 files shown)


tools/yaml-templates/ios-test.yml
```diff
@@ -18,7 +18,10 @@ steps:
       targetType: inline
       script: |
         brew install xctesthtmlreport
```
AE-MS, line 20: If this fails (I assume this is the command that is failing intermittently?) is there a way to get more verbose logs about what the failure is?

tools/yaml-templates/ios-test.yml
```diff
@@ -148,7 +151,7 @@ steps:
       echo "ls -lR ${output}"
       ls -lR "${output}"
     displayName: Preparations for publishing results
-    condition: always()
+    condition: eq(variables['xctestHtmlReportInstalled'], 'Yes')
```
AE-MS, line 154: What effect will this have? Will the test results not be published if installation failed? Does that mean that any test failures will be overlooked if the html reporter didn't get installed?

AE-MS (changes requested):
:clock1:

tools/yaml-templates/ios-test.yml
KangxuanYe: I think they typically hide logs for us when we install this. Basically the error log it throws to console is the detailed one. Not sure if we can get more detailed log on VM.

tools/yaml-templates/ios-test.yml
KangxuanYe: This condition is to say, only when `xctestHtmlReportInstall` task successfully be executed, this task will then be executed to publish results to the artifact.

Any test failures will NOT be overlooked. It just affects devs to figure out why test cases fail on CI pipeline with detailed screenshots and logs.

KangxuanYe (commented):
Replied to questions.

tools/yaml-templates/ios-test.yml
```diff
@@ -18,7 +18,10 @@ steps:
       targetType: inline
       script: |
         brew install xctesthtmlreport
+
+        echo "##vso[task.setvariable variable=xctestHtmlReportInstalled]Yes"
```
AE-MS, line 22: Can you add a comment or two explaining what this does and why it is here?

tools/yaml-templates/ios-test.yml
AE-MS: Can you add a comment explaining that? Thanks!

maglims: just wanted to make sure we are on the same page:

This does look like improvement, i.e, if we don't have `xctestHtmlReport` tool installed correctly, we shouldn't try to use the tool in later phase. The PR is not fixing existing environment issue, but rather than a pipeline task improvement. is my understanding correct ? @KangxuanYe

KangxuanYe: Yes, 1000 out of 100 @maglims 

@AE-MS I'll add comment to explain the change over tasks.

tools/yaml-templates/ios-test.yml
KangxuanYe: Comments added. Please take a look.

tools/yaml-templates/ios-test.yml
KangxuanYe: comments added.

jekloudaMSFT (approved):

AE-MS (approved):
:shipit:

maglims (approved):

PR #2621, opened 2024-11-15
noahdarveau/2.31.0 fixes

noahdarveau-MSFT:
There are 2 bugs with the current release of 2.31.0:

1. The `EduType` is not exported in the index file, so references to it are failing. Adding it to the index file will resolve the issue.
2. With rolling up the types into one file, the enums are no longer being exported as enums. Rollup is doing some strange type aliasing for them instead. To resolve this, we will no longer rollup the types into one file and instead will use the tsc generated types.

### DIFF
--- a/apps/typed-dependency-tester/package.json
+++ b/apps/typed-dependency-tester/package.json
@@ -5,7 +5,7 @@
   "description": "A tester to check if types in the dependency array will cause errors",
   "version": "0.0.1",
   "scripts": {
-    "build": "pnpm i && pnpm copy && pnpm tsc && pnpm clean",
+    "build": "pnpm i && pnpm copy && pnpm clean",
     "copy": "cp -r ./node_modules/@microsoft/teams-js/dist/esm/packages/teams-js/dts ./ || xcopy .\\node_modules\\@microsoft\\teams-js\\dist\\esm\\packages\\teams-js\\dts .\\ /s /Y",
     "clean": "rimraf node_modules && rimraf MicrosoftTeams.d.ts"
   },
--- a/apps/typed-dependency-tester/package.json
+++ b/apps/typed-dependency-tester/package.json
@@ -6,7 +6,7 @@
   "version": "0.0.1",
   "scripts": {
     "build": "pnpm i && pnpm copy && pnpm tsc && pnpm clean",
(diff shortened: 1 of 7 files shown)


TrevorJoelHarris (dismissed):

AE-MS:
For #1: I can see how maybe our tests didn't consume that particular enum so we wouldn't have caught it
For #2: would this break any enum consumption? Does our test app consume enums... at all?

noahdarveau-MSFT:
> For #1: I can see how maybe our tests didn't consume that particular enum so we wouldn't have caught it For #2: would this break any enum consumption? Does our test app consume enums... at all?

For 1, `EduType` is being exported from the interfaces file, however it was not exported through the index file. The old webpack bundling didn't care and would export it out of the package anyways, Rollup will not export it out of the package if it's not exported through the index file. We do have a test that references `EduType`, however it is referenced directly from the file not the index file.

For 2, There is no problem referencing the members of the enums, only the enum objects themselves. For example, this issue was brought to my attention with someone trying to assign a variable to the enum object ex. `let foo = app.FailedReason`. This is where the type aliasing that the rollup bundler does fails as referencing the enum object's values still works fine, ie. `let foo = app.Failedreason.AuthFailed`. I assume we don't have any tests that try to assign any variables to any enum objects themselves.

TrevorJoelHarris:
> > For #1: I can see how maybe our tests didn't consume that particular enum so we wouldn't have caught it For #2: would this break any enum consumption? Does our test app consume enums... at all?
> 
> For 1, `EduType` is being exported from the interfaces file, however it was not exported through the index file. The old webpack bundling didn't care and would export it out of the package anyways, Rollup will not export it out of the package if it's not exported through the index file. We do have a test that references `EduType`, however it is referenced directly from the file not the index file.
> 
> For 2, There is no problem referencing the members of the enums, only the enum objects themselves. For example, this issue was brought to my attention with someone trying to assign a variable to the enum object ex. `let foo = app.FailedReason`. This is where the type aliasing that the rollup bundler does fails as referencing the enum object's values still works fine, ie. `let foo = app.Failedreason.AuthFailed`. I assume we don't have any tests that try to assign any variables to any enum objects themselves.

I feel like this must be a dumb question for some reason I can't think of but...why are they assigning an entire enum in our library to a variable?

jadahiya-MSFT (approved):

TrevorJoelHarris (approved):

PR #2623, opened 2024-11-15
Added optional FeatureSet parameter to eligibility interface

lakhveerk:

## Description
Added optional parameter featureSet to the AppEligibilityInformation interface.

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |


packages/teams-js/src/public/interfaces.ts
```diff
@@ -1148,6 +1148,26 @@ export interface AppEligibilityInformation {
...
+ */
+export interface FeatureSet {
+  /**
+   * UX Feature set
+   */
+  uxFeatures: string[];
+  /**
+   * Server Feature set
+   */
+  serverFeatures: string[];
```
erinha, line 1170: Consider alphabetizing? (This would apply to host sdk changes as well)

packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -319,6 +338,28 @@ describe('copilot', () => {
 
         await expect(promise).rejects.toThrowError('Error deserializing eligibility information');
       });
+
+      it('getEligibilityInfo should throw if AppEligibilityInformation.featureSet.serverFeatures or uxFeatures is undefined', async () => {
```
erinha, line 342: Suggest adding two separate unit tests, one for `serverFeatures` and one for `uxFeatures` being undefined so it's clearer if there was a failure, one which caused it.

erinha (approved):

erinha (dismissed):

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1150,6 +1154,22 @@ export interface AppEligibilityInformation {
...
+ */
+export interface FeatureSet {
+  /**
+   * Server Feature set
+   */
+  serverFeatures: string[];
+  /**
+   * UX Feature set
+   */
+  uxFeatures: string[];
```
AE-MS, line 1170: Should these be declared as `ReadonlyArray<string>` to ensure they aren't accidentally changed and to express the semantic that they're an "input only" property (e.g. they are immutable properties from the host)?

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1132,6 +1132,10 @@ export interface AppEligibilityInformation {
    * A user will be in at most one cohort.
    */
   cohort: Cohort | null;
+  /**
+   * Feature Sets
+   */
+  featureSet?: FeatureSet;
```
AE-MS, line 1138: >? [](http://example.com/codeflow?start=12&length=1)

What would be the semantic difference between having this property be `undefined` vs defined but with two empty array properties?

I'm guessing the answer is something like "this has to have the `?` because older version of the host won't know to send this property so it will show up as undefined then." If that's true, then consider adding documentation for the property that says:
"If this property is undefined, it indicates that the host is an older version that doesn't support this property."

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/interfaces.ts
erinha: My understanding is it's the latter, so I think adding a comment is the way to go here.

lakhveerk: yep it is for back compatibility. Added the comment!

packages/teams-js/src/public/interfaces.ts
lakhveerk: ReadonlyArray makes sense. Updated the code

AE-MS (approved):
:shipit:

PR #2631, opened 2024-11-20
Disable iOS E2E pipelines until they can be made stable

TrevorJoelHarris:
## Description

These jobs are failing near-deterministically due to issues outside of our control having to do with the Mac image we've been given. Disabling these for now, and will open a bug for someone to re-enable them once they can be made stable.
### Main changes in the PR:

Disabled 2 iOS E2E jobs in the pipeline

### Validation performed:

None

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

No, change doesn't affect code we ship to customers

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -218,26 +218,26 @@ extends:
                   shardNum: 2
                   shardIndex: 1
 
-          - job: E2ETestIOS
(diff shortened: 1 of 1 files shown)


erinha (approved):

AE-MS (approved):
:shipit:

PR #2647, opened 2024-12-04
add ics filter params

YinHang2515:

## Description

re-add otherAppState and Store e2e tests & add store size interface & add ics filter params

### Main changes in the PR:

1. e2e-test-data: otherAppState.json & store.json
2. store.ts

### Validation performed:

1. use store.json e2e test file

### Unit Tests added:

No. No need for this change

### End-to-end tests added:

Yes

### Change file added:

Yes


YinHang2515:
pending until hub-sdk pr checked in

packages/teams-js/src/private/store.ts
```diff
@@ -39,14 +40,25 @@ export enum StoreDialogType {
...
+
 /**
  * @beta
  * @hidden
  * Interface of open full store, copilot store and in-context-store function parameter
  * @internal
  * Limited to Microsoft-internal use
  */
-export interface OpenFullStoreAndICSParams {
+export interface OpenFullStoreAndICSParams extends StoreSizeInfo {
```
TrevorJoelHarris, line 61: If I remember correctly, this interface can be used to either open the full store (a full screen experience) or the ICS store (a sort of dialog).

I don't think developers should be allowed to set a dialog size if they are opening a full store, that value wouldn't have any meaning. Do you agree?

packages/teams-js/src/private/store.ts
```diff
@@ -150,6 +162,7 @@ export async function openStoreExperience(openStoreParams: OpenStoreParams): Pro
       openStoreParams.dialogType,
       (openStoreParams as OpenAppDetailParams).appId,
       (openStoreParams as OpenSpecificStoreParams).collectionId,
```
TrevorJoelHarris, line 164: this function should be validating that the size being passed in is legal values. Off the top of my head, this is probably at least that neither size is < 0, but you may want to put other restrictions on it such as more reasonable minimums and maximums. Without that a dev could ask you to open a 10000x10000px dialog

packages/teams-js/src/private/store.ts
```diff
@@ -39,14 +40,25 @@ export enum StoreDialogType {
...
 
+/**
+ * @beta
+ * @hidden
+ * Interface of store dialog size
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export interface StoreSizeInfo {
+  size?: DialogSize;
```
TrevorJoelHarris, line 51: I actually don't have a problem with you re-using DialogSize here, but you should be aware that that object lets devs either specify specific numbers for the size *or* a more abstract small/medium/large. If you want to support the more abstract sizes that's fine, just be aware that it means the host will need to decide how big to make the dialogs in those cases.

packages/teams-js/src/private/store.ts
```diff
@@ -150,6 +162,7 @@ export async function openStoreExperience(openStoreParams: OpenStoreParams): Pro
       openStoreParams.dialogType,
       (openStoreParams as OpenAppDetailParams).appId,
       (openStoreParams as OpenSpecificStoreParams).collectionId,
+      JSON.stringify(openStoreParams.size),
```
TrevorJoelHarris, line 165: There should be new unit tests for this change

packages/teams-js/src/private/store.ts
```diff
@@ -39,14 +40,25 @@ export enum StoreDialogType {
...
 }
 
+/**
+ * @beta
+ * @hidden
+ * Interface of store dialog size
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export interface StoreSizeInfo {
```
TrevorJoelHarris, line 50: Because this is optional you should document what happens if the size is omitted (for example, you could say something like "the host will choose an appropriate size for you".

change/@microsoft-teams-js-dda2304f-2384-4778-83a4-caa10d929abf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "re-add otherAppState and store e2e tests & add store size info interface",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Add ability to specify size of store dialogs",
```

packages/teams-js/src/private/store.ts
YinHang2515: Actually a 'full store' is also a dialog, which means opening a store without navigation to a specific collection in a dialog

packages/teams-js/src/private/store.ts
YinHang2515: added

packages/teams-js/src/private/store.ts
YinHang2515: updated

packages/teams-js/src/private/store.ts
YinHang2515: I've gone through the whole flow of dialog service and the dialog size will be handled in the implementation of dialog service on each hub/host. Let's take Harmony as an example.
```
// this is the implementation of dialog service in Harmony
const TaskModuleDialogControl = (size: DialogSize) => {
        // windowHeight is the height of current browser window
        const { height: windowHeight, width: windowWidth } =
            useSingletonWindowSize();

        const [sizes, setSizes] = useState<DialogSize>({
            height: getHeight(windowHeight, size.height),
            width: getWidth(windowWidth, size.width),
        });
}
export const getHeight = (
    windowHeight: number,
    height: number | DialogDimension | undefined,
): number | DialogDimension => {
    if (!height) return DialogDimension.Medium;
    if (typeof height === 'number') {
        return clampDimension(windowHeight, height);
    }

    switch (height as DialogDimension) {
        case DialogDimension.Small:
            return windowHeight * 0.2;
        case DialogDimension.Medium:
            return windowHeight * 0.5;
        case DialogDimension.Large:
            return windowHeight * 0.66;
    }
};
```

And when implementing the openStore api in storeService, it will finally call and pass the size info to dialog.open() in dialog service, which means the dialog service in each host will take care of the size info in the end, no matter the size is number out of bound or an abstract size.
```
export class StoreService implements IStoreService {
  constructor(protected dialogService: IDialogService) {}

  public async openStoreExperience(params: OpenStoreParams): Promise<OpenStoreResults> {
    const { dialogType, size} = params;
    const url = getUrlFromDialogType(dialogType);
    const dialogInfo: DialogWithUrl = {
      url: urlFormatted,
      type: DialogContentType.Url,
    };
    await this.dialogService!.urlService!.open(dialogInfo);
  }
}
```

TrevorJoelHarris: We can certainly let hosts decide what to cap dialog sizes at, although I worry a little bit that many of them might not think of that case and could run into bugs where they have to cap the size because someone asked for one that was too large.

It *does* seem though that teams-js should at least validate that the dialog size is > 0. It won't ever be valid to ask for a dialog that is 0px high or wide, and if we catch that in teams-js we can fail faster and let the developer know that the dialog won't be opening asap without them having to wait for a round trip to the host.

TrevorJoelHarris (changes requested):
Please validate that dialog sizes aren't too small. It will never be valid for a developer to pass up a size <= 0 for either height or width.

packages/teams-js/src/private/store.ts
YinHang2515: added unit test & e2e test

packages/teams-js/src/private/store.ts
```diff
@@ -110,6 +110,15 @@ export interface OpenSpecificStoreParams extends StoreSizeInfo {
...
 export type OpenStoreParams = OpenFullStoreAndICSParams | OpenAppDetailParams | OpenSpecificStoreParams;
 
+/**
+ * @beta
+ * @hidden
+ * error message when getting illegal store dialog size
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export const errorInvalidDialogSize = 'Invalid store dialog size';
```
TrevorJoelHarris, line 120:
```suggestion
const errorInvalidDialogSize = 'Invalid store dialog size';
```

No need to export this. I'd actually like this to be moved into the function it's used in as well. Since we aren't using this string anywhere else there's no need to put it in the global namespace

TrevorJoelHarris: This was resolved but no change was made and no explanation was given as to why no change was made.

packages/teams-js/src/private/store.ts
```diff
@@ -42,15 +43,64 @@ export enum StoreDialogType {
...
+  /**
+   * The application meta capabilities (e.g., ["copilotPlugins", "copilotExtensions"]).
+   */
+  appMetaCapabilities?: string[];
+
+  /**
+   * The installation scope (e.g., "Personal" | "Team").
+   * Defaults to "Personal".
+   */
+  installationScope?: string;
```
TrevorJoelHarris, line 98: Things like appCapability, appMetaCapabilities, installationScope, and filteredOutAppIds should be more strongly typed than just strings and arrays of strings - just using strings here will make it very easy for consumers of this sdk to pass in invalid input - for example, what happens if they set `installationScope` to "personal" or "personal " or "Person"? Generally, you want to make app developer facing things as strongly typed as possible so that you can help them eliminate those kind of invalid input errors. When you that, depending on *how* you do it you may need to make those types serializable so that they can be stringified and passed to the host successfully.

Off the top of my head, the parameters you are adding here fall into two buckets: 

1) `appCapability`, `appMetaCapabilities`, and `installationScope` should all be enums (like `StoreDialogType` above). This will make sure that your code in the host only receives values that it is expecting. Going this way means that introducing *new* values will require future changes to this code.

2) filteredOutAppIds should use the `AppId` class, defined in `appId.ts`. That file includes serialization and validation logic for AppIds.

TrevorJoelHarris (changes requested):
You've added some new functionality to this since I last reviewed it, so left some updated requests around typing.

packages/teams-js/src/private/store.ts
YinHang2515: This error message is be veried in unit test, so I make it export, like other error scenario.

packages/teams-js/src/private/store.ts
YinHang2515: Updated

TrevorJoelHarris (dismissed):

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +92,103 @@ export interface OpenAppDetailParams {
...
  * @beta
  * @hidden
- * error message when getting wrong collection id or missing collection id
+ * Api to open an in-context-store dialog
  * @internal
  * Limited to Microsoft-internal use
  */
-export const errorMissingCollectionId =
-  'No Collection Id present, but CollectionId needed to open a store specific to a collection';
+export async function openInContextStore(params: OpenInContextStoreParams): Promise<void> {
```
TrevorJoelHarris, line 156: Since this is a private function I don't feel *strongly* about it, but right now every parameter in `OpenInContextStoreParams` is optional. To me, that means it would make sense if `params` was optional here as well. That way a caller who wanted *all* of the default values could write code like this:

```typescript
await openInContextStore();
```

instead of this:

```typescript
await openInContextStore({});
```

Something to consider.

TrevorJoelHarris (changes requested):
Couple very small comments/questions about optional parameters but looks mostly good

packages/teams-js/src/private/store.ts
YinHang2515: updated

packages/teams-js/src/private/store.ts
```diff
@@ -12,46 +14,57 @@ import { runtime } from '../public/runtime';
...
+  size?: DialogSize;
+
+  /**
+   * The application capability (e.g., "Tab", "Bot", "Messaging", "Connector", "CUSTOMBOT").
+   * Defaults to "Tab".
+   */
+  appCapability?: string;
+
+  /**
+   * The application meta capabilities (e.g., ["copilotPlugins", "copilotExtensions"]).
```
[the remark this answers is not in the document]
YinHang2515: updated

change/@microsoft-teams-js-ea7c4e5c-5826-4103-a2b8-f539b122546f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "separate store service message handler for each type of store & add new handler for ics using filter parameters & add dialog size parameter & enable store service e2e tests",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Added functions for opening different types of stores along with ics filtering",
```

TrevorJoelHarris (approved):

packages/teams-js/src/private/store.ts
```diff
@@ -12,46 +14,58 @@ import { runtime } from '../public/runtime';
...
  * @hidden
- * Interface of open full store, copilot store and in-context-store function parameter
+ * Interface for opening the in-context store function parameters
  * @internal
  * Limited to Microsoft-internal use
  */
-export interface OpenFullStoreAndICSParams {
+export interface OpenInContextStoreParams {
+  /**
+   * the store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size
```
jekloudaMSFT, line 41: Nit:
```suggestion
   * The store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size.
```

packages/teams-js/src/private/store.ts
```diff
@@ -12,46 +14,58 @@ import { runtime } from '../public/runtime';
...
- * Enum of store dialog type
+ * Interface for opening the full store function parameters
  * @internal
  * Limited to Microsoft-internal use
  */
-export enum StoreDialogType {
+export interface OpenFullStoreParams {
   /**
-   * open a store without navigation
+   * the store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size
```
jekloudaMSFT, line 27: Nit:
```suggestion
   * The store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size.
```

packages/teams-js/src/private/store.ts
```diff
@@ -60,15 +74,17 @@ export interface OpenFullStoreAndICSParams {
...
-   * need to be app detail type, defined by {@link StoreDialogType}
-   */
-  dialogType: StoreDialogType.AppDetail;
   /**
    * app id of the dialog to open
    */
   appId: AppId;
+
+  /**
+   * the store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size
```
jekloudaMSFT, line 83:
```suggestion
   * The store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size.
```

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
...
-   * need to be specific store type, defined by {@link StoreDialogType}
-   */
-  dialogType: StoreDialogType.SpecificStore;
   /**
    * collection id of the plugin store to open
    */
   collectionId: string;
+
+  /**
+   * the store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size
```
jekloudaMSFT, line 102:
```suggestion
   * The store dialog size, defined by {@link DialogSize}, if not present, the host will choose an appropriate size.
```

packages/teams-js/src/private/store.ts
```diff
@@ -163,3 +199,25 @@ export async function openStoreExperience(openStoreParams: OpenStoreParams): Pro
 export function isSupported(): boolean {
   return ensureInitialized(runtime) && !!runtime.supports.store;
 }
+
+function ensureStoreReady(): void {
```
jekloudaMSFT, line 203: Why do we need this separate function for checking if we're in the right contexts? Can this logic just be moved into `isSupported`?

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
...
  */
 export interface OpenSpecificStoreParams {
-  /**
-   * need to be specific store type, defined by {@link StoreDialogType}
-   */
-  dialogType: StoreDialogType.SpecificStore;
   /**
    * collection id of the plugin store to open
    */
   collectionId: string;
```
jekloudaMSFT, line 99: Is there any format we can validate here (e.g. UUID, AppID), or can this be any string?

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
  * Limited to Microsoft-internal use
  */
 export interface OpenSpecificStoreParams {
-  /**
-   * need to be specific store type, defined by {@link StoreDialogType}
-   */
-  dialogType: StoreDialogType.SpecificStore;
   /**
    * collection id of the plugin store to open
```
jekloudaMSFT, line 97:
```suggestion
   * Collection ID of the plugin store to open.
```

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
...
+const errorInvalidDialogSize = 'Invalid store dialog size';
+const errorMissingAppId = 'No App Id present, but AppId needed to open AppDetail store';
+const errorMissingCollectionId =
+  'No Collection Id present, but CollectionId needed to open a store specific to a collection';
 
 /**
  * @beta
  * @hidden
- * error message when getting invalid store dialog type
+ * Api to open a full store without navigation
```
jekloudaMSFT, line 116:
```suggestion
 * API to open a full store without navigation
```

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
...
+    [serializeValidSize(size)],
+    getApiVersionTag(StoreVersionTagNum, ApiName.Store_OpenFullStore),
+  );
+}
+
 /**
  * @beta
  * @hidden
- * error message when getting wrong app id or missing app id
+ * Api to open an app detail dialog
```
jekloudaMSFT, line 133:
```suggestion
 * API to open an app detail dialog
```

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
...
+    [serializeValidSize(size), appId],
+    getApiVersionTag(StoreVersionTagNum, ApiName.Store_OpenAppDetail),
+  );
+}
+
 /**
  * @beta
  * @hidden
- * error message when getting wrong collection id or missing collection id
+ * Api to open an in-context-store dialog
```
jekloudaMSFT, line 153:
```suggestion
 * API to open an in-context store dialog
```

packages/teams-js/src/private/store.ts
```diff
@@ -77,83 +93,103 @@ export interface OpenAppDetailParams {
...
+}
+
 /**
  * @beta
  * @hidden
- * Api to open a store
- *
- * @param openStoreParams - params to call openStoreExperience
- *
+ * Api to open an store with navigation to a specific collection
```
jekloudaMSFT, line 176:
```suggestion
 * API to open a store with navigation to a specific collection
```

packages/teams-js/src/private/store.ts
YinHang2515: As the store service is accessibile under the context of FrameContexts.content, FrameContexts.sidePanel, FrameContexts.meetingStage, while in other context the error is not expected be thrown, thus it's separated

jekloudaMSFT: So we want `isSupported` to return true (and not throw) in all contexts, but the store functions themselves should throw when they're called in the contexts other than content sidePanel and meetingStage? Is that right?

YinHang2515: yes, and similar code pattern can be found in other services

jekloudaMSFT: Yep, looks good, thanks!

packages/teams-js/src/private/store.ts
YinHang2515: the collection id is a string like 'copilotextensions' offered from Teams, not like a real app id. and for sure on Teams side, they have a way to validate so that the reason im doing that

jekloudaMSFT (approved):

PR #2649, opened 2024-12-04
Add custom child iframe in teams test app for nested app auth e2e tests

baljesingh:
## Description

> Currently, the same test app is displayed within a nested iframe, which can lead to issues like timeouts for mobile SDKs during testing. Additionally, it may introduce future complications due to initialization steps. This PR removes the dependency on the parent app and replaces it with a custom iframe featuring a minimal design. The custom iframe includes two buttons: one for sending messages to the bridge and another for sending messages to the top window.

### Main changes in the PR:

1. Added naa_childIframe.html.
2. Updated the nested child iframe source to naa_childIframe.html.

### Validation performed:

Locally with other SDKs

### Unit Tests added:

N/A

### End-to-end tests added:

N/A

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| Before | After|

### DIFF
--- a/apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
+++ b/apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
@@ -175,6 +175,20 @@ const NestedAppAuthAPIs = (): ReactElement => {
 
       iframeContainer.appendChild(childIframe);
       setIframeAdded(true);
+
+      // Send payload to the iframe after it loads
+      childIframe.onload = () => {
+        const payloads = {
+          defaultPayloadForBridge: NestedAppAuthRequest,
+          defaultPayloadForTopWindow: JSON.stringify({
+            id: '2',
+            func: 'nestedAppAuth.execute',
+            args: [],
+            data: NestedAppAuthRequest,
+          }),
+        };
+        childIframe.contentWindow?.postMessage(payloads, window.location.origin);
+      };
     };
 
     return (
--- a/apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
+++ b/apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
@@ -152,26 +152,25 @@ const NestedAppAuthAPIs = (): ReactElement => {
       }),
     });
 
-  const AddChildIframeSection = (): React.ReactElement | null => {
+  const AddChildIframeSection = (): React.ReactElement => {
     const [iframeAdded, setIframeAdded] = useState(false);
 
     const addChildIframe = (): void => {
       if (iframeAdded) {
-        console.log('Iframe already added.');
         return;
       }
 
       const iframeContainer = document.getElementById('nestedChildIframeContainer');
       if (!iframeContainer) {
-        console.error('Container not found: nestedChildIframeContainer');
+        console.error('Iframe container not found');
         return;
       }
 
       const childIframe = document.createElement('iframe');
-      childIframe.src = `${window.location.href}?appInitializationTest=true&groupedMode=NestedAppAuthAPIs`;
+      childIframe.src = '/naa_childIframe.html';
       childIframe.id = 'nestedAuthChildIframe';
-      childIframe.width = '100%';
-      childIframe.height = '400px';
+      childIframe.style.width = '100%';
+      childIframe.style.height = '400px';
       childIframe.style.border = 'none';
 
       iframeContainer.appendChild(childIframe);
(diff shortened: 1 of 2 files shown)


apps/teams-test-app/src/public/naa_childIframe.html
```diff
@@ -0,0 +1,219 @@
...
+
+      // Utility function to display a response in a given element
+      const displayResponse = (element, message) => {
+        element.innerText = message;
+      };
+
+      // Predefined payloads
+      const defaultPayloadForBridge =
+        '{"messageType":"NestedAppAuthRequest","method":"GetToken","sendTime":1732269811006,"clientLibrary":"testOS","clientLibraryVersion":"1.0.0","requestId":"684352c2-7ab7-4def-b7e1-XXXXXXXXXXX","tokenParams":{"correlationId":"39dc85fe-9054-11ed-a1eb-XXXXXXXXXXX"}}';
+      const defaultPayloadForTopWindow =
```
TrevorJoelHarris, line 119: With this approach, if the NAA payload changes shape at all (name changes, property additions, etc.) how will someone know how to find and update these hardcoded strings? It might be easier to use TeamsJS directly the same way that the rest of the teams-test-app does

TrevorJoelHarris (changes requested):
I'm worried about hardcoding post message payloads into the html file as opposed to calling libraries that will create those paylods for you (like TeamsJS)

apps/teams-test-app/src/public/naa_childIframe.html
```diff
@@ -0,0 +1,219 @@
...
+          const bridge = window.nestedAppAuthBridge;
+          if (!bridge) {
+            displayResponse(responseElement, 'NAA Bridge not available');
+            return;
+          }
+
+          bridge.postMessage(JSON.stringify(payload));
+          displayResponse(responseElement, 'Message sent to Bridge. Awaiting response...');
+
+          bridge.addEventListener('message', (response) => {
```
jekloudaMSFT, line 165: Is there any risk of missing the response if we wait to add the event listener after posting the message to the NAA bridge?

baljesingh: That's a good callout. I will move the addEventListener before the postMessage call. Thanks

maglims:
Thanks for the detailed PR descriptions!

I am wondering if there is any need to keep the scenario/code support of "nested child app which is also a metaos app", just for some reversal testing later for hub sdk.

apps/teams-test-app/src/public/naa_childIframe.html
TrevorJoelHarris: Okay, I see that you've just taken the default payload from NestedAppAuthAPIs.tsx (after you pointed it out to me). Could you pass the payload from NestedAppAuthAPIs.tsx into here so that we don't have duplicate definitions?

baljesingh: Thanks. As discussed offline, the payload is now passed from the parent to the child, eliminating the need for hardcoded payloads in the child.

baljesingh:
> Thanks for the detailed PR descriptions!
> 
> I am wondering if there is any need to keep the scenario/code support of "nested child app which is also a teams app", just for some reversal testing later for host sdk.

That's a good question. I'm not sure if I fully understand the term "reversal testing" for the host SDK correctly but currently, there are no scenarios where a nested child uses TeamsJS, and TeamsJS neither supports nor encourages NAA in nested iframes. Including this section without a concrete use case might lead to confusion. I’ll consider adding it later if such a use case arises, though the likelihood is minimal.

TrevorJoelHarris (approved):

jekloudaMSFT (approved):

apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
```diff
@@ -152,36 +152,58 @@ const NestedAppAuthAPIs = (): ReactElement => {
...
+
+      // Send payload to the iframe after it loads
+      childIframe.onload = () => {
+        const payloads = {
+          defaultPayloadForBridge: NestedAppAuthRequest,
+          defaultPayloadForTopWindow: JSON.stringify({
+            id: '2',
+            func: 'nestedAppAuth.execute',
+            args: [],
+            data: NestedAppAuthRequest,
```
KangxuanYe, line 187: Hmmm generally speaking, all of functions will have an api version tag and yours also break the rule here because you postMessage directly to the iframe using `postMessage` directly. Could you please add a version tag?

KangxuanYe (approved):

apps/teams-test-app/src/components/NestedAppAuthAPIs.tsx
KangxuanYe: Okay @baljesingh  says that he would create a ticket for tracking adding the tag. Because if it is not added, telemetry on this API will have `undefined_tag` in our database, which doesn't make sense.

baljesingh: As discussed offline, Api version tag need to be added in NAA flow as well. I will handle this later. Filling the bug now.

PR #2651, opened 2024-12-05
Fix publishBuildArtifact task issue

KangxuanYe:
**Summary**

This pull request addresses an issue with the PublishBuildArtifacts@1 task in the pipeline by zipping the logs directory before publishing. And we changed the logic to only publishing zip files for debugging purpose when E2E test failed.

Based on our observations and experience, we can confirm that when we use new version of Xcode, after E2E test completes and if there is any error, it will create a lot of files all under .xcresult as below (.xcresult has a lot of hidden files on MacOS) 

```
`DebugLogs/output.xcresult/Debug description for `StaticText (Elements matching predicate 'value CONTAINS "Error Error InvalidAuthUrl"')`_1_533C9087-ED7E-4A94-8048-26DD268E2FBA.txt`
```

When we try to publish the artifacts, the task will scan each of them and they may have problems figuring out
1. files with super long name
2. files whose name has some special characters, i.e. `:`

And then they will get stuck here

```
Exception: TF10123: The path 'DebugLogs/output.xcresult/Debug description for `StaticText (Elements matching predicate 'value CONTAINS "Error Error InvalidAuthUrl"')`_1_533C9087-ED7E-4A94-8048-26DD268E2FBA.txt' contains the character '"'. Remove the '"' and try again. 
```

We can neither use lower version of Xcode, nor control the file names be created. Xcode version is coupled with MacOS system and file names are all created by Xcode automatically. However, by zipping this file, it seems like we can prevent them scanning files within .xcresult file and upload them as a whole zip file so that they would not fail. ( However, I still cannot confirm this because the team who built this has not replied to me yet.)

jobs/templates/common-tasks.yml: Added a step to zip the logs directory and updated the PublishBuildArtifacts task to publish the zipped logs file.

### DIFF
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -6,6 +6,12 @@ parameters:
     default: iosE2ETestPlan
     type: string
 
+variables:
+  - name: logsDirectory
+    value: '$(agent.buildDirectory)/Logs/$(agent.JobName)'
+  - name: debugLogsZip
+    value: '$(agent.JobName)_DebugLogs.zip'
+
 steps:
   - checkout: self
   - checkout: ${{ parameters.iOSAppHostingSdkGitPath }}
@@ -137,26 +143,26 @@ steps:
 
   - bash: |
       output="$(agent.buildDirectory)/Logs/$(agent.JobName)"
-      rm -rf "${output}" > /dev/null
+      rm -rf "$(logsDirectory)" > /dev/null
 
-      mkdir -p "${output}"
+      mkdir -p "$(logsDirectory)"
 
-      cp -r $(Agent.BuildDirectory)/iOSHost/TestResults "${output}/output.xcresult"
-      xchtmlreport -r ${output}/output.xcresult -j
+      cp -r $(Agent.BuildDirectory)/iOSHost/TestResults "$(logsDirectory)/output.xcresult"
+      xchtmlreport -r $(logsDirectory)/output.xcresult -j
 
 
-      echo "ls -lR ${output}"
-      ls -lR "${output}"
+      echo "ls -lR $(logsDirectory)"
+      ls -lR "$(logsDirectory)"
 
       echo "Zipping logs directory..."
-      cd "$(agent.buildDirectory)/Logs/$(agent.JobName)"
-      zip -r "$(agent.JobName)_DebugLogs.zip" .
+      cd "$(logsDirectory)"
+      zip -r "$(debugLogsZip)" .
     displayName: Preparations for publishing results
     condition: always()
 
   - task: 1ES.PublishPipelineArtifact@1
     inputs:
-      path: '$(agent.buildDirectory)/Logs/$(agent.JobName)/$(agent.JobName)_DebugLogs.zip'
+      path: '$(logsDirectory)/$(debugLogsZip)'
       artifact: iOSDebugLogs - ${{ parameters.testPlan }} - Attempt $(System.JobAttempt)
     condition: always()
     continueOnError: true
--- a/tools/yaml-templates/ios-test.yml
+++ b/tools/yaml-templates/ios-test.yml
@@ -6,12 +6,6 @@ parameters:
     default: iosE2ETestPlan
     type: string
 
-variables:
-  - name: logsDirectory
-    value: '$(agent.buildDirectory)/Logs/$(agent.JobName)'
-  - name: debugLogsZip
-    value: '$(agent.JobName)_DebugLogs.zip'
-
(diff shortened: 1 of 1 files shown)


baljesingh (dismissed):
LGTM

tools/yaml-templates/ios-test.yml
```diff
@@ -147,11 +147,16 @@ steps:
 
       echo "ls -lR ${output}"
       ls -lR "${output}"
+
+      echo "Zipping logs directory..."
+      cd "$(agent.buildDirectory)/Logs/$(agent.JobName)"
```
AE-MS, line 152: >$(agent.buildDirectory)/Logs/$(agent.JobName) [](http://example.com/codeflow?start=10&length=45)

It looks like there is a variable already set to this above (`output`), can that variable be reused here?

tools/yaml-templates/ios-test.yml
```diff
@@ -147,11 +147,16 @@ steps:
...
+      zip -r "$(agent.JobName)_DebugLogs.zip" .
```
AE-MS, line 153: >$(agent.JobName)_DebugLogs.zip [](http://example.com/codeflow?start=14&length=30)

Can this string be set to a YAML variable/constant so you don't have to duplicate it here and then again on line 159?

Perhaps you could do the same with the path? E.g.
```
variables:
  logsDirectory: "$(agent.buildDirectory)/Logs/$(agent.JobName)"
  debugLogsZip: "$(agent.JobName)_DebugLogs.zip"
...
      echo "Zipping logs directory..."
      cd "${{ variables.logsDirectory }}"
      zip -r "${{ variables.debugLogsZip }}" .
...
  - task: 1ES.PublishPipelineArtifact@1
    inputs:
      path: '${{ variables.logsDirectory }}/${{ variables.debugLogsZip }}'
```

...or something like that?

AE-MS (changes requested):
:clock1:

AE-MS:
Can you add some detail to the PR description that describes why zipping the logs directory fixes the problem?

tools/yaml-templates/ios-test.yml
KangxuanYe: We can have variables but that will be defined in `azure-pipelines.yml`. I feel like in that way it doesn't make too much sense to other yml files. You can check here that I made a change under iOS yml file and it cannot be parsed correctly. So maybe I'll keep it as is?

https://office.visualstudio.com/ISS/_build/results?buildId=33089221&view=results

tools/yaml-templates/ios-test.yml
KangxuanYe: Yes, it will be reused.

KangxuanYe:
> Can you add some detail to the PR description that describes why zipping the logs directory fixes the problem?

Sure I can write some of them. This would be a guess and maybe I shouldn't say it is a fix because we cannot confirm it really fix the issue . 

Could you please take another look to see if description makes sense to you?

AE-MS:
> > Can you add some detail to the PR description that describes why zipping the logs directory fixes the problem?
> 
> Sure I can write some of them. This would be a guess and maybe I shouldn't say it is a fix because we cannot confirm it really fix the issue .
> 
> Could you please take another look to see if description makes sense to you?

This description is great! Very clear!

tools/yaml-templates/ios-test.yml
AE-MS: I didn't know that YAML imposes such a restriction on where variables can be defined. Seems weird. I guess my encouragement then would be "reuse as much as you can to avoid duplication within the limits of what YAML allowsl."

AE-MS (approved):
:shipit:

tools/yaml-templates/ios-test.yml
KangxuanYe: YAML file restrictions are weird. It looks like you are writing some type of code but actually it doesn't. Even if you take look on their documentation, their examples show that you can add `variables` under the job. But only when you write something and give it a try, you know that `variables` can only be defined in the most upper level of yml file.

jadahiya-MSFT (approved):

maglims:
Thanks for writing the detailed description.

I understand there are apple's issue... but it feels very strange how they randomly define the file name(based on the failures it looks) without providing a way to customize that... Does it still worth to create a follow up work item to check back later on this to see if Apple makes any improvements? @baljesingh  @KangxuanYe

KangxuanYe:
@maglims, it is good to follow with Apple or Artifact team to see whether they can fix it to make our life easier. It will depend on how much bandwidth/time/effort we have.

PR #2654, opened 2024-12-10
Added customTelemetry subcapability under copilot 

ndangudubiyyam:

## Description
Enable the Copilot app to send custom telemetry data to host apps for tracking UI element loads within Biz Chat.

### Main changes in the PR:

1. Added a sub capability under copilot called customTelemetry.sendCustomTelemetryData()
2. Export the above capability in copilot.ts file.

### Validation performed:

1. Added Unit tests
2. Added UI validation to test app

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

No

### Change file added: Yes

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
@@ -1,9 +1,8 @@
-import { copilot } from '@microsoft/teams-js';
+import { copilot, UUID } from '@microsoft/teams-js';
 import React, { ReactElement } from 'react';
 
 import { ApiWithoutInput, ApiWithTextInput } from '../utils';
 import { ModuleWrapper } from '../utils/ModuleWrapper';
-import { UUID } from 'crypto';
 
 const CopilotAPIs = (): ReactElement => {
   const CheckCopilotEligibilityCapability = (): ReactElement =>
@@ -35,11 +34,17 @@ const CopilotAPIs = (): ReactElement => {
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
         validateInput: (_input) => {},
         submit: async (input) => {
-          const result = await copilot.customTelemetry.sendCustomTelemetryData(input.name, input.timestamp);
+          const result = await copilot.customTelemetry.sendCustomTelemetryData(
+            input.stageNameIdentifier,
+            input.timestamp,
+          );
           return JSON.stringify(result);
         },
       },
-      defaultInput: JSON.stringify({ name: copilot.Stage.STAGE_E, timestamp: Date.now() }),
+      defaultInput: JSON.stringify({
+        stageNameIdentifier: new UUID('805a4340-d5e0-4587-8f04-0ae88219699f'),
+        timestamp: Date.now(),
+      }),
     });
 
   return (
--- a/apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
@@ -3,6 +3,7 @@ import React, { ReactElement } from 'react';
 
 import { ApiWithoutInput, ApiWithTextInput } from '../utils';
 import { ModuleWrapper } from '../utils/ModuleWrapper';
+import { UUID } from 'crypto';
 
 const CopilotAPIs = (): ReactElement => {
   const CheckCopilotEligibilityCapability = (): ReactElement =>
@@ -25,7 +26,7 @@ const CopilotAPIs = (): ReactElement => {
 
   const SendCustomTelemetryData = (): ReactElement =>
     ApiWithTextInput<{
-      name: copilot.Stage;
+      stageNameIdentifier: UUID;
       timestamp: number;
     }>({
       name: 'sendCustomTelemetryData',
--- a/apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
+++ b/apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
@@ -1,7 +1,7 @@
 import { copilot } from '@microsoft/teams-js';
 import React, { ReactElement } from 'react';
 
-import { ApiWithoutInput } from '../utils';
+import { ApiWithoutInput, ApiWithTextInput } from '../utils';
 import { ModuleWrapper } from '../utils/ModuleWrapper';
 
 const CopilotAPIs = (): ReactElement => {
@@ -23,11 +23,34 @@ const CopilotAPIs = (): ReactElement => {
       },
     });
 
+  const SendCustomTelemetryData = (): ReactElement =>
+    ApiWithTextInput<{
+      name: copilot.Stage;
+      timestamp: number;
+    }>({
+      name: 'sendCustomTelemetryData',
+      title: 'sendCustomTelemetryData',
+      onClick: {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        validateInput: (_input) => {},
+        submit: async (input) => {
+          const result = await copilot.customTelemetry.sendCustomTelemetryData(input.name, input.timestamp);
+          return JSON.stringify(result);
+        },
+      },
+      defaultInput: JSON.stringify({ name: copilot.Stage.STAGE_E, timestamp: Date.now() }),
+    });
+
   return (
-    <ModuleWrapper title="Copilot.Eligibility">
-      <CheckCopilotEligibilityCapability />
-      <GetEligibilityInfo />
-    </ModuleWrapper>
+    <>
+      <ModuleWrapper title="Copilot.Eligibility">
+        <CheckCopilotEligibilityCapability />
+        <GetEligibilityInfo />
+      </ModuleWrapper>
+      <ModuleWrapper title="Copilot.CustomTelemetry">
+        <SendCustomTelemetryData />
+      </ModuleWrapper>
+    </>
   );
 };
 
--- a/change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
+++ b/change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Added `customTelemetry` capability under `copilot` to send app loading data to the host. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
+  "comment": "Added `customTelemetry` capability under `copilot` to send app loading data to the host.",
   "packageName": "@microsoft/teams-js",
   "email": "niharikad@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
+++ b/change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `sendCustomTelemetry` capability under copilot for Biz Chat to send app loading data to Harmony. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
+  "packageName": "@microsoft/teams-js",
+  "email": "niharikad@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
+++ b/change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Added `sendCustomTelemetry` capability under copilot for Biz Chat to send app loading data to Harmony. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
+  "comment": "Added `customTelemetry` capability under `copilot` to send app loading data to the host. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
   "packageName": "@microsoft/teams-js",
   "email": "niharikad@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/teams-js/src/index.ts
+++ b/packages/teams-js/src/index.ts
@@ -1,3 +1,2 @@
 export * from './private/index';
 export * from './public/index';
-export { UUID } from './internal/uuidObject';
--- a/packages/teams-js/src/index.ts
+++ b/packages/teams-js/src/index.ts
@@ -1,2 +1,3 @@
 export * from './private/index';
 export * from './public/index';
+export { UUID } from './internal/uuidObject';
--- a/packages/teams-js/src/internal/uuidObject.ts
+++ b/packages/teams-js/src/internal/uuidObject.ts
@@ -4,14 +4,29 @@ import { generateGUID, validateUuid } from './utils';
  * @internal
  * Limited to Microsoft-internal use
  *
- * UUID object
+ * Represents a UUID (Universally Unique Identifier) object.
+ * This class provides a way to generate, validate, and represent UUIDs as strings.
  */
 export class UUID {
+  /**
+   * Creates an instance of the UUID class.
+   * If no UUID string is provided, a new UUID is generated.
+   *
+   * @param {string} [uuid=generateGUID()] - The UUID string. Defaults to a newly generated UUID.
+   * @throws {Error} - Throws an error if the provided UUID is invalid.
+   */
   public constructor(private readonly uuid: string = generateGUID()) {
     validateUuid(uuid);
   }
 
+  /**
+   * Returns the UUID as a string.
+   *
+   * @returns {string} - The UUID string.
+   */
   public toString(): string {
     return this.uuid;
   }
 }
+
+export default UUID;
--- a/packages/teams-js/src/private/copilot/customTelemetry.ts
+++ b/packages/teams-js/src/private/copilot/customTelemetry.ts
@@ -16,7 +16,7 @@ import { runtime } from '../../public/runtime';
 const copilotTelemetryVersionNumber: ApiVersionNumber = ApiVersionNumber.V_2;
 const copilotLogger = getLogger('copilot');
 
-enum Stage {
+export enum Stage {
   STAGE_E = 'E',
 }
 /**
--- a/packages/teams-js/src/private/copilot/customTelemetry.ts
+++ b/packages/teams-js/src/private/copilot/customTelemetry.ts
@@ -22,8 +22,8 @@ const copilotLogger = getLogger('copilot');
  *
  * @param { UUID } stageNameIdentifier - The stageName UUID identifier for the telemetry data.
  * @param { number } [timestamp=getCurrentTimestamp() ?? Date.now()] - The timestamp of the telemetry data. Defaults to the current timestamp.
- * @returns { Promise<void> } - A promise that resolves when the telemetry data has been sent.
- * @throws { Error } - Throws an error if the app has not been successfully initialized.
+ * @returns { Promise<void> } - promise resolves when the hubsdk acknowledges that it has received the message.
+ * @throws { Error } - Throws an error if the app has not been successfully initialized or the host-sdk returns an error as a response to this call
  *
  * @hidden
  * @internal
@@ -35,7 +35,11 @@ export async function sendCustomTelemetryData(
   timestamp: number = getCurrentTimestamp() ?? Date.now(),
 ): Promise<void> {
   ensureInitialized(runtime);
-  copilotLogger('Sending custom telemetry data to host. to record timestamp: %s', timestamp);
+  copilotLogger(
+    'Sending custom telemetry data to host for stage: %s to record timestamp: %s',
+    stageNameIdentifier,
+    timestamp,
+  );
   return callFunctionInHost(
     ApiName.Copilot_CustomTelemetry_SendCustomTelemetryData,
     [stageNameIdentifier.toString(), timestamp],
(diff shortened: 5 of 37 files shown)


packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -384,4 +384,22 @@ describe('copilot', () => {
...
+  describe('copilot.customTelemetry', () => {
+    describe('sendCustomTelemetryData', () => {
+      it('sendCustomTelemetryData should throw if called before initialization', async () => {
+        expect.assertions(1);
+        utils.uninitializeRuntimeConfig();
+        await expect(copilot.customTelemetry.sendCustomTelemetryData()).rejects.toThrowError(
+          new Error(errorLibraryNotInitialized),
+        );
+      });
+      it('sendCustomTelemetryData should throw if called before initialization', async () => {
```
erinha, line 397: Copy/paste error? This test case is the successful case, right?

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export function sendCustomTelemetryData(
+  name: Stage = Stage.STAGE_E,
+  timestamp: number = getCurrentTimestamp() ?? Date.now(),
+): Promise<void> {
+  ensureInitialized(runtime);
+  copilotLogger('Sending custom telemetry data to host. to record timestamp: %s', timestamp);
```
erinha, line 40: This log statement tripped me up a bit. I think the intent is to log what we are sending to the host? If so, can we include the `name` as well as the `timestamp`?

Probably would capitalize the 't' in "to" but you might reword this if including `name` as well?

packages/teams-js/src/public/runtime.ts
```diff
@@ -236,6 +236,7 @@ interface IRuntimeV4 extends IBaseRuntime {
     readonly conversations?: {};
     readonly copilot?: {
       readonly eligibility?: {};
+      readonly customTelemetry?: {};
```
erinha, line 239: nit: alphabetize

change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `sendCustomTelemetry` capability under copilot for Biz Chat to send app loading data to Harmony. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
```
erinha, line 3:
```suggestion
  "comment": "Added `customTelemetry` capability under `copilot` to send app loading data to the host. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
```

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @returns {Promise<void>} - A promise that resolves when the telemetry data has been sent.
+ * @throws {Error} - Throws an error if the app has not been successfully initialized.
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export function sendCustomTelemetryData(
+  name: Stage = Stage.STAGE_E,
```
TrevorJoelHarris, line 36: I'd prefer that this was a UUID instead of an enum for a few reasons:

- If we use an enum, any time a new stage is added the enum will have to be updated
- If they want to gather telemetry for something that is *not* stage related, we will need to update this

A UUID will allow them to pass in any UUID they want and define what it means for themselves. It also ensures that they don't inadvertently pass personal information that can't be added to telemetry in because we will validate that the UUID is a UUID and not an arbitrary string.

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export enum Stage {
+  STAGE_E = 'E',
+}
+/**
+ * Sends custom telemetry data to the host.
+ *
+ * @param {Stage} [name=Stage.STAGE_E] - The stage of the telemetry data. Defaults to Stage.STAGE_E.
+ * @param {number} [timestamp=getCurrentTimestamp() ?? Date.now()] - The timestamp of the telemetry data. Defaults to the current timestamp.
+ * @returns {Promise<void>} - A promise that resolves when the telemetry data has been sent.
```
TrevorJoelHarris, line 27: This should probably say that the promise resolves when the host acknowledges that it has received the telemetry.

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @throws {Error} - Throws an error if the app has not been successfully initialized.
```
TrevorJoelHarris, line 28: Note: if the host-sdk returns an error as a response to this call then `callFunctionInHost` will throw an error containing that data.

apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
```diff
@@ -23,11 +23,34 @@ const CopilotAPIs = (): ReactElement => {
       },
     });
 
+  const SendCustomTelemetryData = (): ReactElement =>
+    ApiWithTextInput<{
+      name: copilot.Stage;
```
maglims, line 28: nit: how do you think of copilot.Telemetry or copilot.Telemetry.Stage?

It feels copilot.Stage is not obvious enough to be correlated with telemetry by looking at the test app perspective.

apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
```diff
@@ -23,11 +23,34 @@ const CopilotAPIs = (): ReactElement => {
...
-    <ModuleWrapper title="Copilot.Eligibility">
-      <CheckCopilotEligibilityCapability />
-      <GetEligibilityInfo />
-    </ModuleWrapper>
+    <>
+      <ModuleWrapper title="Copilot.Eligibility">
+        <CheckCopilotEligibilityCapability />
+        <GetEligibilityInfo />
+      </ModuleWrapper>
+      <ModuleWrapper title="Copilot.CustomTelemetry">
```
maglims, line 50: nit: for some reason, I am not a fan of the term `Custom` to be honest. There are too many terms contains "Custom" in hubsdk...but feel free to use as this if it is proper.

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ensureInitialized } from '../../internal/internalAPIs';
+import { ApiName, ApiVersionNumber, getApiVersionTag, getLogger } from '../../internal/telemetry';
+import { getCurrentTimestamp } from '../../internal/utils';
+import { runtime } from '../../public/runtime';
+
+const copilotTelemetryVersionNumber: ApiVersionNumber = ApiVersionNumber.V_2;
+const copilotLogger = getLogger('copilot');
+
+export enum Stage {
+  STAGE_E = 'E',
```
maglims, line 20: Adding some comments for StageE? 

Also I am not sure if StageE is good naming... it looks only Copilot folks will know what StageE is .. is it possible to use some generic naming like "NotifyUIUpdateFromAppToHost", or something that reflect StageE more close?

packages/teams-js/src/private/copilot/customTelemetry.ts
ndangudubiyyam: Lets say we switch to UUID design. The Biz chat is sending customTelemetry for both stage E and some other stages(X, Y, Z) in their code. How will the hubs corelate the UUID to stage E , X, Y and Z?

erinha: I could be misunderstanding, but Copilot and the hub would communicate which UUID means what so they can correlate it and understand what the timestamp represents.

erinha: Since the meaning of the `name` isn't strictly relevant to the SDK, this allows the app and host to work together to define well-understood values that they can use to correlate any type of timestamp data in a meaningful way, without having to make SDK updates.

ndangudubiyyam: @TrevorJoelHarris I have updated the PR with the changes. Please take a look and let me know if that is the correct way.

apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
ndangudubiyyam: They are custom telemetry data points, as they don't follow the existing apis.

packages/teams-js/test/private/copilot.spec.ts
ndangudubiyyam: resolved

packages/teams-js/src/private/copilot/customTelemetry.ts
ndangudubiyyam: fixed

packages/teams-js/src/public/runtime.ts
ndangudubiyyam: @erinha  I have a question about this, does it need to be the part of runtime capability?
I mean if the hub doesn't support it, they will ignore these telemetry messages.

packages/teams-js/src/private/copilot/customTelemetry.ts
ndangudubiyyam: fixed it.

packages/teams-js/src/private/copilot/customTelemetry.ts
ndangudubiyyam: Oh. Fixed it

packages/teams-js/src/public/runtime.ts
ndangudubiyyam: done

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -11,36 +11,38 @@ import { callFunctionInHost } from '../../internal/communication';
...
 /**
  * Sends custom telemetry data to the host.
  *
- * @param {Stage} [name=Stage.STAGE_E] - The stage of the telemetry data. Defaults to Stage.STAGE_E.
- * @param {number} [timestamp=getCurrentTimestamp() ?? Date.now()] - The timestamp of the telemetry data. Defaults to the current timestamp.
- * @returns {Promise<void>} - A promise that resolves when the telemetry data has been sent.
- * @throws {Error} - Throws an error if the app has not been successfully initialized.
+ * @param { UUID } stageNameIdentifier - The stageName UUID identifier for the telemetry data.
+ * @param { number } [timestamp=getCurrentTimestamp() ?? Date.now()] - The timestamp of the telemetry data. Defaults to the current timestamp.
+ * @returns { Promise<void> } - promise resolves when the hubsdk acknowledges that it has received the message.
```
erinha, line 25:
```suggestion
 * @returns { Promise<void> } - promise resolves when the host SDK acknowledges that it has received the message.
```

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -11,36 +11,38 @@ import { callFunctionInHost } from '../../internal/communication';
...
+ * @throws { Error } - Throws an error if the app has not been successfully initialized or the host-sdk returns an error as a response to this call
```
erinha, line 26:
```suggestion
 * @throws { Error } - Throws an error if the app has not been successfully initialized or the host SDK returns an error as a response to this call
```

change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `customTelemetry` capability under `copilot` to send app loading data to the host. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
```
erinha, line 3: @TrevorJoelHarris do we update the capability matrix for private capabilities? Should we remove the link from this message?

erinha (dismissed):

packages/teams-js/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './private/index';
 export * from './public/index';
+export { UUID } from './internal/uuidObject';
```
jadahiya-MSFT, line 3: Wouldn't exporting this `internal` function defeat the purpose of it being `internal`? Should we consider moving this function to public instead?

apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
```diff
@@ -23,11 +23,40 @@ const CopilotAPIs = (): ReactElement => {
...
+        submit: async (input) => {
+          const result = await copilot.customTelemetry.sendCustomTelemetryData(
+            input.stageNameIdentifier,
+            input.timestamp,
+          );
+          return JSON.stringify(result);
+        },
+      },
+      defaultInput: JSON.stringify({
+        stageNameIdentifier: new UUID('805a4340-d5e0-4587-8f04-0ae88219699f'),
```
jadahiya-MSFT, line 45: Since the `UUID` constructor takes in a string input and this object is internal to our uses, I am of the opinion that we should try to keep the functions in the test app as agnostic towards how the internals of the SDK function. 
This way if changes happen to the internal UUID object, we don't break testing and if we do we will know it's probably for a good reason.

I am not committed to this and am open to hear what you and others think.

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @returns { Promise<void> } - promise resolves when the host SDK acknowledges that it has received the message.
+ * @throws { Error } - Throws an error if the app has not been successfully initialized or the host SDK returns an error as a response to this call
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export async function sendCustomTelemetryData(
+  stageNameIdentifier: UUID,
```
jadahiya-MSFT, line 34: Adding `UUID` to this exported function now means that any user that wants to use this function must now also import `UUID` from the internal `UUID` function to make it work. This may lead to confusion or conflicts if they have their own implementation of a `UUID`.

Consider instead to have `stageNameIdentifier` be passed in as a string that then gets cast to a `UUID`. This may lead to incorrect strings being passed in to `stageNameIdentifier`, but those will fail for not using a valid uuid during the cast.

change/@microsoft-teams-js-05b6fda8-1215-45ac-b8c5-338fe7d7f77e.json
erinha, line 3:
```suggestion
  "comment": "Added `customTelemetry` capability under `copilot` to send app loading data to the host",
```

packages/teams-js/src/internal/uuidObject.ts
```diff
@@ -4,14 +4,29 @@ import { generateGUID, validateUuid } from './utils';
...
- * UUID object
+ * Represents a UUID (Universally Unique Identifier) object.
+ * This class provides a way to generate, validate, and represent UUIDs as strings.
  */
 export class UUID {
+  /**
+   * Creates an instance of the UUID class.
+   * If no UUID string is provided, a new UUID is generated.
+   *
+   * @param {string} [uuid=generateGUID()] - The UUID string. Defaults to a newly generated UUID.
```
AE-MS, line 15: >{string} [uuid=generateGUID()] [](http://example.com/codeflow?start=12&length=30)

Can you say more about this typedoc syntax? I haven't seen it elsewhere in this codebase and I'm not totally sure what it does. I generated the docs from your change I see this:


...I don't know what those docs will look like when generated by the Microsoft doc system, but it seems like if your intention was to highlight the default value when not specified, it's not coming through in the docs.

AE-MS:
> 

Can you say more about why we need this as the default export? Is exporting the class above enough?

Default exports are generally [a bad idea](https://www.bing.com/search?pglt=129&q=don%27t+use+javascript+default+exports&cvid=3c64b332f6c2430ea54b59e6ab3237ab&gs_lcrp=EgRlZGdlKgYIABBFGDkyBggAEEUYOTIGCAEQABhAMgYIAhAAGEAyBggDEAAYQDIGCAQQABhAMgYIBRAAGEAyBggGEAAYQDIGCAcQABhAMgYICBAAGEAyCAgJEOkHGPxV0gEINTQ1NGowajGoAgCwAgA&FORM=ANNAB1&PC=U531) in JavaScript and using them has caused bugs in the host SDK.

---
Refers to: packages/teams-js/src/internal/uuidObject.ts:33 in 90ca82b. [](commit_id = 90ca82bac8ca6b253aeab7caac1f2881e0bdd4ed, deletion_comment = False)

packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -384,4 +387,24 @@ describe('copilot', () => {
       });
     });
   });
+
+  describe('copilot.customTelemetry', () => {
+    describe('sendCustomTelemetryData', () => {
```
AE-MS, line 392: >describe [](http://example.com/codeflow?start=4&length=8)

Should there be tests to validate that the timestamp is recorded and sent properly when it is not provided? And that it is passed-on when it is provided?

packages/teams-js/test/private/copilot.spec.ts
```diff
@@ -384,4 +387,24 @@ describe('copilot', () => {
...
+      it('sendCustomTelemetryData should throw if called before initialization', async () => {
+        expect.assertions(1);
+        utils.uninitializeRuntimeConfig();
+        await expect(copilot.customTelemetry.sendCustomTelemetryData(new UUID())).rejects.toThrowError(
+          new Error(errorLibraryNotInitialized),
+        );
+      });
+
+      it('sendCustomTelemetryData message should not be null', async () => {
```
AE-MS, line 401: >should not be null [](http://example.com/codeflow?start=42&length=18)

Should this test validate that the contents of the message is actually correct and not just non-null (e.g., contains the provided UUID?)

packages/teams-js/src/private/copilot/customTelemetry.ts
```diff
@@ -0,0 +1,48 @@
...
+  timestamp: number = getCurrentTimestamp() ?? Date.now(),
+): Promise<void> {
+  ensureInitialized(runtime);
+  copilotLogger(
+    'Sending custom telemetry data to host for stage: %s to record timestamp: %s',
+    stageNameIdentifier,
+    timestamp,
+  );
+  return callFunctionInHost(
```
AE-MS, line 43: >callFunctionInHost [](http://example.com/codeflow?start=9&length=18)

I love that you're using this new function! Yay!

AE-MS (changes requested):
:clock1:

erinha:
Moving the uuidObject.ts file to the public folder in this PR: https://github.com/OfficeDev/microsoft-teams-library-js/pull/2671

So we can use it in this PR correctly.

packages/teams-js/src/internal/uuidObject.ts
erinha: I will look into this after merging this PR. We don't generate docs for the private folder, so this shouldn't have any public documentation-facing impact.

apps/teams-test-app/src/components/privateApis/CopilotAPIs.tsx
erinha: The UUID object is now public, and I have exported it as part of this PR, so I think this should be okay now. We should not make changes to the UUID object that would cause anything to break. :)

packages/teams-js/src/private/copilot/customTelemetry.ts
erinha: I think it is better to accept a stronger type than accept a string and subsequently fail. Since we're exporting the UUID type now, it should be relatively straightforward for users of this API to create the `UUID` type.

packages/teams-js/test/private/copilot.spec.ts
erinha: Yes, it should. I tried adding some unit tests, but they require a bit more time to do correctly. I propose adding them in a subsequent change so that this PR can make the next release.

packages/teams-js/test/private/copilot.spec.ts
erinha: Yes, there should. I tried adding some unit tests, but they require a bit more time to do correctly. I propose adding them in a subsequent change so that this PR can make the next release.

erinha:
> > 
> 
> Can you say more about why we need this as the default export? Is exporting the class above enough?
> 
> Default exports are generally [a bad idea](https://www.bing.com/search?pglt=129&q=don%27t+use+javascript+default+exports&cvid=3c64b332f6c2430ea54b59e6ab3237ab&gs_lcrp=EgRlZGdlKgYIABBFGDkyBggAEEUYOTIGCAEQABhAMgYIAhAAGEAyBggDEAAYQDIGCAQQABhAMgYIBRAAGEAyBggGEAAYQDIGCAcQABhAMgYICBAAGEAyCAgJEOkHGPxV0gEINTQ1NGowajGoAgCwAgA&FORM=ANNAB1&PC=U531) in JavaScript and using them has caused bugs in the host SDK.
> 
> Refers to: packages/teams-js/src/internal/uuidObject.ts:33 in [90ca82b](https://github.com/OfficeDev/microsoft-teams-library-js/commit/90ca82bac8ca6b253aeab7caac1f2881e0bdd4ed). [](commit_id = [90ca82b](https://github.com/OfficeDev/microsoft-teams-library-js/commit/90ca82bac8ca6b253aeab7caac1f2881e0bdd4ed), deletion_comment = False)

Looks like the default export line can be removed and it will still compile fine.

erinha (approved):

AE-MS (approved):
:shipit:

PR #2661, opened 2024-12-17
noahdarveau/debug polyfill bug fix

noahdarveau-MSFT:
With the changes from webpack to rollup, we need to use the rollup-plugin-node-polyfills plugin to include the polyfill for the `buffer` in the final bundle. This plugin scrapes the pakcage for any node.js built-ins and includes the polyfills for them. This has been causing an issue with the `debug` package as it has separate node.js and browser implementations, so it does not need to be polyfilled. It determines which implementation to use by checking if the `process` object is defined or not. The polyfilling plugin is detecting the `process` reference and polyfilling it accordingly, which is breaking the implementation decision. A full write-up about this bug can be found in 9619161 (I recommend giving it a read, it's actually a pretty interesting problem). I noticed in our previous bundles, prior to switching to rollup that this problem was not present as the webpack bundler was always resolving to the code to the browser implementation and removing the node.js implementation all together. As such we can directly reference the browser implementation of `debug` when importing the library to remove the issue stated above. 

In this PR I also removed the `globals` field in the rollup config. That configuration option is only relevant to bundling to umd and iife formats, which we are no longer doing, so it is just dead code that I am removing for cleanliness.

### DIFF
--- a/change/@microsoft-teams-js-355651b0-c6d4-47bb-b923-426b9a9a3bae.json
+++ b/change/@microsoft-teams-js-355651b0-c6d4-47bb-b923-426b9a9a3bae.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Explicitly use browser implementation of debug to resolve polyfill issue.",
+  "comment": "Explicitly use browser implementation of `debug` package to resolve polyfill issue.",
   "packageName": "@microsoft/teams-js",
   "email": "109628470+noahdarveau-MSFT@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-teams-js-355651b0-c6d4-47bb-b923-426b9a9a3bae.json
+++ b/change/@microsoft-teams-js-355651b0-c6d4-47bb-b923-426b9a9a3bae.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Explicitly use browser implementation of debug to resolve polyfill issue.",
+  "packageName": "@microsoft/teams-js",
+  "email": "109628470+noahdarveau-MSFT@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/rollup.config.mjs
+++ b/packages/teams-js/rollup.config.mjs
@@ -36,7 +36,7 @@ export default [
       typescript(),
       json(),
       commonjs(),
-      nodePolyfills({ include: './src/**/*' }),
+      nodePolyfills(),
     ],
     treeshake: {
       moduleSideEffects: [
--- a/packages/teams-js/rollup.config.mjs
+++ b/packages/teams-js/rollup.config.mjs
@@ -22,12 +22,6 @@ export default [
       entryFileNames: '[name].js',
       sourcemap: false,
       plugins: [terser()],
-      globals: {
-        buffer: 'Buffer',
-        tty: 'tty',
-        util: 'util',
-        os: 'os',
-      },
     },
     preserveEntrySignatures: 'strict',
     plugins: [
@@ -42,7 +36,7 @@ export default [
       typescript(),
       json(),
       commonjs(),
-      nodePolyfills(),
+      nodePolyfills({ include: './src/**/*' }),
     ],
     treeshake: {
       moduleSideEffects: [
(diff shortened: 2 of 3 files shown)


packages/teams-js/src/internal/telemetry.ts
```diff
@@ -1,4 +1,4 @@
-import { debug as registerLogger, Debugger } from 'debug';
+import { debug as registerLogger, Debugger } from 'debug/src/browser';
```
AE-MS, line 1: Do we also have to care about the reference to this library from microsoft-teams-library-js\packages\teams-js\src\internal\hostToAppTelemetry.ts ?

AE-MS, line 1: >'debug/src/browser'; [](http://example.com/codeflow?start=50&length=20)

Can you leave a comment here explaining to any naive engineer reading this code why we aren't using the root level package?

packages/teams-js/rollup.config.mjs
```diff
@@ -22,12 +22,6 @@ export default [
       entryFileNames: '[name].js',
       sourcemap: false,
       plugins: [terser()],
-      globals: {
-        buffer: 'Buffer',
-        tty: 'tty',
-        util: 'util',
-        os: 'os',
-      },
```
AE-MS, line 30: Maybe this will come in the pending PR description but I'd love to learn more about the motivation and effect for the changes in this file.

change/@microsoft-teams-js-355651b0-c6d4-47bb-b923-426b9a9a3bae.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Explicitly use browser implementation of debug to resolve polyfill issue.",
```
AE-MS, line 3: >debug [](http://example.com/codeflow?start=55&length=5)

Consider either "`debug`" (with backticks) or "debug package"

AE-MS, line 3: > polyfill issue [](http://example.com/codeflow?start=71&length=15)

Thinking about how someone not involved might read this in the changelog, is it worth calling out that it was a `process` polyfill issue? Or was the issue more broad than just that one? Or would devs not have seen "process" listed in the error message so it doesn't matter if you call it out here?

packages/teams-js/src/internal/telemetry.ts
AE-MS, line 1: >debug/src/browser [](http://example.com/codeflow?start=51&length=17)

Out of curiosity is this way of consuming the package "supported" by debug? Or are we kind of cheating?

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/telemetry.ts
noahdarveau-MSFT: No, we do not as the issue only lies in the `debug` object's implementation (the named import, not the package). In `hostToAppTelemetry` we are only importing the `Debugger` interface which comes from the typings file.

change/@microsoft-teams-js-355651b0-c6d4-47bb-b923-426b9a9a3bae.json
noahdarveau-MSFT: This bug actually manifested itself in a couple different ways, that on first glance would seem unrelated, so I figured it was best to just keep it generic. I'm not really sure there is a concise way of covering all manifestations of the bug that would be appropriate for a changelog.

packages/teams-js/src/internal/telemetry.ts
noahdarveau-MSFT: That kind of depends on what you mean by "supported." Do they have any official documentation pointing to this as a something they recommend? No. Though it's not like we're doing some crazy evil bit hacking here or anything. All we are doing is choosing our own entry point to bypass their index file. All the index file does is redirect either to the browser or node implementation since we are having an oddly specific bug that is causing problems with the implementation selection.

packages/teams-js/rollup.config.mjs
noahdarveau-MSFT: See PR description.

TLDR: It's dead code not relevant anymore so I'm removing it for cleanliness.

AE-MS (approved):
:shipit:

erinha (approved):
This is very cool. Thank you for the detailed information!

PR #2663, opened 2024-12-19
[Task 9636919] Add an additional parameter for Handoff ID in composeMailWithHandoffId API

MengyiGong:
## Description

Add an additional parameter for Handoff ID in composeMailWithHandoff API

### Main changes in the PR:

1. Added new composeMailWithHandoff API under Mail capability
2. Added unit tests for the composeMailWithHandoff API

### Unit Tests added:

<Yes>

### End-to-end tests added:

<No>
Will be added in a separate PR after web changes gets in.
### Change file added:

<Yes>

### Screenshots:
Screenshot 2024-12-20 at 2 07 20 PM
Screenshot 2024-12-20 at 2 07 13 PM

### DIFF
--- a/apps/teams-test-app/e2e-test-data/mail.handoff.json
+++ b/apps/teams-test-app/e2e-test-data/mail.handoff.json
@@ -1,65 +1,72 @@
 {
   "name": "Handoff ComposeMail",
   "platforms": "*",
-  "version": ">2.33.0",
-  "testCases": [
+  "featureTests": [
     {
-      "title": "checkMailWithHandoffCapability API Call - Success",
-      "type": "callResponse",
-      "boxSelector": "#box_checkMailWithHandoffCapability",
-      "expectedTestAppValue": "MailWithHandoff module is supported"
-    },
-    {
-      "title": "mail.handoff.composeMail Feature Test - Success",
-      "type": "callResponse",
-      "boxSelector": "#box_composeMailWithHandoff",
-      "inputValue": {
-        "composeMailParams": {
-          "type": "new",
-          "subject": "Test Mail",
-          "toRecipients": ["sam@example.com", "sam1@example.com"],
-          "ccRecipients": ["sam2@example.com"],
-          "bccRecipients": ["sam3@example.com"],
-          "message": "This mail has been sent from the Teams Test App"
-        },
-        "handoffId": "abc123"
+      "feature": {
+        "id": "mail.handoff",
+        "version": 1
       },
-      "expectedAlertValue": "composeMailParamsWithHandoff called with {\"type\":\"new\",\"subject\":\"Test Mail\",\"toRecipients\":[\"sam@example.com\",\"sam1@example.com\"],\"ccRecipients\":[\"sam2@example.com\"],\"bccRecipients\":[\"sam3@example.com\"],\"message\":\"This mail has been sent from the Teams Test App\"} and handoffId: abc123",
-      "expectedTestAppValue": "Completed"
-    },
-    {
-      "title": "mail.handoff.composeMail Feature Test with invalid handoffId- Failure",
-      "type": "callResponse",
-      "boxSelector": "#box_composeMailWithHandoff",
-      "inputValue": {
-        "composeMailParams": {
-          "type": "new",
-          "subject": "Test Mail",
-          "toRecipients": ["sam@example.com", "sam1@example.com"],
-          "ccRecipients": ["sam2@example.com"],
-          "bccRecipients": ["sam3@example.com"],
-          "message": "This mail has been sent from the Teams Test App"
+      "testCases": [
+        {
+          "title": "checkMailWithHandoffCapability API Call - Success",
+          "type": "callResponse",
+          "boxSelector": "#box_checkMailWithHandoffCapability",
+          "expectedTestAppValue": "MailWithHandoff module is supported"
         },
-        "handoffId": ""
-      },
-      "expectedTestAppValue": "Error: Error: handoffId should not be null or empty string."
-    },
-    {
-      "title": "mail.handoff.composeMail Feature Test with invalid email - Failure",
-      "type": "callResponse",
-      "boxSelector": "#box_composeMailWithHandoff",
-      "inputValue": {
-        "composeMailParams": {
-          "type": "new",
-          "subject": "Test Mail",
-          "toRecipients": ["mockEmail", "sam1@example.com"],
-          "ccRecipients": ["sam2@example.com"],
-          "bccRecipients": ["sam3@example.com"],
-          "message": "This mail has been sent from the Teams Test App"
+        {
+          "title": "mail.handoff.composeMail Feature Test - Success",
+          "type": "callResponse",
+          "boxSelector": "#box_composeMailWithHandoff",
+          "inputValue": {
+            "composeMailParams": {
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["sam@example.com", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": "abc123"
+          },
+          "expectedAlertValue": "composeMailParamsWithHandoff called with {\"type\":\"new\",\"subject\":\"Test Mail\",\"toRecipients\":[\"sam@example.com\",\"sam1@example.com\"],\"ccRecipients\":[\"sam2@example.com\"],\"bccRecipients\":[\"sam3@example.com\"],\"message\":\"This mail has been sent from the Teams Test App\"} and handoffId: abc123",
+          "expectedTestAppValue": "Completed"
         },
-        "handoffId": "mockId"
-      },
-      "expectedTestAppValue": "Error: Error: Input email address does not have the correct format."
+        {
+          "title": "mail.handoff.composeMail Feature Test with invalid handoffId- Failure",
+          "type": "callResponse",
+          "boxSelector": "#box_composeMailWithHandoff",
+          "inputValue": {
+            "composeMailParams": {
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["sam@example.com", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": ""
+          },
+          "expectedTestAppValue": "Error: Error: handoffId should not be null or empty string."
+        },
+        {
+          "title": "mail.handoff.composeMail Feature Test with invalid email - Failure",
+          "type": "callResponse",
+          "boxSelector": "#box_composeMailWithHandoff",
+          "inputValue": {
+            "composeMailParams": {
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["mockEmail", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": "mockId"
+          },
+          "expectedTestAppValue": "Error: Error: Input email address does not have the correct format."
+        }
+      ]
     }
   ]
 }
--- a/apps/teams-test-app/e2e-test-data/mail.handoff.json
+++ b/apps/teams-test-app/e2e-test-data/mail.handoff.json
@@ -1,0 +1,76 @@
+{
+  "name": "Handoff ComposeMail",
+  "version": ">2.0.0-beta.0",
+  "platforms": "*",
+  "featureTests": [
+    {
+      "feature": {
+        "id": "mail.handoff",
+        "version": 1
+      },
+      "testCases": [
+        {
+          "title": "checkMailWithHandoffCapability API Call - Success",
+          "type": "callResponse",
+          "boxSelector": "#box_checkMailWithHandoffCapability",
+          "expectedTestAppValue": "MailWithHandoff module is supported"
+        },
+        {
+          "title": "mail.handoff.composeMail Feature Test - Version 1",
+          "type": "callResponse",
+          "version": ">2.0.0-beta.0 && <=2.17.0",
+          "boxSelector": "#box_composeMailWithHandoff",
+          "inputValue": {
+            "composeMailParams": {
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["sam@example.com", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": "abc123"
+          },
+          "expectedAlertValue": "composeMailParamsWithHandoff called with {\"type\":\"new\",\"subject\":\"Test Mail\",\"toRecipients\":[\"sam@example.com\",\"sam1@example.com\"],\"ccRecipients\":[\"sam2@example.com\"],\"bccRecipients\":[\"sam3@example.com\"],\"message\":\"This mail has been sent from the Teams Test App\"} and handoffId: abc123",
+          "expectedTestAppValue": "Completed"
+        },
+        {
+          "title": "mail.handoff.composeMail Feature Test with invalid handoffId- Failure",
+          "type": "callResponse",
+          "version": ">2.0.0-beta.0 && <=2.17.0",
+          "boxSelector": "#box_composeMailWithHandoff",
+          "inputValue": {
+            "composeMailParams": {
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["sam@example.com", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": ""
+          },
+          "expectedTestAppValue": "Error: Error: handoffId should not be null or empty string."
+        },
+        {
+          "title": "mail.handoff.composeMail Feature Test with invalid email - Failure",
+          "type": "callResponse",
+          "version": ">2.0.0-beta.0 && <=2.17.0",
+          "boxSelector": "#box_composeMailWithHandoff",
+          "inputValue": {
+            "composeMailParams": {
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["mockEmail", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": "mockId"
+          },
+          "expectedTestAppValue": "Error: Error: Input email address does not have the correct format."
+        }
+      ]
+    }
+  ]
+}
--- a/apps/teams-test-app/e2e-test-data/mail.handoff.json
+++ b/apps/teams-test-app/e2e-test-data/mail.handoff.json
@@ -1,5 +1,6 @@
 {
   "name": "Handoff ComposeMail",
+  "version": ">2.0.0-beta.0",
   "platforms": "*",
   "featureTests": [
     {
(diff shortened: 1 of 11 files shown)


packages/teams-js/src/public/mail.ts
```diff
@@ -155,3 +176,31 @@ export type ComposeMailParams =
...
+ */
+export interface ComposeMailParamsWithHandoff {
+  /**
+   * Base parameters for composing a mail item.
+   */
+  composeMailParams: ComposeMailParams;
+  /**
+   * Use this endpoint to retrieve the handoff payload when BizChat creates an email draft for external handoff.
+   */
+  handoffId?: string;
```
AE-MS, line 198: >string [](http://example.com/codeflow?start=14&length=6)

Is this ID really any possible string? Is "MengyiGong" a valid handoff Id?

packages/teams-js/src/public/mail.ts
```diff
@@ -63,6 +65,25 @@ export function composeMail(composeMailParams: ComposeMailParams): Promise<void>
   });
 }
 
+/**
+ * Compose a new email in the user's mailbox, opening it in the drafts UX instead of the standard email.
+ *
+ * @param composeMailParamsWithHandoff - Object that specifies the type of mail item to compose and the details of the mail item.
+ *
```
AE-MS, line 72: >* [](http://example.com/codeflow?start=1&length=1)

Can you add documentation for the `@returns` value?

packages/teams-js/src/public/mail.ts
AE-MS, line 198: >? [](http://example.com/codeflow?start=11&length=1)

Is it really valid for this to be optional? When would someone use the `composeMailWithHandoff` function if they didn't have a handoff id to provide?

change/@microsoft-teams-js-61f74ab4-102e-4a46-88bb-55dbf245fbc1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Removed Beta/Preview tag on `{namespace}` capability. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
```
AE-MS, line 3: >Removed Beta/Preview tag on `{namespace}` capability. To track availability of this capability across different hosts see https://aka.ms/capmatrix", [](http://example.com/codeflow?start=14&length=148)

This doesn't seem right

AE-MS (changes requested):
:clock1:

packages/teams-js/src/public/mail.ts
```diff
@@ -63,6 +65,25 @@ export function composeMail(composeMailParams: ComposeMailParams): Promise<void>
...
+ */
+export function composeMailWithHandoff(composeMailParamsWithHandoff: ComposeMailParamsWithHandoff): Promise<boolean> {
+  ensureInitialized(runtime, FrameContexts.content);
+  if (!isSupported()) {
+    throw new Error('Not supported');
+  }
+  return callFunctionInHostAndHandleResponse<boolean, boolean>(
```
AE-MS, line 79: >callFunctionInHostAndHandleResponse [](http://example.com/codeflow?start=9&length=35)

Is it important that this function returns a `boolean` from the host SDK? Or can it just return void and receive an `SdkError` instead?

I believe it would allow you to use the simpler `callFunctionInHost` instead.

change/@microsoft-teams-js-61f74ab4-102e-4a46-88bb-55dbf245fbc1.json
MengyiGong: Thanks for pointing it out! The changes were in local and not pushed yet. Updated!

packages/teams-js/src/public/mail.ts
MengyiGong: Confirmed with @jenniferyellow , any string should be valid for `handoffId`.
Hopeful this link helps: https://eng.ms/docs/experiences-devices/opg/office-ai/app-copilot/onboarding/handoff/handoff

MengyiGong: Thanks, and this is a good point. We are making a new sub-capability `composeMailWithHandoff` API , and the `handoffId` should be required. Updated to required.

packages/teams-js/src/public/mail.ts
MengyiGong: Thanks for your suggestion! That’s a great point about the `boolean` not being used anywhere in the host SDK. I’ve updated it to use `callFunctionInHost`.

AE-MS:
I'm signing off on this because the code looks okay but one of the questions I had on your host SDK PR about whether or not it is going to be a breaking API change might affect how this teamsjs PR is structured.

AE-MS (dismissed):

MengyiGong:
> I'm signing off on this because the code looks okay but one of the questions I had on your host SDK PR about whether or not it is going to be a breaking API change might affect how this teamsjs PR is structured.

Thanks @AE-MS  I will hold off this PR until I figure how we will structure the web APIs.

packages/teams-js/src/public/mail.ts
```diff
@@ -63,6 +64,25 @@ export function composeMail(composeMailParams: ComposeMailParams): Promise<void>
   });
 }
 
+/**
+ * Compose a new email in the user's mailbox, opening it in the drafts UX instead of the standard email.
+ *
+ * @param composeMailParamsWithHandoff - Object that specifies the type of mail item to compose and the details of the mail item.
+ *
+ */
+export function composeMailWithHandoff(composeMailParamsWithHandoff: ComposeMailParamsWithHandoff): Promise<void> {
```
TrevorJoelHarris, line 73: This looks like it would be a breaking change for any hosts that currently implement the mail capability, because we expect all functions that check the same isSupported to work (or not work) at the same time. If this is checked in as-is, then hosts that already implement the mail capability would keep responding to the existing functions but this new one would be ignored by them even though we told the app developer that it worked.

In order to make this change in a non-breaking way you would need to make a new file for your new capability (see the geolocation folder for an example) and then you'd have to have that function check a different (new) section of runtime in order to check if this new subcapability is supported on the current host.

I hope this makes sense, but if it would be helpful I'm happy to sit down and show you what I mean. Just let me know!

packages/teams-js/src/public/mail.ts
```diff
@@ -63,6 +64,25 @@ export function composeMail(composeMailParams: ComposeMailParams): Promise<void>
   });
 }
 
+/**
+ * Compose a new email in the user's mailbox, opening it in the drafts UX instead of the standard email.
```
TrevorJoelHarris, line 68: Any new function we add should be marked with the @beta tag so that developers know that the function may change or be broken with little or no warning. There should be other examples in the library you can check.

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * v2 APIs telemetry file: All of APIs in this capability file should send out API version v2 ONLY
+ */
+const mailTelemetryVersionNumber: ApiVersionNumber = ApiVersionNumber.V_2;
+
+/**
+ * Compose a new email in the user's mailbox, opening it in the drafts UX instead of the standard email.
+ *
+ * @param composeMailParamsWithHandoff - Object that specifies the type of mail item to compose and the details of the mail item.
+ *
```
TrevorJoelHarris, line 23: There should be a `@returns` section here which outlines that the function returns a promise and states when the promise resolves (for example, is it when the host has opened a compose mail window, or only after that window is closed?) Also should outline what errors the Promise can reject with.

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @see {@link ComposeReplyOrForwardParams} for reply or forward-specific parameters.
+ * @see {@link ComposeMailType} for supported mail operation types.
+ */
+export interface ComposeMailParamsWithHandoff {
+  /**
+   * Base parameters for composing a mail item.
+   */
+  composeMailParams: ComposeMailParams;
+  /**
+   * Use this endpoint to retrieve the handoff payload when BizChat creates an email draft for external handoff.
```
TrevorJoelHarris, line 64: Might be something to check with @jenniferyellow 

Where does a developer get a handoff id? This just being a string makes it tricky (at least, without any extra explanation) for a developer to understand what a "valid" handoffId is.

TrevorJoelHarris: (If this can be more strongly typed than string that would also be great, but hard to say without knowing more about what the handoffId is)

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * Used to interact with mail capability, including opening and composing mail.
```
TrevorJoelHarris, line 2: The `@module` comment and every exported item in this file should include the following comment:

```typescript
 * @beta
```

TrevorJoelHarris: This will print a blurb in the dev docs on this capability that looks like this: 

image

You can work with PM to decide when the @beta tag gets removed

change/@microsoft-teams-js-61f74ab4-102e-4a46-88bb-55dbf245fbc1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add an additional parameter for Handoff ID in `composeMailWithHandoffId` API ",
```
TrevorJoelHarris, line 3:
```suggestion
  "comment": "Added the`mail.handoff.composeMailWithHandoffId` function that allows passing a handoff id to the compose mail window.",
```

You may want to check with PM to find out what specific language to write here since this string will show up in the patch notes for developers.

apps/teams-test-app/src/components/MailAPIs.tsx
```diff
@@ -11,6 +11,13 @@ const CheckMailCapability = (): React.ReactElement =>
     onClick: async () => `Mail module ${mail.isSupported() ? 'is' : 'is not'} supported`,
   });
 
+const CheckMailWithHandoffCapability = (): React.ReactElement =>
```
TrevorJoelHarris, line 14: You will need to add new E2E tests for this function (see mail.json for the existing e2e tests).

You should also check with @alexneyman-MSFT - he recently introduced a new, simpler way to author e2e tests that could help you out a lot!

packages/teams-js/test/public/mail.spec.ts
```diff
@@ -265,6 +264,68 @@ describe('mail', () => {
...
+          }
+          await utils.initializeWithContext(frameContext);
+          expect.assertions(1);
+          expect(() => mail.handoff.composeMailWithHandoff(composeMailParamsWithHandoff)).toThrowError(
+            `This call is only allowed in following contexts: ["content"]. Current context: "${frameContext}".`,
+          );
+        });
+      });
+
+    it('should not allow calls if runtime does not support mail', async () => {
```
TrevorJoelHarris, line 297: Maybe should also have a test where mail is supported but mail.handoff is not

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+const mailTelemetryVersionNumber: ApiVersionNumber = ApiVersionNumber.V_2;
+
+/**
+ * Compose a new email in the user's mailbox, opening it in the drafts UX instead of the standard email.
+ *
+ * @param composeMailParamsWithHandoff - Object that specifies the type of mail item to compose and the details of the mail item.
+ *
+ */
+export function composeMailWithHandoff(composeMailParamsWithHandoff: ComposeMailParamsWithHandoff): Promise<void> {
```
TrevorJoelHarris, line 25: Is there any validation of the input params that you could do here to make sure that only legal values are passed to the host? Any invalid props you can catch here are ones the host won't have to deal with in their layer!

TrevorJoelHarris (changes requested):
This is looking good! Left some thoughts and comments, let me know if you have any questions.

packages/teams-js/src/public/mail.ts
MengyiGong: Sorry for the confusion, updated to sub-capability.

apps/teams-test-app/src/components/MailAPIs.tsx
MengyiGong: Please correct me if I'm wrong, should we add e2e test after web PR changes get in otherwise e2e will fail?

packages/teams-js/test/public/mail.spec.ts
MengyiGong: Thanks, updated the test with ` mail is supported but mail.handoff is not` case.

packages/teams-js/src/public/mail/handoff.ts
MengyiGong: Thanks for your suggestion! I’ve updated the changes to include email validation when the input contains email addresses.

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Validates an array of email addresses.
+ *
+ * @param emails - An optional array of email addresses to validate.
+ * @returns `true` if all email addresses are valid or the array is empty/undefined, otherwise `false`.
+ *
+ * @beta
+ */
+function validateEmails(emails?: string[]): boolean {
```
TrevorJoelHarris, line 52: We already have an email validation function that you can use: https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/packages/teams-js/src/internal/emailAddressValidation.ts

That function is less strict that what you have here; all it does is validate that the string isn't empty and that it contains an '@' character followed by a '.' character. The reason it's less strict is that email address validation for all possible valid email addresses using regex is very nearly impossible. Here's a stack overflow post that explains the complexity: https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

Rather than try to support a regex that is hundreds of lines long we have this happy medium!

packages/teams-js/src/public/mail/handoff.ts
TrevorJoelHarris: "promise resolves when the host SDK succeeds" is it possible to be more specific that this? My default assumption would be that the promise resolves after the compose window has opened successfully. Is that correct?

packages/teams-js/src/public/mail/handoff.ts
MengyiGong: Thanks for the ref! Updated to using `validateEmailAddress`.

packages/teams-js/src/public/mail/handoff.ts
MengyiGong: Thanks for the suggestion to make the comment more specific! Updated.

packages/teams-js/src/public/mail/handoff.ts
MengyiGong: Hi @TrevorJoelHarris Based on the discussion with the PMs, requiring a non-empty string is reasonable, and we don't need to perform any validation on the handoffID for now. I updated with the `handoffId` empty string check and also updated the unit test with this test case. Thanks for bringing this up!

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,125 @@
...
+ */
+function validateHandoffComposeMailParams(param: ComposeMailParamsWithHandoff): void {
+  if (!param.handoffId || param.handoffId.trim().length == 0 || param.handoffId.trim() === '') {
+    throw new Error('handoffId should not be null or empty string.');
+  }
+  const composeMailParams = param.composeMailParams;
+  if (composeMailParams.type === ComposeMailType.New) {
+    validateEmails(composeMailParams.toRecipients) &&
+      validateEmails(composeMailParams.ccRecipients) &&
+      validateEmails(composeMailParams.bccRecipients);
```
AE-MS, line 81: What are the semantics of calling `composeMailWithHandoff` with `ComposeMailType.New` and passing empty arrays for `toRecipients`, `ccRecipients`, and `bccRecipients`? The current validation code would allow that case. Should that be an error?

MengyiGong: This is the behavior of the composeMail API, and I will double-check with the PM. (Based on my understanding, using ComposeMailType.New without specifying toRecipients allows the user to save the email as a draft and edit the toRecipients, ccRecipients, and bccRecipients later.)

MengyiGong: Confirmed with PM and this is a valid pattern.

AE-MS (dismissed):
:shipit:

apps/teams-test-app/e2e-test-data/mail.handoff.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "Handoff ComposeMail",
```
KangxuanYe, line 2: ComposeMail is not part of subCapability Name right?

apps/teams-test-app/e2e-test-data/mail.handoff.json
```diff
@@ -0,0 +1,72 @@
...
+              "type": "new",
+              "subject": "Test Mail",
+              "toRecipients": ["sam@example.com", "sam1@example.com"],
+              "ccRecipients": ["sam2@example.com"],
+              "bccRecipients": ["sam3@example.com"],
+              "message": "This mail has been sent from the Teams Test App"
+            },
+            "handoffId": ""
+          },
+          "expectedTestAppValue": "Error: Error: handoffId should not be null or empty string."
```
KangxuanYe, line 50: I think "Error: Error: ..." bug has been fixed? @TrevorJoelHarris , don't we?

apps/teams-test-app/e2e-test-data/mail.handoff.json
MengyiGong: The sub capability named `mail.handoff`. `composeMail` is the API name under `mail.handoff` sub-capability. For example, `geoLocation.map.chooseLocation` is one of the APIs under `geoLocation.map` sub-capability, and `geoLocation.map` is the sub-capability of `geoLocation`.

MengyiGong: Updated to `Mail Handoff`

KangxuanYe: Please follow naming convention like this file: https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/apps/teams-test-app/e2e-test-data/pages.fullTrust.json

(page.fullTrust as an example)

There is a strict check in iOS backward compatible side if you don't name it by this.

KangxuanYe (changes requested):

AE-MS (approved):
:shipit:

apps/teams-test-app/e2e-test-data/mail.handoff.json
TrevorJoelHarris: As far as I know this still shows up in the Test App :(

TrevorJoelHarris: (And it's sort of hard to fix because of how many tests have been written assuming it's there)

KangxuanYe: Okay, I will leave them to be existed still... I thought we have fixed them.

KangxuanYe (approved):

packages/teams-js/src/public/mail/handoff.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Validates an array of email addresses.
+ * For ComposeMailType.New, it is valid to pass empty arrays for toRecipients, ccRecipients, and bccRecipients.
+ * This will result in a new email with handed-off (pre-populated) body content but no pre-populated recipients.
+ *
+ * @param emails - An optional array of email addresses to validate.
+ * @throws Error with a message describing if the email address is invalid.
+ *
+ * @beta
+ */
+function validateEmails(emails?: string[]): boolean {
```
TrevorJoelHarris, line 55: this is very minor but I don't think this function needs to return a boolean. If it had a return type of void that would probably be more correct - it would return nothing, but if any email addresses are invalid it throws an error. Alternatively, you could have it return a boolean and then catch errors in here and return false if one occurs. I'd do option 1, but it's up to you

TrevorJoelHarris (approved):

PR #2666, opened 2025-01-02
Moving CDN PreFetch to kick off prior to initialize call and added timeout logic

jadahiya-MSFT:

## Description

This change was made to address customer feedback that the network call to acquire CDN resources was leading to initialization time outs. We aim to address this concern by decoupling that call.

### Main changes in the PR:

1. Decoupled CDN prefetch from initialization flow by moving logic into `prefetch.ts` file to execute a global fetch upon library load. Moved into a separate file since global functions cause side effects that can't be treeshakeable.
2. Added a timeout of 1.5 seconds for CDN resource acquisition before defaulting back to fallback list.
3. Fixed a typo in `jest-setup.cjs` file
4. Removed `fetch` polyfill rom `test\utils.ts` file and moved into a new `setupTest.ts` file that is used during jest setup.

### Validation performed:

1.  Tested locally using test app with Orange
5. Ran E2E and Unit tests
6. Monitored network timings of moving prefetch to Global space versus during the initialization flow and increased speeds for CDN asset acquisition on lower end networks with 3G and slow 4G speeds. Taking now at most 600ms to complete on both of those network configurations.
7. Confirmed that timeout logic for CDN acquisition works by intentionally blocking network call for 1.5 seconds.

### Unit Tests added:

1. Added `setupTests.ts` file to tests to enable global fetch polyfill for Jest tests.


packages/teams-js/src/public/app/app.ts
```diff
@@ -27,6 +27,8 @@ import {
 import { version } from '../version';
 import * as lifecycle from './lifecycle';
 
+prefetchOriginsFromCDN();
```
TrevorJoelHarris, line 30: Now that this is called in the global namespace this file has a side effect that it didn't used to have (loading the file in memory now automatically makes a call to the CDN). Can you verify with Noah that we don't need to mark this file in any particular way to prevent it from being treeshaken out?

TrevorJoelHarris (approved):

packages/teams-js/src/public/app/app.ts
jadahiya-MSFT: Good call, I'll follow up with Noah on this and if there's any issue I'll move this call into a separate funciton.

jadahiya-MSFT: Followed up with Noah, this will in fact introduce an unintended side effect that would make `app.ts` not treeshake-able. I'll be pushing out a change that will mitigate this or lessen the impact of the side effect.

packages/teams-js/jest.config.cjs
```diff
@@ -1,11 +1,12 @@
...
 const packageVersion = require('./package.json').version;
 
 module.exports = {
   ...commonSettings,
   globals: {
     PACKAGE_VERSION: packageVersion,
+    fetch: global.fetch,
   },
-  //setupFilesAfterEnv: ['./jest-setup.cjs'],
+  setupFilesAfterEnv: ['./test/setupTest.ts'],
```
TrevorJoelHarris, line 11: Why is this change needed?

packages/teams-js/src/public/app/prefetch.ts
```diff
@@ -0,0 +1,3 @@
+import { prefetchOriginsFromCDN } from '../../internal/validOrigins';
+
+prefetchOriginsFromCDN();
```
TrevorJoelHarris, line 3: Could this file be in the internal folder instead of private/public? Doesn't seem like we intended for anyone to call it and are just depending on it calling a function in the global namespace

TrevorJoelHarris (commented):
Left a few questions

packages/teams-js/src/public/app/prefetch.ts
jadahiya-MSFT: Yup I could move this there, I would just need to import it somewhere for it to be picked up as a module.

packages/teams-js/jest.config.cjs
jadahiya-MSFT: This change is needed since we need to execute the setupTest file as setup prior to any test running. That file allows us to create the `fetch` polyfill.

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -5,6 +5,7 @@ import { inServerSideRenderingEnvironment, isValidHttpsURL } from './utils';
 
 let validOriginsCache: string[] = [];
 const validateOriginLogger = getLogger('validateOrigin');
+const ORIGIN_LIST_TIMEOUT = 1500;
```
AE-MS, line 8: >ORIGIN_LIST_TIMEOUT [](http://example.com/codeflow?start=6&length=19)

Can you add a unit to this symbol name? E.g., `ORIGIN_LIST_TIMEOUT_IN_MS`

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -34,7 +35,11 @@ async function getValidOriginsListFromCDN(): Promise<string[]> {
         });
       })
       .catch((e) => {
-        validateOriginLogger('validOrigins fetch call to CDN failed with error: %s. Defaulting to fallback list', e);
+        if (e.name === 'TimeoutError') {
+          validateOriginLogger('validOrigins fetch call to CDN failed due to Timeout. Defaulting to fallback list', e);
```
AE-MS, line 39: >Timeout [](http://example.com/codeflow?start=77&length=7)

To make it easier for humans who read the logs, can you also add the timeout that was used when calling `fetch` (e.g., the value of `ORIGIN_LIST_TIMEOUT`) to the log statement?

AE-MS, line 39: >, e [](http://example.com/codeflow?start=114&length=3)

Does this do anything without the `%s` in the string?

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -19,7 +20,7 @@ async function getValidOriginsListFromCDN(): Promise<string[]> {
     return validOriginsCache;
   }
   if (!inServerSideRenderingEnvironment()) {
-    return fetch(validOriginsCdnEndpoint)
+    return fetch(validOriginsCdnEndpoint, { signal: AbortSignal.timeout(ORIGIN_LIST_TIMEOUT) })
```
AE-MS, line 23: Can we add logging, before `fetch` is called, to indicate that we're about to make the CDN call?

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -19,7 +20,7 @@ async function getValidOriginsListFromCDN(): Promise<string[]> {
...
       .then((response) => {
         if (!response.ok) {
```
AE-MS, line 25: Can we add a log statement in the `then` call to indicate that `fetch` has completed?

The idea is that between the two new log statements I was requesting, you could tell how long this call is taking by reading the logs.

AE-MS:
> async function getValidOriginsListFromCDN(): Promise<string[]> {

If this function gets called a second time while the first `fetch` is still waiting, will it kick off a second CDN network request? Is that... okay? I can't think of a reason why it would cause "damage" but it is maybe unnecessary?
```
¯\_(ツ)_/¯
```

---
Refers to: packages/teams-js/src/internal/validOrigins.ts:18 in 74a6e89. [](commit_id = 74a6e89d8f652d51f393b5320ff0a58ef8975106, deletion_comment = False)

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -34,7 +35,11 @@ async function getValidOriginsListFromCDN(): Promise<string[]> {
         });
       })
       .catch((e) => {
-        validateOriginLogger('validOrigins fetch call to CDN failed with error: %s. Defaulting to fallback list', e);
+        if (e.name === 'TimeoutError') {
+          validateOriginLogger('validOrigins fetch call to CDN failed due to Timeout. Defaulting to fallback list', e);
+        } else {
```
AE-MS, line 40: Can you add new unit test(s) to `validOrigins.spec.ts` that validate this new timeout behavior?

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -129,3 +134,5 @@ export function validateOrigin(messageOrigin: URL): Promise<boolean> {
     return false;
   });
 }
+
+prefetchOriginsFromCDN();
```
AE-MS, line 138: Can you add new unit tests to `validOrigins.spec.ts` that validate the fetch happens when the library is loaded?

change/@microsoft-teams-js-b7743f88-f4a7-4485-a6b8-2a7512bb9d83.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Updated various package versions",
+  "packageName": "@microsoft/teams-js",
+  "email": "jadahiya@microsoft.com",
+  "dependentChangeType": "none"
+}
```
AE-MS, line 7: Is this changefile supposed to be in here?

change/@microsoft-teams-js-ceda76cc-75d3-40de-8b4b-cb392a240254.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Decoupled prefetch of CDN resources from the initialization workflow and added timeout based fail and fallback path",
```
AE-MS, line 3: >Decoupled prefetch of CDN resources from the initialization workflow and added timeout based fail and fallback path [](http://example.com/codeflow?start=14&length=115)

Consider:
"Decoupled prefetch of CDN resources from the initialization workflow and added fallback when resources could not be fetched in 1.5 seconds"

AE-MS (changes requested):
:clock1:

packages/teams-js/src/internal/validOrigins.ts
jadahiya-MSFT: Great suggestion, I've added it in

packages/teams-js/src/internal/validOrigins.ts
jadahiya-MSFT: Nope, good catch. I've removed it

packages/teams-js/src/internal/validOrigins.ts
jadahiya-MSFT: I'm unsure how I would do this since now every unit test actually runs through the fetch code. This is why I needed to add the setupTests.ts file to mock fetch earlier in the jest pipeline otherwise it would complain about `fetch` not existing.

change/@microsoft-teams-js-b7743f88-f4a7-4485-a6b8-2a7512bb9d83.json
jadahiya-MSFT: Nope this was for another PR that got jumbled while I was pushing changes to both. I've removed it now

change/@microsoft-teams-js-ceda76cc-75d3-40de-8b4b-cb392a240254.json
jadahiya-MSFT: I really like this and have made that change.

packages/teams-js/src/internal/validOrigins.ts
jadahiya-MSFT: Hmm I might need to think about this more, but because we are already mocking the `fetch` call since `jest` doesn't have access to that function as it is a Browser API. 

Wouldn't mocking `AbortSignal.timeout` at the same time introduce uncertainty in the validity of the test, since we are now relying on two mocks working correctly together to yield a result that may not be indicative of the real world scenarios?

I could instead have the fetch mock reject with a 'TimeoutError' to ensure that we enter the timeout error logic.

packages/teams-js/test/setupTest.ts
```diff
@@ -0,0 +1,16 @@
+import { validOriginsFallback } from '../src/internal/constants';
+
+/**
+ * We currently run a fetch call to acquire CDN assets as soon as TeamsJS is loaded.
+ * Since fet ch is supported in both browser and Node environments, but not supported in jest/jsdom,
```
AE-MS, line 5: >fet ch [](http://example.com/codeflow?start=9&length=6)

fetch

packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -14,27 +15,40 @@ function isValidOriginsCacheEmpty(): boolean {
   return validOriginsCache.length === 0;
 }
 
-async function getValidOriginsListFromCDN(): Promise<string[]> {
-  if (!isValidOriginsCacheEmpty()) {
+async function getValidOriginsListFromCDN(disableCache?: boolean): Promise<string[]> {
```
AE-MS, line 18: >disableCache?: boolean [](http://example.com/codeflow?start=42&length=22)

Consider:
1. Naming this `shouldDisableCache` makes the intent clearer
2. Set a default value for this, e.g., `shouldDisableCache? :boolean = false` so that the default behavior is clear to callers and the consumers in the function don't have to to be concerned with any `undefined weirdness.

packages/teams-js/test/internal/validOrigins.spec.ts
```diff
@@ -409,102 +412,185 @@ describe('validOrigins', () => {
...
+        throw new Error('AbortError');
+      });
+
+      timeoutSpy = jest.spyOn(global, 'setTimeout');
+      abortSpy = jest.spyOn(AbortController.prototype, 'abort');
+
+      global.fetch = jest.fn(
+        () =>
+          new Promise((resolve) => {
+            jest.advanceTimersByTime(1600);
```
AE-MS, line 537: >1600 [](http://example.com/codeflow?start=37&length=4)

Can you (export and) reference the `ORIGIN_LIST_FETCH_TIMEOUT_IN_MS` constant here rather than duplicating its value?

AE-MS (approved):
:shipit:

packages/teams-js/src/public/app/app.ts
TrevorJoelHarris: Hmmm, that's a good question. I'll think about it (and maybe ask @noahdarveau-MSFT). In general, any function calls in the global namespace are "side effects" if they cause some state somewhere to change because those functions will just get called assuming the entire module doesn't get tree-shaken out and removed. Off the top of my head I'm not sure there's a way to "detect" that.

TrevorJoelHarris (approved):

packages/teams-js/src/public/app/app.ts
TrevorJoelHarris: Okay, I am pretty sure that we could make a linter rule to detect global function calls that would show a warning saying something like "This file should probably be added to the sideEffects array in package.json. Once you've done that (or verified that you don't need to) please suppress this warning."

PR #2673, opened 2025-01-09
2.33.0 Prerelease

TrevorJoelHarris:
Result of `node tools/cli/prerelease.js` command

### Minor changes

- Added `customTelemetry` capability under `copilot` to send app loading data to the host.

### Patches

- Explicitly use browser implementation of `debug` package to resolve polyfill issue.
- Moved uuidObject.ts from internal to public folder in preparation for exporting it.


AE-MS (approved):
:shipit:

PR #2685, opened 2025-01-17
Update authentication.notifySuccess function documentation with secure usage guidance

AE-MS:
## Description

The `authentication.notifySuccess` function did not explicitly discuss secure usage guidance and this PR adds it.

Separately, I am working with the docs team to get the prose documentation pages ([here](https://learn.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/authentication/auth-flow-tab) and [here](https://learn.microsoft.com/en-us/microsoftteams/platform/messaging-extensions/how-to/add-authentication)) updated to be more explicit about this.

The final docs now look like this, after feedback from @JoshuaPartlow:
{6F05D66E-FA63-42CD-B16D-DCB67BDC35E5}
This is obviously the local generation of the docs -- they will be formatted slightly differently on the official site.

### DIFF
--- a/change/@microsoft-teams-js-23b4de1d-348c-4748-90ac-78b4fa7158c7.json
+++ b/change/@microsoft-teams-js-23b4de1d-348c-4748-90ac-78b4fa7158c7.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated `authentication.notifySuccess` documentation with security guidance",
+  "packageName": "@microsoft/teams-js",
+  "email": "36546967+AE-MS@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/teams-js/src/public/authentication.ts
+++ b/packages/teams-js/src/public/authentication.ts
@@ -459,18 +459,18 @@ function startAuthenticationWindowMonitor(): void {
  * authentication request was successful.
  *
  * @remarks
- * This function is usable only from the authentication window.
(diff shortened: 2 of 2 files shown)


noahdarveau-MSFT (dismissed):

noahdarveau-MSFT (approved):

TrevorJoelHarris (dismissed):

JoshuaPartlow:
@AE-MS - Given the size of the info, our thought is that it might go better with the remarks section. Since `result` is the only parameter it doesn't seem like it would get lost there or anything. For the content itself, no major concerns, but our suggestion with a few edits would be the following:

**Note**: The `result` parameter should **never** contain the token that was received from the identity provider, because a malicious app (rather than your own app) might have opened the authentication window. If that was the case, passing the token in this parameter would leak it to them. More secure methods for completing the authentication flow include:
- For a purely browser-based experience (e.g., a personal app/tab app), you could store the token in browser local storage and then have your personal app retrieve it once this `notifySuccess` call is received.
- For a server-based experience (e.g., a message extension), your authentication window could store the token on your service and then generate a unique code passed via this `result` parameter. The caller can then use the unique code to retrieve the token from your service.

@erikadoyle  as an fyi

JoshuaPartlow (dismissed):
Reviewed and provided feedback in comments.

JoshuaPartlow (approved):

noahdarveau-MSFT (approved):

PR #2691, opened 2025-01-22
noahdarveau/size-limit

noahdarveau-MSFT:
Adding the Size-Limit library to the repo. This package can be used to determine the size of the library by running the command `pnpm size`. In a later PR I plan to use this package in conjunction with a github actions workflow to determine library size changes in PRs as compared to main, however before that can be checked into the repo, size-limit needs to be added to main as the github workflow runs `pnpm size` on the main branch so all of the changes cannot be added in one PR.


lakhveerk (approved):
Thank you for adding it!

TrevorJoelHarris (approved):

PR #2694, opened 2025-01-22
[Task] Programmatically access the "real" App ID and Manifest Version 

MengyiGong:

## Description
Adding the `appId` and `manifestVersion` to `AppInfo` so app can get  `appId` and `manifestVersion`  from `getContext()` API.

### Main changes in the PR:

1. Added `appId?: AppId`
2. Add `manifestVersion?: string`

Screenshot 2025-01-23 at 4 18 59 PM

### DIFF
--- a/packages/teams-js/src/public/app/app.ts
+++ b/packages/teams-js/src/public/app/app.ts
@@ -816,6 +816,8 @@ function transformLegacyContextToAppContext(legacyContext: LegacyContext): Conte
         ringId: legacyContext.ringId,
       },
       appLaunchId: legacyContext.appLaunchId,
+      appId: legacyContext.appId ? new AppId(legacyContext.appId) : undefined,
+      manifestVersion: legacyContext.manifestVersion,
     },
     page: {
       id: legacyContext.entityId,
--- a/packages/teams-js/src/public/app/app.ts
+++ b/packages/teams-js/src/public/app/app.ts
@@ -15,6 +15,7 @@ import { ensureInitializeCalled } from '../../internal/internalAPIs';
 import { ApiName, ApiVersionNumber, getApiVersionTag, getLogger } from '../../internal/telemetry';
 import { inServerSideRenderingEnvironment } from '../../internal/utils';
 import * as messageChannels from '../../private/messageChannels/messageChannels';
+import { AppId } from '../appId';
 import { ChannelType, FrameContexts, HostClientType, HostName, TeamType, UserTeamRole } from '../constants';
 import {
   ActionInfo,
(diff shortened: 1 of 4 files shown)


packages/teams-js/src/public/app/app.ts
```diff
@@ -186,6 +187,20 @@ export interface AppInfo {
    * ID for the current visible app which is different for across cached sessions. Used for correlating telemetry data.
    */
   appLaunchId?: string;
+
+  /**
+   * This ID is the unique identifier assigned to the application after deployment and is critical for ensuring the correct app instance is recognized across hosts.
+   *
+   * @beta
+   */
+  appId?: AppId;
```
AE-MS, line 196: Since this property is an object of type `AppId` (which it should be! 😉), does there need to be code somewhere which deserializes the value when it comes from the host SDK (as a string) and turns it into an `AppId` object?

AE-MS (changes requested):
:clock1:

AE-MS (changes requested):
:clock1:

AE-MS:
Are the tests that can be added to validate that the app id on the context is getting set correctly based on what the host SDK returns?

packages/teams-js/src/public/app/app.ts
MengyiGong: Thanks! That's a great point, updated!

MengyiGong:
> Are the tests that can be added to validate that the app id on the context is getting set correctly based on what the host SDK returns?

That's a good point! I checked with @Kangxuan-Ye , we can only test in web side. I will add e2e test to cover this part in web and mobile.

KangxuanYe (dismissed):

packages/teams-js/src/public/app/app.ts
```diff
@@ -805,6 +816,8 @@ function transformLegacyContextToAppContext(legacyContext: LegacyContext): Conte
         ringId: legacyContext.ringId,
       },
       appLaunchId: legacyContext.appLaunchId,
+      appId: legacyContext.appId ? new AppId(legacyContext.appId) : undefined,
```
TrevorJoelHarris, line 819: I think it would be good to expand the unit tests in `app.spec.ts` to test for these new properties (especially app id which is getting converted into a new type)

MengyiGong: Thanks for the guidency! Updated the unit test in `app.spec.ts`.

MengyiGong:
> Are the tests that can be added to validate that the app id on the context is getting set correctly based on what the host SDK returns?

Hi @AE-MS , thanks for pointing this put. I updated the unit test in `app.spec.ts` to verfiy the `appId` and `manifestVersion` are as expected.

TrevorJoelHarris (approved):

packages/teams-js/test/public/app.spec.ts
```diff
@@ -568,6 +568,8 @@ describe('Testing app capability', () => {
             appLaunchId: 'appLaunchId',
             userDisplayName: 'someTestUser',
             teamSiteId: 'someSiteId',
+            appId: 'mock.m365testapp.test',
```
AE-MS, line 571: >'mock.m365testapp.test' [](http://example.com/codeflow?start=19&length=23)

Can you make this a constant that is referenced in all the places you duplicate it today?

(also for the version string)

Thanks!

MengyiGong: Updated both to constant!

AE-MS (approved):
:shipit:

PR #2698, opened 2025-01-24
Disable CodeQL from pipeline level and add it in job level

maglims:

## Description

**What is codeql, and what the persistent issues it has currently?**
From what I understand in a briefly way,  it is static analysis tool, runs on source codes for all kind of languages, and also traces the build process during to build, to capture any security  vulnerability issues. It creates a DB during its run, which I believe records the scan result, and the DB will be uploaded in the pipeline to be source of different reporting/dashboards.
CodeQL currently has some issues arm64 macos image, which has been known for a while and in the back of codeQL team. This is not super clear what is the issues to me, and there is an existing ICM on codeql

**When it runs in 1ES template?**
CodeQL is injected task if the pipeline is on 1ES template. From our observations, 1ES always injects the run  at the beginning of each task, i.e, before it runs any other tasks like tool installing. When the tool is installing, it may be interferred with codeql, causing the failure (Not exactly know the reason)

**How manual run is different from codeql auto-injection task?**
There is no difference between auto-injected codeQL and manual codeQL task themselves. The only difference is with manual, we/our team can decide when to inject. The workaround is to inject after the tool setup has completed, to let codeql have less interruptions with the build process. A recommended order from CodeQL team:
image

**What is our solution here?**
Disable from pipeline level, and then add manual task at each job level. Please note, Codeql doesn't support job level disable as of today, so in order to disable, we have to do it at pipeline level.

**Will we receive the S360 alerts?**
Given we are disabling the CodeQL at pipeline/repo level. Will we receive S360 alerts? 
The answer should be No. We have confirmed with CodeQL team.  As long as DB is uploaded, S360 will have some results to process, meeting the requirement.  It doesn’t track where the DB is created (from 1ES template, or manual run doesn't matter)

### Main changes in the PR:

1. Disable CodeQL from the pipeline
2. Manually add the task at job level

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -64,7 +64,7 @@ extends:
       codeql:
         compiled:
           enabled: false
-          justificationForDisabling: 'CodeQL has some know issues with arm64 macos. Disabling auto-injection and using manual task'
+          justificationForDisabling: 'CodeQL has some known issues with arm64 macos. Disabling auto-injection and using manual task instead'
     customBuildTags:
       - ES365AIMigrationTooling
     stages:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -61,6 +61,10 @@ extends:
           - repository: AppHostingSdkV4
           - repository: AndroidAppHostingSdk
           - repository: IOSAppHostingSdk
+      codeql:
+        compiled:
+          enabled: false
+          justificationForDisabling: 'CodeQL has some know issues with arm64 macos. Disabling auto-injection and using manual task'
     customBuildTags:
       - ES365AIMigrationTooling
     stages:
--- a/tools/yaml-templates/android-test.yml
+++ b/tools/yaml-templates/android-test.yml
@@ -60,10 +60,7 @@ steps:
 
   - task: CodeQL3000Init@0
     inputs:
-      ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
         Enabled: true
-      ${{ else }}:
-        Enabled: false
         
   - task: Bash@3
     displayName: 'Install Emulator'
--- a/tools/yaml-templates/android-test.yml
+++ b/tools/yaml-templates/android-test.yml
@@ -58,6 +58,13 @@ steps:
       script: 'nohup pnpm start-test-app &'
       workingDirectory: '$(ClientSdkProjectDirectory)'
 
+  - task: CodeQL3000Init@0
+    inputs:
+      ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
+        Enabled: true
+      ${{ else }}:
+        Enabled: false
+        
   - task: Bash@3
     displayName: 'Install Emulator'
     # This task is flaky, but when it works it almost always completes in under 5 minutes.
@@ -88,6 +95,9 @@ steps:
       script: 'chmod u+x ./gradlew && chmod u+x e2eTest.sh && ./e2eTest.sh ${{ parameters.shardNum }} ${{ parameters.shardIndex }}'
       workingDirectory: '$(Agent.BuildDirectory)/androidHost/apps/orangeandroid'
 
+  - task: CodeQL3000Finalize@0
+    condition: always()
+    
   - task: Bash@3
     displayName: 'Pull Test Artifacts'
     inputs:
--- a/tools/yaml-templates/build-app-host.yml
+++ b/tools/yaml-templates/build-app-host.yml
@@ -75,10 +75,7 @@ steps:
 
   - task: CodeQL3000Init@0
     inputs:
-      ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
-        Enabled: true
-      ${{ else }}:
-        Enabled: false
+      Enabled: true
 
   - script: |
       pnpm build
--- a/tools/yaml-templates/build-app-host.yml
+++ b/tools/yaml-templates/build-app-host.yml
@@ -73,6 +73,13 @@ steps:
     workingDirectory: '$(AppHostingSdkProjectDirectory)'
     condition: eq(variables.CYPRESS_CACHE_RESTORED, 'false')
 
+  - task: CodeQL3000Init@0
+    inputs:
+      ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
+        Enabled: true
+      ${{ else }}:
+        Enabled: false
+
   - script: |
       pnpm build
     displayName: 'Build app hosting sdk'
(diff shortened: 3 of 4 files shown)


tools/yaml-templates/ios-test.yml
```diff
@@ -87,6 +87,13 @@ steps:
       script: 'nohup pnpm start-test-app --server-options-cert $(localhostSSLCertificate.secureFilePath) --server-options-key $(localhostSSLKey.secureFilePath) &'
       workingDirectory: '$(ClientSdkProjectDirectory)'
 
+  - task: CodeQL3000Init@0
+    inputs:
+      ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
```
maglims, line 92: According to codeQL, it should only run default branch (master or main)

noahdarveau-MSFT:
To clarify my understanding, we will still be running codeQL, just at a later point in the pipeline? Will this have any impacts on its scanning capabilities?

azure-pipelines.yml
```diff
@@ -61,6 +61,10 @@ extends:
           - repository: AppHostingSdkV4
           - repository: AndroidAppHostingSdk
           - repository: IOSAppHostingSdk
+      codeql:
```
KangxuanYe, line 64: Why we need to disable from all-platform level? Isn't iOS only because it is the only platform using ARM64 VM?

tools/yaml-templates/android-test.yml
```diff
@@ -58,6 +58,13 @@ steps:
       script: 'nohup pnpm start-test-app &'
       workingDirectory: '$(ClientSdkProjectDirectory)'
 
+  - task: CodeQL3000Init@0
```
KangxuanYe, line 61: Same thing here because Android is using MacOS 15. Not sure it is ARM based or Intel based or mixed.

tools/yaml-templates/build-app-host.yml
```diff
@@ -73,6 +73,13 @@ steps:
     workingDirectory: '$(AppHostingSdkProjectDirectory)'
     condition: eq(variables.CYPRESS_CACHE_RESTORED, 'false')
 
+  - task: CodeQL3000Init@0
```
KangxuanYe, line 76: Web is running on Linux and I think it is fine to keep it as-is.

tools/yaml-templates/ios-test.yml
KangxuanYe: I don't mind setting it explicitly but I think codeQL by default checks branch. That is to say, in your branch, codeQL would not be triggered (because it isn't running on main) maybe and this inputs setting can be saved?

Also I cannot think of other way we can do something except checking this in and then monitor main branch to see if it fails or not. Hopefully it is not.

azure-pipelines.yml
```diff
@@ -61,6 +61,10 @@ extends:
...
+        compiled:
+          enabled: false
+          justificationForDisabling: 'CodeQL has some know issues with arm64 macos. Disabling auto-injection and using manual task'
```
jadahiya-MSFT, line 67: nit: `known`

tools/yaml-templates/build-app-host.yml
jadahiya-MSFT: @KangxuanYe is right here, this change is unnecessary.

tools/yaml-templates/ios-test.yml
jadahiya-MSFT: CodeQL will by default check for changes to `main`. Using the inputs tag here is to add additional branches. I would recommend modifying this to look for both `main` and `release/*` branches instead of just main.

tools/yaml-templates/build-app-host.yml
maglims: We need this change here, otherwise the codeQL will not run for web repo because the pipeline level codeQL skipping. It is also the same reason as I explained on "CodeQL is disabled at pipeline level" for the other two comments.

This test build should be able to explain better. https://office.visualstudio.com/ISS/_build/results?buildId=34120234&view=logs&j=dd9cb201-f7ed-50bd-a8c9-517b9a22558e&t=0adf847b-690e-5ae5-3253-69f3bf6c88d1 
As we can see, "CodeQl initialize" will always be skipped. The the only way to run scan on web repo is to add this task manually, which you can see the job "CodeQL3000Init@0" in the screen shot
image

maglims:
> To clarify my understanding, we will still be running codeQL, just at a later point in the pipeline? Will this have any impacts on its scanning capabilities?

@noahdarveau-MSFT , this is a really good question, and also the core of changes. I have added some details in the PR description, let me see if I can add more details.

1. Yes, we are still running CodeQL. As you can see, "CodeQL initialize" is the 1ES template injected task, which has been skipped now (Unfortunatly we cannot only skip this only for iOS. That is also why you see that we are adding it back to each job/repo that needs CodeQL scan)
image. 

Down there there is "CodeQL3000Init@0" task, which is the manual task we are adding now, to ensure the coverage.

2. What is the difference between "CodeQL Initialize" and "CodeQL3000Init@0"? 
The tasks themselves are no difference on what they do, i.e, they do the same scan/job. With "CodeQL3000Init@0" task, we can decide where to put it or when we want to run the job. So as you can see, right now, we are putting it after tooling setup (specially you can see that it is running after "Setup pnpm"). That is also the reason why manual task  should be able to fix the issue, because 'setup pnpm' already finishes and there is no way codeQL can mess around with it.

3. Do we lose any scanning capability by running it in later stage? 
We are adding the tasks per CodeQL's suggestions. So there should be no concern from their point of view. Here is what CodeQL suggests how to run codeQL manually, i.e, it just needs to cover compile/build phase.
image

This ensures the scan coverage, without interferring other stages.

As a comparsion of how auto-injected task works, 1ES template always injects at very beginning of the job, which we don't have control. So you can image the order is something like below. So any tasks/tool setting up between "CodeQL init" and "CodeQL Finalize", can be interfered by CodeQL in an unexpected way. This is also the reason why we see intermittent failure.
Job:
steps:
**- CodeQL Init**
- install dotnet
- install python
- run script
**- setup pnpm**
- other non-compile tasks
- compile task
- other non-compile tasks
**- - CodeQL Finalize**

tools/yaml-templates/ios-test.yml
maglims: ok, I think you guys have a point and I also observed the skipping of run on non-default branch. Let me remove it since it doesn't seem doing anything better.

noahdarveau-MSFT (approved):

KangxuanYe (approved):

jadahiya-MSFT (approved):

PR #2702, opened 2025-01-27
noahdarveau/size limit github workflow

noahdarveau-MSFT:
Added a github workflow that will automatically run size-limit on PRs to compare the size changes in the PR to the main branch.


noahdarveau-MSFT:
I updated the path so that it only references the index.js file, rather than all of the files in the dist folder. The reported sizes should still come out the same, however it will make the comment it leaves on the PR much smaller.

lakhveerk (approved):

AE-MS (approved):

PR #2707, opened 2025-01-28
Noahdarveau/global polyfill

noahdarveau-MSFT:
Lately, we have had a number of issues popping up with vite users saying that `global` is undefined. This is due to the fact that vite the vite environment does not define the `global` object by default. This is a commonly known issue and a common solution is for vite app developers to define `global` in their applications or vite configurations. With the string of issues that have been coming in, we've been offering solutions that work for a large majority of people, though we are commonly getting new users chiming in that their slightly unique setups are still having issues. While unncessary on our part, I'm proposing adding a `global` polyfill to the package as I believe it will be a more robust solution to a wider range of vite users. I can't think of any immediate issues that would come to mind with the inclusion of a `global` polyfill, though with the messy nature of the javascript ecosystem, it is hard to be 100% certain it will not cause any new issues to arise for other library consumers.

### DIFF
--- a/packages/teams-js/rollup.config.mjs
+++ b/packages/teams-js/rollup.config.mjs
@@ -7,7 +7,7 @@ import replace from '@rollup/plugin-replace';
 import terser from '@rollup/plugin-terser';
 import typescript from '@rollup/plugin-typescript';
 import { readFileSync } from 'fs';
-import nodePolyfills from 'rollup-plugin-node-polyfills';
+import nodePolyfills from 'rollup-plugin-polyfill-node';
 
 const packageJson = JSON.parse(readFileSync('./package.json', 'utf-8'));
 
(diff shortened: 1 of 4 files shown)


packages/teams-js/rollup.config.mjs
```diff
@@ -7,7 +7,7 @@ import replace from '@rollup/plugin-replace';
 import terser from '@rollup/plugin-terser';
 import typescript from '@rollup/plugin-typescript';
 import { readFileSync } from 'fs';
-import nodePolyfills from 'rollup-plugin-node-polyfills';
+import nodePolyfills from 'rollup-plugin-polyfill-node';
```
jadahiya-MSFT, line 10: Is the global polyfill included in this upgraded package?

noahdarveau-MSFT: Yes, due to some differences in how these two plugins function, using the new plugin will result in the global polyfill being included in the bundle.

TrevorJoelHarris (approved):

PR #2708, opened 2025-01-28
Add field to tag proxied messages.

juanscr:
## Description
This pull requests add a new field to tag messages proxied from child, in order for the Host to differentiate between parent messages and child originated ones.

### Unit Tests added:
Yes

### End-to-end tests added:
No

### Change file added:
Yes

### DIFF
--- a/change/@microsoft-teams-js-eef39838-66eb-4b2d-8333-7b4194fb0b47.json
+++ b/change/@microsoft-teams-js-eef39838-66eb-4b2d-8333-7b4194fb0b47.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added field to tag child proxied messages.",
+  "packageName": "@microsoft/teams-js",
+  "email": "jcardenasr123@gmail.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-teams-js-eef39838-66eb-4b2d-8333-7b4194fb0b47.json
+++ b/change/@microsoft-teams-js-eef39838-66eb-4b2d-8333-7b4194fb0b47.json
@@ -1,5 +1,5 @@
 {
-  "type": "minor",
+  "type": "patch",
   "comment": "Added field to tag child proxied messages.",
   "packageName": "@microsoft/teams-js",
   "email": "jcardenasr123@gmail.com",
--- a/packages/teams-js/src/internal/communication.ts
+++ b/packages/teams-js/src/internal/communication.ts
@@ -1179,7 +1179,7 @@ function getMessageIdsAsLogString(
  * @internal
  * Limited to Microsoft-internal use
  */
-function proxyChildMessageToParent(message: MessageRequest): void {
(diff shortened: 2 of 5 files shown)


change/@microsoft-teams-js-eef39838-66eb-4b2d-8333-7b4194fb0b47.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
TrevorJoelHarris, line 2:
```suggestion
  "type": "patch",
```

packages/teams-js/src/internal/communication.ts
```diff
@@ -927,6 +930,7 @@ function handleIncomingMessageFromChild(evt: DOMMessageEvent): void {
             sendMessageResponseToChild(message.id, message.uuid, args, isPartialResponse);
           }
         },
+        true,
```
TrevorJoelHarris, line 933: Either adding a comment on this line or making this boolean a named variable will help the readability here - it took me a couple of seconds to figure out exactly what this boolean was doing

packages/teams-js/src/internal/communication.ts
```diff
@@ -430,6 +431,7 @@ export function sendMessageToParent(
   actionName: string,
   argsOrCallback?: any[] | Function,
   callback?: Function,
+  isProxiedFromChild?: boolean,
```
TrevorJoelHarris, line 434: Something to consider - adding this property to an exported function like this means that other consumers in the teamsjs library are technically able to set `isProxiedFromChild` to true even if it isn't. Since we don't currently document what any of these parameters do, it seems likely some people will set it wrong without realizing what effect it has.

At the same time, this function is marked as deprecated and I *think* we should expect few (no?) people to call it directly.

Depending on how concerned you are about well-meaning engineers accidentally calling this and setting the value wrong, you could leave `isProxiedFromChild` off of `sendMessageToParent` and call a new, non-exported function that lets you set `isProxiedFromChild` from handleIncomingMessageFromChild

TrevorJoelHarris: (It may be fair to say "that function is deprecated and this isn't worth worrying about". Your call)

TrevorJoelHarris (changes requested):

packages/teams-js/src/internal/communication.ts
juanscr: I can see both ways, but it is a reality that people may add new code using this function, as it still widely used throughout the codebase even if it's deprecated. Would prefer to avoid any issue later down the line with confusion with this parameter! I'll create a new internal function

packages/teams-js/src/internal/communication.ts
juanscr: Seeing the diff in github makes it even weirder lol... I'll add the comment

AE-MS (approved):
:shipit:

TrevorJoelHarris (approved):

PR #2717, opened 2025-02-06
noahdarveau/size-limit test

noahdarveau-MSFT:
Added a configuration to size-limit that will build with `app`, `authentication`, and `pages`. The size limit threshold is set to the size of these capabilities. If any of the other capabilities stop treeshaking, then the size of the build will increase above this threshold and fail the build. However, if someone is making changes to `app`, `authentication`, or `pages`, then they will need to adjust the size limit threshold accordingly. The size-limit step runs as part of the build step, so if there is a size issue, it will fail at build time.


package.json
```diff
@@ -158,6 +158,12 @@
     "uuid": "^9.0.1"
   },
   "size-limit": [
+    {
+      "brotli": false,
+      "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
+      "import": "{ app, authentication, pages }",
+      "limit": "58.033 KB"
```
AE-MS, line 165: >"limit": "58.033 KB" [](http://example.com/codeflow?start=6&length=20)

Is it important that this be so tightly constrained to the current size? E.g., do we want to allow for a little bit of slack in the limit (to allow for some refactoring/etc.)?

AE-MS (changes requested):
See comments

package.json
noahdarveau-MSFT: That is something we could do. I ran into a bit of a problem trying to figure out _how much_ wiggle room we should allow, that would both allow for refactoring and small changes but still catch treeshaking breakages.

noahdarveau-MSFT:
> What will the developer experience be like when someone makes a change that exceeds the limit?

If a developer makes a change that causes the size limit to be exceeded, their build will fail with it throwing an error that the size limit has been increased. It will then be up to the developer to determine if this is a warranted size increase, and either increase the size limit accordingly or resolve the cause of the size increase.

AE-MS:
> > What will the developer experience be like when someone makes a change that exceeds the limit?
> 
> If a developer makes a change that causes the size limit to be exceeded, their build will fail with it throwing an error that the size limit has been increased. It will then be up to the developer to determine if this is a warranted size increase, and either increase the size limit accordingly or resolve the cause of the size increase.

Do we have any control over what the message says? E.g., how will the developer know where to update the size? How will the developer know what to consider when deciding whether they've "broken" tree shaking? At the very least can we put a comment near the size definition to say "Hey if you are changing this you better be sure you didn't break tree shaking; don't just update this because your build failed. Here's a link that describes tree shaking and how you can know whether you broke it: etc. etc."

packages/teams-js/package.json
```diff
@@ -12,15 +12,15 @@
   "typings": "./dist/esm/packages/teams-js/dts/index.d.ts",
   "module": "./dist/esm/packages/teams-js/src/index.js",
   "scripts": {
-    "build": "pnpm clean && pnpm lint && pnpm build-rollup && pnpm build-webpack && pnpm docs:validate",
+    "build": "pnpm clean && pnpm lint && pnpm build-rollup && pnpm build-webpack && pnpm docs:validate && size-limit",
```
jadahiya-MSFT, line 15: Do you anticipate that the below `size` command will be used in a more complex manner in the future? If not, we could simplify future code changes by calling `pnpm size` in the build script and then only have to update the `size` command below whenever we make an intended change to functionality.

package.json
```diff
@@ -158,10 +158,27 @@
     "uuid": "^9.0.1"
   },
   "size-limit": [
+    {
+      "brotli": false,
```
jadahiya-MSFT, line 162: Hmmm this is interesting, alongside our treeshaking capabilities should we also look into brotli compressing our files in the future if we're not already. This comment is moreso to start a conversation for the future of some size reduction changes we could make.

package.json
```diff
@@ -158,10 +158,27 @@
...
+    },
     {
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "*"
     }
+  ],
+  "size-limit-explanation": [
+    "There are 2 configurations for size-limit. The first configuration is used to test treeshakability in teams-js. If treeshaking is broken in any capability other than `app`, `authentication`, or `pages`, then the size of the build",
+    "will increase above the set 58.033KB. The limit is currently set to the exact size of the 3 specified files when built. If your build ever fails because of this size-limit check there are a few scenarios you may fall into:",
```
jadahiya-MSFT, line 175: Is this a strict limit or do we allow for very minor size changes like <1% as an example?

packages/teams-js/package.json
noahdarveau-MSFT: I'm not sure I fully understand what you're saying here. Do you mean not running `pnpm size` as part of `pnpm build` and instead running it as a pipeline step in the pipeline?

package.json
noahdarveau-MSFT: Since the point of the limit is to detect treeshaking breakages, I wanted to have it fairly close to the actual size of the 3 files listed, but if you think it would be less of a headache for developers, I'm open to increasing the limit slightly to allow for a bit of wiggle room. I just want to be cautious about creating a habit of permissiveness when it comes to increasing the size limit as I feel it could lose some of its effectiveness.

package.json
noahdarveau-MSFT: I don't think compressing our library would be particularly useful, since it would need to be decompressed by the library consumer when they consume it. It would be possible as something for them to do but would most definitely be a breaking change. Compressing the package would be a more beneficial optimization for an app, rather than the library.

package.json
jadahiya-MSFT: I think my question is moreso can we guarantee that these 3 files will always build to that exact size if no changes are made?

Additionally, this is an upper limit? What happens if we make major strides to lower the size of these files is there any suggestions for when we should lower this upper limit in the future?

packages/teams-js/package.json
jadahiya-MSFT: Sorry, I what I meant was why do we have `&& size-build` instead of `&& pnpm size` in the `build` command.

noahdarveau-MSFT: Oh yeah, that was an oopsie on my part. Fixed!

package.json
noahdarveau-MSFT: So as long as no changes are made to these 3 files, or any of their dependencies the size won't change. I went ahead and bumped up the size limit slightly, but it's really hard to know a good 'buffer' amount without just putting it into practice and seeing. If the size goes down, we will need to adjust the limit to re-correspond with the new size.

jadahiya-MSFT (approved):

jekloudaMSFT (approved):

PR #2718, opened 2025-02-07
noahdarveau/repo cleanup

noahdarveau-MSFT:
Just doing a few cleanup changes to the repo:

1. The treeshaking test app was still referencing a test npm library owned by Trevor. It doesn't make much sense to have that in the repo, especially now that he's leaving the team.
2. The uint8array-extras was a package I was using for some buffer polyfill removal stuff. Now that it's been determined we can't use this alternative due to it not being supported in jsdom causing jest tests using the library to fail, I'm removing the extraneous code.


jadahiya-MSFT (approved):

AE-MS (approved):
:shipit:

jadahiya-MSFT (approved):

PR #2724, opened 2025-02-18
noahdarveau/dynamic imports fix

noahdarveau-MSFT:
Teams-JS currently has a bug that occurs when library consumers import the library dynamically, such as when using the syntax 
`import('@microsoft/teams-js').then(({ app, pages }) => { //DO STUFF });`, instead of the commonly used static import syntax, `import { app, pages } from '@microsoft/teams-js'`. For some reason that I still am not entirely sure the reason of (and I doubt I'll ever definitively be able to pinpoint, as it pertains to the innerworkings of how ecmascript async module resolution functions), when using rollup to build an app with the dynamic import syntax, it causes an error stating: `Uncaught (in promise) ReferenceError: Cannot access 'backStack' before initialization`. This error is occurring because for some reason, when rollup is generating the bundle, the `backstack` object is being defined _after_ the `pages` object, despite the fact that the `pages` object references the `backstack` object, so when the bundle is parsed by the browser at runtime, it reaches the `backstack` reference in the `pages` object, and sees that it is undefined and throws the error. From what I could find in my investigation, this is most likely due to the large number of circular dependencies in our library, especially surrounding the `pages` capability. I was able to determine that the line in the `handlers.ts` file that calls `pages.backstack._initialize()` was problematic as it was a major contributing factor to the circular dependencies and commenting out that line appeared to alleviate the issue. In order to remove the `pages` reference from `handlers.ts`, I refactored the `backstack` initialization code to be inside a helper file in the `pagesHelpers.ts` file. This way we can have `handlers.ts` call the helper function directly and import `pagesHelpers.ts` instead of `pages.ts`. Since the `_initialization` function in `backstack` is technically an exported api by the library, we have to leave its calling signature unchanged to maintain back compat, so I refactored it to just call the helper function, even if it isn't being used anywhere else in the library anymore after the helper refactor.

### DIFF
--- a/packages/teams-js/src/internal/pagesHelpers.ts
+++ b/packages/teams-js/src/internal/pagesHelpers.ts
@@ -10,13 +10,16 @@ import {
 import * as pages from '../public/pages/pages';
 import { runtime } from '../public/runtime';
 import {
+  Communication,
   sendAndHandleStatusAndReason,
   sendAndHandleStatusAndReasonWithDefaultError,
   sendAndUnwrap,
+  sendMessageEventToChild,
   sendMessageToParent,
 } from './communication';
+import { registerHandler } from './handlers';
 import { ensureInitialized } from './internalAPIs';
-import { ApiVersionNumber } from './telemetry';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from './telemetry';
 
 /**
  * v2 APIs telemetry file: All of APIs in this capability file should send out API version v2 ONLY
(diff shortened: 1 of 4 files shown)


jekloudaMSFT (approved):

packages/teams-js/src/internal/pagesHelpers.ts
```diff
@@ -183,3 +186,35 @@ export function convertAppNavigationParametersToNavigateToAppParams(
     webUrl: params.webUrl ? params.webUrl.toString() : undefined,
   };
 }
+
+//Back Stack Helpers
+
+let backButtonPressHandler: (() => boolean) | undefined;
+
+export function getBackButtonPressHandler(): (() => boolean) | undefined {
```
juanscr, line 194: Is this function needed? I don't see any code calling this function.

packages/teams-js/src/internal/pagesHelpers.ts
```diff
@@ -183,3 +186,35 @@ export function convertAppNavigationParametersToNavigateToAppParams(
...
+function handleBackButtonPress(): void {
+  if (!backButtonPressHandler || !backButtonPressHandler()) {
+    if (Communication.childWindow) {
+      // If the current window did not handle it let the child window
+      sendMessageEventToChild('backButtonPress', []);
+    } else {
+      pages.backStack.navigateBack();
+    }
+  }
+}
```
juanscr, line 220: Minor nit pick but I think it would be good to have the exported functions tagged as @internal as these are new exported functions available in the whole teams-js codebase.

juanscr (changes requested):

packages/teams-js/src/internal/pagesHelpers.ts
noahdarveau-MSFT: The `backButtonPressHandler` variable used to be in the `backstack` file and I moved it to the `pagesHelpers` file. It used to be referenced directly but since you can't assign to exported variable, I gave it a getter and a setter. The function that reads the value of `backButtonPressHandler` also moved to the `pagesHelpers` file so it can still reference it directly, but I wanted to give the variable a getter to indicate to future devs that it can be used elsewhere, like in the original backstack file, even though it's not there anymore. I didn't want to export the variable directly as I felt that might lead to some confusion with trying to assign to it. But if you feel handling it differently is better, I'm fine with changing it.

packages/teams-js/src/internal/pagesHelpers.ts
noahdarveau-MSFT: All of the exported functions in pagesHelpers file are only available for use within teams-js, they are not exported through the index files, so you cannot invoke them as a library consumer, so tagging them seems kind of unnecessary.

packages/teams-js/src/internal/pagesHelpers.ts
juanscr: Okey dokey, wanted to make sure :)

juanscr (approved):

jadahiya-MSFT (approved):

PR #2728, opened 2025-02-20
Refactor child communication to own module

juanscr:
## Description
This pull request refactors all the child communication to an isolated module. Although this allows for clear and independent separation for how child proxying works, this is mostly done so in a next PR we can tree shake this module with a feature flag. There is no particular functionality change introduced in this pull request.

### Main changes in the PR:

1. Refactor child communication to an isolated module.
2. Added proper unit tests to detail how child proxying and child event relaying works.

### Validation performed:

1. Ran set of tests.
2. Ran child app to see if everything worked as usual.

### Unit Tests added:
Yes

### End-to-end tests added:

No

### Change file added:
Yes

### Next/remaining steps:
A follow-up PR will make add a feature flag for using this module, which would then make it so all child proxying code is removed in the final bundle if not needed.

### DIFF
--- a/packages/teams-js/src/internal/childCommunication.ts
+++ b/packages/teams-js/src/internal/childCommunication.ts
@@ -62,7 +62,7 @@ type SetCallbackForRequest = (uuid: MessageUUID, callback: Function) => void;
  * @internal
  * Limited to Microsoft-internal use
  */
-export function shouldMessageBeProxiedToParent(messageSource: Window, messageOrigin: string): boolean {
+export function shouldProcessChildMessage(messageSource: Window, messageOrigin: string): boolean {
   if (!ChildCommunication.window || ChildCommunication.window.closed || messageSource === ChildCommunication.window) {
     ChildCommunication.window = messageSource;
     ChildCommunication.origin = messageOrigin;
@@ -83,7 +83,7 @@ export function shouldMessageBeProxiedToParent(messageSource: Window, messageOri
  * @internal
  * Limited to Microsoft-internal use
  */
-export async function proxyChildMessageToParent(
+export async function handleIncomingMessageFromChild(
   evt: DOMMessageEvent,
   messageSource: Window,
   sendMessageToParentHelper: SendMessageToParentHelper,
@@ -98,7 +98,7 @@ export async function proxyChildMessageToParent(
   flushMessageQueue(ChildCommunication.window, ChildCommunication.origin, ChildCommunication.messageQueue, 'child');
 
   // Handle the message
-  handleIncomingMessageFromChild(evt, sendMessageToParentHelper, setCallbackForRequest);
+  handleIncomingMessage(evt, sendMessageToParentHelper, setCallbackForRequest);
 }
 
 const handleIncomingMessageFromChildLogger = communicationLogger.extend('handleIncomingMessageFromChild');
@@ -107,7 +107,7 @@ const handleIncomingMessageFromChildLogger = communicationLogger.extend('handleI
  * @internal
  * Limited to Microsoft-internal use
  */
-function handleIncomingMessageFromChild(
+function handleIncomingMessage(
   evt: DOMMessageEvent,
   sendMessageToParentHelper: SendMessageToParentHelper,
   setCallbackForRequest: SetCallbackForRequest,
--- a/packages/teams-js/src/internal/childCommunication.ts
+++ b/packages/teams-js/src/internal/childCommunication.ts
@@ -1,0 +1,255 @@
+/* eslint-disable @typescript-eslint/ban-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { UUID as MessageUUID } from '../public/uuidObject';
+import { flushMessageQueue, getMessageIdsAsLogString } from './communicationUtils';
+import { callHandler } from './handlers';
+import { DOMMessageEvent } from './interfaces';
+import {
+  deserializeMessageRequest,
+  MessageID,
+  MessageRequest,
+  MessageRequestWithRequiredProperties,
+  MessageResponse,
+  SerializedMessageRequest,
+  serializeMessageResponse,
+} from './messageObjects';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from './telemetry';
+import { getLogger } from './telemetry';
+
+const communicationLogger = getLogger('childProxyingCommunication');
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+class ChildCommunication {
+  public static window: Window | null;
+  public static origin: string | null;
+  public static messageQueue: MessageRequest[] = [];
+}
+
+/**
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function uninitializeChildCommunication(): void {
+  ChildCommunication.window = null;
+  ChildCommunication.origin = null;
+  ChildCommunication.messageQueue = [];
+}
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function shouldEventBeRelayedToChild(): boolean {
+  return !!ChildCommunication.window;
+}
+
+type SendMessageToParentHelper = (
+  apiVersionTag: string,
+  func: string,
+  args?: any[],
+  isProxiedFromChild?: boolean,
+) => MessageRequestWithRequiredProperties;
+
+type SetCallbackForRequest = (uuid: MessageUUID, callback: Function) => void;
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function shouldMessageBeProxiedToParent(messageSource: Window, messageOrigin: string): boolean {
+  if (!ChildCommunication.window || ChildCommunication.window.closed || messageSource === ChildCommunication.window) {
+    ChildCommunication.window = messageSource;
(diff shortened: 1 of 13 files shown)


packages/teams-js/src/internal/childCommunication.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  return ChildCommunication.window === messageSource;
+}
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export async function proxyChildMessageToParent(
```
jekloudaMSFT, line 86: I'm a little confused by the name of this function, doesn't it post a reply to the child window when there's a registered handler, and then only if there is no handler it sends the message to the parent?

packages/teams-js/src/internal/childCommunication.ts
```diff
@@ -0,0 +1,255 @@
...
+type SetCallbackForRequest = (uuid: MessageUUID, callback: Function) => void;
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function shouldMessageBeProxiedToParent(messageSource: Window, messageOrigin: string): boolean {
+  if (!ChildCommunication.window || ChildCommunication.window.closed || messageSource === ChildCommunication.window) {
+    ChildCommunication.window = messageSource;
```
jekloudaMSFT, line 67: Should we do anything with the ChildCommunication message queue here? Or in the condition below where we clean up pointers to child windows?

packages/teams-js/src/internal/childCommunication.ts
```diff
@@ -0,0 +1,255 @@
...
+) => MessageRequestWithRequiredProperties;
+
+type SetCallbackForRequest = (uuid: MessageUUID, callback: Function) => void;
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function shouldMessageBeProxiedToParent(messageSource: Window, messageOrigin: string): boolean {
```
jekloudaMSFT, line 65: Same question about if this is the right name here, this seems like a more general "should we handle this message from the child?" than "should I send this on to the parent window?"

packages/teams-js/src/internal/childCommunication.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  // If we have any messages in our queue, send them now
+  flushMessageQueue(ChildCommunication.window, ChildCommunication.origin, ChildCommunication.messageQueue, 'child');
+
+  // Handle the message
+  handleIncomingMessageFromChild(evt, sendMessageToParentHelper, setCallbackForRequest);
+}
+
+const handleIncomingMessageFromChildLogger = communicationLogger.extend('handleIncomingMessageFromChild');
```
jekloudaMSFT, line 104: Do you know why we define loggers in the global namespace like this? I see that it's an existing pattern, just curious why!

packages/teams-js/test/utils.ts
```diff
@@ -136,6 +136,9 @@ export class Utils {
         return;
       },
       closed: false,
+      location: {
```
jekloudaMSFT, line 139: Why did `location` need to be added here?

packages/teams-js/src/internal/communication.ts
```diff
@@ -567,11 +565,21 @@ async function processIncomingMessage(evt: DOMMessageEvent): Promise<void> {
     }
     // Update our parent and child relationships based on this message
```
jekloudaMSFT, line 566: Just parent relationship now, I think?

packages/teams-js/src/internal/communication.ts
```diff
@@ -705,28 +713,21 @@ function updateRelationships(messageSource: Window, messageOrigin: string): void
...
     Communication.parentOrigin = null;
   }
-  if (Communication.childWindow && Communication.childWindow.closed) {
-    Communication.childWindow = null;
-    Communication.childOrigin = null;
-  }
 
   // If we have any messages in our queue, send them now
-  flushMessageQueue(Communication.parentWindow);
-  flushMessageQueue(Communication.childWindow);
```
jekloudaMSFT, line 729: Is there any change in behavior we should be aware of with not flushing the child queue every time we flush the parent queue?

juanscr: That's a great question! I think there isn't any behavioral change for two reasons that apply both before and after my PR:

1. The child message queue only gets filled with messages if a child window is not set.
2. The moment a first child window gets configured, all messages that were sent beforehand are flushed.

As such, the message queues don't get flushed at the same time, even before this PR (which is tricky because they are called one after the other). They are only flushed at the exact moment the communication layer _knows_ which window to set the message, which only happens one at a time.

packages/teams-js/src/internal/communication.ts
juanscr: Good catch! Yes, will update comment

packages/teams-js/test/utils.ts
juanscr: It doesn't precisely need to be here, but I think it makes it a bit more concise, and it aligns with the existing Window interface. I just wanted a way for the child window to have an origin associated with it easily so I can access it through the same window object.

Nevertheless, it is just for testing and I could remove it.

packages/teams-js/src/internal/childCommunication.ts
juanscr: Yess, it is done so it easily parsed when we want to know where the error come from more easily. This is the example from their docs, which you can see each extend adds a different prefix to the logger :)
image

packages/teams-js/src/internal/childCommunication.ts
juanscr: I agree, will definitely change name :)

packages/teams-js/src/internal/childCommunication.ts
juanscr: Good question! I think is more of what makes more sense, as they would be exactly the same theoretically. Would love to hear your opinion, do you find it confusing that the message gets flushed in the other function?

IMO, it makes this function more reusable as it only checks if messages can be handled by a child window but doesn't have more side effects, making it a bit purer.

packages/teams-js/src/internal/childCommunication.ts
juanscr: Yeah I can understand your point. Let me correct some of the namings.

packages/teams-js/src/internal/childCommunication.ts
juanscr: Name changed!

packages/teams-js/src/internal/childCommunication.ts
juanscr: Name changed!

noahdarveau-MSFT (commented):
Just a couple nits and questions to clarify my own understanding, but otherwise looks good to me.

packages/teams-js/src/private/privateAPIs.ts
```diff
@@ -57,8 +58,7 @@ export function sendCustomMessage(actionName: string, args?: any[], callback?: (
 
 /**
  * @hidden
- * Sends a custom action MessageEvent to a child iframe/window, only if you are not using auth popup.
- * Otherwise it will go to the auth popup (which becomes the child)
+ * Sends a custom action MessageEvent to a child iframe/window.
```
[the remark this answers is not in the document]
juanscr: Sure! On a previous PR (https://github.com/OfficeDev/microsoft-teams-library-js/pull/2710), I removed it, so authentication never gets launched as a child app of the parent but instead relays all authentication to the host. In this manner, this comment is no longer true as an authentication now doesn't rely on child proxying anymore.

packages/teams-js/src/internal/childCommunication.ts
```diff
@@ -0,0 +1,255 @@
...
+  ChildCommunication.origin = null;
+  ChildCommunication.messageQueue = [];
+}
+
+/**
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export function shouldEventBeRelayedToChild(): boolean {
```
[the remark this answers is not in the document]
juanscr: Yes! This is something that will change in a future PR with the feature flag I am adding but it was the existing behavior. What happens is if the parent app isn't registered for a specific event it _may_ happen that the child app has and there is no way to be sure. So, in this manner, if the parent app doesn't handle the event sent by the host it passes that event to the child if any,

packages/teams-js/src/internal/communication.ts
```diff
@@ -696,37 +704,30 @@ async function shouldProcessIncomingMessage(messageSource: Window, messageOrigin
  * Limited to Microsoft-internal use
  */
 function updateRelationships(messageSource: Window, messageOrigin: string): void {
-  // Determine whether the source of the message is our parent or child and update our
+  // Determine whether the source of the message is our parent and update our
```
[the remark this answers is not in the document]
juanscr: yup, its clearer using `if`.

noahdarveau-MSFT (approved):

packages/teams-js/src/internal/childCommunication.ts
jekloudaMSFT: Yeah that makes sense, thanks!

jekloudaMSFT (approved):

PR #2746, opened 2025-03-11
Disable child proxying by default and add feature flag scaffolding

juanscr:
## Description
This pull request aligns with the on-going effort of removing and deprecating child proxying in teams-js. This pull request disables child proxying by default and makes it so any app must opt-in through the `activateChildProxyingCommunication` feature flag function.  Additionally, a security issue which could possibly leak data was fixed and also guarded through a feature flag in case any app still needs this pattern which is unlikely.

Although theoretically this would be considered a "breaking change", I find it unreasonable to bump to `3.0.0` as this is mostly considered a security fix that most apps won't have to do anything. As such, will leave the changefile as minor as the pattern can still be used through feature flags.

### Main changes in the PR:

1. Added scaffolding for build and runtime feature flags in teams-js. Build feature flags are supposed to be static flags that make it so apps that are not using them, do not include that code in their bundle through tree-shaking. Meanwhile, runtime feature flags can be activated at any moment and will not remove the code from the final bundle.
2. Added the `childProxyingCommunication` build feature flag, in order for apps that do not need child proxying to not include that code in their bundle.
3. Added the runtime feature flag `disableEnforceOriginMatchForChildResponses` for disabling the security patch for the child origin match.

### Validation performed:

Through test app and unit testing.

### Unit Tests added:

Yes.

### End-to-end tests added:

No.

### Change file added:

Yes.

### DIFF
--- a/packages/teams-js/src/internal/childCommunication.ts
+++ b/packages/teams-js/src/internal/childCommunication.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+import { isChildProxyingEnabled } from '../public/featureFlags';
 import { UUID as MessageUUID } from '../public/uuidObject';
 import { flushMessageQueue, getMessageIdsAsLogString } from './communicationUtils';
 import { callHandler } from './handlers';
@@ -45,6 +46,9 @@ export function uninitializeChildCommunication(): void {
  * Limited to Microsoft-internal use
  */
 export function shouldEventBeRelayedToChild(): boolean {
+  if (!isChildProxyingEnabled()) {
+    return false;
+  }
   return !!ChildCommunication.window;
 }
 
@@ -63,6 +67,10 @@ type SetCallbackForRequest = (uuid: MessageUUID, callback: Function) => void;
  * Limited to Microsoft-internal use
  */
 export function shouldProcessChildMessage(messageSource: Window, messageOrigin: string): boolean {
+  if (!isChildProxyingEnabled()) {
+    return false;
+  }
+
   if (!ChildCommunication.window || ChildCommunication.window.closed || messageSource === ChildCommunication.window) {
     ChildCommunication.window = messageSource;
     ChildCommunication.origin = messageOrigin;
--- a/packages/teams-js/src/internal/childCommunication.ts
+++ b/packages/teams-js/src/internal/childCommunication.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { isChildProxyingEnabled } from '../public/featureFlags';
+import { getCurrentFeatureFlagsState, isChildProxyingEnabled } from '../public/featureFlags';
 import { UUID as MessageUUID } from '../public/uuidObject';
 import { flushMessageQueue, getMessageIdsAsLogString } from './communicationUtils';
 import { callHandler } from './handlers';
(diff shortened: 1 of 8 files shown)


packages/teams-js/src/public/featureFlags.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Disables origin validation for responses to child windows. When enabled, this flag bypasses security checks that verify the origin of child window that receives the response.
+   *
+   * Default: false
+   */
+  disableEnforceOriginMatchForChildResponses: boolean;
+}
+
+// Default runtime feature flags
+export const defaultFeatureFlags: RuntimeFeatureFlags = {
```
jekloudaMSFT, line 46: Why are the default runtime feature flags exported?  Do the default build feature flags need to be exported too?

packages/teams-js/src/public/featureFlags.ts
```diff
@@ -0,0 +1,76 @@
+// All build feature flags are defined inside this object. Any build feature flag must have its own unique getter and setter function. This pattern allows for client apps to treeshake unused code and avoid including code guarded by this feature flags in the final bundle. If this property isn't desired, use the below runtime feature flags object.
```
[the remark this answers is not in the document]
juanscr: Yes! I am planning to create a wiki to explain how this works. But all that if statement gets removed if someone never calls the `activateChildProxyingCommunication` function: Here I have two minimal examples that show that behavior:

No activating feature flag:
image

Activating feature flag:
image

This helps a lot as child proxying is insecure by default and allowing apps to automatically not include this insecure module if not needed, protects way more the App as it is harder to manipulate.

packages/teams-js/src/public/featureFlags.ts
juanscr: Ohh, yeah. It's not needed... I actually didn't export default build feature flags because I created a testing function for resetting the state in unit tests, which is the main usage of this. I'll change this export, thanks!

juanscr: Fixed!

packages/teams-js/src/public/featureFlags.ts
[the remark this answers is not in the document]
jekloudaMSFT: Neat! Thanks!

packages/teams-js/src/internal/childCommunication.ts
```diff
@@ -157,17 +165,32 @@ function sendChildMessageToParent(
...
   setCallbackForRequest(request.uuid, (...args: any[]): void => {
-    if (ChildCommunication.window) {
-      const isPartialResponse = args.pop();
+    if (!ChildCommunication.window) {
+      return;
+    }
+
+    if (
+      !getCurrentFeatureFlagsState().disableEnforceOriginMatchForChildResponses &&
+      requestChildWindowOrigin !== ChildCommunication.origin
```
jekloudaMSFT, line 178: Is this right? Wouldn't we want requestChildWindowOrigin to match ChildCommunication.origin when enforceOriginMatchForChildResponses is enabled?

packages/teams-js/test/public/pages.spec.ts
```diff
@@ -33,9 +35,11 @@ describe('Testing pages module', () => {
     beforeEach(() => {
       utils = new Utils();
       utils.messages = [];
+      activateChildProxyingCommunication();
```
jekloudaMSFT, line 38: Are there specific tests that require child communication here? I would like to see everything that doesn't require child communication tested with the default flags. And if the behavior is different between feature flag on/off we should have tests for both cases

juanscr: Good point! I'll refactor some of the tests to make it clear how it works with feature flag on/off :)

juanscr: Hi @jekloudaMSFT! Fixed this, took some time as I wanted to write good tests and fix typing issues already in those tests. But now they are clearly separated between default behavior and child proxying on.

jekloudaMSFT (approved):

noahdarveau-MSFT:
> Although theoretically this would be considered a "breaking change", I find it unreasonable to bump to 3.0.0 as this is mostly considered a security fix that most apps won't have to do anything. As such, will leave the changefile as minor as the pattern can still be used through feature flags.

So, to clarify, most apps will not be affected by this change and won't need to make any changes to their code? Do we know how many customers this is going to impact as a breaking change, i.e. how many customers are going to need to make changes to their codebase once this is pushed?

noahdarveau-MSFT (approved):

PR #2750, opened 2025-03-17
Standalone NAA(Nested App Auth) bridge for nested child app

baljesingh:

## Description

> This PR introduces a standalone bridge that enables a child app (embedded in an iframe) to communicate with the top-level window for deeply nested auth scenarios.
The child app can use this bridge as a standalone module by extracting the nestedAppAuth logic from TeamsJS and initializing the bridge with the required parameters.

### Main changes in the PR:

1. Added `NestedAppAuthBridge.ts`, which includes the bridge initialization logic.
2. Added 16 unit tests to cover possible edge cases in `nestedAppAuthBridge.spec.ts`.
3. Exported `nestedAppAuthBridge` in `index.ts`
4. Added a size limit of 4 KB in `package.json`.

### Validation performed:

1. Manual validation was performed for this initial version. We plan to automate this in a future update.
2. Temporarily updated the tryPolyfillWithNestedAppAuthBridge() method in nestedAppAuthUtils.ts to initialize this new bridge with the required parameters and return immediately. 
3. Used the existing Teams test app. Navigated to the Nested App Auth section and clicked on "Send NAA Message to NestedAppAuth Bridge" after selecting the default payload.
4. Verified that a response was received as expected.

### Unit Tests added:

Added

### End-to-end tests added:

N/A


packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
```diff
@@ -0,0 +1,301 @@
+import { v4 as generateUUID } from 'uuid';
+
+/**
+ * Interface representing a request structure.
```
maglims, line 4: Will/do we have any public documentation(at teamsjs doc level) on how this script can be used? @baljesingh @vikramtha

maglims (approved):

packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
baljesingh: Not yet, but we’ll add it soon. @vikramtha , could you please take care of it?

package.json
```diff
@@ -164,6 +164,12 @@
       "import": "{ app, authentication, pages }",
       "limit": "57 KB"
     },
+    {
+      "brotli": false,
```
jekloudaMSFT, line 168: Does this run automatically now? Or do we need to add another step to the pipeline for it?

jekloudaMSFT (approved):

package.json
baljesingh: It runs automatically whenever teamsJS builds. It's part of the build command.

PR #2758, opened 2025-03-25
TJS unique instance id

KangxuanYe:

## Description

This PR is to address the issue that when multiple bundles with Teams-JS library send initialization to web host sdk, web host sdk has no way to track each of bundle's app initialization state unless each of TJS instance provides its id. This scenario could happen to authentication where both parent app and auth window are talking to the same web host sdk instance, nested app where both parent app and child app talk to the same web host instance, etc. 

### Main changes in the PR:

1. use UUID as Teams-JS library instance id
2. put Teams-JS library instance id inside of message request which is sent to host sdks

### Validation performed:

1. while running against on testing host using test app in Teams-JS library, you would be able to observe an additional field inside of message request in host sdk (you may need to parse the request when host sdk received it to see the field).

### Unit Tests added:

Yes, typically test a scenario when parent app receives message and tries to relay it to host sdk, parent app would not change the teams-js library id within child message request

### Change file added: Yes


change/@microsoft-teams-js-006a0221-7575-4e5e-9e92-ab7de1245d23.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Set a unique Teams-JS instance id when Teams-JS library is used and appended this unique id to message request sent to host sdk.",
```
juanscr, line 3: Not sure these clearly highlights for app sdk users the importance of this change. Maybe re-write to emphasize this allows us to identify messages from nested apps? You can disregard this comment if you feel is good enough 😄

juanscr (approved):

package.json
```diff
@@ -162,7 +162,7 @@
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "{ app, authentication, pages }",
-      "limit": "57 KB"
+      "limit": "190 KB"
```
noahdarveau-MSFT, line 165: Please only bump this limit by a small amount (0.5KB or less)

noahdarveau-MSFT (dismissed):

package.json
KangxuanYe: Sorry it exceeded 119 bytes and I read it as 119 kb. I revert the change and updated it to be 57.15kb. Please take another look.

change/@microsoft-teams-js-006a0221-7575-4e5e-9e92-ab7de1245d23.json
KangxuanYe: Teams-JS users shouldn't expect anything with this change because it shouldn't affect anything they are using.

pnpm-lock.yaml
juanscr, line 1: Why is the `pnpm-lock` updated?? The only change was due to the size limit of the bundle and updating this will probably update many of the dependency tree which could have any unintentional side effect.

juanscr (changes requested):

juanscr (dismissed):

pnpm-lock.yaml
KangxuanYe: Updated a `prettierIgnore` file to make sure those files would not be changed because of format in the future.

packages/teams-js/test/internal/childCommunication.spec.ts
```diff
@@ -217,6 +217,91 @@ describe('childCommunication', () => {
...
+        expect(sendMessageToParentHelper).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.anything(),
+          undefined,
+          true,
+          expectedTeamsJsInstanceId,
+        );
+      });
+
+      it("if parent app receives a message request from child app without teamsJs isntance id and 'handleIncomingmessageFromChild' function gets called, the teamsJs instance id be relayed by parent app would be the same as the one from child when passed it to 'sendMessageToParentHelper'", async () => {
```
jekloudaMSFT, line 263:
```suggestion
      it("if parent app receives a message request from child app without teamsJs isntance id and 'handleIncomingmessageFromChild' function gets called, the teamsJs instance id be relayed by parent app would be undefined", async () => {
```

jekloudaMSFT (approved):

juanscr (approved):

PR #2770, opened 2025-03-31
Added `apiVersion` tag in NAA request and removed `isNAAChannelRecommended` check in `isDeeplyNestedAuthSupported` api

baljesingh:

## Description

> This PR includes minor NAA updates:
1. Added the apiVersion tag to the NAA request.
2. Added a version constant to track the bridge version.
3. Removed the isNAAChannelRecommended check in the isDeeplyNestedAuthSupported API.

### Main changes in the PR:

1. Added the apiVersion tag to the NAA request in both the default bridge and the Nested App Auth standalone bridge.
2. Removed the isNAAChannelRecommended() check, as it's not needed.
3. Added a constant for the version in the Nested App Auth bridge to track changes at a granular level.
4. Updated unit tests.

### Validation performed:

1. Yes, through teams test app.

### Unit Tests added:

Updated

### End-to-end tests added:

M/A

### DIFF
--- a/packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
+++ b/packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
@@ -1,6 +1,8 @@
 import { v4 as generateUUID } from 'uuid';
 
 /**
+ * @beta
+ * @hidden
  * Local version of the Nested App Auth Bridge module.
  *
  * This version is specific to this standalone module and is not tied to the overall TeamsJS SDK version.
@@ -13,6 +15,9 @@ import { v4 as generateUUID } from 'uuid';
  *   if (nestedAppAuthBridge.version.startsWith('1.')) {
  *     // Safe to use with current logic
  *   }
+ *
+ * @internal
+ * Limited to Microsoft-internal use
  */
 export const version = '1.0.0';
 
@@ -99,10 +104,15 @@ let topOriginForNAA: string | null = null;
 let isNAALoggerEnabled = false;
 
 /**
+ * @beta
+ * @hidden
  * Initializes the Nested App Auth Bridge.
  * @param window The window object where the bridge will be attached.
  * @param topOrigin The origin of the top-level frame.
  * @param enableLogging - Optional flag to enable internal debug and error logging. Defaults to false.
+ *
+ * @internal
+ * Limited to Microsoft-internal use
  */
 export function initialize(window: Window | null, topOrigin: string, enableLogging = false): void {
   isNAALoggerEnabled = enableLogging;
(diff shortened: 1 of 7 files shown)


packages/teams-js/src/public/nestedAppAuth.ts
```diff
@@ -85,7 +85,7 @@ export function canParentManageNAATrustedOrigins(): boolean {
  * @beta
  */
 export function isDeeplyNestedAuthSupported(): boolean {
-  return (ensureInitialized(runtime) && isNAAChannelRecommended() && runtime.isDeeplyNestedAuthSupported) ?? false;
+  return (ensureInitialized(runtime) && runtime.isDeeplyNestedAuthSupported) ?? false;
```
baljesingh, line 88: The isNAAChannelRecommended() check is unnecessary because there won't be any case where isNAAChannelRecommended() is false but isDeeplyNestedAuthSupported() is true. It's better to keep the logic clean and simple, so the check has been removed.

packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
```diff
@@ -260,6 +264,7 @@ function createNestedAppAuthRequest(data: string): NestedAppAuthRequest {
     uuid: generateUUID(),
     func: 'nestedAppAuth.execute',
     timestamp: timestamp,
+    apiVersionTag: 'v2_nestedAppAuth.execute',
```
KangxuanYe, line 267: Could you please comment the reason for using hardcoded string here. We do not recommend having hardcoded string for `apiVersionTag` but in your case, you have special reason to do so.

packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
```diff
@@ -1,5 +1,8 @@
 import { v4 as generateUUID } from 'uuid';
```
KangxuanYe, line 1: Please verify request with tag in Orange test app to make sure it works on host. Make sure host can receive message with tag and process it correctly (generally speaking it shouldn't break anything)

baljesingh: Already tested.

packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
baljesingh: Thanks. I added comment. Let me know if you have any other concerns.

packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
```diff
@@ -1,5 +1,8 @@
...
 
+// Initial version of standalone nested app auth bridge
```
maglims, line 3: Let's update the comment on how this version can be used by app developers?

KangxuanYe (approved):

packages/teams-js/src/private/nestedAppAuth/nestedAppAuthBridge.ts
baljesingh: Updated. Please let me know if you have any further concern.

maglims (dismissed):

KangxuanYe (approved):

KangxuanYe (approved):

maglims (approved):

PR #2803, opened 2025-05-14
Supporting isDeeplyNestedAuthSupported flag for Legacy Teams Mobile Scenario

prpatwa:

## Description

The current isDeeplyNestedAuthSupported function that app developers will call, checks the value of the isDeeplyNestedAuthSupported flag in the runtime config. For hosts which are using the host sdks, the config object is sent during app initialization which sets the value of the is isDeeplyNestedAuthSupported flag in the runtime config correctly. For legacy Teams (i.e. Teams app not using the host SDK), the config object is not sent during the app initialization. The runtime config object is created on the JS SDK side. Therefore, a solution is required to return the value of the isDeeplyNestedAuthSupported correctly for the legacy Teams scenarios.

### Main changes in the PR:

Added the nestedAppAuth.deeplyNestedAuth capability against the new version in the mapTeamsVersionToSupportedCapabilities in runtime.ts which is used along with versionAndPlatformAgnosticTeamsRuntimeConfig object to create the runtime object for legacy teams scenarios.
Along with the check of isDeeplyNestedAuthSupported value in the runtime config, an additional condition is added to check the following:
the host is Android or iOS or iPadOs
the value of isLegacyTeams flag in the runtime
Is the nestedAppAuth?.deeplyNestedAuth capability present in the runtime.supports

### Validation performed:

1. Validated the isDeeplyNestedAuthSupported is returning correct value for legacy teams using Teams-test-app

### Unit Tests added:

Yes
Added unit tests to test the isDeeplyNestedAuthSupported against different conditions.

### End-to-end tests added:

<Yes/No>

### Change file added:

Yes

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | 
Screenshot_2025-05-14-23-41-15-96_c93d8b93098b4b30ab038c53d04444c3
 |

### DIFF
--- a/packages/teams-js/src/public/runtime.ts
+++ b/packages/teams-js/src/public/runtime.ts
@@ -277,7 +277,9 @@ interface IRuntimeV4 extends IBaseRuntime {
       readonly dataLayer?: {};
     };
     readonly monetization?: {};
-    readonly nestedAppAuth?: {};
+    readonly nestedAppAuth?: {
+      readonly deeplyNestedAuth?: {};
+    };
     readonly notifications?: {};
     readonly otherAppStateChange?: {};
     readonly pages?: {
@@ -608,6 +610,12 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
   '2.1.1': [
     {
       capability: { nestedAppAuth: {} },
+      hostClientTypes: [HostClientType.android, HostClientType.ios, HostClientType.ipados, HostClientType.visionOS],
+    },
+  ],
+  '2.1.2': [
+    {
+      capability: { nestedAppAuth: { deeplyNestedAuth: {} } },
       hostClientTypes: [HostClientType.android, HostClientType.ios, HostClientType.ipados, HostClientType.visionOS],
     },
   ],
--- a/packages/teams-js/test/public/nestedAppAuth.spec.ts
+++ b/packages/teams-js/test/public/nestedAppAuth.spec.ts
@@ -115,7 +115,7 @@ describe('nestedAppAuth', () => {
           await utils.initializeWithContext(FrameContexts.content, hostClient);
           const runtimeConfig: Runtime = {
             apiVersion: 4,
-            supports: { nestedAppAuth },
+            supports: { nestedAppAuth: {} },
             isNAAChannelRecommended: false,
             isLegacyTeams: true,
           };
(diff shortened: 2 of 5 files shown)


packages/teams-js/src/public/runtime.ts
```diff
@@ -611,6 +616,8 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
       hostClientTypes: [HostClientType.android, HostClientType.ios, HostClientType.ipados, HostClientType.visionOS],
     },
   ],
+  //teamsMobileVersionLegacyForDeeplyNestedAuth 2.1.2 is for deeplyNestedAppAuth but
+  //doesn't have any new capability added into it.
 };
```
baljesingh, line 621: We should use the exact object name as defined in runtime object to avoid ambiguity.  
```
'2.1.1': [
     // isDeeplyNestedAuthSupported: true (based on const legacyTeamsMobileVersionForDeeplyNestedAuth)
    {
      capability: { nestedAppAuth: {} },
      hostClientTypes: [HostClientType.android, HostClientType.ios, HostClientType.ipados, HostClientType.visionOS],
    },
  ],
```

packages/teams-js/test/public/nestedAppAuth.spec.ts
```diff
@@ -125,84 +129,142 @@ describe('nestedAppAuth', () => {
...
+      utils.setRuntimeConfig(runtimeConfig);
+      expect(nestedAppAuth.isDeeplyNestedAuthSupported()).toBeFalsy();
+    });
+
+    describe('should return true if isDeeplyNestedAuthSupported is false and isLegacyTeams is true in runtimeConfig for following clients that supports nestedAppAuth', () => {
+      const hostClients = [HostClientType.ipados, HostClientType.ios, HostClientType.android];
+      hostClients.forEach((hostClient) => {
+        it(`for ${hostClient} client`, async () => {
+          await utils.initializeWithContext(FrameContexts.content, hostClient);
+          utils.setClientSupportedSDKVersion(teamsMobileVersionLegacyForDeeplyNestedAuth);
```
baljesingh, line 217: Please don’t use teamsMobileVersionLegacyForDeeplyNestedAuth directly in tests.
If the version constant changes later, the test will silently follow the new value and may still pass, even though the behavior has changed.

Instead, hardcode the version. If you expect '2.1.2' to be supported, write both:

a positive test with '2.1.2'
a negative test with '2.1.1'

This way, if someone updates the constant in the future, the test will fail; making the change visible and reviewable.

prpatwa: I have updated the test for 2.1.1

packages/teams-js/src/public/runtime.ts
baljesingh: Sorry, the version number is incorrect. I meant to say, can we add a placeholder here and then add a comment that isDeeplyNestedAuthSupported is true? 

For example:
```
  '2.1.2': [
    // isDeeplyNestedAuthSupported: true (based on const legacyTeamsMobileVersionForDeeplyNestedAuth)
  ],

```

packages/teams-js/test/public/nestedAppAuth.spec.ts
baljesingh: Here is the scenario:

If in the future we realize a missing functionality in order to support deeply nested app and update teamsMobileVersionLegacyForDeeplyNestedAuth from 2.1.2 to 2.X (where X > 1), Your current test will still pass even though the behavior has changed because it uses the updated value directly.

That means the test won't catch the change, and no one will know something needs review.
But if we hardcode the value (like '2.1.1' and '2.1.0'), then the test will fail when the constant changes. That failure forces someone to review the test and confirm that the behavior is still correct.

By hardcoding both the values like '2.1.0' (negative case) and '2.1.1' (positive case), we're clearly stating what we expect:

2.1.0 should not support deeply nested auth
2.1.1 should support it

So if someone changes the version, they also have to update the unit test. That makes the change more visible and easier to track.

Using the constant hides changes. Hardcoding makes expectations clear and keeps the test meaningful.

packages/teams-js/src/public/runtime.ts
```diff
@@ -544,6 +544,11 @@ export const upgradeChain: IRuntimeUpgrade[] = [
   },
 ];
 
+/**
+ * Teams Mobile Sdk version that supports Deeply Nested App Auth
+ */
+export const teamsMobileVersionLegacyForDeeplyNestedAuth = '2.1.2';
```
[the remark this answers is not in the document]
prpatwa: updated

packages/teams-js/src/public/runtime.ts
```diff
@@ -611,6 +617,10 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
       hostClientTypes: [HostClientType.android, HostClientType.ios, HostClientType.ipados, HostClientType.visionOS],
     },
   ],
+  '2.1.2': [
```
jekloudaMSFT, line 620: Can you add a test here to make sure an empty record like this doesn't break `generateVersionBasedTeamsRuntimeConfig`?

prpatwa: What is the expectation here with the new test case?

prpatwa: I'm not modifying the logic of `generateVersionBasedTeamsRuntimeConfig` for which I could add a new test case. 

If the test cases would have been using a different `mapTeamsVersionToSupportedCapabilities ` with no empty record. I would have updated that `mapTeamsVersionToSupportedCapabilities ` in the test to make sure empty value is not breaking any of the existing functionality of `generateVersionBasedTeamsRuntimeConfig `.

But since the test is using the same value of `mapTeamsVersionToSupportedCapabilities` from `runtime.ts`.
All the tests of `generateVersionBasedTeamsRuntimeConfig ` are referring to the map with empty record. 
And since all the test cases are passing, it implies that `generateVersionBasedTeamsRuntimeConfig ` is working as expected even with empty record.

jekloudaMSFT (approved):

PR #2794, opened 2025-05-01
Added `renderingSurface` property to `{app.Page.Context}`

lakhveerk:
For more information about how to contribute to this repo, visit [this page](https://github.com/OfficeDev/microsoft-teams-library-js/blob/main/CONTRIBUTING.md).

## Description

Added `renderingSurface` property to `{app.Page.Context}` capability. The `Renderingsurfaces' enum for this property only contains 'copilotSidePanel' for now, but in future more values will be added.

This PR increased the teams-js size bundle size from 57.204Kb to 57.33kb. The size increase is justified as we are adding new code to app.ts file. This change does not break the tree shakability.

> If this Pull Request should close/resolve any issues when merged, use [the special syntax for that](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) here.

PR #2809, opened 2025-05-20
noahdarveau/updated size limit to unblock pipeline

noahdarveau-MSFT:
Upping the size limit to unblock the pipeline.

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,7 +162,7 @@
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "{ app, authentication, pages }",
-      "limit": "57.33 KB"
+      "limit": "57.40 KB"
     },
     {
       "brotli": false,


jekloudaMSFT (approved):

baljesingh (approved):

package.json
```diff
@@ -162,7 +162,7 @@
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "{ app, authentication, pages }",
-      "limit": "57.33 KB"
+      "limit": "57.40 KB"
```
maglims, line 165: Just curious, do we know why the size has increased?

maglims (approved):

package.json
noahdarveau-MSFT: The size increase is from [this pr](https://github.com/OfficeDev/microsoft-teams-library-js/pull/2794) that merged in earlier today. It made some changes to the app file, which is one of the baseline files for the size check, so we needed a (completely valid) bump to the size check. For some reason, when the library builds in the PR, vs when it builds in the prod pipeline, there is like a 3 byte difference in the build sizes. Probably due to some string pipeline variable or something. The size limit is supposed to be a little bit higher than the current build size, but it only got increased to the PR build size. So now when it builds in the pipeline, it is 3 bytes over the size limit causing it to fail.

PR #2837, opened 2025-06-18
Try fallback valid domains list before CDN list

jekloudaMSFT:

## Description

This PR reduces latency for the app initialization scenario by relying on the local valid domain list before checking the list on the CDN. It also resolves an issue where teams-js was making multiple CDN fetch requests if it tried to access the valid domains list before prefetching completes.

### Main changes in the PR:

1. `validateOrigin` checks the fallback domain list before using the CDN list if there is no cached list available.
2. `getValidOriginsListFromCDN` will no longer create a new fetch request if one is already in progress.

### Validation performed:

1. Unit tests added

### Unit Tests added:

Yes

### End-to-end tests added:

N/A

### DIFF
--- a/packages/teams-js/src/internal/validOrigins.ts
+++ b/packages/teams-js/src/internal/validOrigins.ts
@@ -22,6 +22,7 @@ async function getValidOriginsListFromCDN(shouldDisableCache: boolean = false): 
     return validOriginsCache;
   }
   if (validOriginsPromise) {
+    // Fetch has already been initiated, return the existing promise
     return validOriginsPromise;
   }
   if (!inServerSideRenderingEnvironment()) {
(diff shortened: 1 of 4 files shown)


packages/teams-js/src/internal/validOrigins.ts
```diff
@@ -5,9 +5,12 @@ import { inServerSideRenderingEnvironment, isValidHttpsURL } from './utils';
 
 let validOriginsCache: string[] = [];
 const validateOriginLogger = getLogger('validateOrigin');
+let validOriginsPromise: Promise<string[]> | undefined;
```
maglims, line 8: Pasted Graphic 1

Pasted Graphic 2

maglims: I asked the copilot to review the added code and here is what it suggests. Feel free to see if it makes sense.

jekloudaMSFT: `validOriginsPromise` is not rejected - on error it returns the fallback list. That being said, it may make sense to revise the logic here so that we retry instead, since we check the fallback list _before_ trying the CDN, so there's no reason to return the fallback list after failing to fetch from CDN

jekloudaMSFT: And resetValidOriginsCache is only used for testing, I'll try to make this more clear!

jadahiya-MSFT (approved):

maglims (approved):

PR #2856, opened 2025-07-16
Added copilot view API for closing the side panel

juanscr:
## Description
This pull requests adds a new API and two new rendering surface values, to address some copilot accessibility issues.

### Main changes in the PR:
1. Adds a new API for requesting the host for closing the copilot side panel. This is only available for the copilot app and is intended to be used for adding hotkeys to this behavior inside of copilot.
2. Added two new rendering surface values to cover two additional surfaces used in Teams for rendering copilot.

## Size Limit increased
This was increased as the runtime new value and the two new enum values increased the bundle size by 6 bytes. As these changes are reasonable between the scope of this change, the size limit was increased by 7 bytes to support this.

### Validation performed:

1. Ran in a test host and validated the service is called appropriately.
2. Ran unit testing and end-to-end tests.

### Unit Tests added:

Yes

### End-to-end tests added:

Yes.

### Change file added:
Yes

### DIFF
--- a/packages/teams-js/src/private/copilot/view.ts
+++ b/packages/teams-js/src/private/copilot/view.ts
@@ -22,15 +22,14 @@ export function isSupported(): boolean {
 /**
  * Closes the side panel that is hosting the copilot app.
  *
- * @returns { Promise<Record<sting, never>> } - empty object to indicate successful closure of the side panel
- * @throws { SdkError } - Throws a SdkError if host SDK returns an error as a response to this call
+ * @throws { Error } - Throws a Error if host SDK returns an error as a response to this call
  *
  * @hidden
  * @beta
  * @internal
  * Limited to Microsoft-internal use
  */
-export async function closeSidePanel(): Promise<Record<string, never>> {
+export async function closeSidePanel(): Promise<void> {
   ensureInitialized(runtime);
   await callFunctionInHostAndHandleResponse(
     ApiName.Copilot_View_CloseSidePanel,
@@ -38,7 +37,6 @@ export async function closeSidePanel(): Promise<Record<string, never>> {
     new CloseSidePanelResponseHandler(),
     getApiVersionTag(copilotTelemetryVersionNumber, ApiName.Copilot_View_CloseSidePanel),
   );
-  return {};
 }
 
 class CloseSidePanelResponseHandler extends ResponseHandler<Record<string, never>, Record<string, never>> {
--- a/packages/teams-js/src/private/copilot/view.ts
+++ b/packages/teams-js/src/private/copilot/view.ts
@@ -1,0 +1,52 @@
+import { callFunctionInHostAndHandleResponse } from '../../internal/communication';
+import { ensureInitialized } from '../../internal/internalAPIs';
+import { ResponseHandler } from '../../internal/responseHandler';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from '../../internal/telemetry';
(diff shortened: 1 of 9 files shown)


packages/teams-js/src/private/copilot/view.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @returns { Promise<Record<sting, never>> } - empty object to indicate successful closure of the side panel
+ * @throws { SdkError } - Throws a SdkError if host SDK returns an error as a response to this call
+ *
+ * @hidden
+ * @beta
+ * @internal
+ * Limited to Microsoft-internal use
+ */
+export async function closeSidePanel(): Promise<Record<string, never>> {
```
jekloudaMSFT: Why does this need a return type? Can we return `Promise<void>` and have the Promise resolve to indicate a successful call?

packages/teams-js/src/private/copilot/view.ts
```diff
@@ -0,0 +1,52 @@
...
+  ensureInitialized(runtime);
```
jekloudaMSFT, line 33: Should this only run in the SidePanel rendering surface? If it is, is the app expected to check the rendering surface before calling it?

packages/teams-js/src/private/copilot/view.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export function isSupported(): boolean {
+  return ensureInitialized(runtime) && !!runtime.supports.copilot?.view;
+}
+
+/**
+ * Closes the side panel that is hosting the copilot app.
+ *
+ * @returns { Promise<Record<sting, never>> } - empty object to indicate successful closure of the side panel
+ * @throws { SdkError } - Throws a SdkError if host SDK returns an error as a response to this call
```
jekloudaMSFT: Not necessarily in scope for this PR, but I see that `callFunctionInHostAndHandleResponse` can throw `Errors` (not just `SdkError`, is that a bug in the communication code? And for this PR, should we document that this can throw regular `Error` too?

juanscr: Actually, now that you mentioned it, the internal function _only_ returns `Error` object not `SdkError`. Hmmm I probably misunderstood that, I thought it throwed any reportable error lol. Anyway, yeah, we should document it properly, I'll update it.

packages/teams-js/src/private/copilot/view.ts
juanscr: Yeah we can definitely do that. I was returning an object to scale if we wanted to add more info down the line but I agree that it can be fully internally and just don't return anything for the app as it may be more confusing.

packages/teams-js/src/private/copilot/view.ts
juanscr: I think is more for the app to decide. We can check here but not sure, I didn't do that based on how the `sidePanel` copilot APIs work here which they don't force the rendering surface to be any value although those APIs are only for the side panel rendering surface.

jekloudaMSFT (approved):

packages/teams-js/src/private/copilot/view.ts
jekloudaMSFT: That makes sense, it's really up to the host to say when this is supported via the "supports" object. It just seems strange to have a "closeSidePanel" API that isn't really dependent on the app being in a SidePanel, but I agree that's correct

PR #2857, opened 2025-07-17
Add Outlook DoD domains to validDomains

salaman:

## Description

Add Outlook GccH + DoD domains to validDomains.json. Not a code change.


package.json
```diff
@@ -162,7 +162,7 @@
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "{ app, authentication, pages }",
-      "limit": "57.69 KB"
+      "limit": "58 KB"
```
maglims: This PR already updated this value. https://github.com/OfficeDev/microsoft-teams-library-js/pull/2869/files#diff-3b50f79a0d8093f5bd66b9841f9f65ea4a98b60b431e3382dedc6b54854998c4

This either needs to be reverted, or it will overwrite the change. @marianamu-MSFT @sunbrk

salaman: merged main @sunbrk @maglims

sunbrk: @salaman  Can you please update this change?

package.json
```diff
@@ -162,7 +162,7 @@
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "{ app, authentication, pages }",
-      "limit": "58.49 KB"
+      "limit": "59 KB"
```
maglims, line 165: Hmm, in main, it is "58.59" Kb. Is there any reason why we need to increase to 59Kb?

salaman: I think it's a bit silly to constantly be updating this to the exact 0.01kb for every change. It should be %-increase based, or at least prevent large changes (like importing a new library or an accidental import). In a prior iteration I also set it to the exact 0.01kb and it still failed because the build used the exact byte count: https://github.com/OfficeDev/microsoft-teams-library-js/actions/runs/16667552609/job/47176738074

I just set it to the nearest ceiling round KB.

maglims: Ok, I don't have enough insights here. @sunbrk @jekloudaMSFT for awareness for the potential improvement here.

maglims (approved):

PR #2858, opened 2025-07-18
Fixed the typo from `NonAdult` to `NotAdult` in `LegalAgeGroupClassification` Interface

ndangudubiyyam:
Fixed the typo from `NonAdult` to `NotAdult` in `LegalAgeGroupClassification` Interface

### DIFF
--- a/packages/teams-js/src/internal/appHelpers.ts
+++ b/packages/teams-js/src/internal/appHelpers.ts
@@ -155,14 +155,14 @@ function initializeHelper(apiVersionTag: string, validMessageOrigins?: string[])
           // After Teams updates its client code, we can remove this default code.
           try {
             initializeHelperLogger('Parsing %s', runtimeConfig);
-            const givenRuntimeConfig: IBaseRuntime | null = JSON.parse(runtimeConfig);
+            let givenRuntimeConfig: IBaseRuntime | null = JSON.parse(runtimeConfig);
             initializeHelperLogger('Checking if %o is a valid runtime object', givenRuntimeConfig ?? 'null');
             // Check that givenRuntimeConfig is a valid instance of IBaseRuntime
             if (!givenRuntimeConfig || !givenRuntimeConfig.apiVersion) {
               throw new Error('Received runtime config is invalid');
             }
             // Normalize ageGroup value for backward compatibility
-            normalizeAgeGroupValue(givenRuntimeConfig);
+            givenRuntimeConfig = normalizeAgeGroupValue(givenRuntimeConfig);
             runtimeConfig && applyRuntimeConfig(givenRuntimeConfig);
           } catch (e) {
             if (e instanceof SyntaxError) {
@@ -175,7 +175,7 @@ function initializeHelper(apiVersionTag: string, validMessageOrigins?: string[])
                 if (!isNaN(compareSDKVersions(runtimeConfig, defaultSDKVersionForCompatCheck))) {
                   GlobalVars.clientSupportedSDKVersion = runtimeConfig;
                 }
-                const givenRuntimeConfig: IBaseRuntime | null = JSON.parse(clientSupportedSDKVersion);
+                let givenRuntimeConfig: IBaseRuntime | null = JSON.parse(clientSupportedSDKVersion);
                 initializeHelperLogger('givenRuntimeConfig parsed to %o', givenRuntimeConfig ?? 'null');
 
                 if (!givenRuntimeConfig) {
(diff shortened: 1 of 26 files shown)


sunbrk (approved):

sunbrk (approved):

KangxuanYe (approved):

MengyiGong (approved):

jekloudaMSFT (approved):

packages/teams-js/src/public/interfaces.ts
```diff
@@ -1295,7 +1295,7 @@ export enum LegalAgeGroupClassification {
    * United Kingdom, European Union, or South Korea, and the user's age is between a minor and an adult age
    * (as stipulated based on country or region). Generally, this means that teenagers are considered as notAdult in regulated countries.
    */
-  NonAdult = 'nonAdult',
+  NotAdult = 'notAdult',
```
jekloudaMSFT, line 1307: Can you modify the eligibility API to convert 'nonAdult' to 'notAdult' for back compat?

ndangudubiyyam: oh good point. Yes I will do that.

ndangudubiyyam: Do you think it is a good idea to write a unit test for this.?

jekloudaMSFT: Yes, please do! You'll need to do the conversion in the `getEligibilityInfo` function itself too, since this is just passed as a string from the host

ndangudubiyyam: done.

packages/teams-js/src/private/copilot/eligibility.ts
```diff
@@ -73,6 +73,10 @@ export async function getEligibilityInfo(forceRefresh?: boolean): Promise<AppEli
   if (!isEligibilityInfoValid(response)) {
     throw new Error('Error deserializing eligibility information');
   }
+  // convert nonAdult age group to NotAdult
```
lakhveerk, line 76: Don't we need to run the same logic on run time object as well?

packages/teams-js/src/private/copilot/eligibility.ts
```diff
@@ -73,6 +73,10 @@ export async function getEligibilityInfo(forceRefresh?: boolean): Promise<AppEli
...
+  if ((response.ageGroup as unknown) === 'NonAdult') {
+    response.ageGroup = 'NotAdult' as typeof response.ageGroup;
```
jekloudaMSFT: Could we just use the enum directly here? What is the cast doing here?

packages/teams-js/src/private/copilot/eligibility.ts
ndangudubiyyam: oh thanks for pointing that out.I have added the check.

packages/teams-js/src/internal/utils.ts
```diff
@@ -566,3 +567,34 @@ export function isPrimitiveOrPlainObject(value: unknown, depth: number = 0): boo
...
+export function normalizeAgeGroupValue(runtimeConfig: IBaseRuntime): IBaseRuntime {
+  // If no ageGroup exists, return the original config
+  if (!runtimeConfig.hostVersionsInfo?.appEligibilityInformation?.ageGroup) {
+    return runtimeConfig;
+  }
+
+  const ageGroup = runtimeConfig.hostVersionsInfo.appEligibilityInformation.ageGroup as unknown as string;
+
+  // If the ageGroup doesn't need normalization, return the original config
+  if (ageGroup?.toLowerCase() !== 'nonadult') {
```
lakhveerk, line 585: Is this `nonadult` or `nonAdult`?

packages/teams-js/src/internal/appHelpers.ts
```diff
@@ -155,6 +161,8 @@ function initializeHelper(apiVersionTag: string, validMessageOrigins?: string[])
             if (!givenRuntimeConfig || !givenRuntimeConfig.apiVersion) {
               throw new Error('Received runtime config is invalid');
             }
+            // Normalize ageGroup value for backward compatibility
+            normalizeAgeGroupValue(givenRuntimeConfig);
```
jekloudaMSFT: It looks like `normalizeAgeGroupValue` creates a new runtime object rather than modifying the input object, do we need to re-set the runtime here?

packages/teams-js/src/internal/utils.ts
lakhveerk: oh, I did not notice the toLowerCase here.

lakhveerk (approved):
lgtm

packages/teams-js/src/internal/appHelpers.ts
ndangudubiyyam: omg. I thought I pushed that change, thanks for catching it. Fixed it.

lakhveerk (approved):

jekloudaMSFT (approved):

Issue #2867, opened 2025-08-01
RFC: Support Host (Teams) Shortcuts in Apps


YuanboXue-Amber:
# RFC: Support Host Shortcuts in Apps

## Summary

This proposal introduces APIs to support host keyboard shortcuts when pressed within app iframes.

## Problem statement

Apps integrated with host run inside their own iframes. Currently, host shortcuts do not function within these iframes.

> Example: `"GoToPowerBar"` shortcut in Teams host:
>
> - On Mac browsers: `"option+cmd+e"`
> - On Windows browsers: `"ctrl+alt+e"`
>
> This shortcut should focus the search input box at the top of Teams. If the search box is not visible (e.g., in a narrow window), it should be made visible.
>
> Currently, pressing this shortcut inside an app iframe has no effect.

Additionally, apps may need to override certain host shortcuts. For example, `"ctrl+b"` bolds text in the Teams compose box, but if an app has its own compose box, it may want to handle `"ctrl+b"` itself.

## Detailed Design or Proposal

The design consists of two parts:

1. Supporting host shortcuts in apps
2. Allowing apps to override certain host shortcuts

### Part 1 - Support Host Shortcuts

- Introduce a new API: `shortcutRelay.enableShortcutRelayCapability`
- When an app calls this API:

  1. `teams-js` posts a message to request all mapped host shortcuts:

     ```ts
     callFunctionInHostAndHandleResponse(
       ApiName.ShortcutRelay_RequestHostShortcuts, // New function in Hub/Hub SDK
       [],
       new HostShortcutsResponseHandler(),
       getApiVersionTag(ApiVersionNumber.V_2, ApiName.App_RequestTeamsShortcut),
     );
     ```

     The response from the host will be an object like:

     ```jsonc
     {
       "shortcuts": ["ctrl+/", "ctrl+shift+/"]
     }
     ```

     `teams-js` stores this response in memory.

  2. `teams-js` attaches a keydown event handler to the app’s `document`. Using the shortcut map, it checks if the event matches a host shortcut. If so, it forwards the event to host:

     ```ts
     if (matchingShortcut) {
       const payload = serializeKeydownEvent(event);
       callFunctionInHost(
         ApiName.ShortcutRelay_ForwardShortcutEvent, // New function in Hub/Hub SDK
         [payload],
         getApiVersionTag(ApiVersionNumber.V_2, ApiName.ShortcutRelay_ForwardShortcutEvent),
       );
     }
[...]

PR #2868, opened 2025-08-01
Add shortcutRelay capability to support host shortcut in Apps

YuanboXue-Amber:
## Description

Add `shortcutRelay` capability to support host shortcut in Apps.

Please see detailed design in #2867

### DIFF
--- a/packages/teams-js/src/public/shortcutRelay.ts
+++ b/packages/teams-js/src/public/shortcutRelay.ts
@@ -1,0 +1,262 @@
+/**
+ * Allows host shortcuts to function in your application by forwarding keyboard shortcuts to the host.
+ *
+ * This functionality is in Beta.
+ * @beta
+ * @module
+ */
+
+import { callFunctionInHost, callFunctionInHostAndHandleResponse } from '../internal/communication';
+import { ensureInitialized } from '../internal/internalAPIs';
+import { ResponseHandler } from '../internal/responseHandler';
+import { ApiName, ApiVersionNumber, getApiVersionTag } from '../internal/telemetry';
+import { errorNotSupportedOnPlatform } from './constants';
+import { runtime } from './runtime';
+import { ISerializable } from './serializable.interface';
+
+/* ------------------------------------------------------------------ */
+/* Types                                                              */
+/* ------------------------------------------------------------------ */
+
+type Shortcuts = Set<string>; // stores shortcut key combinations, for example: ["ctrl+/", "ctrl+shift+/"]
+
+type HostShortcutsResponse = {
+  shortcuts: Shortcuts;
+  overridableShortcuts: Shortcuts;
+};
+
+export type OverridableShortcutHandlerData = {
+  matchedShortcut: string;
+};
+
+/**
+ * A handler must return `true` when it wants to **consume** the shortcut
+ * (i.e. prevent forwarding to host). Any other return value forwards it.
+ */
+export type OverridableShortcutHandler = (event: KeyboardEvent, data: OverridableShortcutHandlerData) => boolean;
+
+/* ------------------------------------------------------------------ */
+/* Utils                                                              */
+/* ------------------------------------------------------------------ */
+
+class SerializableKeyboardEvent implements ISerializable {
+  public constructor(private event: KeyboardEvent) {}
+  public serialize(): object | string {
+    return {
+      altKey: this.event.altKey,
+      bubbles: this.event.bubbles,
+      cancelBubble: this.event.cancelBubble,
+      charCode: this.event.charCode,
+      code: this.event.code,
+      composed: this.event.composed,
+      ctrlKey: this.event.ctrlKey,
+      // currentTarget: skipped
+      defaultPrevented: this.event.defaultPrevented,
+      detail: this.event.detail,
+      eventPhase: this.event.eventPhase,
+      isComposing: this.event.isComposing,
+      isTrusted: this.event.isTrusted,
+      key: this.event.key,
+      keyCode: this.event.keyCode,
+      location: this.event.location,
+      metaKey: this.event.metaKey,
+      // path - skipped,
+      repeat: this.event.repeat,
+      returnValue: this.event.returnValue,
+      shiftKey: this.event.shiftKey,
+      // sourceCapabilities - skipped,
+      // srcElement - slipped.
+      // target - skipped.
+      timeStamp: this.event.timeStamp,
+      type: this.event.type,
+      // view - skipped
+      which: this.event.which,
+    };
+  }
+}
+
+/**
+ * Normalizes a shortcut string to a canonical form.
+ */
+function normalizeShortcut(shortcut: string): string {
+  return shortcut.toLowerCase().split('+').sort().join('+');
+}
+
+/**
+ * Build a canonical, lower-case “ctrl+shift+x” representation of the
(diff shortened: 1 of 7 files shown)


packages/teams-js/src/public/shortcutRelay.ts
```diff
@@ -0,0 +1,287 @@
...
+  private onSuccess(response: HostShortcutsResponse): void {
+    hostShortcuts.clear();
+    response.shortcuts.forEach((shortcut: string) => {
+      hostShortcuts.add(normalizeShortcut(shortcut));
+    });
+    overridableShortcuts.clear();
+    response.overridableShortcuts.forEach((shortcut: string) => {
+      overridableShortcuts.add(normalizeShortcut(shortcut));
+    });
+  }
```
juanscr: I think is best to move what the `onSuccess` function is doing to happen inside the `enableShortcutRelayCapability`. I think is easier to debug and read when this side effect is made explicit in the function, instead of relying on the response handler as it is implicit to the code flow.

As such, the code would be exactly the same, just a bit more localized:

```typescript
// inside enableShortcutRelayCapability
const response = await callFunctionInHostAndHandleResponse(
  ApiName.ShortcutRelay_GetHostShortcuts,
  [],
  new HostShortcutsResponseHandler(),
  getApiVersionTag(ApiVersionNumber.V_2, ApiName.ShortcutRelay_GetHostShortcuts),
);

// Explicitly set the host shortcuts in global object
hostShortcuts.clear();
response.shortcuts.forEach((shortcut: string) => {
  hostShortcuts.add(normalizeShortcut(shortcut));
});
overridableShortcuts.clear();
response.overridableShortcuts.forEach((shortcut: string) => {
  overridableShortcuts.add(normalizeShortcut(shortcut));
});
```
You can move that to a function as well if you want. This also makes this response handler a bit more reusable, as you may need the same validation for the update shortcuts event and you probably want a different behavior than this.

Moreover, this forces you to handle errors when calling the host as we are not currently awaiting the response from the host.

packages/teams-js/src/public/shortcutRelay.ts
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * Enable capability to support host shortcuts.
+ *
+ * @beta
+ */
+export function enableShortcutRelayCapability(): void {
+  if (isShortcutRelayCapabilityEnabled) {
+    return;
+  }
+  isShortcutRelayCapabilityEnabled = true;
```
juanscr: There is a minor bug in this code, as the `isShortcutRelayCapabilityEnabled` function is set to `true` when the function gets called, not when the event listener gets attached to the document. This may as well in some cases prevent an app for getting this relay, due to calling it during initialization or some other exception that may raise when they call the function (preventing re-try calls when failure).

I would move this check to the point where the listener gets attached. It should be something like:

```typescript
const response = await callFunctionInHostAndHandleResponse(
  ApiName.ShortcutRelay_GetHostShortcuts,
  [],
  new HostShortcutsResponseHandler(),
  getApiVersionTag(ApiVersionNumber.V_2, ApiName.ShortcutRelay_GetHostShortcuts),
);

// ... the code to handle the response per the comment above ...

// Avoiding race conditions when multiple instances were called beforehand
if (!shortCutRelayCapabilityEnabled) {
  document.addEventListener('keydown', keydownHandler, { capture: true });
}
isShortcutRelayCapabilityEnabled = true;
```

packages/teams-js/src/public/shortcutRelay.ts
```diff
@@ -0,0 +1,287 @@
...
+
+  ensureInitialized(runtime);
+
+  if (!isSupported()) {
+    throw errorNotSupportedOnPlatform;
+  }
```
juanscr: I think you only need the `isSupported` check, as the `ensureInitialized` is called during this function.

packages/teams-js/src/public/shortcutRelay.ts
```diff
@@ -0,0 +1,287 @@
...
+      overridableShortcuts.add(normalizeShortcut(shortcut));
+    });
+  }
+}
+
+function keydownHandler(event: KeyboardEvent): void {
+  // Skip if the event target is within an element that has the `data-disable-shortcuts-forwarding` attribute
+  if ((event.target as HTMLElement).closest(`[${DISABLE_SHORTCUT_FORWARDING_ATTRIBUTE}]`)) {
+    return;
+  }
```
juanscr, line 174: I am not fully convincing this actually fixes anything, as this allows apps to add this tag everywhere allowing to override non-overridable key binds. I don't _feel_ this threatens security, as this event handler will only relay to the host matched shortcuts not any combination of keystrokes, therefore only exposing pre-defined keys.

I just don't see how this implementation could expose sensitive data. Any thoughts on this @lakhveerk?

juanscr (changes requested):
Good job!

I think the only thing that would be missing is adding the capability to the test app, you can just use the existing code for other services. And adding the event for updating the shortcuts but not sure if you want to keep this for a PR after this one.

packages/teams-js/src/public/shortcutRelay.ts
YuanboXue-Amber: I updated it. This makes `enableShortcutRelayCapability` async but I think it's fine

packages/teams-js/src/public/shortcutRelay.ts
YuanboXue-Amber: Agree, I updated it.

packages/teams-js/src/public/shortcutRelay.ts
YuanboXue-Amber: I think it's good to have a way for Apps to opt out of supporting host shortcuts within subtrees, not only for potential security concerns.

packages/teams-js/src/public/shortcutRelay.ts
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * register a handler to be called when shortcuts are updated in the host.
+ */
+function registerOnHostShortcutChangedHandler(handler: (hostShortcuts: HostShortcutsResponse) => void): void {
+  registerHandler(
+    getApiVersionTag(ApiVersionNumber.V_2, ApiName.ShortcutRelay_HostShortcutChanged),
+    ApiName.ShortcutRelay_HostShortcutChanged,
+    handler,
+  );
+}
```
YuanboXue-Amber, line 168: The reason for sending the same `HostShortcutsResponse` object when the host updates shortcut:

In Teams, users can do more than just reassign a single shortcut. 

They also have the option to reset all shortcuts to the default settings or reassign all shortcuts to match the Slack or Zoom shortcut model.

juanscr:
> > Good job!
> > I think the only thing that would be missing is adding the capability to the test app, you can just use the existing code for other services. And adding the event for updating the shortcuts but not sure if you want to keep this for a PR after this one.
> 
> @juanscr is it better to add test together or separately? If there's no difference, I can do it in a follow up PR.
> 
> I remember we talked about running integration tests with test apps after teams-js changes, and at the same time as making changes in the hub, but I might be mistaken.
> 
> I've confirmed locally that these changes work as expected by testing them in the copilot app, after replacing node_modules/teams-js with my updates.

Yeah this is fine, eventually we have to add E2E tests but that would usually come in an effort afterwards so we can do it in a later pR.

PR #2893, opened 2025-09-04
noahdarveau/ OpenFilePreview in sidePanel Frame Context

noahdarveau-MSFT:
Added `FrameContexts.sidePanel` to `openFilePreview`'s ensureInitialized call to allow for the use in SidePanel.


maglims:
@noahdarveau-MSFT , thanks for the quick fix.

Adding @sbtron for awareness. We have similar APIs which in the same boat on if they should be extended into sidePanel so let's check once.

juanscr (approved):

PR #2894, opened 2025-09-04
Set precache flow for test app

KangxuanYe:

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/apps/teams-test-app/src/App.tsx
+++ b/apps/teams-test-app/src/App.tsx
@@ -1,6 +1,6 @@
 import './App.css';
 
-import { app, appInitialization, initialize } from '@microsoft/teams-js';
+import { app, appInitialization, initialize, ResumeContext } from '@microsoft/teams-js';
 import React, { ReactElement } from 'react';
 import { BrowserRouter, Route, Routes } from 'react-router-dom';
 
@@ -17,7 +17,7 @@ const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? u
 const validMessageOrigins: string[] | undefined = getOriginsParam ? getOriginsParam.split(',') : undefined;
 
 // This is added for custom initialization when app can be initialized based upon a trigger/click.
-if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
+if (!urlParams.has('customInit') || !urlParams.get('customInit') || !urlParams.get('precacheApp')) {
   if (isTestBackCompat()) {
     initialize(undefined, validMessageOrigins);
   } else {
@@ -41,6 +41,30 @@ if (
     app.notifyAppLoaded();
     app.notifySuccess();
   }
+}
+
+if (urlParams.has('precacheApp') && urlParams.get('precacheApp')) {
+  app.initialize(validMessageOrigins).then(() => {
+    app.notifyAppLoaded();
+    app.lifecycle.registerBeforeSuspendOrTerminateHandler(() => {
+      return new Promise<void>((resolve) => {
+        resolve();
+      });
+    });
+    app.lifecycle.registerOnResumeHandler((context: ResumeContext): void => {
+      // get the route from the context
+      console.log(context.contentUrl);
+      // navigate to the correct path based on URL
+      // navigate(route.pathname);
+      if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
+        app.notifySuccess();
+      }
+    });
+    app.notifySuccess();
+  });
+
+  window.addEventListener('message', handleMessageFromMockedHost);
+  console.info('Test app listener is set to receive messages for precached app to load');
 }
 
 function handleMessageFromMockedHost(msg: MessageEvent): void {


apps/teams-test-app/src/App.tsx
```diff
@@ -17,7 +17,7 @@ const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? u
 const validMessageOrigins: string[] | undefined = getOriginsParam ? getOriginsParam.split(',') : undefined;
 
 // This is added for custom initialization when app can be initialized based upon a trigger/click.
-if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
+if (!urlParams.has('customInit') || !urlParams.get('customInit') || !urlParams.get('precacheApp')) {
```
jekloudaMSFT: do we need to check `urlParams.has` before calling `urlParams.get`?

KangxuanYe: Wait why it is still here. I have removed this.

KangxuanYe: Nvmd yes we need. I thought it was another line. Thank you for pointing it out.

KangxuanYe: Updated!

juanscr: I re-iterate this, why do we have `!urlParams.has('customInit') || !urlParams.get('customInit')`? If "customInit" is not present in the map get will just return `undefined` therefore entering the condition. Or am I missing something?

apps/teams-test-app/src/App.tsx
```diff
@@ -43,6 +51,25 @@ if (
   }
 }
 
+if (urlParams.has('precacheApp') && urlParams.get('precacheApp')) {
```
juanscr, line 54: Same here, can't we just use `urlParams.get('precacheApp') !== undefined` or something like that

juanscr: And if its needed, can't we add all this checks into a unique function? Something like `isParamPresent` or something like that? Its seem to be like a lot of duplicate code.

KangxuanYe: No, we cannot because `urlParams` need to make sure it has param and then get it.

As for duplication, sometime we do !has && !get, sometimes !has || !get. You would fall into another scenario that we created multiple functions.

My thought for improvement is to sort this whole file to provide three different kinds of test app we are using. But that would be later work.

apps/teams-test-app/src/App.tsx
KangxuanYe: It avoids https://localhost:4000?customInit, which force you to write as `https://localhost:4000?customInit=true`. SURPRISE?!

juanscr: Got it... Alright, I definitely won't block this PR for this, this seems like a nitpick to me.

Just for future work or if you want to make it cleaner, in cases like this one http://localhost:4000?customInit the search parameters for the key `customInit` is an empty string not `undefined`. Therefore, the usual check to avoid doing the map operation twice is doing `urlParams.get('customInit') !== undefined` as if it's in the format without a mapped value it will just return an empty string. It only returns `undefined` when the parameter does not exist in the URL.

Buut, like I said, it is a minor nitpick and its also valid to the `has`/`get` check if you want to support the empty string case.

juanscr (approved):

apps/teams-test-app/src/App.tsx
juanscr: Yeah, see comment above for explanation. It's definitely something we can write shorter but not need now, just for future reference. And yeah, I agree, making a function doesn't add much.

apps/teams-test-app/src/App.tsx
```diff
@@ -17,7 +17,10 @@ const getOriginsParam = urlParams.has('origins') && urlParams.get('origins') ? u
 const validMessageOrigins: string[] | undefined = getOriginsParam ? getOriginsParam.split(',') : undefined;
 
 // This is added for custom initialization when app can be initialized based upon a trigger/click.
-if (!urlParams.has('customInit') || !urlParams.get('customInit')) {
+if (
```
baljesingh, line 20: I think readability can be improved. How about using variables to avoid repeated checks?

```
const hasCustomInit = urlParams.has('customInit') && urlParams.get('customInit');
const hasPrecacheApp = urlParams.has('precacheApp') && urlParams.get('precacheApp');
const hasAppInitTest =
  urlParams.has(appInitializationTestQueryParameter) &&
  urlParams.get(appInitializationTestQueryParameter);

const noCustomInit = !hasCustomInit;
const noPrecacheApp = !hasPrecacheApp;

if (noCustomInit && noPrecacheApp) {
  if (isTestBackCompat()) {
    initialize(undefined, validMessageOrigins);
  } else {
    app.initialize(validMessageOrigins);
  }
}

// for AppInitialization tests we need a way to stop the Test App from sending these
// we do it by adding appInitializationTest=true to query string
if (
  hasCustomInit ||
  (hasAppInitTest && noPrecacheApp)
) {
  window.addEventListener('message', handleMessageFromMockedHost);
  console.info('Not calling appInitialization because part of App Initialization Test run');
} else if (noPrecacheApp && noCustomInit) {
  if (isTestBackCompat()) {
    appInitialization.notifyAppLoaded();
    appInitialization.notifySuccess();
  } else {
    app.notifyAppLoaded();
    app.notifySuccess();
  }
}

if (hasPrecacheApp) {
  app.initialize(validMessageOrigins).then(() => {
    app.notifyAppLoaded();
    app.lifecycle.registerBeforeSuspendOrTerminateHandler(() => {
      return new Promise<void>((resolve) => {
        resolve();
      });
    });
    app.lifecycle.registerOnResumeHandler((context: ResumeContext): void => {
      // get the route from the context
      console.log(context.contentUrl);
      if (noCustomInit) {
        app.notifySuccess();
      }
    });
    app.notifySuccess();
  });
}

```

PR #2923, opened 2025-11-13
APIs to support Open AI widgets

ndangudubiyyam:

## Description

This change consists of set of APIs that the widget hosting metaos app will be calling to support the open AI widgets.

### Validation performed:

1. Unit tests
2. Integration tests

### Unit Tests added:

yes

### End-to-end tests added:

no


juanscr (changes requested):
Everything looks good. I only have one question, which is why are the widget host APIs private? Shouldn't we want them to make this APIs available to the public as this API should be accessible to any widget developer?

PR #2924, opened 2025-11-13
Handle error when returned from the host

juanscr:
## Description

Beforehand, this API would assume it would always resolve successfully. This is not really true in some cases, where the API can fail, which would mix up successful results with errors in the response. This pull requests adds error handling and verification through the updated `callFunctionInHostAndHandleResponse` in order to have visible errors for the consumers of this API.

### Main changes in the PR:

1. Changed `getConfigSetting` teams full trust API to handle the error.

### Validation performed:

1. Used testing host, unit testing and e2e tests.

### Unit Tests added:

Yes.

### End-to-end tests added:

No.  Existing tests are enough.

### Change file added:

Yes.


samelliottdlt (approved):
LGTM ✨

PR #2926, opened 2025-12-02
Add register memory metrics handler for memory heartbeat from host

juanscr:
## Description

This pull request adds an event handler in order for apps to receive memory metrics periodically. This will allow any app to create score cards or monitor memory usage of an app overtime, in order to diagnose any possible memory leak or regression.

### Main changes in the PR:

1. Added memory event handler that app registers to it in order to get heartbeats of memory from the host.

### Validation performed:

1. Tested in testing host with test app and unit tests.

### Unit Tests added:

Yes.

### End-to-end tests added:

No.

### Change file added:

Yes

### DIFF
--- a/tools/releases/powershell-commands/update-npmrc-with-npm-token.ps1
+++ b/tools/releases/powershell-commands/update-npmrc-with-npm-token.ps1
@@ -1,0 +1,3 @@
+param ($PIPELINE_WORKSPACE, $NPM_TOKEN)
+
+Set-Content -Path $PIPELINE_WORKSPACE/microsoft-teams-library-js-pipeline/NPMFeed/.npmrc -Value "//registry.npmjs.org/:_authToken=$NPM_TOKEN"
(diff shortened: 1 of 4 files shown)


packages/teams-js/src/public/app/app.ts
```diff
@@ -771,6 +778,24 @@ export function registerHostToAppPerformanceMetricsHandler(handler: HostToAppPer
...
+/**
+ * @beta
+ * Registers a function to handle memory metrics heartbeat sent from the host periodically.
+ *
+ * @remarks
+ * Only one handler can be registered at a time. A subsequent registration replaces an existing registration.
+ *
+ * @param handler - The handler to invoke each time memory metrics heartbeat is received from the host.
+ */
+export function registerHostMemoryMetricsHandler(handler: HostMemoryMetricsHandler): void {
```
ydogandjiev: How will apps know if this API is supported? Should we move it under a separate namespace and its own capability - e.g. app.memory.registerHostMemoryMetricsHandler? That way an app can call the isSupported function under that namespace before trying to use the API (or get an error when it tries to call it)?

juanscr: Addressed

jekloudaMSFT (approved):

PR #2938, opened 2025-12-08
Add host ancestors to context

samelliottdlt:
## Description

Adds host ancestor to context

### Main changes in the PR:

This pull request introduces support for tracking ancestor hosts in the app context, allowing the system to represent the hierarchy of hosts that an app is embedded within. The changes add a new `ancestors` field to `AppHostInfo`, update context transformation logic to populate this field, and deprecate the older `hostAncestors` field in favor of the new approach.

### Context and Type Updates

* Added an `ancestors` field (array of strings) to the `AppHostInfo` interface in `app.ts`, representing the list of ancestor hosts for an app.
* Updated the `transformLegacyContextToAppContext` function to map the legacy `hostAncestors` field into the new `ancestors` field in the app context.

### Change file added:

Yes

### DIFF
--- a/change/@microsoft-teams-js-02238656-008f-4fa8-a8ec-2719ce15f945.json
+++ b/change/@microsoft-teams-js-02238656-008f-4fa8-a8ec-2719ce15f945.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated valid domains list for M365 Copilot SSR end points.",
+  "packageName": "@microsoft/teams-js",
+  "email": "niharikad@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/package.json
+++ b/package.json
@@ -163,7 +163,7 @@
       "brotli": false,
       "path": "./packages/teams-js/dist/esm/packages/teams-js/src/index.js",
       "import": "{ app, authentication, pages }",
-      "limit": "59 KB"
+      "limit": "60 KB"
     },
     {
       "brotli": false,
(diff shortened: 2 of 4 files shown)


change/@microsoft-teams-js-bf300f9d-59c3-48d7-8699-dd16963346a1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
juanscr: I don't think this should be a `patch` fix, could you change it to `minor`?

change/@microsoft-teams-js-bf300f9d-59c3-48d7-8699-dd16963346a1.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "add host ancestor to context",
```
juanscr: Also minor nitpick but use past tense in changefile. "Added host ancestors to app context."

packages/teams-js/src/public/app/app.ts
```diff
@@ -235,6 +235,11 @@ export interface AppHostInfo {
    * Current ring ID
    */
   ringId?: string;
+
+  /**
+   * List of ancestor hosts
+   */
```
juanscr, line 243: Could you add a more relevant comment here? I like the one you use int he deprecated context:

```typescript
/**
   * An array representing the hierarchy of ancestor hosts that the app is embedded inside of.
   * The array is ordered from immediate parent to root host.
   * For example, if Bizchat is running in Calendar in Teams, this would be ["Calendar", "Teams"].
   */
```

This is useful because this context is actually the one they apps are consuming nowadays, so this will provide better inline relevant documentation.

juanscr (changes requested):

jimchou-dev (approved):

PR #2961, opened 2026-01-15
Updated SSR tests app

vhuseinova-msft:
## Description

- Added routes for SSR test app for better testing
- Enabled SSR certs generation for testing the pages in the Orange app.
- Updated teams-js interfaces

### Validation performed:
Local testing, UI tests

### Unit Tests added:

No new functionality for teams-js

### Change file added:

Yes

### DIFF
--- a/apps/ssr-test-app/.gitignore copy
+++ b/apps/ssr-test-app/.gitignore copy
@@ -1,2 +1,0 @@
-# Generated certificates
-certs/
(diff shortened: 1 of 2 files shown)


apps/ssr-test-app/.gitignore copy
juanscr, line 1: Why is this file needed? hehe

vhuseinova-msft: we use certs only in this app, so adding this file to ignore certs folder only here, instead of doing this for the whole repo

marianamu-MSFT (approved):

PR #2971, opened 2026-01-26
Increasing the shard numbers for android tests as current tests time out too frequently

aeunms:
The tests currently time out an estimated 50% of the time currently. This simple PR tries increasing the number of shards to see if the lessened workload on each shard will be enough to fix the flakiness, even by a bit.

### DIFF
--- a/change/@microsoft-teams-js-58fad0e7-f58a-433b-9684-c15f3988fe2a.json
+++ b/change/@microsoft-teams-js-58fad0e7-f58a-433b-9684-c15f3988fe2a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added MAC domains to valid domains for CDN",
+  "packageName": "@microsoft/teams-js",
+  "email": "niharikad@microsoft.com",
+  "dependentChangeType": "patch"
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -234,7 +234,7 @@ extends:
               - template: tools/yaml-templates/android-test.yml@self
                 parameters:
                   androidAppHostingSdkGitPath: AndroidAppHostingSdk
-                  shardNum: 2
+                  shardNum: 3
```
jekloudaMSFT, line 237: We'll need a `shardIndex: 2` job here too, right?

jekloudaMSFT: Not sure how fancy yaml lets you get, but maybe there's a way to configure all the jobs just based on the number of shards?

aeunms: You're totally right! Let me fix that. Thank you for catching it!

aeunms: Just pushed. I spent some time trying to see if I could do some sort of loop. But I couldn't figure it out too quickly, so I just hard coded it for now.

PR #2979, opened 2026-02-05
Added SSR app build step

vhuseinova-msft:
## Description
Added SSR app build step so CI tests can pass for tests with names starting with S

### Main changes in the PR:

1. Added SSR app build step
2. Added SSR demo app certification creation step

### Validation performed:

1. Ci

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

No

### DIFF
--- a/tools/yaml-templates/web-e2e-tests-job.yml
+++ b/tools/yaml-templates/web-e2e-tests-job.yml
@@ -56,6 +56,11 @@ jobs:
               value: 'test'
             ${{ else }}:
               value: 'localScriptTag'
+          - name: ssrFlag
+            ${{ if eq(testPrefixPattern, '{[S],[s]}') }}:
+              value: '--ssr'
+            ${{ else }}:
+              value: ''
         steps:
           - template: build-app-host.yml
             parameters:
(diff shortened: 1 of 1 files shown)


juanscr (approved):

tools/yaml-templates/web-e2e-tests-job.yml
```diff
@@ -56,6 +56,11 @@ jobs:
               value: 'test'
             ${{ else }}:
               value: 'localScriptTag'
+          - name: ssrFlag
+            ${{ if eq(testPrefixPattern, '{[S],[s]}') }}:
```
aeunms, line 60: Would this make it so that the a new test in the future that also starts with S might be triggered by this to be fed an ssr flag when not expected?

vhuseinova-msft: This is triggered for all tests which names starts with S or s (including ssr tests) once because of how our test patterns used. This flag just adds a step to start ssr test app, so it won't break anything, just increase time for tests to run

aeunms: I think I should have left this comment on line 102 but I think it's the same answer for that, sounds good!

PR #2984, opened 2026-02-17
Support for valid domains using a nested wildcard

juanscr:

## Description
Beforehand, we only allowed wildcard on the first subdomain of a valid origin. Although this was an easy abstraction, we have found that new apps need to add more valid origins that have wildcards in different part of the domains. As such:

1. Wildcards can be part in any part of the subdomain that is not the top-level domain (TLD). The reason why we are not allowing it in the TLD is to force apps to properly configure which domains are available instead of allowing too much unintended domains.
2. Only one wildcard is allowed in the whole domain, which keeps the old restriction still true while allowing more flexibility. The only wildcard is also for security reasons, so app need to fully enumerate the domains they want avoiding not-wanted domains.

### Main changes in the PR:

1. Change domain validation to support wildcards in other parts of the subdomain.

### Validation performed:

1. Unit testing.

### Unit Tests added:

Yes.

### End-to-end tests added:

No.

### Change file added:

No.


jekloudaMSFT (approved):

packages/teams-js/src/internal/urlPattern.ts
vhuseinova-msft, line 24: minor: Could you add description for what the func does?

packages/teams-js/src/internal/urlPattern.ts
```diff
@@ -28,20 +28,40 @@ const userOriginUrlValidationRegExp = /^[A-Za-z][A-Za-z\d+.-]*:\/\//;
...
+    if (patternSegments[i] === hostSegments[i]) {
+      continue;
+    }
+
+    if (patternSegments[i] !== '*') {
+      return false;
+    }
+
+    // Wildcard in the last segment (TLD position) is not allowed for security reasons.
+    if (i === patternSegments.length - 1 || hasUsedWildcard) {
```
vhuseinova-msft, line 58: minor, could you add a comment here that the second part of the check (`|| hasUsedWildcard`) is needed because only 1 wildcard is allowe in the pattern?

vhuseinova-msft (approved):

PR #2990, opened 2026-02-26
Pre-release 2.49.0

juanscr:
## 2.49.0

Thu, 26 Feb 2026 20:18:33 GMT

### Minor changes

- Added support for nested wildcard subdomains. The valid origins are still restricted to the usage of only one wildcard in the valid origin.
- Added support for non-https URLs for opt-in app valid origins in `app.initialize()`.

### Patches

- Updated stageView.StageViewParams interface to make threadId optional.


juanscr:
Abandoning this PR as I will wait for another PR to be merged. Will re-open a new one with that work included.

PR #3001, opened 2026-03-09
Plugins API

ndangudubiyyam:

## Description

Add support for the plugins API

### Main changes in the PR:

1. Added pluginService.ts
2. Added 'deprecated' tag in the customMessage in privateAPI

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/apps/teams-test-app/e2e-test-data/profile.json
+++ b/apps/teams-test-app/e2e-test-data/profile.json
@@ -9,6 +9,9 @@
     {
       "title": "showProfile API Call - Success",
       "type": "callResponse",
+      "hostSdkVersion": {
+        "web": "<=9.15.0"
+      },
       "boxSelector": "#box_showProfile",
       "inputValue": {
         "modality": "Card",
@@ -26,6 +29,30 @@
         "triggerType": "MouseClick"
       },
       "expectedAlertValue": "showProfile called with {\"modality\":\"Card\",\"persona\":{\"identifiers\":{\"Smtp\":\"test@microsoft.com\"}},\"targetElementBoundingRect\":{\"x\":243,\"y\":0,\"width\":0,\"height\":0,\"top\":0,\"right\":243,\"bottom\":0,\"left\":243},\"triggerType\":\"MouseClick\"}"
+    },
+    {
+      "title": "showProfile API Call - Success",
+      "type": "callResponse",
+      "hostSdkVersion": {
+        "web": ">10.0.0"
+      },
+      "boxSelector": "#box_showProfile",
+      "inputValue": {
+        "modality": "Card",
+        "persona": {
+          "identifiers": {
+            "Smtp": "test@microsoft.com"
+          }
+        },
+        "targetElementBoundingRect": {
+          "x": 0,
+          "y": 0,
+          "width": 0,
+          "height": 0
+        },
+        "triggerType": "MouseClick"
+      },
+      "expectedAlertValue": "regex:showProfile called with .*\"Smtp\":\"test@microsoft.com\".*\"triggerType\":\"MouseClick\""
     }
   ]
 }


packages/teams-js/test/internal/urlPattern.spec.ts
```diff
@@ -19,7 +19,7 @@ describe('urlPattern', () => {
     });
 
     it('returns true for an http pattern', () => {
-      expect(isValidPatternUrl('http://example.com')).toBe(true);
+      expect(isValidPatternUrl('https://example.com')).toBe(true);
```
vhuseinova-msft: this should be http (see the tests description)

packages/teams-js/src/public/runtime.ts
```diff
@@ -626,6 +627,7 @@ export const mapTeamsVersionToSupportedCapabilities: Record<string, Array<ICapab
     // isDeeplyNestedAuthSupported: true (based on const legacyTeamsMobileVersionForDeeplyNestedAuth)
     // for hostClientTypes: [HostClientType.android, HostClientType.ios, HostClientType.ipados, HostClientType.visionOS]
   ],
+  // TODO: Need to add plugins here if we old teams hosts decide to use plugins.
```
vhuseinova-msft: Is this still needed?

packages/teams-js/src/private/pluginService.ts
```diff
@@ -0,0 +1,220 @@
...
+ * @remarks
+ * Plugin messages are serialized before transport to the host, so payload data
+ * is constrained to JSON-safe types.
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export type JsonValue = string | number | boolean | null | JsonValue[] | { [key: string]: JsonValue };
```
vhuseinova-msft: Should we just have string|object here to simplify the type?

packages/teams-js/src/private/plugin.md
```diff
@@ -0,0 +1,257 @@
+# Implementing Plugins
+
+This guide covers how to create plugins for the Hub SDK.
+
+## Dependencies
+
+The Hub SDK should be a **dev dependency** (`devDependencies`) in your plugin package:
```
vhuseinova-msft: Is this still needed? Also, we don't want to include web hub sdk info in teams-js

packages/teams-js/src/private/pluginService.ts
```diff
@@ -0,0 +1,220 @@
...
+ * @beta
+ */
+export async function sendMessage(message: PluginMessage): Promise<void> {
+  ensureInitialized(runtime);
+
+  if (!message.func) {
+    throw new Error('func is required in PluginMessage.');
+  }
+  if (!message.pluginId) {
+    throw new Error('pluginId is required in PluginMessage.');
```
vhuseinova-msft: Why do we need plugin id here?
Plugin id shouldn't be a requirement for plugin message or for sending messages and it should be used for access to plugin in plugin service.

packages/teams-js/src/private/plugin.md
ndangudubiyyam: good catch. Will remove it

packages/teams-js/src/private/pluginService.ts
juanscr: I _think_ it seems important if the plugin system expands, doesn't it? Accidentally, two different plugins may register a same message and we need to be able to separate this, correct? Or enforce this runtime wise on the Host but it could be a bit tricky.

vhuseinova-msft: ` Accidentally, two different plugins may register a same message and we need to be able to separate this, correct?`
We shouldn’t restrict registering multiple plugins for the same message.I'm not sure why the app or hub needs to distinguish which plugin produced the message (similar to how custom messages work, we don't really care which part of the app sent it)

apps/teams-test-app/src/components/utils/catalyst-plugin-contract.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * Catalyst Plugin Contract
+ *
+ * This file represents what would be a standalone shared types package possibly in 1js
+ * (e.g. `@1js/bizchat-catalyst-plugins-metaos-hub`) in production.
```
vhuseinova-msft: not sure if it's okay to mention 1JS repo here.

apps/teams-test-app/src/components/utils/catalyst-plugin-contract.ts
```diff
@@ -0,0 +1,62 @@
...
+ *   - Host (hub): uses it to type outbound send calls and inbound listener args
+ *   - App (iframe): uses it to know what messages to handle and what to send back
+ *
+ * Neither side owns this file — it is the source of truth for the wire contract.
+ */
+
+// ── Plugin identifiers ───────────────────────────────────────────────────────
+// Routed through every PluginMessage — host and app must agree on these.
+
+export const CatalystPluginIds = {
```
vhuseinova-msft: Same questions as in other repo :)

packages/teams-js/test/internal/urlPattern.spec.ts
```diff
@@ -96,7 +97,7 @@ describe('urlPattern', () => {
     });
 
     it('returns a verifier for an http pattern', () => {
-      const verifier = createURLVerifier('http://example.com', logger);
+      const verifier = createURLVerifier('https://example.com', logger);
```
vhuseinova-msft: Should it be http here?

packages/teams-js/src/private/pluginService.ts
ndangudubiyyam: @juanscr I think we can just use the funcName as a differentiatior for now. In future we can add different fields to the PluginMessage interface, that can add a secondary differentiator like `pluginId` or `schemaVersion` or `messageType`

packages/teams-js/src/private/pluginService.ts
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @throws Error if SDK initialization has not completed, if the host returns
+ * an error response, or if `func` is missing.
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export async function sendMessage(message: PluginMessage): Promise<void> {
```
vhuseinova-msft: Instead of asking customers to call sendMessage / receiveMessage APIs directly, I was thinking we could introduce a structure similar to the Web Hub SDK.

Specifically, we could define an IPlugin interface that plugins implement, and then allow them to register or create a bridge via a plugin service. This would be similar to [plugin.ts in microsoft‑teams‑library‑js](https://github.com/OfficeDev/microsoft-teams-library-js/blob/22c0ba96304ac1a2189d5f8573f9cbdea87b1ee9/packages/teams-js/src/internal/plugin.ts).

As a result, the plugin interfaces exposed to API consumers would look very similar between Teams JS and the Web Hub SDK.

vhuseinova-msft: For future refference:
"Thanks for the design suggestion — the IPlugin / PluginMessagingBridge pattern is a great fit for SDKs that own plugin lifecycle, and I can see the appeal of aligning with the Web Hub SDK interface.
I'm going to keep the functional API for now, for a few reasons:
Consistency with the rest of Teams JS — Every capability in this SDK app pages, meeting, etc.) exposes plain functions. Introducing a class-based interface here would be an outlier, and I'd rather not create a second pattern for consumers to learn.
Versioning and deprecation flexibility — With a functional API, I can add new functions alongside old ones and deprecate them independently. With IPlugin, any rename or addition of a required method is a breaking change for every plugin implementer. The functional model gives us more room to evolve without breaking callers.
Tree-shaking — A consumer that only needs sendMessage will import just that one function. With the bridge/interface approach, the full interface shape tends to pull in more than needed."

vhuseinova-msft (dismissed):

juanscr (dismissed):

packages/teams-js/src/private/pluginService.ts
```diff
@@ -0,0 +1,172 @@
...
+ * @remarks
+ * Plugin messages are serialized before transport to the host, so payload data
+ * is constrained to JSON-safe types.
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export type JsonValue = string | number | boolean | null | JsonValue[] | { [key: string]: JsonValue };
```
jekloudaMSFT: We already have `SimpleType` and `ISerializable` that we use to pass messages over the iframe boundary, could we reuse those here?

packages/teams-js/src/private/pluginService.ts
```diff
@@ -0,0 +1,172 @@
...
+ * @param handler - Callback invoked for each incoming plugin message payload.
+ *
+ * @throws Error if plugin messaging is not supported by the current host.
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export function receivePluginMessage(handler: ReceiveMessageHandler): void {
```
jekloudaMSFT: Should this be something like `onReceivePluginMessage` or `registerReceivePluginMessageHandler`? I think that's generally the convention for functions that register event handlers

packages/teams-js/src/private/pluginService.ts
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export type PluginMessage = {
+  func: string;
+  args?: JsonValue;
+  correlationId?: string; // Useful in the future for correlating requests and responses between host and plugin, but currently unused.
```
jekloudaMSFT: Are we going to implement `correlationId` passing? If it's going to be part of the initial implementation we probably don't have to say it's unused here. And if it is going to be unsupported that should be made explicit in the TSDoc comment for consumers

packages/teams-js/src/private/pluginService.ts
ndangudubiyyam: Sure let me change it

packages/teams-js/src/private/pluginService.ts
ndangudubiyyam: Thanks for the input I modified the comments

packages/teams-js/src/private/plugins.ts
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export function receivePluginMessage(handler: ReceiveMessageHandler): void {
+  registerHandlerHelper(
+    getApiVersionTag(pluginTelemetryVersionNumber, ApiName.Plugins_ReceiveMessage),
+    ApiName.Plugins_ReceiveMessage,
```
jekloudaMSFT, line 126: It looks like this only allows one handler per app, is that expected? Like the app cannot register multiple handlers for multiple plugins/messages, it all needs to be handled on one handler?

packages/teams-js/src/private/plugins.ts
```diff
@@ -0,0 +1,172 @@
+import { callFunctionInHost } from '../internal/communication';
```
jekloudaMSFT, line 1: nit: I think `plugin.ts` or `plugins.ts` would follow the general naming convention for a capability file. I would like to see a consistent naming between this file, the test app component, and the runtime object ("plugin" vs "plugins" vs "pluginService")

packages/teams-js/src/public/runtime.ts
jekloudaMSFT: Please remove this comment or link a work item for this

packages/teams-js/test/internal/urlPattern.spec.ts
```diff
@@ -19,6 +19,7 @@ describe('urlPattern', () => {
     });
 
     it('returns true for an http pattern', () => {
+      // eslint-disable-next-line @microsoft/sdl/no-insecure-url
```
jekloudaMSFT, line 22: Why do we need these eslint comments now?

packages/teams-js/test/private/pluginService.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      expect(handler).toHaveBeenCalledWith(
+        expect.objectContaining({
+          func: 'catalyst.contextUpdate',
+          args: { key: 'theme', value: 'dark' },
+          correlationId: 'corr-999',
+        }),
+      );
+    });
+
+    it('should only be callable from content context', async () => {
```
jekloudaMSFT: This API isn't only limited to the content context, right?

apps/teams-test-app/e2e-test-data/profile.json
```diff
@@ -26,6 +29,30 @@
         "triggerType": "MouseClick"
       },
       "expectedAlertValue": "showProfile called with {\"modality\":\"Card\",\"persona\":{\"identifiers\":{\"Smtp\":\"test@microsoft.com\"}},\"targetElementBoundingRect\":{\"x\":243,\"y\":0,\"width\":0,\"height\":0,\"top\":0,\"right\":243,\"bottom\":0,\"left\":243},\"triggerType\":\"MouseClick\"}"
+    },
+    {
+      "title": "showProfile API Call - Success",
```
jekloudaMSFT: Why is this profile test included in this PR?

apps/teams-test-app/src/components/utils/shared-plugin-contract.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Catalyst Plugin Contract
+ *
+ * This file represents what would be a standalone shared types package
+ * (e.g. `@Shared/plugin-contract`) in production.
+ *
+ * Both sides of the messaging boundary import from it:
+ *   - Host (hub): uses it to type outbound send calls and inbound listener args
+ *   - App (iframe): uses it to know what messages to handle and what to send back
+ *
+ * Neither side owns this file — it is the source of truth for the wire contract.
```
jekloudaMSFT, line 11: Does the host actually use this file, or do they have to maintain a duplicate of it?

change/@microsoft-teams-js-ae58228c-281e-4bfd-82b9-552968455d51.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `plugin` capability that will help directly interact with the host bypassing API registration in hubsdk. The capability is still awaiting support in one or most host applications. To track availability of this capability across different hosts see https://aka.ms/capmatrix",
```
jekloudaMSFT: Please change this to `pluginService` if that's the capability name we're using here

packages/teams-js/test/internal/urlPattern.spec.ts
juanscr: Those are needed, I think I messed up when adding the tests hehe

packages/teams-js/src/private/pluginService.ts
juanscr: Agree with this! I think we can extend simple type to include the recursive ` { [key: string]: SimpleType }`. We are having a bunch of serialization classes only just to pass the object as-is

apps/teams-test-app/src/components/privateApis/PluginAPI.tsx
```diff
@@ -0,0 +1,74 @@
+import { plugins } from '@microsoft/teams-js';
```
jekloudaMSFT, line 1: nit: I think this should probably be `PluginsAPI.tsx` now?

packages/teams-js/src/private/plugins.ts
```diff
@@ -0,0 +1,159 @@
...
+ * @param handler - Callback invoked for each incoming plugin message payload.
+ *
+ * @throws Error if plugin messaging is not supported by the current host.
+ *
+ * @hidden
+ * @internal
+ * Limited to Microsoft-internal use
+ * @beta
+ */
+export function registerPluginMessage(handler: ReceiveMessageHandler): void {
```
jekloudaMSFT, line 123:
```suggestion
export function registerPluginMessageHandler(handler: ReceiveMessageHandler): void {
```

packages/teams-js/src/internal/responseHandler.ts
```diff
@@ -22,7 +22,7 @@ export abstract class ResponseHandler<SerializedReturnValueFromHost, Deserialize
   public abstract deserialize(response: SerializedReturnValueFromHost): DeserializedReturnValueFromHost;
 }
 
-export type SimpleType = string | number | boolean | null | undefined | SimpleType[];
+export type SimpleType = string | number | boolean | null | undefined | SimpleType[] | { [key: string]: SimpleType };
```
jekloudaMSFT, line 25: This type needs to be exposed to consumers if they're expected to call the plugins APIs directly. Maybe we could export a type alias from the plugins package?

PR #3003, opened 2026-03-12
Add new `appState` parameter to deprecate old fields.

juanscr:
## Description

Added new field called `appState` to indicate the current state of the app's lifecycle. This replaces the old `isCaching` and `isPrecaching` field, as they are booleans that can't be true at once (mark as deprecated to not break existing apps). This fields will still be populated by hosts in order to preserve backwards compatibility with any version of the SDK for now.

### Main changes in the PR:

1. Added new field `appState`.
2. Deprecate `isCaching` and `isPrecaching`.

### Validation performed:

1. Tested raising events in test host and E2E tests.

### Unit Tests added:

No, schema change only.

### End-to-end tests added:

Modified to be aligned with newer schema.

### Change file added:

Yes.

### Related PRs:

https://github.com/OfficeDev/microsoft-teams-library-js/pull/2994, https://github.com/OfficeDev/microsoft-teams-library-js/pull/2926,

### DIFF
--- a/change/@microsoft-teams-js-5455b35d-b329-401c-947b-3ad9a3371e8c.json
+++ b/change/@microsoft-teams-js-5455b35d-b329-401c-947b-3ad9a3371e8c.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Added `appState` parameter to indicate current state of the app's lifecycle. Deprecate `isCaching` and `isPrecaching` fields to use the new structure.",
+  "comment": "Added `appState` parameter to `HostMemoryMetrics` to indicate current state of the app's lifecycle. Deprecated `isCached` and `isPrecached` fields to use the new structure.",
   "packageName": "@microsoft/teams-js",
   "email": "31258166+juanscr@users.noreply.github.com",
   "dependentChangeType": "patch"


change/@microsoft-teams-js-5455b35d-b329-401c-947b-3ad9a3371e8c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added `appState` parameter to indicate current state of the app's lifecycle. Deprecate `isCaching` and `isPrecaching` fields to use the new structure.",
```
jekloudaMSFT:
```suggestion
  "comment": "Added `appState` parameter to `HostMemoryMetrics` to indicate current state of the app's lifecycle. Deprecated `isCached` and `isPrecached` fields to use the new structure.",
```

apps/teams-test-app/e2e-test-data/appPerformanceMetrics.json
```diff
@@ -13,8 +13,9 @@
       "boxSelector": "#box_appPerformanceMetrics_registerHostMemoryMetricsHandler",
       "eventName": "memoryUsageHeartbeat",
       "eventData": {
+        "appState": "active",
         "isCached": false,
-        "isPrecached": true,
+        "isPrecached": false,
```
jekloudaMSFT, line 18: Why is this test value changed? Shouldn't the host continue to return the same values for the deprecated parameters?

juanscr: Good question! The updated hosts will auto-populate the deprecated fields based on the value of the `appState` to preserve compatibility. As such, as I added `appState` as active, both deprecated fields have a `false` value; if I wanted to keep the test intact, would have to put `appState: "precache"` in order to garantee this test makes sense.

PR #3016, opened 2026-03-24
Add new content type to shareWebContent SDK schema

snandy-msft:
## Description

This PR extends the `sharing.shareWebContent` API to support sharing **File** content (e.g. SharePoint/OneDrive document URLs) in addition to the existing URL content type. Previously, `shareWebContent` only accepted `IURLContent` items. With this change, callers can now pass `IFileContent` items to share a file URL directly into a Teams conversation.

No breaking changes — existing URL sharing behavior is unchanged. The new `IFileContent` interface mirrors the shape of `IURLContent` (`url`, `message?`, `preview?`) with `type: 'File'` as the discriminator.

### Main changes in the PR:

1. **`sharing.ts`** — Added `IFileContent` interface, extended `ContentType` union to `'URL' | 'File'`, extended `IShareRequestContentType` to `IURLContent | IFileContent`, and added File branch in `validateContentForSupportedTypes`.
2. **`SharingAPIs.tsx`** — Updated the Share Web Content test app component to accept and validate both URL and File content types.
3. **`sharing.json`** (e2e-test-data) — Added E2E test case for File type sharing.
4. **`sharing.spec.ts`** — Added unit tests for File content success, File missing URL error, and URL+File mixed-type error scenarios.

### Validation performed:

1. Ran full unit test suite (`pnpm test`) — all 9200 tests pass across 85 test suites.
2. Verified build succeeds with no errors (`pnpm build` in `packages/teams-js`).
3. Verified input validation rejects File content missing `url`, mixed URL+File content, and unsupported types.

### Unit Tests added:

Yes — added for both v1 (callback) and v2 (Promise) API variants:
- File content success (v1 + v2)
- File content missing `url` error (v1 + v2)
- Mixed URL + File content type error (v1 + v2)

### End-to-end tests added:

Yes — added a File type test case to `apps/teams-test-app/e2e-test-data/sharing.json`.

### Change file added:

Yes — `change/@microsoft-teams-js-b9b5c932-d1ba-40cb-9fee-326eb2ea24d0.json` (type: `minor`).

### DIFF
--- a/apps/teams-test-app/src/components/SharingAPIs.tsx
+++ b/apps/teams-test-app/src/components/SharingAPIs.tsx
@@ -20,17 +20,13 @@ const ShareWebContent = (): React.ReactElement =>
         if (!input.content || input.content.length === 0) {
           throw new Error('content is required');
         }
+        const supportedTypes: string[] = ['URL', 'FILE'];
         for (const contentItem of input.content) {
-          if (contentItem.type === 'URL') {
-            if (!(contentItem as sharing.IURLContent).url) {
-              throw new Error('Each URL content item must have a url property set.');
-            }
-          } else if (contentItem.type === 'FILE') {
-            if (!(contentItem as sharing.IFileContent).url) {
-              throw new Error('Each File content item must have a url property set.');
-            }
-          } else {
-            throw new Error('Unsupported content type');
+          if (!supportedTypes.includes(contentItem.type)) {
+            throw new Error(`Unsupported content type: ${contentItem.type}`);
+          }
+          if (!contentItem.url) {
+            throw new Error(`Each ${contentItem.type} content item must have a url property set.`);
           }
         }
       },
--- a/packages/teams-js/src/public/sharing/sharing.ts
+++ b/packages/teams-js/src/public/sharing/sharing.ts
@@ -189,31 +189,22 @@ function validateTypeConsistency(shareRequest: IShareRequest<IShareRequestConten
   }
 }
 
-function validateContentForSupportedTypes(shareRequest: IShareRequest<IShareRequestContentType>): void {
-  let err: SdkError | undefined;
-  if (shareRequest.content[0].type === 'URL') {
-    if (shareRequest.content.some((item) => !item.url)) {
-      err = {
-        errorCode: ErrorCode.INVALID_ARGUMENTS,
-        message: 'URLs are required for URL content types',
-      };
-      throw err;
+function validateContentItem(item: IShareRequestContentType): void {
+  if (item.type === 'URL') {
+    if (!item.url) {
+      throw { errorCode: ErrorCode.INVALID_ARGUMENTS, message: 'URLs are required for URL content types' };
     }
-  } else if (shareRequest.content[0].type === 'FILE') {
-    if (shareRequest.content.some((item) => !item.url)) {
-      err = {
-        errorCode: ErrorCode.INVALID_ARGUMENTS,
-        message: 'File URLs are required for File content types',
-      };
-      throw err;
+  } else if (item.type === 'FILE') {
+    if (!item.url) {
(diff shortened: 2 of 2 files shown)


apps/teams-test-app/src/components/SharingAPIs.tsx
```diff
@@ -21,11 +21,16 @@ const ShareWebContent = (): React.ReactElement =>
...
+          if (contentItem.type === 'URL') {
+            if (!(contentItem as sharing.IURLContent).url) {
+              throw new Error('Each URL content item must have a url property set.');
+            }
+          } else if (contentItem.type === 'FILE') {
+            if (!(contentItem as sharing.IFileContent).url) {
+              throw new Error('Each File content item must have a url property set.');
+            }
+          } else {
+            throw new Error('Unsupported content type');
```
vhuseinova-msft: This could be simplified:
```suggestion
for (const contentItem of input.content) {
   if (contentItem.type !== 'URL' && contentItem.type !== 'FILE') {
     throw new Error(`Unsupported content type: ${contentItem.type}`);
   }
 
   if (!contentItem.url) {
     throw new Error(`Each ${contentItem.type} content item must have a url property set.`);
   }
 }
```

packages/teams-js/src/public/sharing/sharing.ts
```diff
@@ -178,6 +199,14 @@ function validateContentForSupportedTypes(shareRequest: IShareRequest<IShareRequ
       };
       throw err;
     }
+  } else if (shareRequest.content[0].type === 'FILE') {
+    if (shareRequest.content.some((item) => !item.url)) {
```
vhuseinova-msft: This change looks like a duplication of lines 195–200, let’s move it out into a function.

packages/teams-js/src/public/sharing/sharing.ts
```diff
@@ -178,6 +199,14 @@ function validateContentForSupportedTypes(shareRequest: IShareRequest<IShareRequ
       };
       throw err;
     }
+  } else if (shareRequest.content[0].type === 'FILE') {
```
juanscr: Why are we only checking the first item of the array? Shouldn't this parse all types in the content array or am I missing something?

packages/teams-js/src/public/sharing/sharing.ts
```diff
@@ -178,6 +199,14 @@ function validateContentForSupportedTypes(shareRequest: IShareRequest<IShareRequ
...
+    if (shareRequest.content.some((item) => !item.url)) {
+      err = {
+        errorCode: ErrorCode.INVALID_ARGUMENTS,
+        message: 'File URLs are required for File content types',
+      };
+      throw err;
+    }
```
juanscr: I would tidy up this a lot. Jus thave one function to check an item of the content array, which cheks valid type and URL presence. In this function we just call that function for all items and that's it.

apps/teams-test-app/src/components/SharingAPIs.tsx
snandy-msft: Fixed. Please re-review.

packages/teams-js/src/public/sharing/sharing.ts
snandy-msft: Fixed.

packages/teams-js/src/public/sharing/sharing.ts
snandy-msft: The original code checked `content[0].type` to branch, then `used .some()` to validate URLs across all items, but the type check itself only looked at the first element, so mixed-type arrays (e.g., ['URL', 'FILE', 'INVALID']) would silently skip validation for later items. Fixed by extracting validateContentItem which checks both type and URL presence per item, then calling it for every item via forEach.

packages/teams-js/src/public/sharing/sharing.ts
snandy-msft: Fixed.

PR #3095, opened 2026-07-24
[CI] Pipeline security isolation CFS issues

jimchou-dev:

## Description

- Address ICM 839806238 security isolation for the following pipelines flagged with violations
  - OfficeDev/microsoft-teams-library-js/azure-pipelines.yml
  - OfficeDev/microsoft-teams-library-js/azure-pipelines.yml
  - OfficeDev/microsoft-teams-library-js/tools/releases/build-release.yml

### Main changes in the PR:

1. Configure and authenticate the CFS-backed npm registry before installing pnpm or restoring workspace dependencies across Windows, macOS, Android, iOS, and app-hosting jobs.
2. Generate pipeline-only NuGet configuration and explicitly restore the Blazor test app from the authenticated CFS feed.
3. Remove the Blazor Fluent UI dependency and externally hosted web-component script, replacing the Fluent loading indicator with the app's local spinner.
4. Disable Next.js telemetry and optional Cypress binary downloads during CI builds and tests.
5. Run the SSR test app in production mode when a production build exists, preventing Next.js development-mode version checks against npmjs.org.
6. Install `mkcert` through the agent package manager and invoke Playwright through pnpm instead of downloading or resolving tools directly from public endpoints.
7. Replace test-only `contoso.com` URLs with approved Microsoft-owned domains so existing unit and E2E scenarios remain compatible with network isolation.
8. Enable deterministic size-limit timing during isolated CI builds.

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

No

### End-to-end tests added:

No

### Change file added:

No

### Next/remaining steps:

- [ ] TODO for look into how to report test result without calling `brew install xctesthtmlreport`

### Screenshots:

image


aeunms (dismissed):

palatter (commented):
Reviewed the full diff plus the branch source. The approach is right — routing restore through the internal feed, killing telemetry callouts, and dropping the 3P CDN dependencies is the correct shape for this. A few things I think need another look before this goes in.

## Blockers

### 1. `.spinner` has no CSS definition — the loading indicator is now invisible

`Components/Loading.razor:15` and `Components/Welcome.razor:11` swap `<FluentProgressRing/>` for `<div class="spinner"></div>`, but I can't find a rule that defines it anywhere on this branch:

- `grep -rn spinner apps/blazor-test-app/` returns only the two usages, no definition
- `wwwroot/css/site.css` has no `.spinner`
- there's no `Loading.razor.css`, and `Welcome.razor.css` doesn't define it
- I pulled the externally-referenced `fabric.min.css` (v11.0.0, ~339 KB) that `_Host.cshtml` loads and it has zero matches for `.spinner{` and no `ms-Spinner` rules at all — it's the core-only bundle, no component CSS

So both loading states should now render a zero-size empty div. Unless I'm missing an import, the spinner is gone rather than restyled. Probably wants an actual keyframe rule in `site.css` or a scoped `Loading.razor.css`.

### 2. `server.js` silently switches local dev into production mode

```js
const hasProductionBuild = fs.existsSync(path.join(__dirname, '.next', 'BUILD_ID'));
const dev = process.env.NODE_ENV !== 'production' && !hasProductionBuild;
```

`dev:https` is `node server.js`, so any developer who has ever run `pnpm build` loses HMR from that point on — no log line, no opt-out — and can be served a stale `.next` build while editing source. That's a rough one to debug when it bites.

Could we gate on something explicit instead, e.g. `process.env.CI` or a named `NEXT_USE_PROD_BUILD=1` set in the pipeline, and log which mode was picked?

## Isolation gaps — I don't think these are fully closed yet

### 3. The Playwright browser binary still comes from a public CDN

`npx playwright install` → `pnpm exec playwright install` only changes how the *CLI* is resolved. The chromium binary itself still downloads from `cdn.playwright.dev` / `playwright.azureedge.net`. I grepped and there's no `PLAYWRIGHT_DOWNLOAD_HOST` set anywhere in the repo. The header comment was updated to match the new command, which makes this read as handled when I don't think it is.

### 4. `apt-get` still reaches the public Ubuntu archive

`sudo apt-get update && sudo apt-get install -y libnss3-tools mkcert` runs on `Azure-Pipelines-1ESPT-ExDShared` / `ubuntu-latest`. Unless that pool has an internal apt mirror configured, this swaps `dl.filippo.io` for `archive.ubuntu.com` — still a public endpoint. Worth confirming whether apt is in scope for this policy. Functionally it should work, `mkcert` is in Ubuntu universe.

### 5. `ios-test.yml` looks orphaned

`grep -rn "ios-test.yml"` across the repo returns nothing — no pipeline appears to reference it. If that's right, the `brew install xctesthtmlreport` TODO is sitting in dead code and the iOS changes here are inert. Worth either wiring it up or noting that explicitly.

For what it's worth, `tools/releases/build-release.yml` **is** covered transitively via `build-test-publish.yml@self`, so that part of the description checks out.

## Maintainability

**6. The same config is duplicated 4–6 times.** The nuget.config XML block is repeated verbatim across `android-test.yml`, `ios-test.yml`, `build-app-host.yml`, and `build-test-publish.yml`, and the registry URL is hardcoded in six places. Could we extract a `tools/yaml-templates/configure-cfs-feeds.yml` step template plus a pipeline variable for the feed URL? Right now moving the feed is a six-file change.

**7. Three different mechanisms edit the same `.npmrc`.** android/ios use a bare `echo >>`, app-hosting uses `sed -i '/^registry=/d'` + `printf`, build-test-publish uses PowerShell `Add-Content`. The bare append only works today because the root `.npmrc` happens to end in a newline — I checked, it does. Drop that trailing newline in some future edit and you silently get `strict-peer-dependencies=trueregistry=...`. Worth standardizing on one.

**8. `pnpm-lock.yaml` is pure churn** — the only change is removing the trailing newline. Probably worth reverting.

**9. `--frozen-lockfile` is added in exactly one place** (client-sdk install in `build-app-host.yml`). Should that be consistent across the other `pnpm install` calls, or dropped?

## Smaller things

- The unit-test `contoso.com` → `microsoft.com` changes in the three spec files are jest tests against mocked windows, so no network is involved — is this needed for a static string scan? Asking because `contoso.sharepoint.com` still survives in `apps/teams-test-app/e2e-test-data/sharing.json`, so the rule isn't applied uniformly either way.
- `e2e-test-data/app.json` sets `expectedWindowOpenTarget` to `https://microsoft.com/l/entity/...`. If that E2E genuinely calls `window.open`, this trades one real outbound request for another — a non-resolving or localhost target might be safer.
- `_Host.cshtml:15` still pulls `https://static2.sharepointonline.com/.../fabric.min.css`. Microsoft-owned so likely allowlisted, but the unpkg script was removed for a similar reason — worth confirming the SPO CDN is explicitly fine.
- `SIZE_LIMIT_FAKE_TIME=1` degrades the size-limit timing signal. Is the size gate still meaningful with it on, and is it only needed because the timing measurement pulls a Chrome binary?
- Replacing `<FluentDesignSystemProvider AccentBaseColor="#6264A7">` with a bare `<div>` leaves a no-op DOM node and quietly drops the Teams theming. Fine for a test app, just flagging it's a behavior change and not purely mechanical.
- `always-auth=true` is legacy and ignored by npm 9+; `npmAuthenticate@0` writes the scoped token that actually matters. Harmless, just noise.
- After `npmAuthenticate@0` the sources-dir `.npmrc` holds a registry credential — could someone confirm no step publishes `$(Build.SourcesDirectory)` as a pipeline artifact? The `$(Pipeline.Workspace)/.pnpm-store` cache looks fine.

## Summary

Items 1 and 2 look like genuine regressions that are independent of the isolation work, so I'd want those sorted first. Items 3–5 are the ones that decide whether this actually clears the underlying compliance ask rather than just moving the endpoints around. The rest is cleanup that'll pay for itself the first time the feed URL moves.

One process note: the incident ID referenced in the description may not be the one currently tracking this work — worth double-checking before this gets linked for closure.

Happy to be wrong on any of these, particularly #3 and #4 if the pool config already handles them.

jimchou-dev:
> Reviewed the full diff plus the branch source. The approach is right — routing restore through the internal feed, killing telemetry callouts, and dropping the 3P CDN dependencies is the correct shape for this. A few things I think need another look before this goes in.

Fixed

• #1 Spinner — added a  .spinner  rule +  @keyframes  in  site.css  (Teams accent  #6264A7 ). Loading indicator renders again.
• #2 server.js — replaced the implicit dev/prod switch with an explicit gate: honors  NODE_ENV=production / NEXT_USE_PROD_BUILD=1  and  NEXT_FORCE_DEV=1 , falls back to on-disk  .next/BUILD_ID  detection, and logs the chosen mode. This is also the actual root cause of the 25 violations — Next.js dev server fetches  registry.npmjs.org/-/package/next/dist-tags , and the hub-sdk spawns the SSR app with a stripped env, so it defaulted to dev.
• #7 .npmrc — standardized all appends to start on a fresh line ( printf '\n…'  and PowerShell  "`n…" ), removing the trailing-newline dependency.
• #8 Lockfile — reverted;  pnpm-lock.yaml  is now 0-diff vs main.

Won't change

• #3 Playwright CDN —  PLAYWRIGHT_DOWNLOAD_HOST  was tried previously and breaks the pipeline, so leaving CLI resolution as-is. The binary download isn't being flagged.
• #4 apt/mkcert — mkcert is already present on the pool, so this isn't triggering a violation.
• #5 iOS — pipeline isn't flagging the  brew install  call; removed the speculative TODO. Left the file wired as-is.
• #6 Duplication — agreed it's real debt; tracking as a separate follow-up rather than widening this PR's blast radius.
• #9  --frozen-lockfile  — kept scoped to client-sdk; expanding it is unrelated behavior risk.

palatter (dismissed):

palatter (approved):

tools/yaml-templates/android-test.yml
```diff
@@ -43,20 +56,51 @@ steps:
...
 
   - task: CmdLine@2
     displayName: 'Build client sdk'
     inputs:
       script: |
         pnpm build-force-blazor
       workingDirectory: '$(ClientSdkProjectDirectory)'
+    env:
+      NEXT_TELEMETRY_DISABLED: '1'
+      SIZE_LIMIT_FAKE_TIME: '1'
```
jekloudaMSFT, line 94: What does "SIZE_LIMIT_FAKE_TIME" do?

jekloudaMSFT (approved):

tools/yaml-templates/android-test.yml
jimchou-dev: The  @size-limit/time  plugin measures not just bundle size but estimated JS execution time. To do that, it uses  estimo , which launches a headless Chrome (Puppeteer) to actually run the JS and profile it.

async function getTime(file, throttling = 1) {
  if (process.env.SIZE_LIMIT_FAKE_TIME) {
    return throttling * parseFloat(process.env.SIZE_LIMIT_FAKE_TIME)   // ← short-circuit
  }
  ...
  perf = await estimo(file)   // ← otherwise: run Puppeteer/Chrome
}

When  SIZE_LIMIT_FAKE_TIME  is set,  getTime()  returns a fake fixed number immediately and never calls  estimo  — so headless Chrome is never launched.

How that prevents the network call

Two network-triggering paths get skipped:

1. Chromium binary download. Puppeteer (via  estimo ) needs a Chrome binary. If it isn't cached locally, Puppeteer downloads it from  storage.googleapis.com  / the Chrome-for-Testing CDN — a public endpoint that trips CFSClean. Faking the time means Puppeteer is never invoked, so nothing downloads.
2. Calibration reference. The plugin also profiles a baseline file  https://unpkg.com/react@18.2.0/...  ( EXAMPLE ) to compute a device throttling factor. That  unpkg.com  fetch is also bypassed when the time is faked.

So  SIZE_LIMIT_FAKE_TIME=1  = "don't actually measure runtime, just plug in a constant," which removes the entire headless-Chrome path and its associated public-CDN traffic.

PR #3105, opened 2026-08-06
Scope the SSR failureSuccess POST counter to the app session

palatter:
The counter that decides whether this page calls notifyFailure or notifySuccess was a single module-level variable, so it was shared by every request the page ever served. Four tab definitions in the hub's app catalog point at this same page (plain, customInit, withMessage, and both), and the count reset on every second request, which made the behaviour parity-dependent: an even count yields notifyFailure, an odd one yields notifySuccess.

That only holds together while requests arrive in exact pairs. Any extra POST -- a Cypress retry, a host-initiated reload, a refresh -- flips the parity, and the next test then receives notifySuccess where it asserts notifyFailure. The following attempt flips it back and passes, which is what makes this present as flakiness rather than as a broken test.

Key the count by the app session id from the POST body instead. That is the correct scope: the host regenerates the session id when the app, entity or frame context changes, so a page load or tab switch starts a fresh count, while the notifyFailure -> reload -> notifySuccess cycle keeps the same one, which is precisely the sequence this counter exists to track.

Verified against the hub: orange-web derives the session id from a uuid per page load and regenerates it in an effect keyed on appName, entityId and frameContext, none of which change during the reload cycle.

The map is bounded, since this server outlives many test runs, and a missing or unreadable session id falls back to a shared key so a body-shape change degrades to the previous behaviour rather than breaking the page.

Covered by 14 assertions over the logic: the normal failure-then-success cycle, isolation between sessions, the interrupted-then-retried case that produced the flake, interleaved sessions, malformed and missing session ids, and the map bound.

## Description

### Main changes in the PR:

1. <Change 1>
2. <Change 2>

### Validation performed:

1. <Step 1>
2. <Step 2>

### Unit Tests added:

<Yes/No>

### End-to-end tests added:

<Yes/No>

### Change file added:

<Yes/No>

### Related PRs:

> Remove this section if n/a

### Next/remaining steps:

> Remove this section if n/a

- [ ] Item 1
- [ ] Item 2

### Screenshots:

> Remove this section if n/a

| Before     | After      |
| ---------- | ---------- |
| < image1 > | < image2 > |

### DIFF
--- a/apps/ssr-test-app/pages/failureSuccessTest.tsx
+++ b/apps/ssr-test-app/pages/failureSuccessTest.tsx
@@ -36,8 +36,8 @@ export interface FailureSuccessTestPageProps {
 // one, which is the sequence this counter exists to track.
 const postRequestCountBySession = new Map<string, number>();
 
-// This server outlives many test runs, so the map is bounded. Entries are only a
-// small integer each, and the oldest session is always the least interesting.
+// This server outlives many test runs, so the map is bounded, evicting the least
+// recently used session. Entries are one small integer each.
 const maxTrackedSessions = 100;
 
 // Used when the session id cannot be read. Falls back to the previous shared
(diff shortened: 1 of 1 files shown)


Leah-Xia-Microsoft (approved):

_Automated review._

PR #3114, opened 2026-08-14
Add fixture-driven capability contract tests (calendar + chat)

vhuseinova-msft:
This is the teams-js half of moving our shared capability tests from the browser E2E suite into fast Jest contract tests. Nothing is removed from E2E here — these are additive producer tests.

Each test drives a real teams-js API and checks the wire message it emits, with the expected values read from the shared capability fixtures (so a contract change fails the test). Two families to start:
- calendar.openCalendarItem
- chat.openGroupChat — the interesting case, since teams-js transforms the input ({users}) into a different wire shape ({members}), so it validates we assert the emitted wire, not the raw input.

These pair with the Hub SDK provider tests that read the same fixtures, so neither repo can change the wire shape without a failing test on one side.

Paired Hub SDK PR: #5593136

Test-only — no src changes.

### DIFF
--- a/packages/teams-js/test/contract/loadFixtureCase.ts
+++ b/packages/teams-js/test/contract/loadFixtureCase.ts
@@ -1,0 +1,43 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+export interface FixtureCase<TInputValue = unknown> {
+  title: string;
+  inputValue: TInputValue;
+  expectedAlertValue: string;
+}
+
+interface FixtureCaseRecord {
+  title?: string;
+  inputValue?: unknown;
+  expectedAlertValue?: string;
+}
+
+interface FixtureFeatureRecord {
+  testCases?: FixtureCaseRecord[];
+}
+
+interface FixtureRecord {
+  testCases?: FixtureCaseRecord[];
+  featureTests?: FixtureFeatureRecord[];
+}
+
+export function loadFixtureCase<TInputValue = unknown>(family: string, title: string): FixtureCase<TInputValue> {
+  const fixturePath = path.resolve(__dirname, '../../../../apps/teams-test-app/e2e-test-data', `${family}.json`);
+  const fixture = JSON.parse(fs.readFileSync(fixturePath, 'utf8')) as FixtureRecord;
+  const testCases = [
+    ...(fixture.testCases ?? []),
+    ...(fixture.featureTests ?? []).flatMap((featureTest) => featureTest.testCases ?? []),
+  ];
+  const fixtureCase = testCases.find((testCase) => testCase.title === title);
+
+  if (!fixtureCase) {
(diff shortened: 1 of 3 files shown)


packages/teams-js/test/public/chat.contract.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(fixtureCase.expectedWirePayload).toBeDefined();
+
+    const openGroupChatPromise = chat.openGroupChat(fixtureCase.inputValue);
+    const message = utils.findMessageByFunc('chat.openChat');
+
+    expect(message).not.toBeNull();
+    if (!message) {
+      throw new Error('chat.openChat message not found');
+    }
+    expect(message.args).toEqual([fixtureCase.expectedWirePayload]);
```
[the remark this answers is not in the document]
vhuseinova-msft: I think this one's actually fine as-is — toEqual (unlike toStrictEqual) ignores undefined properties, so {members, message} matches {members, message, topic: undefined} and the test passes (green locally and in CI).

packages/teams-js/test/contract/loadFixtureCase.ts
```diff
@@ -0,0 +1,94 @@
...
+    return undefined;
+  }
+}
+
+export function loadFixtureCase<TInputValue = unknown, TExpectedWirePayload = unknown>(
+  family: string,
+  title: string,
+  options: LoadFixtureCaseOptions = {},
+): FixtureCase<TInputValue, TExpectedWirePayload> {
+  const fixturePath = path.resolve(__dirname, '../../../../apps/teams-test-app/e2e-test-data', `${family}.json`);
```
palatter, line 60: Small thing, and you may already have this in the follow-up plan: this loader is a second implementation of the one in the paired Hub SDK PR, so the same parse contract is now written twice. They have already diverged cosmetically, `expectedWirePayload` here vs `wirePayload` there, marker `' called with '` vs `'called with '`, fixed `../../../../` vs `TEAMS_JS_ROOT` with a tmp fallback, and the ambiguous-title guard is the same logic in both files rather than one copy. The fixtures cannot drift since both sides read the same JSON, but a change to the fixture format now has to land in two parsers, and the silent-fallback bug that existed only on the Hub SDK side is that already happening once. Since teams-js owns the fixtures, would exporting the parser from here be worth it while there are only two families?

vhuseinova-msft: Thanks! Yes, this is tracked in my plan as the next steps and I'll prioritise it to be added in the next PR

jekloudaMSFT (approved):

palatter (approved):
Sounds good, happy to leave the shared parser to the follow-up since this is additive test code and nothing depends on it yet. Approving.

PR #3116, opened 2026-08-18
Own the fixture parse contract and flag cases with unit coverage

vhuseinova-msft:
This is the teams-js half of moving shared capability tests from the browser E2E suite into fast Jest contract tests. Nothing is removed from E2E — everything still runs in the browser exactly as before.

**Two new schema fields**

- `unitCovered` — the case contract is also proven by fast unit/integration tests. Nine cases are flagged (calendar, chat, appInstallDialog, appEntity, sharing).
- `e2eScope` — `"pr"` (default) or `"shadow"`, the future switch for moving a case off the PR critical path once it has replacement coverage and evidence. Nothing is flipped yet.

Marking coverage in the fixtures rather than in a list on the Hub SDK side means the Hub tests iterate whatever is flagged, so adding a case to a covered family becomes a flag flip with no code change there. It also means both the Cypress and Playwright engines can read it later.

**One parser instead of two**

Per Patrick review on #3114: the fixture parse contract was implemented separately in each repo and had already drifted — different marker string, different field name, a duplicated ambiguous-title guard, and placeholder/flag support on only one side. Since this repo owns the fixtures, it now owns how they are read: `apps/teams-test-app/fixture-contract` exports case selection, covered-case filtering, and wire payload parsing, and `loadFixtureCase` uses it. The Hub SDK reads the same module, so breaking the parser here now fails the Hub tests.

It sits beside `e2e-test-data` rather than inside it, since the Hub SDK copies that directory wholesale and it should stay data only.

Paired Hub SDK PR: # 5610399

Related: #3114


Leah-Xia-Microsoft (dismissed):

_Automated review._

apps/teams-test-app/fixture-contract/index.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  if (!payload.startsWith('{') && !payload.startsWith('[')) {
+    return undefined;
+  }
+
+  try {
+    return JSON.parse(payload);
+  } catch {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
vhuseinova-msft: The primitive restriction was already present in the original loader. Keeping primitive support out of this fix: the dialog alert describes the result argument, while tasks.completeTask sends [result, appIds]. Parsing the string alone would not establish that wire contract.

packages/teams-js/test/public/chat.contract.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(fixtureCase.expectedWirePayload).toBeDefined();
+
+    const openGroupChatPromise = chat.openGroupChat(fixtureCase.inputValue);
+    const message = utils.findMessageByFunc('chat.openChat');
+
+    expect(message).not.toBeNull();
+    if (!message) {
+      throw new Error('chat.openChat message not found');
+    }
+    expect(message.args).toEqual([fixtureCase.expectedWirePayload]);
```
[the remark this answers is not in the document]
vhuseinova-msft: Jest toEqual ignores object properties whose value is undefined. The original chat contract test passes with the assertion unchanged, so no JSON roundtrip is needed here.

apps/teams-test-app/fixture-contract/index.ts
```diff
@@ -0,0 +1,163 @@
...
+ * on `getMedia` — rather than alternative wordings of one call.
+ */
+export function normalizeAlertValues(expectedAlertValue: unknown): string[] {
+  if (typeof expectedAlertValue === 'string') {
+    return [expectedAlertValue];
+  }
+
+  if (Array.isArray(expectedAlertValue)) {
+    return expectedAlertValue.filter((alert): alert is string => typeof alert === 'string');
+  }
```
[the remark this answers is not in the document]
vhuseinova-msft: Fixed. Arrays are now accepted only when every entry is a string; otherwise normalization returns []. Added regressions showing that a parseable alert followed by 123 neither survives normalization nor produces a payload.

apps/teams-test-app/fixture-contract/index.ts
Leah-Xia-Microsoft, line 1: Nothing in CI checks this file. Verified:

- `apps/teams-test-app/tsconfig.json` → `"include": ["src"]`, so the app build never type-checks it
- `apps/teams-test-app/package.json` → `"lint": "pnpm eslint ./src --max-warnings 0 --fix --ext .tsx"` — wrong directory *and* `.tsx` only, so a `.ts` file here wouldn't match even if the path did
- `packages/teams-js/tsconfig.eslint.json` extends a tsconfig rooted at `packages/teams-js`, so it can't reach a file four directories up either
- no repo-wide `prettier --check` or `tsc --noEmit` in `.github/workflows/` or the Azure pipeline
- net: the only thing that compiles this file is ts-jest, via `jest.config.common.js` → `{ strictNullChecks: false, target: 'ES2015' }`

Here's the proof rather than the theory: our root `.eslintrc.js` extends `plugin:prettier/recommended`, which makes formatting an ESLint *error*. This file currently violates `trailingComma: 'all'` at **line 137** (`options: SelectFixtureCaseOptions = {}`) and **line 153** (the ambiguity `throw` template) — and CI is green. So we know for certain that no ESLint run in CI covers this path.

Could we add `fixture-contract` to the app's tsconfig `include` and widen its lint script to `./src ./fixture-contract --ext .ts,.tsx`? A module compiled with `strictNullChecks` off is a shaky foundation for the thing two repos parse their fixtures with.

apps/teams-test-app/capabilities.schema.json
```diff
@@ -66,7 +66,16 @@
                 "type": "array",
                 "items": { "type": "string" }
               },
-              "skipForCallbackBasedRuns": { "type": "boolean" }
+              "skipForCallbackBasedRuns": { "type": "boolean" },
+              "unitCovered": {
```
Leah-Xia-Microsoft, line 70: This description says "this case's contract is also proven by fast unit/integration tests," but nothing can verify that, and I checked all nine flagged cases: only two (calendar `openCalendarItem`, chat `openGroupChat`) are actually consumed by a fixture-driven test. The other seven point at hand-written specs that share no data with the fixture — so in practice the flag means "this family has unit tests somewhere," which is a much weaker statement than the description makes.

That's fine while the flag is informational, but the `e2eScope` description directly below makes it the gate for moving cases off the PR pipeline. A flag that can silently drift is a risky input to a coverage-removal decision.

Two options I'd be happy with:

- a test that iterates `selectCoveredCases` and fails when a flagged case has no fixture-driven spec consuming it — that makes the flag mechanically true, and gives `selectCoveredCases` its first caller
- make the flag carry its evidence (`unitCovered: { producer: "chat.contract.spec.ts" }`) so staleness is greppable

Otherwise, could we flag only the two cases this repo proves today and let the paired Hub SDK PR add its own?

apps/teams-test-app/e2e-test-data/calendar.json
```diff
@@ -26,7 +27,8 @@
         "content": "content"
       },
       "expectedAlertValue": "composeMeeting called with ##JSON_INPUT_VALUE##",
-      "expectedTestAppValue": "Completed"
+      "expectedTestAppValue": "Completed",
+      "unitCovered": true
```
Leah-Xia-Microsoft, line 31: Concrete example of the gap I described on `capabilities.schema.json`. This case's `inputValue` populates `attendees`, `startTime`, `endTime`, `subject`, and `content` — but the hand-written test it's leaning on, `packages/teams-js/test/public/calendar.spec.ts:7`, is:

```ts
const composeMeetingParams: calendar.ComposeMeetingParams = {};
```

An empty object. So the "unit coverage" backing this flag exercises none of the five fields the E2E case covers, and the two can't diverge detectably because they were never connected in the first place.

Same pattern in the others I checked:

| fixture | fixture data | hand-written spec |
| --- | --- | --- |
| `appInstallDialog.json` | appId `957f8a7e-fbcd-411d-b69f-acb7eb58b515` | `appInstallDialog.spec.ts:13` → `appId: '0'` |
| `sharing.json` | `https://bing.com` | `sharing.spec.ts:78` → `https://www.microsoft.com` |
| `appEntity.json` | structured object w/ `threadId`/`categories`/`subEntityId` | `appEntity.spec.ts:5` → `('threadID', [], '', () => {})` |

vhuseinova-msft: Added a fixture-driven composeMeeting producer test using all five populated fields. It checks the emitted payload and host response.

apps/teams-test-app/capabilities.schema.json
vhuseinova-msft: All nine flagged cases are consumed by the Hub provider tests; broader Teams producer coverage is already planned. Clarified the schema so the flag does not imply paired coverage or E2E removal.

apps/teams-test-app/fixture-contract/index.ts
vhuseinova-msft: The shared parser is now included in the strict app tsconfig and an explicit .ts lint target. The existing src lint scope is unchanged.

palatter (dismissed):

packages/teams-js/test/public/calendar.contract.spec.ts
```diff
@@ -19,6 +19,24 @@ describe('calendar contract', () => {
...
+    const fixtureCase = loadFixtureCase<calendar.ComposeMeetingParams>('calendar', 'composeMeeting API Call - Success');
+
+    const composeMeetingPromise = calendar.composeMeeting(fixtureCase.inputValue);
+    const message = utils.findMessageByFunc('calendar.composeMeeting');
+
+    expect(message).not.toBeNull();
+    if (!message) {
+      throw new Error('calendar.composeMeeting message not found');
+    }
+    expect(message.args).toEqual([fixtureCase.inputValue]);
```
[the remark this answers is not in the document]
vhuseinova-msft: Updated the composeMeeting test to check that expectedWirePayload is defined and compare the emitted arguments against it. This also catches changes to the fixture’s expected payload.

apps/teams-test-app/fixture-contract/index.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  return [];
+}
+
+/** Flattens a fixture file's top-level cases together with any nested feature-test cases. */
+export function getFixtureCases(fixture: FixtureFileRecord): FixtureCaseRecord[] {
+  return [
+    ...(fixture.testCases ?? []),
+    ...(fixture.featureTests ?? []).flatMap((featureTest) => featureTest.testCases ?? []),
```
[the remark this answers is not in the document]
vhuseinova-msft: The contract tests pass with the current Jest configuration: the Node type declarations include the library needed for flatMap. I’ll keep the existing implementation unless there’s a failing configuration we’re missing.

vhuseinova-msft:
I’ve added the selector regression tests from [this review](https://github.com/OfficeDev/microsoft-teams-library-js/pull/3116#pullrequestreview-5182305543) to the plan for the next follow-up PR: nested flagged/unflagged cases, ambiguous titles, and version selection.

Leah-Xia-Microsoft (approved):

Leah-Xia-Microsoft (approved):

_Automated review._

palatter (approved):

PR #3123, opened 2026-08-20
Cover the mobile and version guards in internalAPIs.ts

edwardlee-msft:
## Description

`src/internal/internalAPIs.ts` had no spec of its own. `test/internal/internalAPI.spec.ts` (singular) sounds like the right home but doesn't test this file at all — it only covers `isFollowingApiVersionTagFormat` from `telemetry.ts`.

This adds `test/internal/internalAPIs.spec.ts` (25 tests), deliberately scoped to the genuinely uncovered branches. Two of the module's exports are already well exercised indirectly and are intentionally **not** duplicated here:

- `processAdditionalValidOrigins` — covered by `validOrigins.spec.ts`
- `throwExceptionIfMobileApiIsNotSupported`'s `OLD_PLATFORM` path — covered by `location.spec.ts`

### What's newly covered

| Gap | Tests |
| --- | --- |
| `isHostClientMobile` across all four mobile client types — `ipados` and `visionOS` are the newer additions and were untested | ✅ all 4 mobile types, all 9 non-mobile types, and `undefined` |
| `throwExceptionIfMobileApiIsNotSupported`'s `NOT_SUPPORTED_ON_PLATFORM` branch when the host isn't mobile | ✅ desktop host and undefined host |
| `isCurrentSDKVersionAtLeast` returning `false` when `compareSDKVersions` yields `NaN` on a malformed version string | ✅ malformed client version, malformed required version, unset client version |

Also included: the standard at-least/equal/below and default-`requiredVersion` comparisons, plus the happy path for `throwExceptionIfMobileApiIsNotSupported`.

## Drive-by fix

Line 129 of `internalAPIs.ts` read `specifuied` — corrected to `specified`. This is a doc-comment-only change on an `@internal` function, so a `patch` beachball change file is included.

## Validation

- `npx jest --testPathPattern "internal/internalAPIs.spec.ts"` → 25 passed
- `eslint` and `prettier --check` clean on both changed files

No production behavior changes.

### DIFF
--- a/change/@microsoft-teams-js-194726a3-306a-4889-b066-427997eafd72.json
+++ b/change/@microsoft-teams-js-194726a3-306a-4889-b066-427997eafd72.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed a typo in the internal `processAdditionalValidOrigins` documentation comment.",
+  "packageName": "@microsoft/teams-js",
+  "email": "edwardlee@microsoft.com",
+  "dependentChangeType": "none"
+}
--- a/packages/teams-js/src/internal/internalAPIs.ts
+++ b/packages/teams-js/src/internal/internalAPIs.ts
@@ -126,7 +126,7 @@ export function throwExceptionIfMobileApiIsNotSupported(
 
 /**
  * @hidden
- * Processes the valid origins specifuied by the user, de-duplicates and converts them into a regexp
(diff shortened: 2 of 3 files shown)


jimchou-dev (approved):

palatter (approved):
Good call giving the comment typo its own changefile rather than letting it ride along silently in a test PR. The guard coverage reads clean too, every case asserts a concrete return value or a specific error. Approving.

PR #3125, opened 2026-08-21
Cover the tryPolyfillWithNestedAppAuthBridge guards, including the nested-iframe check

edwardlee-msft:
## Description

`tryPolyfillWithNestedAppAuthBridge` (`src/internal/nestedAppAuthUtils.ts`) has five early-return guards and no spec of its own. This PR adds `test/internal/nestedAppAuthUtils.spec.ts` to cover them directly.

Being precise about what was and wasn't already covered:

- **Already covered indirectly** via `test/internal/communication.spec.ts` — the happy path (bridge gets created and wired up) and the frameless-window guard.
- **Not covered anywhere under `test/`** — and the reason for this PR:
  - the **nested-iframe guard** (`window.parent !== window.top`, line 135), a security-relevant check that stops default NAA bridge injection when the app is not the top-most frame;
  - the **"bridge already exists, skip polyfill"** branch;
  - the **`JSON.parse` failure** path (malformed `clientSupportedSDKVersion`);
  - the **missing `supports.nestedAppAuth`** path.

None of those branches were reachable from the existing tests, because `communication.spec.ts` always drives the function through a top-most mock window with a well-formed capability payload.

## Changes

Adds 9 tests against a purpose-built minimal window:

| Test | Guard |
| --- | --- |
| polyfills on a top-most window when the host supports NAA | happy path |
| frameless window → no bridge | `GlobalVars.isFramelessWindow` |
| `null` window → does not throw | `!window` |
| **nested iframe → no bridge** | `window.parent !== window.top` |
| single-level iframe (`parent === top`) → bridge created | boundary of the same guard |
| invalid JSON version string → no bridge | `JSON.parse` catch |
| no `supports` object → no bridge | `!parsedClientSupportedSDKVersion.supports?.nestedAppAuth` |
| `supports` without `nestedAppAuth` → no bridge | same |
| existing bridge is left untouched | `extendedWindow.nestedAppAuthBridge` |

The nested-iframe case also pins the boundary: an iframe directly under the top window still reports `parent === top`, so injection is allowed there — only genuinely nested frames are blocked.

## Validation

- `jest test/internal/nestedAppAuthUtils.spec.ts` — 9/9 passing.
- Mutation check: deleting the `window.parent !== window.top` guard from the source makes the nested-iframe test fail, and only that test — confirming it actually pins the behavior rather than passing vacuously.
- `eslint --max-warnings 0` and `prettier --check` clean.
- No change file needed: `beachball.config.js` ignores `**/test/**`.

Test-only change; no production code or public API is touched.

### DIFF
--- a/packages/teams-js/test/internal/nestedAppAuthUtils.spec.ts
+++ b/packages/teams-js/test/internal/nestedAppAuthUtils.spec.ts
@@ -1,0 +1,143 @@
+import { GlobalVars } from '../../src/internal/globalVars';
+import {
+  NestedAppAuthMessageEventNames,
+  NestedAuthExtendedWindow,
+  tryPolyfillWithNestedAppAuthBridge,
+} from '../../src/internal/nestedAppAuthUtils';
+
+/**
+ * These tests cover the early-return guards in tryPolyfillWithNestedAppAuthBridge directly.
+ * The happy path is exercised indirectly through communication.spec.ts, but the guards -
+ * in particular the nested iframe check, which prevents bridge injection into a non-top-most
+ * app - are only reachable by calling the function with a purpose-built window.
+ */
+describe('nestedAppAuthUtils', () => {
+  describe('tryPolyfillWithNestedAppAuthBridge', () => {
+    const supportedSDKVersion = JSON.stringify({ supports: { nestedAppAuth: {} } });
+
+    let handlers: {
+      onMessage: jest.Mock;
+      sendPostMessage: jest.Mock;
+    };
+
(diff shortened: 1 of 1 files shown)


JoshuaLai (approved):

palatter (approved):
Good catch that the nested-iframe guard had no coverage. The pair of tests either side of it is the useful part, since the passing case pins `parent === top` for a single-level iframe rather than only asserting the negative, so a regression that widened the check would still fail. Approving.

PR #3132, opened 2026-08-26
docs: fix broken @example block in search.registerHandlers

edwardlee-msft:
## Description

Two small doc bugs in the `search.registerHandlers` TSDoc `@example`:

1. **Wrong quote type.** The `onClosedHandler` line used single quotes around a string containing `${query.searchTerm}`, so the published example prints the placeholder verbatim instead of interpolating it. The other two handlers in the same example already use backticks.

2. **Malformed comment block.** The lines inside the fenced code block were missing their leading `*` continuation markers, which breaks the fenced code block in the generated API docs.

Adding the missing `*` markers means Prettier now treats the whole JSDoc block as a normal comment and normalizes its indentation (it was indented 3 spaces instead of 1), which accounts for the rest of the diff.

Comment-only change — no runtime behavior is affected.

## Validation

- `tsc --noEmit` on `packages/teams-js` passes
- `eslint` and `prettier --check` are clean on the modified file
- Beachball change file included (`patch`, documentation comment)

### DIFF
--- a/change/@microsoft-teams-js-7a7c8ccb-76ca-4dcf-917c-fa7aaa249656.json
+++ b/change/@microsoft-teams-js-7a7c8ccb-76ca-4dcf-917c-fa7aaa249656.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated documentation for `search` capability.",
+  "packageName": "@microsoft/teams-js",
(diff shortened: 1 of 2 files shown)


Leah-Xia-Microsoft (approved):

_Automated review._

jimchou-dev (approved):

PR #3136, opened 2026-08-26
Add safe TeamsJS release preparation and maintainer guidance

palatter:

## Description

I think the release runbook should enforce its safety claims through tested preparation commands rather than ask an agent to mutate and restore a developer checkout. This keeps the release skill and its contributor-facing policy explanation together, and consolidates the corrected parts of #3158 into this PR. Publication verification and immutable final metadata remain the separate lifecycle change in #3156.

### Main changes in the PR:

1. Pin the approved source and preview in a disposable detached worktree. Explicit status checks include hidden untracked files, failures stop the operation, and cleanup touches only the helper-owned candidate.
2. Prepare and verify the full pending change set using the actual Beachball CLI and existing preparation entrypoint. Source must belong to fetched `origin/main`; common publishing/write-token environment variables are not inherited by build children. This is not a sandbox for untrusted source.
3. Provide a controlled maintainer-approved exceptional version intent, applied only inside the isolated candidate. Restore the original contributor config byte-for-byte before review/staging, assert the exact version, and reject grouped or flat changes that would be masked by a lower prerelease-prefix intent. No relaxation of `main` or policy-restoration PR is required.
4. Check the generated version carriers, changelog, consumed change files, and both SRI carriers against built bytes. Scope staging, verify the index again before commit, and stop on branch/commit/push failures.
5. Correct the contributor menu and explain that Beachball validates disallowed files before bumping, not by silently demoting them. The current policy's allowance for hand-authored `premajor`, `preminor`, and `prepatch` is documented, not silently changed.

This is one preparation/runbook contract. The gate's size and path-group warnings are acknowledged: about 650 added source lines in five source files across guidance, root contributor documentation, tools, and CI wiring. No SDK runtime package is changed. Splitting the runbook from the helper would leave executable safety claims without their implementation.

### Validation performed:

1. Ran all 24 preparation tests under Node 18.20.8 and Node 20.20.1 with the actual installed Beachball 2.62.0 CLI.
2. Exercised isolated repositories/worktrees, hidden untracked files, failed status/lockfile/build paths, complete major and prerelease candidate preparation, mixed and grouped pending changes, exact-version rejection, preserved contributor config, index scope, wrong built bytes, and fail-closed runbook fences.
3. Inspected the installed CLI/config APIs and actual interactive choices, rather than inferring CLI behavior from a change-type helper alone.
4. Used the existing preparation entrypoint with explicit synthetic package-install/build doubles. This proves orchestration and Beachball behavior, not a real full dependency build or publication.
5. Formatted with the repository-pinned Prettier and reviewed the exact public diff before pushing.

### Unit Tests added:

Yes. The existing Node 18/20 CI matrix runs `tools/cli/prepare-release.test.js`.

### End-to-end tests added:

No live build publication, approvals, or remote release mutations. Tests execute actual local CLI entrypoints with synthetic isolated git state and build outputs.

### Change file added:

No. Only release tooling, documentation, and CI wiring change.

### Related PRs:

- #3158 is the public predecessor. Its accurate contributor correction is consolidated here; the invalid silent-demotion and restore-on-main procedure is replaced, not retained.
- #3156 supplies the separate public plan/receipt and immutable-finalization lifecycle.

### Next/remaining steps:

- [ ] Land the reviewed #3156 tools and integrate the existing publisher's final-byte export plus independent approval handoff before actual publication or finalization.
- [ ] Retain operational approval and complete publication evidence before announcements or manual promotion. The legacy downstream merge notification is unchanged and is not shipment proof.
- [ ] Semantic prerelease preparation does not authorize publishing to an unsupported destination. A separate approved destination adapter is still required; stable-version production verification and a GitHub prerelease label are distinct.
- [ ] Any decision to change the contributor policy for hand-authored `premajor`, `preminor`, or `prepatch` remains a separate maintainer decision.


.github/skills/release-teamsjs/SKILL.md
```diff
@@ -0,0 +1,210 @@
...
+### 1. Read the version at runtime
+
+Never bake a version into this file, a script, or a comment.
+
+```bash
+# What beachball would bump to, from the pending change files.
+pnpm beachball bump
+node -p "require('./packages/teams-js/package.json').version"
+# Undo the local bump once you have read it; the release branch does the real one.
+git checkout -- .
```
Leah-Xia-Microsoft: `git checkout -- .` discards every uncommitted tracked change in the working tree, not just the bump an agent following this step just made — should it be scoped to the files `beachball bump` writes?

_Automated review._

Leah-Xia-Microsoft (commented):
1 thing to look at before merge — see inline comment.

_Automated review._

.github/skills/release-teamsjs/SKILL.md
```diff
@@ -0,0 +1,210 @@
...
+
+### 2. Cut the release branch
+
+```bash
+git checkout main && git pull
+git checkout -b "release/<x.y.z>"
+git push --set-upstream origin "release/<x.y.z>"
+```
+
+Push it with **no changes**. The bump lands via PR in the next step, so the branch history stays reviewable.
```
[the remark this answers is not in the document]
palatter: Good catch. I now distinguish the public workflow from the manual sequence. The workflow text follows `.github/workflows/prerelease.yml`: `preRelease.js` runs first, the generated files remain uncommitted when `release/<x.y.z>` is pushed, and `create-pull-request` commits them on its working branch before opening the bump PR. The manual path is labeled separately and keeps the same remote branch result.

.github/skills/release-teamsjs/SKILL.md
palatter: Good catch. I replaced the broad checkout with a clean-tree guard and an `EXIT` trap that restores only the tracked paths Beachball can change here: the TeamsJS package manifest and changelog, `pnpm-lock.yaml`, and `change/`. I also tested both the clean rollback and dirty-tree failure paths; unrelated tracked work is preserved.

.github/skills/release-teamsjs/SKILL.md
```diff
@@ -0,0 +1,210 @@
...
+## How versioning works
+
+Versioning is [beachball](https://microsoft.github.io/beachball/), configured in `beachball.config.js`:
+
+- `publish: false` and `push: false`, so beachball computes versions and writes the changelog but never publishes. The pipelines do that.
+- `scope: ['packages/teams-js']`
+- `disallowedChangeTypes: ['major', 'prerelease']`
+- `ignorePatterns` includes `*.md`, so a docs-only change needs no change file.
+
+Contributors add change files with `pnpm changefile` (see `CONTRIBUTING.md`). A release consumes every pending change file, folds them into `CHANGELOG.md`, and deletes them.
```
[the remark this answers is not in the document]
palatter: Good catch. I updated every ambiguous reference to the exact public path, `packages/teams-js/CHANGELOG.md`, including the versioning overview, PR-description guidance, and reviewer checklist.

.github/skills/release-teamsjs/SKILL.md
```diff
@@ -0,0 +1,230 @@
...
+
+- every pending change file deleted
+- `packages/teams-js/CHANGELOG.md` with a new version section holding those entries, matching the PR description
+- `packages/teams-js/package.json` at the new version
+- `packages/teams-js/README.md` script `src` and integrity hash pointing at the new version
+- the teams-test-app CDN html and its `package.json` likewise
+
+### 4. Create the GitHub release, marked prerelease
+
+Tag `v<x.y.z>`, target `release/<x.y.z>`, title `v<x.y.z>`, body = the new changelog section. Tick **prerelease**. Do not attach binaries; publishing adds them.
```
Leah-Xia-Microsoft: Nothing between opening the bump PR in step 3 and tagging here says that PR has to merge first — should this step state that `release/<x.y.z>` must already carry the bump before the tag is cut?

_Automated review._

Leah-Xia-Microsoft (commented):
1 thing to look at before merge — see inline comment.

_Automated review._

.github/skills/release-teamsjs/SKILL.md
palatter: Good catch. I added a hard ordering gate: the bump PR must merge into `release/<x.y.z>` before tagging. The step now force-fetches the exact remote release ref and reads `packages/teams-js/package.json`; a missing ref, unreadable manifest, or version mismatch stops the workflow before the tag instructions.

.github/skills/release-teamsjs/README.md
```diff
@@ -0,0 +1,62 @@
...
+URLs, approval-system names, individual names or aliases, and internal distribution lists. When a
+release teaches something that cannot be written down without one of those, record it in the
+internal runbook and leave a neutral pointer here.
+
+## What is in this folder
+
+| File | Purpose |
+|---|---|
+| `SKILL.md` | The workflow the agent follows, failure modes, and hard-learned rules. |
+| `README.md` | This file. |
```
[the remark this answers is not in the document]
palatter: I looked into this one and I think the table may actually be fine as written. The file has no `||` sequences, and lines 59-62 are a standard header row, a `|---|---|` separator, and two body rows, which GitHub renders as a table. I may be missing something, so if you are seeing it render incorrectly in a specific view, let me know and I will take another pass.

.github/skills/release-teamsjs/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+
+**Write it the way the rest of this file is written:** evergreen and imperative, stating the invariant a future run must hold rather than the incident that revealed it. "A green pipeline is not proof the version reached the CDN" still reads correctly in a year; "the release failed last Tuesday" does not.
+
+Mechanics. This is a standalone PR touching skill files only; never fold it into a release branch, whose diff has to stay limited to the version carriers.
+
+```bash
+git fetch origin main
+git switch -c "<alias>/release-skill-learnings-$(date +%Y%m%d)" origin/main
+# Edit only .github/skills/release-teamsjs/**
+git add -A && git commit -m "Record what the <version> release taught the release skill"
```
JoshuaLai: The instructions require this PR to touch only `.github/skills/release-teamsjs/**`, but `git add -A` stages every modified and untracked path carried into this branch. After the earlier bump/build steps, this can accidentally include generated residue in the skill-only PR. Please scope staging to the skill directory (for example, `git add -- .github/skills/release-teamsjs`) and/or assert a clean worktree before switching branches.

palatter: Good catch, and this is a real hole rather than a theoretical one. I reproduced it: with typical post-release residue in the tree (a bumped `packages/teams-js/package.json`, a generated `change/*.json`, and build output), the old block staged three non-skill files into a PR whose stated contract is skill files only. Fixed in d107beae.

I took both halves of your suggestion, and one thing your comment made me realize goes further. Scoping the add alone is not sufficient: `git add -- <dir>` does not unstage anything, so a file staged before this block still reaches the commit. So the block now enforces the scope instead of printing a list for a human to eyeball, and refuses to commit any staged path outside the skill directory. Rename detection is disabled there because a `git mv` out of the directory otherwise reports only the destination and hides the deletion.

On the clean-tree assert, I used `--untracked-files=all`. That turned out to matter: with `status.showUntrackedFiles=no` set locally, plain `git status --porcelain` reports zero changes and the guard would have waved the residue straight through. The failure message now also points at `git stash push --include-untracked`, since plain `git stash` leaves untracked files behind and would send you back into the same guard.

One change beyond what you raised, flagged so it is not a surprise: the branch example moves from `%Y%m%d` to `%Y%m%d-%H%M%S`. Two learning PRs on the same day currently collide on the branch name. It fails loudly rather than silently so it is minor, and the three internal sibling skills already carry that fix, so leaving the public one divergent seemed likely to reintroduce it. Happy to split it back out if you would rather keep this commit strictly to the staging defect.

Verified: the old block leaks 3 files and the new one leaks 0 on the same residue; pre-staged file, cross-directory rename, and nothing-staged all refuse; `bash -n` passes on all 9 blocks.

.github/skills/release-teamsjs/SKILL.md
```diff
@@ -0,0 +1,294 @@
...
+State the version, the branch, and where you will stop (the internal pipeline step). One short block, before touching anything.
+
+### 1. Read the version at runtime
+
+Never bake a version into this file, a script, or a comment.
+
+```bash
+set -e
+
+if [ -n "$(git status --porcelain)" ]; then
```
JoshuaLai, line 53: This clean-tree guard omits `--untracked-files=all`, unlike the later guard. With `status.showUntrackedFiles=no`, an untracked file under `change/` is hidden here; `pnpm beachball bump` can then consume/delete it, while the EXIT restore from `HEAD` cannot recover an untracked file. Use `git status --porcelain --untracked-files=all` here as well to avoid losing an uncommitted change file.

palatter: Good catch. The preview no longer bumps or restores the developer checkout. It explicitly checks status with --untracked-files=all, pins the source, and runs Beachball in a disposable detached worktree. A failed status command stops before installation or bumping. The regression covers hidden untracked change files and preserves them, alongside failure cleanup and scoped staging.

tools/cli/prepare-release.test.js
```diff
@@ -0,0 +1,639 @@
...
+after(() => fs.rmSync(testRoot, { recursive: true, force: true }));
+
+function run(command, args, options = {}) {
+  const result = spawnSync(command, args, {
+    cwd: options.cwd,
+    encoding: 'utf8',
+    env: options.env || process.env,
+    input: options.input,
+    maxBuffer: 10 * 1024 * 1024,
+  });
```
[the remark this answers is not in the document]
palatter: Good catch. I separated the Bash-fence execution from the argv-only helper runner. The fences now go to a literal bash -s command through stdin, rather than passing their text through the generic command-argument path. The preparation tests also run from a temporary path containing spaces and an apostrophe. No alert suppression or production helper change is involved.

tools/cli/prepare-release.js
```diff
@@ -0,0 +1,418 @@
...
+  return run('git', ['-C', cwd, 'rev-parse', '--show-toplevel']).trim();
+}
+
+function parseArguments(argv) {
+  const [command, ...rest] = argv;
+  if (!['preview', 'prepare', 'verify', 'stage', 'check-staged'].includes(command)) {
+    throw new Error(
+      'Usage: prepare-release.js <preview|prepare|verify> --source-commit <SHA> [--expected-version <semver>] [--intent-file <approved.json>]',
+    );
+  }
```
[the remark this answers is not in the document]
palatter: Good catch. The accepted command list now also drives the usage text, so stage and check-staged are included. The missing-version diagnostic names the actual command. The argument regression covers both messages.

tools/cli/prepare-release.js
```diff
@@ -0,0 +1,419 @@
...
+  if (expectedVersion && !SEMVER_PATTERN.test(expectedVersion)) {
+    throw new Error('--expected-version must be an exact semantic version');
+  }
+  if (command !== 'preview' && !expectedVersion) {
+    throw new Error(`--expected-version is required for ${command}`);
+  }
+  if (['verify', 'stage', 'check-staged'].includes(command) && values['--intent-file'])
+    throw new Error('Verification does not consume an exception intent');
+  return { command, sourceCommit, expectedVersion, intentFile: values['--intent-file'] };
```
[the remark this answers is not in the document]
palatter: I checked this against the actual readIntent implementation, and I think the TypeError path is already blocked. The exact-field check requires expectedVersion in the JSON, then intent.expectedVersion !== expectedVersion rejects an omitted CLI value before the includes call. A missing JSON field fails the earlier field-set check instead. Executing that path returns the intentional source/version/channel mismatch error, with no mutation. stage and check-staged are both verification operations and intentionally cannot consume an exception intent. I am leaving the fail-closed behavior unchanged rather than treating this as an exception-gate bypass.

palatter:
I think this is better picked up later. Closing without merging and keeping the branch and commits intact for a future restart.

PR #3143, opened 2026-08-28
Release pipeline (4/4): automate GitHub pre-release creation on release PR merge- #3141

jekloudaMSFT:
## Description

Part 4 of the release-pipeline streamlining stack. This PR automates **GitHub pre-release creation** when a release PR merges, removing the manual "create the GitHub Release on the Releases page" step from the happy path. It also folds the old `postrelease.yml` (TeamsFx dispatch) into a single orchestration workflow.

Stacked directly on #3090 (`pr1-changelog-util`) — independent of the validation PRs, so it can be reviewed and merged in parallel with them.

### Main changes in the PR:

1. Add `tools/cli/create-github-release.js` — idempotent `gh`-based pre-release creation/update. Sources release notes from the shared changelog extractor (#3090), so re-runs update rather than duplicate the release.
2. Add `.github/workflows/release-orchestrate.yml` — on release-PR merge, generates notes and creates the idempotent GitHub **pre-release**, and preserves the existing TeamsFx dispatch.
3. Delete `.github/workflows/postrelease.yml` — its behavior is folded into `release-orchestrate.yml`.

### Validation performed:

1. Ran `create-github-release.js` logic against a fixture changelog to confirm the correct section is extracted and the create/update path is idempotent.
2. Confirmed the TeamsFx dispatch step carried over from `postrelease.yml` is preserved in `release-orchestrate.yml`.

### Unit Tests added:

No direct unit test for the `gh`-invoking script (it shells out to the GitHub CLI/API); its changelog-extraction dependency is covered by the tests in #3090. The workflow is validated by execution on release-PR merge.

### End-to-end tests added:

No — validated via the workflow run on a real release PR merge.

### Change file added:

No — CI/release tooling only; not a published-package change, so beachball does not apply.

### Related PRs:

- Depends on #3090 (`pr1-changelog-util`) for the shared changelog extractor — must merge first.
- Sibling of the `pr2`/`pr3` validation stack (independent; can merge in parallel).

### Next/remaining steps:

- [ ] Phase 2: ADO → GitHub callback to promote the pre-release to a full release after prod publish.


tools/cli/create-github-release.js
```diff
@@ -0,0 +1,99 @@
...
+  const tag = `v${version}`;
+  const title = prerelease ? `v${version} (pending prod publish)` : `v${version}`;
+
+  if (releaseExists(tag)) {
+    console.log(`Release ${tag} already exists; updating notes/title.`);
+    const args = ['release', 'edit', tag, '--title', title];
+    if (notesFile) {
+      args.push('--notes-file', notesFile);
+    }
+    args.push('--prerelease', prerelease ? 'true' : 'false');
```
Leah-Xia-Microsoft, line 58: `--prerelease` is a boolean flag on `gh release edit`, so passing `'true'` as a separate argument makes the existing-release path fail with `accepts 1 arg(s), received 2`.

_Automated review._

Leah-Xia-Microsoft (commented):
1 thing to look at before merge — see inline comment.

_Automated review._

Leah-Xia-Microsoft (commented):
Not approving — the open thread on `tools/cli/create-github-release.js` is still unaddressed at this head: `gh release edit <tag> --prerelease true` exits with `accepts 1 arg(s), received 2`, so the existing-release path fails on any re-run.

No new findings beyond the threads already open.

_Automated review._

.github/workflows/release-orchestrate.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
+
+      - name: Extract version
+        id: extract_version
+        run: echo "version=${GITHUB_BASE_REF#release/}" >> "$GITHUB_OUTPUT"
+
+      - name: Generate release notes from changelog
+        run: node tools/cli/extract-changelog-section.js ${{ steps.extract_version.outputs.version }} > release-notes.md
```
JoshuaLai, line 33: **Security (script injection): pass these values through `env:` instead of interpolating `${{ }}` into `run:`.**

This line — and lines 40-41 in the next step — splice GitHub expressions unquoted into a shell command. `steps.extract_version.outputs.version` is `${GITHUB_BASE_REF#release/}`, i.e. the raw base branch name, and `github.base_ref` is that branch name verbatim. Neither is validated. Git ref names allow `$`, backtick, `(`, `)`, `;`, `&`, and `|`, and the `release/*` filter matches any single-segment suffix, so a branch like `release/2.56.0;<cmd>` or `release/$(<cmd>)` executes arbitrary commands on the runner.

Impact is higher here than in a read-only workflow: this job declares `permissions: contents: write` and step 35 puts `GH_TOKEN` in the environment, so injected code inherits a write-scoped token for this repo.

Suggested shape for both steps:

```yaml
      - name: Generate release notes from changelog
        env:
          RELEASE_VERSION: ${{ steps.extract_version.outputs.version }}
        run: node tools/cli/extract-changelog-section.js "$RELEASE_VERSION" > release-notes.md

      - name: Create GitHub pre-release
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          RELEASE_VERSION: ${{ steps.extract_version.outputs.version }}
          RELEASE_TARGET: ${{ github.base_ref }}
        run: |
          node tools/cli/create-github-release.js \
            --version "$RELEASE_VERSION" \
            --target "$RELEASE_TARGET" \
            --notes-file release-notes.md \
            --prerelease
```

Quoting matters independently of injection: `parseArgs` in `create-github-release.js` consumes the next non-`--` token as a value, so if `version` ever expands to empty the unquoted form silently shifts `--target` into the `--version` slot.

tools/cli/create-github-release.js
```diff
@@ -0,0 +1,99 @@
...
+      notesFile: typeof args['notes-file'] === 'string' ? args['notes-file'] : undefined,
+      prerelease: Boolean(args.prerelease),
+    });
+  } catch (e) {
+    console.error(e.message || e);
+    process.exit(1);
+  }
+}
+
+module.exports = { createOrUpdateRelease };
```
JoshuaLai, line 99: **Missing test coverage for `createOrUpdateRelease` — the argument vectors are the whole behavior of this module, and they are currently unverified.**

The PR description justifies the gap with "it shells out to the GitHub CLI/API", but the module is already structured for testing: `createOrUpdateRelease` is exported here, all `gh` invocations go through `execFileSync`, and `tools/cli/test/` already runs jest with three sibling specs (`extract-changelog-section.spec.js`, `validate-release-integrity.spec.js`, `verify-sri.spec.js`). Mocking `child_process.execFileSync` and asserting the exact `argv` is a few lines and needs no network or `gh` binary.

This is not hypothetical: the `--prerelease`/`'true'` defect open on line 58 is precisely an argv-shape bug, and a single assertion on the update path would have caught it before review.

Suggested cases:

```js
jest.mock('child_process');
const { execFileSync } = require('child_process');
const { createOrUpdateRelease } = require('../create-github-release');

// existing release -> edit path, and the edit argv is a valid `gh release edit` invocation
// missing release  -> create path, includes --target and --prerelease
// no notesFile     -> create path falls back to --generate-notes
// releaseExists()  -> `gh release view` failure routes to create, not a thrown error
```

Without this, the only validation of the release-creation path is a live run against a real release PR merge, where a bad argv fails after the merge has already landed.

Leah-Xia-Microsoft (commented):
Not approving — the `--prerelease` argv defect already flagged inline is still present at this head.

_Automated review._

PR #3144, opened 2026-08-28
test: add unit tests for isExternalAppError type guard

edwardlee-msft:
## Description

`isExternalAppError` in `src/private/externalAppErrorHandling.ts` was an exported, untested type guard. This PR adds `test/private/externalAppErrorHandling.spec.ts` covering its full truth table:

- non-object inputs (string, number, boolean, `undefined`, symbol, function) → `false`
- `null` → `false` (the explicit guard that exists because `typeof null === 'object'`)
- a valid `ExternalAppErrorCode` with the message omitted, explicitly `undefined`, or a string (including `''`) → `true`
- a valid `ExternalAppErrorCode` with a non-string message (number, boolean, `null`, object, array, function) → `false`
- unknown / missing / wrong-cased error codes → `false`
- an `Error` instance without a valid `errorCode` → `false`, and one augmented with a valid `errorCode` → `true`
- the guard actually narrows `unknown` to `ExternalAppError` at the type level

Tests only — no product code changes, so no beachball change file (consistent with other test-only PRs such as #3135).

## Validation

```
npx jest test/private/externalAppErrorHandling.spec.ts   # 11 passed
npx eslint test/private/externalAppErrorHandling.spec.ts # clean
npx tsc --noEmit -p tsconfig.json                        # clean
```

### DIFF
--- a/packages/teams-js/test/private/externalAppErrorHandling.spec.ts
+++ b/packages/teams-js/test/private/externalAppErrorHandling.spec.ts
@@ -1,0 +1,77 @@
+import { ExternalAppErrorCode } from '../../src/private/constants';
+import { isExternalAppError } from '../../src/private/externalAppErrorHandling';
+
+describe('externalAppErrorHandling', () => {
+  describe('isExternalAppError', () => {
+    it('should return false for non-object values', () => {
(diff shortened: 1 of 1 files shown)


Leah-Xia-Microsoft (approved):

_Automated review._

jimchou-dev (approved):

PR #3158, opened 2026-09-01
Document the maintainer path for major and prerelease releases

palatter:
I think keeping one preparation runbook is clearer than maintaining a separate exception recipe. The accurate contributor/menu correction and all substantive feedback from this PR are now carried in #3136, which replaces the incorrect silent-demotion and restore-on-main advice with controlled isolated preparation. #3156 remains separate for executable publication verification and immutable metadata. Closing this predecessor without merging; its review history stays here. The open policy choice for premajor/preminor/prepatch and operational publication approvals remain explicit in #3136.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,18 +50,32 @@ The TeamsJS Client library contains a [Change Log](./packages/teams-js/CHANGELOG
 
 Beachball generates JSON change files based on a few simple answers from you:
 
-- Change type: this can be one of four types: Major, Minor, Patch, and None. In addition to the descriptions provided in the prompt, the following can help guide which type to choose:
+- Change type: contributor PRs select from Minor, Patch, and None. In addition to the descriptions provided in the prompt, the following can help guide which type to choose:
 
-  - Major - when you make incompatible API changes,
   - Minor - when you add functionality in a backwards compatible manner,
-  - Patch - when you make backwards compatible bug fixes
+  - Patch - when you make backwards compatible bug fixes,
   - None - when the change does not affect the published package in any way.
+
+  The Major and Prerelease change types are gated by `disallowedChangeTypes` in [beachball.config.js](./beachball.config.js) and are not available from the interactive prompt. See [Cutting a major or prerelease release](#cutting-a-major-or-prerelease-release) for the maintainer procedure.
 
 - Describe changes: Type your own message or choose one of the commit messages. Try to make it descriptive - it will help you if you need to locate the change file later.
   - Please use past tense (e.g., "Added comments to \`app.initialize\`")
   - Enclose function/interface/enum/etc. names in backticks
 
 And that's it! As easy as hitting 'enter' twice. Beachball will automatically commit the change file you've created. All you have to do is run `pnpm changefile` in the monorepo root to do the above change file generation as the last step in your branch to make sure your PR is ready for review. Our pipelines will check to see if you generated a change file and will fail if you forgot. If they do, please create the change file as per the steps listed and update the content accordingly.
+
+### Cutting a major or prerelease release
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -50,19 +50,33 @@ The TeamsJS Client library contains a [Change Log](./packages/teams-js/CHANGELOG
...
 
+### Cutting a major or prerelease release
+
+The `major` and `prerelease` change types are intentionally excluded from the interactive change-file prompt so an ordinary contributor PR cannot introduce a major or prerelease version bump. This gate is enforced by `disallowedChangeTypes` in [beachball.config.js](./beachball.config.js), and `beachball check` will reject a hand-authored change file that uses either type while the gate is in place. Beachball also silently demotes a disallowed type at bump time with no warning: with the current `['major', 'prerelease']` gate, a hand-authored `"type": "major"` file collapses to a `premajor` bump (so a maintainer expecting `3.0.0` would ship `3.0.0-0`), and a `"type": "prerelease"` file collapses to `none` (so the change file is dropped from the release entirely). The gate therefore must be relaxed at the moment `beachball bump` runs, not just when the change file is authored.
```
vhuseinova-msft, line 69: Are we intentionally allowing `premajor`, `preminor`, and `prepatch` here? Beachball accepts them, but the config only blocks `major` and `prerelease`, so a hand-authored `premajor` change can still produce `3.0.0-0`.

🤖 AI-assisted

vhuseinova-msft, line 69: I don’t think the demotion described here happens. `beachball bump` validates first and fails on a disallowed type. If validation is bypassed, direct changes don’t apply `disallowedChangeTypes` and keep their original type. Could we update this explanation?

🤖 AI-assisted

palatter: Good catch. The existing policy does allow hand-authored premajor, preminor, and prepatch files, although they are not choices in the stable-package interactive prompt. #3136 now documents that distinction and requires an independently approved intent for those types in the preparation helper. I have not changed the contributor policy itself; deciding whether to disallow those types remains an explicit maintainer decision carried forward in #3136.

palatter: You're right. Running the actual Beachball 2.62 CLI confirms that it validates and rejects disallowed files before bumping; the helper-only experiment did not establish silent publication. #3136 replaces that explanation and the restore-on-main recipe with tested isolated preparation, original-policy restoration inside the candidate, and an exact whole-set version assertion. Publication and its approval gates remain separate.

CONTRIBUTING.md
```diff
@@ -50,19 +50,33 @@ The TeamsJS Client library contains a [Change Log](./packages/teams-js/CHANGELOG
...
 
+### Cutting a major or prerelease release
+
+The `major` and `prerelease` change types are intentionally excluded from the interactive change-file prompt so an ordinary contributor PR cannot introduce a major or prerelease version bump. This gate is enforced by `disallowedChangeTypes` in [beachball.config.js](./beachball.config.js), and `beachball check` will reject a hand-authored change file that uses either type while the gate is in place. Beachball also silently demotes a disallowed type at bump time, so the gate must be relaxed at the moment `beachball bump` runs, not just when the change file is authored.
+
+To ship a major or prerelease version, a maintainer must:
+
+1. Open a coordinated release PR that removes the relevant entry (`'major'` and/or `'prerelease'`) from `disallowedChangeTypes` in [beachball.config.js](./beachball.config.js).
+2. Add the corresponding change file by hand under [change](./change) with `"type"` set to `"major"` or `"prerelease"`, `"dependentChangeType"` set to `"patch"` (matching the shape beachball generates), the correct `"packageName"`, and a clear description of the breaking change.
```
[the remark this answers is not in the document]
palatter: Good catch. I removed the hand-authored change-file recipe rather than treating patch as a universal dependentChangeType. The replacement in #3136 uses a reviewed source/version intent in an isolated candidate, with a synthetic preparation-only entry whose dependentChangeType is none; it does not describe that as the standard contributor-generated shape. Major stable preparation and semantic prerelease preparation are documented separately.

PR #3168, opened 2026-09-16
Add M365 admin center origins to validDomains.json

Leah-Xia-Microsoft:
## Summary

Adds `admin.microsoft.com` and `m365.admin.cloud.microsoft` to the valid origins list.

## Why

Investigating **ICM 51000001501664** (S500 customer): setting up a Custom Federated (MCP) connector in the M365 admin center fails with *"Authentication for connector was cancelled or closed before completion"* — even though the user never closes the popup.

Traced from a customer browser `net-export` capture (which, unlike a HAR, records popup traffic):

1. Entra sign-in **succeeds** — 302 carrying a valid authorization code.
2. The popup loads `teams.microsoft.com/api/platform/v1.0/oAuthRedirect`, which returns `Cross-Origin-Opener-Policy: unsafe-none`, so `window.opener` is intact.
3. That page calls `app.initialize(validMessageOrigins)` with a hardcoded list containing only M365 Copilot origins.
4. The parent's origin is `https://admin.microsoft.com` — present in **neither** that list nor `validDomains.json`.
5. `validateOrigin()` returns false, the parent's reply is dropped, `app.initialize()` never resolves (there is no timeout), and the popup posts nothing back.
6. The Hub SDK's popup poll then falls through to its default reason, `CancelledByUser`.

Nothing fails at the network layer, which is why earlier HAR captures looked healthy — the check is client-side.

## Coverage check

All four **non-production** rings are already listed (`ignite`, `sdf`, `canary`, `ring0`, plus `admin-sdf.microsoft.com`), but production is absent.

`validateHostAgainstPattern` requires an exact segment-count match and `*` matches exactly one segment, so no existing wildcard covers either host:

| Host | Before | After |
| --- | --- | --- |
| `admin.microsoft.com` | not covered | ✅ covered |
| `m365.admin.cloud.microsoft` | not covered | ✅ covered |
| `admin.cloud.microsoft` | already covered by `*.cloud.microsoft` | unchanged — **not** added |

## Size-limit increase (required explanation)

Per the `size-limit-explanation` guidance in `package.json`, here is why the limit moved:

**Change:** first `size-limit` entry (the treeshakability check for `{ app, authentication, pages }`) raised from **60.8 KB → 61 KB**.

**Why it was necessary.** `validDomains.json` is not CDN-only — it is imported by `src/internal/constants.ts` and re-exported as the offline fallback:

```ts
export const validOriginsFallback = validOriginsLocal.validOrigins;
```

so the list ships inside the bundle. Adding the two origins above pushed the measured build just past the existing limit and failed CI:

```
Package size limit has exceeded by 39 B
Size limit:   60.8 kB
Size:         60.84 kB  with all dependencies and minified
```

**Which scenario this is.** Case **2** in the guidance: the change is to an internal-folder file, the size increase is intentional, and it **does not affect treeshakability**. No capability moved in or out of the treeshaken graph — the only delta is two additional strings in an already-included constant. The second size-limit check (`{ nestedAppAuthBridge }`, 4 KB) was unaffected and still passes at 3.51 kB.

**Why 61 KB.** Smallest clean value above the measured 60.84 kB, leaving ~160 B of headroom. This follows the precedent in #2943, which raised the same limit 59 KB → 60 KB when adding origins for the M365 Copilot SSR endpoints.

The bump cannot be avoided by trimming the list: the two entries are what the fix consists of, and the overage (39 B) accounts for most of what they add to the minified bundle.

## Note on rollout

The CDN serves `validDomains.json` with `cache-control: max-age=31536000` on an unversioned URL. The customer's capture shows their browser read the list **from HTTP cache with no network request**, so this change may take a while to reach existing clients. Worth tracking separately from this PR.

## Related incidents

- **51000001501664** — this one.
- **822163896** — same `oAuthRedirect` page, different trigger (opener severed by a COOP header). Closed as "Customer Error."
- **834062705** — proposes removing the dependency entirely by having the Hub SDK poll the TGS backend instead of relying on the popup messaging back.


Leah-Xia-Microsoft (commented):

_Automated review._

ndangudubiyyam (dismissed):
LGTM

Leah-Xia-Microsoft (commented):
No blocking issues — checked correctness, security, breaking changes, data safety, and test integrity. Not approving because the review account authored this PR.

_Automated review._

vhuseinova-msft (approved):

ndangudubiyyam (approved):
Ship it

